PR #29, opened 2017-05-03
Refactor batch parley, prepare for MultiAgent Batch Parley

jaseweston:


parlai/core/worlds.py
```diff
@@ -75,6 +76,63 @@ def __init__(self, opt, agents=None, shared=None):
...
+    def getID(self):
+        """Return the name of the world, typically the task the world encodes."""
+        return self.id
+
+    def display(self):
+        """Returns a string describing the current state of the world.
+        Useful for monitoring and debugging.
+        By default, display the messages between the agents."""
+        lines = []
+        for index, msg in enumerate(self.acts):
```
Tuzongyu: maybe self.get_acts(), with a `return []` base definition? this could get a runtime error right now since self.acts isn't defined in __init__.

jaseweston: ok done

parlai/core/worlds.py
```diff
@@ -424,6 +428,8 @@ def __init__(self, opt, world):
             opti = copy.deepcopy(opt)
             opti['batchindex'] = i
             self.worlds.append(shared['world_class'](opti, None, shared))
+        self.batch_acts = [ [] ] * len(self.worlds)
```
Tuzongyu: this needs to be: `self.batch_acts = [ [] for _ in self.worlds]`
because:
```python
>>> [ [] ] * 3
[[], [], []]
>>> l = [ [] ] * 3
>>> l[0].append(1)
>>> l
[[1], [1], [1]]
```

jaseweston: done

PR #35, opened 2017-05-03
simplified data class and moved indexing into dialog teacher

Tuzongyu:


parlai/core/data.py
```diff
@@ -3,9 +3,7 @@
 # This source code is licensed under the BSD-style license found in the
```
jaseweston: DialogData and dialog_data.py ?

PR #41, opened 2017-05-03
better check_examples (exits squad after 60s)

Tuzongyu:


Tuzongyu:
also reverted required=True for tasks arg (maybe you want to agents to talk, etc)

PR #54, opened 2017-05-08
Add VQA V2.0 and Visual Dialog V0.9.

jiasenlu:
This pull request contains 
1: add vqa_v2 task loader.
2: add VisDial_v0.9 task loader.
3: disable download COCO image as default 
4: move COCO image path to "--download_path" 

To test the new added task

For VQA 2.0: python examples/display_data.py -t vqa_coco2014_v2 --download-path 'path_to_COCO_img'

For Visual Dialog: python examples/display_data.py -t visdial

Currently, the Visual Dialog inherit from the default "DialogTeacher" class. There is no placeholder for the additional image information. The "DialogData" class with the format [(x, y, r, c), new_episode?], could we extend this to  [(x, y, r, c, i), new_episode?] where "i" is some optional additional information such as image_id? If so, I can send another pull request to modify that.

### DIFF
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -265,7 +265,7 @@ class DialogData(object):
                 table['reward'] = entry[2]
                 if len(entry) > 3:
                     table['label_candidates'] = entry[3]
-                    if len(entry) > 4 and not opt.get('no_images', False):
+                    if len(entry) > 4 and not self.opt.get('no_images', False):
                         table['image'] = load_image(self.opt, entry[4])
 
 
--- a/parlai/tasks/vqa_coco2014/agents.py
+++ b/parlai/tasks/vqa_coco2014/agents.py
@@ -11,7 +11,7 @@ from PIL import Image
 import json
 import random
 import os
-
+import pdb
 
 def _path(opt):
     build(opt)
@@ -36,7 +36,7 @@ def _path(opt):
     annotation_path = os.path.join(opt['datapath'], 'VQA-COCO2014',
         annotation_suffix + '_annotations.json')
 
-    image_path = os.path.join(opt['datapath'], 'VQA-COCO2014', img_suffix)
+    image_path = os.path.join(opt['download_path'], img_suffix)
 
     return data_path, annotation_path, image_path
 
--- a/parlai/tasks/vqa_coco2014/agents.py
+++ b/parlai/tasks/vqa_coco2014/agents.py
@@ -14,6 +14,7 @@ import os
 
 def _path(opt):
     build(opt)
+    buildImage(opt)
     dt = opt['datatype'].split(':')[0]
 
     if dt == 'train':
@@ -35,7 +36,7 @@ def _path(opt):
     annotation_path = os.path.join(opt['datapath'], 'VQA-COCO2014',
         annotation_suffix + '_annotations.json')
 
-    image_path = os.path.join(opt['download_path'], img_suffix)
+    image_path = os.path.join(opt['datapath'], 'COCO-IMG', img_suffix)
 
     return data_path, annotation_path, image_path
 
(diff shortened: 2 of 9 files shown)


alexholdenmiller:
Thanks Jiasen!! I think we'd like to keep the dataset downloading the images automatically (to datapath)--I know it's large, but users should only need to do it once anyways. Everything in downloads is also automatically downloaded (currently just the Memnn github repo).

As far as DialogTeacher/DialogData, I think we want to write a new one instead that includes some of the data loading from your vqa_coco2014 (v1) and prepares us to handle the preprocessing options that we talked about before. We don't want image_id as part of the action/observation dictionary--just images themselves--so we want to include the image loading as part of that data class.

parlai/tasks/vqa_coco2014/agents.py
```diff
@@ -36,7 +35,7 @@ def _path(opt):
     annotation_path = os.path.join(opt['datapath'], 'VQA-COCO2014',
         annotation_suffix + '_annotations.json')
 
-    image_path = os.path.join(opt['datapath'], 'VQA-COCO2014', img_suffix)
```
jiasenlu: I see. Then I think we might not want the image stay in 'VQA-COCO2014' folder, other task such as Visual Dialog may also use the COCO image. How about put the image under `COCO-IMG` ? Then, multiple task can share the image data.

parlai/tasks/vqa_coco2014_v2/agents.py
```diff
@@ -0,0 +1,107 @@
...
+        qa = self.ques['questions'][self.episode_idx]
+        question = qa['question']
+        image_id = qa['image_id']
+        # question_id = qa['question_id']
+
+        if self.datatype != 'test':
+            anno = self.annotation['annotations'][self.episode_idx]
+            answers = [ans['answer'] for ans in anno['answers']]
+        else:
+            answers = ['fake_answer']
```
jiasenlu: ok, thanks. will change this.

parlai/tasks/vqa_coco2014/agents.py
alexholdenmiller: Sounds great! Yeah giving it the most general name for the data file is perfect, and then multiple tasks can depend on it (and it won't rebuild it if it's already there).

parlai/core/dialog_teacher.py
```diff
@@ -265,7 +265,7 @@ def get(self, episode_idx, entry_idx=0):
                 table['reward'] = entry[2]
                 if len(entry) > 3:
                     table['label_candidates'] = entry[3]
-                    if len(entry) > 4 and not opt.get('no_images', False):
+                    if len(entry) > 4 and not self.opt.get('no_images', False):
```
alexholdenmiller: ah good catch thank you

parlai/tasks/vqa_coco2014_v2/agents.py
```diff
@@ -0,0 +1,146 @@
...
+
+        if self.datatype != 'test':
+            print('loading: ' + annotation_path)
+            with open(annotation_path) as data_file:
+                self.annotation = json.load(data_file)
+
+        self.len = len(self.ques['questions'])
+
+class DefaultTeacher(OeTeacher):
+    pass
```
alexholdenmiller: does v2 have a multiple-choice version?

jaseweston:
yes, we need to have at least a simple example of visualizing the images..

On Thu, May 11, 2017 at 11:37 AM, Alexander Miller <notifications@github.com

parlai/tasks/vqa_coco2014_v2/agents.py
jiasenlu: no, VQA v2.o doesn't have the multiple-choice now.

alexholdenmiller:
simple example which prints ascii version of images (after `pip install asciimatics`), or we could write our own similar one

```python
import sys
from asciimatics.renderers import ImageFile
img = ImageFile(sys.argv[1], height=30)
print(img)
```

jiasenlu:
Yeah, I think we should do that. Maybe we can add this in a separate pull request?

alexholdenmiller:
Just pushed this to get that fix in

PR #62, opened 2017-05-09
fixes to BatchWorld

ajfisch:
A number of fixes to batch world logic.

- batch_observe: Each agent in each world gets the right observation for its respective world.

- naming: batch_act takes in batch_observations, and returns batch_actions. batch_observe takes in batch_actions, and returns batch_observations.

- NB: users should check for empty batch items at the end of an epoch (a table with only {'episode_done': True}) inside of batch_act. Discussed with @alexholdenmiller that this is preferable in some cases to truncating observations prior to batch_act.

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -460,8 +460,8 @@ class BatchWorld(World):
         batch_observations = []
         for i, w in enumerate(self.worlds):
             agents = w.get_agents()
-            agents[index].observe(validate(batch_actions[i]))
-            batch_observations.append(agents[index].observation)
+            observation = agents[index].observe(validate(batch_actions[i]))
+            batch_observations.append(observation)
         return batch_observations
 
     def batch_act(self, index, batch_observation):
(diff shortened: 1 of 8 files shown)


ajfisch:
Confirmed that evaluations are the same:

Batchsize 1 (squad):
05/09/2017 02:08:05 PM: EM = 0.59 | F1 = 0.70 | exs = 10570
05/09/2017 02:08:05 PM: [ Done. Time = 379.26 (s) ]

Batchsize 32 (squad):
05/09/2017 02:14:15 PM: EM = 0.70 | F1 = 0.71 | exs = 1000  
05/09/2017 02:14:15 PM: [ Done. Time = 2.22 (s) ]

parlai/core/worlds.py
```diff
@@ -456,32 +456,34 @@ def __next__(self):
...
-    def batch_observe(self, index, batch):
-        for w in self.worlds:
+    def batch_observe(self, index, batch_actions):
+        batch_observations = []
+        for i, w in enumerate(self.worlds):
             agents = w.get_agents()
-            agents[index].observe(validate(batch[index]))
-        return batch
+            observation = agents[index].observe(validate(batch_actions[i]))
+            batch_observations.append(observation)
```
jaseweston: raise an error if none?

ajfisch: Yes. Should we do that in this change, or when changing all the agents' returns?

PR #63, opened 2017-05-10
Images

alexholdenmiller:
Tentative PR for adding image paths to DialogData. Should it be a different class? Maybe not, this isn't that much more complicated.

### DIFF
--- a/parlai/core/build_data.py
+++ b/parlai/core/build_data.py
@@ -23,7 +23,6 @@ def download(path, url, redownload=True):
     download tar file again if it is present (default true).
     """
     if redownload or not os.path.isfile(path):
-        # only download if file hasn't been downloaded already
         filename = wget.download(url, out=path)
         print() # wget prints download status, without newline
 
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -289,7 +289,7 @@ class DialogData(object):
         return table, end_of_data
 
 def load_image(opt, path):
-    if opt.get('no_images', False):
+    if opt.get('no_images', False) or not path:
         return None
     mode = opt.get('image_preprocessor', 'raw')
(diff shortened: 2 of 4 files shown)


parlai/core/build_data.py
```diff
@@ -23,7 +23,6 @@ def download(path, url, redownload=True):
     download tar file again if it is present (default true).
     """
     if redownload or not os.path.isfile(path):
-        # only download if file hasn't been downloaded already
         filename = wget.download(url, out=path)
```
jaseweston: why is the comment gone?

parlai/core/dialog_teacher.py
```diff
@@ -153,12 +153,14 @@ class DialogData(object):
...
+    - x is a query and possibly context
     ... can contain additional fields, specifically
-        y is an iterable of label(s) for that query
-        r is the str reward for getting that query correct
-        c is an iterable of label candidates that the student can choose from
-    new_episode? is a boolean value specifying whether that example is the start
+      - y is an iterable of label(s) for that query
+      - r is the str reward for getting that query correct
+      - c is an iterable of label candidates that the student can choose from
+      - i is a str path to an image on disk, which will be loaded by the data
```
jaseweston: what about processed images?

parlai/core/build_data.py
alexholdenmiller: I figured the function docstring explained it better

parlai/core/dialog_teacher.py
alexholdenmiller: yeah need to confirm exactly how we want to do this, but the basic solution I set up in load_data was that a folder should exist in the same directory as the images which contains the name of the preprocessor ('vgg') which contains a preprocessed version of the images

see load_data below for that

PR #67, opened 2017-05-13
task.shutdown() added to every examples

chetkhatri:
* At examples, task were created but not shutdown called except 1 example, that has been added as a standard practice.

### DIFF
--- a/examples/display_data.py
+++ b/examples/display_data.py
@@ -37,6 +37,6 @@ def main():
             if world.epoch_done():
                 print("EPOCH DONE")
                 break
-
+    world.shutdown()
 if __name__ == '__main__':
     main()
--- a/examples/display_model.py
+++ b/examples/display_model.py
@@ -37,6 +37,7 @@ def main():
             if world.epoch_done():
                 print("EPOCH DONE")
                 break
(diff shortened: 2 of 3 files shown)


chetkhatri:
@alexholdenmiller Can you please review it and merge

examples/display_model.py
```diff
@@ -37,6 +37,7 @@ def main():
             if world.epoch_done():
                 print("EPOCH DONE")
                 break
+    world.shutdown()
```
alexholdenmiller: same with this one

examples/display_data.py
```diff
@@ -37,6 +37,6 @@ def main():
             if world.epoch_done():
                 print("EPOCH DONE")
                 break
-
+    world.shutdown()
```
alexholdenmiller: the `with` statement calls this automatically, so we can remove this one here

chetkhatri: @alexholdenmiller  Okey.

examples/display_model.py
chetkhatri: @alexholdenmiller closing this

PR #77, opened 2017-05-16
Fixed: Documentation page updated as per example guideline 

chetkhatri:
* Documentation page has example code at http://ec2-35-162-199-80.us-west-2.compute.amazonaws.com/static/docs/basic_tutorial.html which had disconnect with given example https://github.com/facebookresearch/ParlAI/blob/master/examples/eval_model.py.
* task created with agent and parser were not calling shutdown and it is without with so no by default shutdown call.

### DIFF
--- a/docs/source/basic_tutorial.rst
+++ b/docs/source/basic_tutorial.rst
@@ -195,6 +195,7 @@ Now that we have our our agent, we'll set up the display loop.
         if world.epoch_done():
             print('EPOCH DONE')
             break
+    world.shutdown()
 
 And that's it! The world.display() automatically cycles through each of the
 world's agents and displays their last action. If you run this on the command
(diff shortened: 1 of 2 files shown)


docs/source/basic_tutorial.rst
```diff
@@ -195,6 +195,7 @@ Now that we have our our agent, we'll set up the display loop.
         if world.epoch_done():
             print('EPOCH DONE')
             break
+    world.shutdown()
```
alexholdenmiller: I'm intentionally leaving this out of the tutorial, as it's only needed in very specific cases, not in the general case (e.g. with a ZMQ connection that should send a shutdown signal for a remote agent).

In a similar vein, this example also doesn't explain more advanced features like the 'share' function which enables batching or hogwild training.

alexholdenmiller:
thanks for the suggestion but I'm going to keep the tutorial a little bit simpler for now!

chetkhatri:
@alexholdenmiller Sure. sounds good.

PR #83, opened 2017-05-16
README.md Order updated Installation->Examples

chetkhatri:
* Order Sequence updated at README.md, Ideally Installation section comes first then Examples comes under the project. It has been updated.

CC/ @alexholdenmiller @jgehring


chetkhatri:
As our documentation and README.md contains Installation and setup steps, I don't think adding setup.sh 
is a best practice CC/ @alexholdenmiller @jgehring

Issue #59, opened 2017-05-09
TF-IDF IR Baseline model performance on bAbI_dialog


chaitjo:
Hey! I'm trying to use the TF_IDF model provided by the framework to get comparable results to the bAbI Dialog Tasks paper. I had some strange results that I didn't understand.

For the eval_model script-
```
$ python eval_model.py -m ir_baseline -t dialog_babi:Task:1 -dt valid

[download_path:/home/chait/ParlAI/downloads]
[parlai_home:/home/chait/ParlAI]
[datatype:valid]
[task:dialog_babi:Task:1]
[model:ir_baseline]
[model_file:]
[datapath:/home/chait/ParlAI/data]
[batchsize:1]
[display_examples:False]
[model_params:]
[numthreads:1]
[num_examples:1000]
IrBaselineAgent
[Agent initializing.]
[length_penalty:0.5]
[parlai_home:/home/chait/ParlAI]
[creating task(s): dialog_babi:Task:1]
[DialogTeacher initializing.]
[loading fbdialog data:/home/chait/ParlAI/data/dialog-bAbI/dialog-bAbI-tasks/dialog-babi-task1-API-calls-dev.txt]
---
{'hits@k': {1: 0.0, 10: 0.0, 50: 0.0, 100: 0.0, 5: 0.0}, 'f1': 0.18181818181818182, 'total': 1, 'accuracy': 0.0}
---
{'hits@k': {1: 0.0, 10: 0.0, 50: 0.0, 100: 0.0, 5: 0.0}, 'f1': 0.09090909090909091, 'total': 2, 'accuracy': 0.0}
---
{'hits@k': {1: 0.0, 10: 0.0, 50: 0.0, 100: 0.0, 5: 0.0}, 'f1': 0.06060606060606061, 'total': 3, 'accuracy': 0.0}
---
{'hits@k': {1: 0.0, 10: 0.0, 50: 0.0, 100: 0.0, 5: 0.0}, 'f1': 0.045454545454545456, 'total': 4, 'accuracy': 0.0}
---
{'hits@k': {1: 0.0, 10: 0.0, 50: 0.0, 100: 0.0, 5: 0.0}, 'f1': 0.03636363636363636, 'total': 5, 'accuracy': 0.0}
---
{'hits@k': {1: 0.0, 10: 0.0, 50: 0.0, 100: 0.0, 5: 0.0}, 'f1': 0.030303030303030304, 'total': 6, 'accuracy': 0.0}
---
[...]

PR #94, opened 2017-05-21
Add TFIDF to IR baseline,  Fix bAbI Dialog single tasks (candidates were missing)

jaseweston:
#59


parlai/agents/ir_baseline/agents.py
```diff
@@ -5,13 +5,15 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 #
 # Simple IR baselines.
-# Implements the following variants:
+# We Plan to implement the following variants:
```
alexholdenmiller: Plan

PR #96, opened 2017-05-22
Seq2Seq model

alexholdenmiller:
First draft of simple seq2seq model


examples/rnn_baselines/train.py
```diff
@@ -0,0 +1,97 @@
+# Copyright (c) 2017-present, Facebook, Inc.
```
jaseweston: i'm confused it is called rnn_baselines  (plural) but then you just have train.py and agents.py, but i guess they could be moved if we add more rnn baselines?  or could just give them more specific names now

examples/rnn_baselines/train.py
```diff
@@ -0,0 +1,97 @@
...
+# All rights reserved.
```
jaseweston: we need to start making a general train program *now*. lets do it after you get this PR in

parlai/agents/rnn_baselines/agents.py
```diff
@@ -0,0 +1,275 @@
...
+import torch
+
+import copy
+import os
+import random
+import time
+
+
+class Seq2SeqAgent(Agent):
+    """Simple agent which uses an LSTM to process incoming text observations."""
```
jaseweston: explain a bit more the architecture please (GRUs, layers etc.)

parlai/core/dict.py
```diff
@@ -315,13 +329,13 @@ def act(self):
...
                         self.add_to_dict(self.tokenize(text))
-        return {}
+        return {'id': 'Dictionary'}
 
     def share(self):
         """Creates shared-memory versions of the internal maps."""
-        self.freq = SharedTable(self.freq)
-        self.tok2ind = SharedTable(self.tok2ind)
-        self.ind2tok = SharedTable(self.ind2tok)
+        # self.freq = SharedTable(self.freq)
```
jaseweston: what happened here?

parlai/core/dict.py
```diff
@@ -294,8 +306,10 @@ def txt2vec(self, text, vec_type=np.ndarray):
                 (self[token] for token in self.tokenize(str(text))),
                 np.int
             )
-        else:
+        elif vec_type == list:
```
jaseweston: need to make sure you are not breaking anything else that uses the dict, e.g. IR baseline etc.

parlai/agents/rnn_baselines/agents.py
```diff
@@ -0,0 +1,275 @@
...
+        argparser.add_arg('-nl', '--s2s-numlayers', type=int, default=2,
+            help='number of hidden layers')
+        argparser.add_arg('-lr', '--s2s-learningrate', type=float, default=0.5,
+            help='learning rate')
+        argparser.add_arg('--cuda', action='store_true', default=False,
+            help='enable GPUs if available')
+        argparser.add_arg('--gpu', type=int, default=-1,
+            help='which GPU device to use')
+
+    def __init__(self, opt, shared=None):
```
jaseweston: be cool if this model also ranked candidates, which we need for many of the parlAI tasks,  i guess it doesnt yet?

parlai/core/dict.py
alexholdenmiller: actually nothing was using this function except remote agent, looks like

parlai/agents/rnn_baselines/agents.py
alexholdenmiller: yeah this doesn't do that yet! I'd rather check it in before adding that functionality

examples/rnn_baselines/train.py
alexholdenmiller: I'll update these! haven't yet

parlai/core/params.py
```diff
@@ -73,11 +73,16 @@ def add_mturk_args(self):
         self.parser.set_defaults(verbose=False)
 
     def add_parlai_args(self):
+        default_log_path = os.path.join(self.parlai_home, 'logs')
         default_downloads_path = os.path.join(self.parlai_home, 'downloads')
         self.parser.add_argument(
             '-t', '--task',
             help='ParlAI task(s), e.g. "babi:Task1" or "babi,cbt"')
         self.parser.add_argument(
+            '--logpath', default=default_log_path,
```
jaseweston: still here?

PR #103, opened 2017-05-25
Docs format fix

yf225:
Docs demo: http://parl.ai/static/docs/agents.html


parlai/core/agents.py
```diff
@@ -333,3 +233,109 @@ def report(self):
...
+    try:
+        # Tries to call the create_agent function in agents.py
+        create_agent = getattr(my_module, 'create_agents')
+        task_agents = create_agent(opt)
+    except AttributeError:
+        # Create_agent not found, so try to create the teacher directly.
+        return create_task_agent_from_taskname(opt)
+    if type(task_agents) != list:
+        task_agents = [task_agents]
+    return task_agents
```
alexholdenmiller: this code is all the same, right?

alexholdenmiller:
awesome!! thank you!

alexholdenmiller:
merging so I can do work on the task_list file

parlai/core/agents.py
yf225: Yes it is - moving them around a little bit so that in the doc all of the classes will show first and then the methods :)

PR #107, opened 2017-05-27
Add Dockerfile. Fix #75

### LINKED ISSUE #75: Add Dockerfile
Add Dockerfile and a docker image which might make trying out ParlAI easier.

jgmize:
This is a basic Dockerfile to get started. The short tests run by the default CMD all pass, but the `tests/check_examples.sh` script fails due to a missing dependency, pytorch. Since pytorch isn't a requirement for the base framework I'd recommend a separate requirements.txt and Dockerfile for the pytorch example and others where appropriate. I'm currently running the `tests/run_tests_long.sh` script in a container built with this Dockerfile; will comment later with results once that is done.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM python:3
+WORKDIR /ParlAI
+ADD . /ParlAI/
+RUN pip install -r requirements.txt
+RUN python setup.py develop
+CMD cd tests && ./run_tests_short.sh


jgmize:
I saw some test failures in the container running `run_tests_long.sh` in the `tests` directory due to running out of memory on the old laptop I was running these on with 8GB of memory. I can try again later on a different machine with more memory. For future reference, here is the relevant output:

```
unpacking wikiqa.tar.gz                                                                                                                                                [134/1803]
building fbformat:train
building fbformat:valid
building fbformat:test
building fbformat:train-filtered
building fbformat:valid-filtered
building fbformat:test-filtered
[DialogTeacher initializing.]
[loading fbdialog data:/tmp/parlai_test_data/WikiQA/train-filtered.txt]
[DialogTeacher initializing.]
[loading fbdialog data:/tmp/parlai_test_data/WikiQA/valid-filtered.txt]
[DialogTeacher initializing.]
[loading fbdialog data:/tmp/parlai_test_data/WikiQA/test-filtered.txt]
E
ERROR: test_qacnn (__main__.TestData)
  File "test_data.py", line 234, in test_qacnn
    teacher = DefaultTeacher(opt)
  File "/ParlAI/parlai/tasks/qacnn/agents.py", line 23, in __init__
    opt['datafile'] = _path(opt)
  File "/ParlAI/parlai/tasks/qacnn/agents.py", line 15, in _path
    build(opt)
  File "/ParlAI/parlai/tasks/qacnn/build.py", line 52, in build
    os.path.join(dpath, 'cnn', 'questions', 'training'))
  File "/ParlAI/parlai/tasks/qacnn/build.py", line 31, in create_fb_format
    for f in os.listdir(inpath):
OSError: [Errno 12] Cannot allocate memory: '/tmp/parlai_test_data/QACNN/cnn/questions/training'
```
[8 other similar tracebacks]
```
Ran 20 tests in 8577.587s

FAILED (errors=9)
At least one test failed. Leaving directory /tmp/parlai_test_data/ with temporary files in place for inspection (only failed tasks remain).
```

ajfisch:
(NB: not very familiar with Dockerfile best practice, so please correct).
This looks fine. But it assumes that ParlAI has already been downloaded from git locally (the COPY)? I think it might be better if the Dockerfile included things like a basic ubuntu install (with git) and downloaded ParlAI as part of the setup -- see for example [TensorFlow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docker/Dockerfile) and [PyTorch](https://github.com/pytorch/pytorch/blob/master/Dockerfile)  Personal preference for Anaconda as the python distribution, although not important.

trishalaneeraj:
Instead of creating a Dockerfile with all of ParlAI requirements, is it not better to inherit from a docker image which already has basic pytorch, anaconda etc. and then, a really small Dockerfile to clone and install ParlAI and the requirements.txt file from GitHub? That way, all that we have to ensure is that the base image which ParlAI is inheriting from is up to date and there will be minimal need to update the entire Dockerfile. Let me know what you think about it?

Ideally we could just inherit from the [pytorch Dockerfile](https://github.com/pytorch/pytorch/blob/master/Dockerfile) but currently, I don't think there is a public image available for this. However, I built and pushed it to my own [docker hub](https://hub.docker.com/r/trishalaneeraj/pytorch/) account.

alexholdenmiller:
closing for now without dockerfile support

PR #111, opened 2017-05-31
train_model example 

jaseweston:
The aim is to get as many models as possible working with this (as well as eval_model).

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -1,0 +1,77 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+"""Train a model.
+   After training, computes validation and test error.
+   Run with, e.g.:
+   python examples/train_model.py -m ir_baseline -t dialog_babi:Task:1 -mf "/tmp/model"
+
+   TODO List:
+   - Add model specific params
+   - Validate & Log while training
+   - Keep best model from validation error, if desired
+"""
+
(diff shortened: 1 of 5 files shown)


examples/train_model.py
```diff
@@ -0,0 +1,77 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+"""Train a model.
+   After training, computes validation and test error.
```
alexholdenmiller: these can be left-justified (no tab)

examples/train_model.py
```diff
@@ -0,0 +1,77 @@
...
+    parser.add_argument('-mtt', '--max-train-time',
+                        type=float, default=float('inf'))
+    opt = parser.parse_args()
+    # Create model and assign it to the specified task
+    agent = create_agent(opt)
+    world = create_task(opt, agent)
+
+    train_time = Timer()
+    print("[training...]")
+    for _ in range(len(world) * opt['num_epochs']):
```
alexholdenmiller: I like it better as...
```
for _ in range(opt['num_epochs']:
    for _ in world:
        world.parley()
```
but fine if you prefer this way for simplicity

jaseweston: added a TODO because it should deal with worlds with no length anyway, so will fix later..

examples/train_model.py
jaseweston: done

examples/train_model.py
jaseweston: ok, changed to yours because maybe it does deal with that case after all..

Issue #101, opened 2017-05-25
Error running DrQA


bittlingmayer:
I ran: `python examples/drqa/train.py -t "#opensubtitles"`

After 10 or 20 minutes:

```
05/25/2017 02:34:31 PM: [ Running validation... ]
Traceback (most recent call last):
  File "examples/drqa/train.py", line 180, in <module>
    main(opt)
  File "examples/drqa/train.py", line 118, in main
    valid_metric = validate(opt, doc_reader, iteration)
  File "examples/drqa/train.py", line 62, in validate
    valid_world.parley()
  File "/Users/bittlingmayer/Desktop/aca/ParlAI/parlai/core/worlds.py", line 221, in parley
    acts[1] = agents[1].act()
  File "/Users/bittlingmayer/Desktop/aca/ParlAI/parlai/agents/drqa/agents.py", line 157, in act
    [ex], null=self.word_dict['<NULL>'], cuda=self.opt['cuda']
  File "/Users/bittlingmayer/Desktop/aca/ParlAI/parlai/agents/drqa/utils.py", line 126, in batchify
    max_length = max([d.size(0) for d in docs])
  File "/Users/bittlingmayer/Desktop/aca/ParlAI/parlai/agents/drqa/utils.py", line 126, in <listcomp>
    max_length = max([d.size(0) for d in docs])
RuntimeError: dimension 0 out of range of 0D tensor at /Users/soumith/code/builder/wheel/pytorch-src/torch/lib/TH/generic/THTensor.c:24
```

bittlingmayer:
Happened with ` --embedding_dim 100 --train_interval 100`, but in a few minutes, so I assume it happens at or near the end.

ajfisch:
Hm. I haven't tried the drqa model on this dataset. From the error it looks like there is an empty example.

The fix would be to skip examples where either the question or passage has 0 tokens:
https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/drqa/agents.py#L226-L227

You can try putting a
```python
if len(inputs['document']) == 0:
    import pdb
    pdb.set_trace()
```
there to investigate.

I can add some more checks to be more robust.

bittlingmayer:
Which datasets should it work with?

I added:

```
        if len(inputs['document']) == 0:
            print("empty document")
            inputs['document'] = self.word_dict.tokenize("<EMPTY>")
        if len(inputs['question']) == 0:
            print("empty question")
[...]

PR #113, opened 2017-05-31
improved error message for drqa

ajfisch:
- Raise runtime error for invalid inputs

- Also added in a fix to always load models on the cpu first (and then transfer if using cuda)


ajfisch:
addresses #101

PR #115, opened 2017-06-01
Add visual datasets to task list in docs

yf225:


docs/source/task_list.inc
```diff
@@ -295,7 +295,46 @@ WikiQA
 
 **Group Tags**: ``#all``, ``#QA``
 
-**Description**: Open domain QA from Wikipedia dataset from Yang, et al. '15. Link: https://www.microsoft.com/en-us/research/publication/wikiqa-a-challenge-dataset-for-open-domain-question-answering/
+**Description**: Open domain QA from Wikipedia dataset from Yang et al. '15. Link: https://www.microsoft.com/en-us/research/publication/wikiqa-a-challenge-dataset-for-open-domain-question-answering/
+
+
+
+VQAv1
```
jaseweston: cool! it would be nice if the docs task list has tabs for qa,chitchat etc. is that easy to do?

yf225: Definitely - I just added the tabs into task list :)

PR #116, opened 2017-06-01
Pull in args from specific model

jaseweston:


parlai/core/params.py
```diff
@@ -106,12 +109,17 @@ def add_model_args(self):
...
             help='model file name for loading and saving models')
+        # Find which model specified, and add its specific arguments.
+        model = None
+        for index, item in enumerate(sys.argv):
+            if item == '-m' or item == '--model':
+                model = sys.argv[index + 1]
+        if model:
+            agent = get_agent_module(model)
+            if hasattr(agent, 'add_cmdline_args'):
+                agent.add_cmdline_args(self)
```
alexholdenmiller: could add `add_cmdline_args` to core/agents.py:Agent so that there's an empty definition instead of checking if the definition exists, but this is fine too

jaseweston: ok, will leave as is, but agnostic

PR #122, opened 2017-06-05
removed wget and asciimatics dependencies

alexholdenmiller:
removed dependencies on `wget` (just using `requests` now) and `asciimatics` (just using `Pillow` now)

### DIFF
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -5,10 +5,7 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 
 from .agents import Teacher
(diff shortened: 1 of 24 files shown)


parlai/core/dialog_teacher.py
```diff
@@ -302,6 +299,22 @@ def get(self, episode_idx, entry_idx=0):
         return table, end_of_data
 
 
+_greyscale = '  .,:;crsA23hHG#98&@'
+
+
+def img_to_ascii(path):
```
jaseweston: can you dump this in utils?

PR #124, opened 2017-06-06
Move seq2seq training into standard train_model.py script

jaseweston:
The plan is to try to move all model training and eval into standard scripts as much as possible.


examples/train_model.py
```diff
@@ -33,36 +36,109 @@ def run_eval(agent, opt, datatype):
...
     parser = ParlaiParser(True, True)
     parser.add_argument('-d', '--display-examples',
                         type='bool', default=False)
-    parser.add_argument('-e', '--num-epochs', default=1)
+    parser.add_argument('-e', '--num-epochs', type=int, default=1)
     parser.add_argument('-mtt', '--max-train-time',
                         type=float, default=float('inf'))
+    parser.add_argument('-lt', '--log-every-n-secs',
+                        type=float, default=1)
+    parser.add_argument('-vt', '--validate-every-n-secs',
```
alexholdenmiller: -vt would be a great param for "validation task" in case you want to validate on something else :)

alexholdenmiller: looks good overall!

PR #126, opened 2017-06-07
Send episode_done=True from local human agent

yf225:


parlai/agents/local_human/local_human.py
```diff
@@ -24,4 +25,11 @@ def act(self):
         reply = {}
         reply['id'] = self.getID()
         reply['text'] = input("Enter Your Reply: ")
+        if reply_text == '[DONE]':
```
jaseweston: looks like something wrong here?

also if you use [DONE] to denote the episode is done,  i guess the text is empty?
we could also allow some text and a [DONE] at the end so we could send a final answer and "leave"

yf225: yes sorry forgot to change it during copy-paste..

yf225:
@jaseweston it seems that ``MultiAgentDialogWorld``'s episode_done() will call each agent's episode_done() to see if an episode has ended (sorry I meant to override episode_done() not epoch_done()). Or are there other ways to check if an episode has ended?

PR #127, opened 2017-06-07
Allow DrQA to be used from the standard  examples/train_model.py  and examples/eval_model.py scripts

jaseweston:


ajfisch:
LGTM

examples/train_model.py
```diff
@@ -101,6 +116,10 @@ def main():
                         type=float, default=1)
     parser.add_argument('-vtim', '--validate-every-n-secs',
                         type=float, default=False)
+    parser.add_argument('-vimp', '--validation-impatience',
```
alexholdenmiller: I've seen other people call this `patience`, could we call it that? this is just a long param name :)

parlai/agents/drqa/drqa.py
```diff
@@ -83,13 +99,21 @@ def add_to_dict(self, tokens):
...
+class DrqaAgent(Agent):
 
     @staticmethod
     def add_cmdline_args(argparser):
         config.add_cmdline_args(argparser)
+        SimpleDictionaryAgent.add_cmdline_args(argparser)
 
-    def __init__(self, opt, shared=None, word_dict=None):
+    @staticmethod
+    def dictionary_class():
```
alexholdenmiller: oh, I was thinking you could just return SimpleDictionaryAgent directly. you like this better?

alexholdenmiller: I guess it does make it so that you can override this param directly from the command line instead of changing this function, which maybe some people would want to do

parlai/agents/drqa/drqa.py
```diff
@@ -83,13 +99,21 @@ def add_to_dict(self, tokens):
...
+        return "parlai.agents.drqa.drqa:SimpleDictionaryAgent"
+
+    def __init__(self, opt, shared=None):
+        # Load dict.
+        if not shared:
+            word_dict = SimpleDictionaryAgent(opt)
```
alexholdenmiller: should this not use the function above to choose which dict class to use? if the other function returned a class then it would be easy, but the string does make it harder...

parlai/core/params.py
```diff
@@ -118,6 +118,10 @@ def add_model_args(self):
             agent = get_agent_module(model)
             if hasattr(agent, 'add_cmdline_args'):
                 agent.add_cmdline_args(self)
+            if hasattr(agent, 'dictionary_class'):
+                self.parser.add_argument(
+                    '-dict_class', default=agent.dictionary_class(), type=str,
```
alexholdenmiller: two hyphens at start, hyphen instead of underscore like we talked about before: `--dict-class`

alexholdenmiller: actually on that note: model_file should be `model-file` or `modelfile` to match all the other args

parlai/agents/drqa/drqa.py
jaseweston: lets fix this in another diff so we can get this one through

PR #135, opened 2017-06-12
Added mnist task

belbs:


jaseweston:
To add a task like this to ParlAI it would need a description somewhere.
Please also call this something like mnist_qa or well something other than mnist so people do not confuse it with the original dataset/task.

parlai/tasks/mnist/agents.py
```diff
@@ -0,0 +1,67 @@
...
+
+import json
+import os
+
+
+def _path(opt):
+    build(opt)
+    dt = opt['datatype'].split(':')[0]
+
+    if dt == 'train':
```
alexholdenmiller: could just use dt here => this check was from another task that was remapping 'dev' => 'valid' I think

parlai/tasks/mnist/agents.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
```
alexholdenmiller: add a description of this task after the license info as a docstring, e.g. 
```
"""Some helpful info
about the task.
"""
```

This will get imported into the documentation automatically.

alexholdenmiller:
But otherwise looks good!

alexholdenmiller:
agree with Jason that it would be good to change the task name from mnist to mnist_qa

belbs:
Added the suggested modifications.

parlai/tasks/mnist_qa/agents.py
```diff
@@ -0,0 +1,66 @@
...
+
+class MnistQATeacher(DialogTeacher):
+    """
+    This version of MNIST inherits from the core Dialog Teacher, which just
+    requires it to define an iterator over its data `setup_data` in order to
+    inherit basic metrics, a `act` function, and enables
+    Hogwild training with shared memory with no extra work.
+    """
+    def __init__(self, opt, shared=None):
+        self.datatype = opt['datatype']
```
alexholdenmiller: actually the split function from before is still needed: it converts e.g. train:ordered to train

parlai/tasks/task_list.py
```diff
@@ -171,25 +171,32 @@
...
+            "display_name": "VisDial",
+            "task": "visdial",
+            "tags": [ "all", "Visual"],
+            "description": "Task which requires agents to hold a meaningful dialog about visual content. From Das et al. '16. Link: https://arxiv.org/abs/1611.08669"
+    },
+    {
+            "id": "MNIST_QA",
+            "display_name": "MNIST_QA",
+            "task": "mnist_qa",
+            "tags": [ "all", "QA", "Visual"],
```
alexholdenmiller: one last thing: since this **requires** the image to answer the question, let's remove the QA tag here

parlai/tasks/mnist_qa/agents.py
belbs: visdial agents has the same line, should I fix it there as well?

belbs:
Updated from comments

parlai/tasks/mnist_qa/build.py
```diff
@@ -0,0 +1,29 @@
...
+    print(dpath)
+
+    if not build_data.built(dpath):
+        print('[building data: ' + dpath + ']')
+        build_data.remove_dir(dpath)
+        build_data.make_dir(dpath)
+
+        # Download the data.
+        fname = 'mnist.tar.gz'
+        url = 'https://s3.amazonaws.com/fair-data/parlai/mnist/' + fname
```
jaseweston: mnist_qa please!!

jaseweston:
just this one i think?
(visdial isn't anything to do with mnist?)

belbs:
I thought it would make sense to leave the dataset itself named simply "mnist", since it is indeed just the standard mnist dataset (just like the images for vqa etc are in a directory named after the dataset COCO-IMG). But if you think name the dataset mnist_qa is better ill change it as well.

parlai/tasks/mnist_qa/agents.py
alexholdenmiller: yes, thanks!

alexholdenmiller:
@jaseweston yeah I think it should be just mnist, since there is nothing in that tar file that is related to QA--we are just rehosting the mnist data broken into separate image files from a build of the mnist data where they were all in one file

jaseweston:
ok

PR #140, opened 2017-06-14
Build versioning -- so we can rebuild if data version is updated

jaseweston:


parlai/core/build_data.py
```diff
@@ -14,10 +14,30 @@
...
+    If a version_string is provided, this has to match, or the version
+    is regarded as not built.
+    """
+    if version_string:
+        fname = os.path.join(path, '.built')
+        if not os.path.isfile(fname):
+            return False
+        else:
+            file = open(fname, 'r') 
+            text = file.read().split('\n')
```
alexholdenmiller: I think this is better as:
```
with open(fname, 'r') as read:
    text = read.read().split('\n')
```

alexholdenmiller: it just makes sure the file is closed properly

parlai/core/build_data.py
```diff
@@ -65,14 +85,6 @@ def make_dir(path):
...
 
 
-def mark_done(path):
-    """Marks the path as done by adding a '.built' file with the current
-    timestamp.
-    """
-    with open(os.path.join(path, '.built'), 'w') as write:
-        write.write(str(datetime.datetime.today()))
-
-
```
alexholdenmiller: killing my alphabetical order! fine...

PR #141, opened 2017-06-14
[build_dict] use train set only

ajfisch:
In case you do some special things in training based on frequency counts or unknown words

### DIFF
--- a/examples/build_dict.py
+++ b/examples/build_dict.py
@@ -18,8 +18,8 @@ def build_dict(opt):
         return
     print('[ setting up dictionary. ]')
     if os.path.isfile(opt['dict_file']):
-        # Dict already built
-        print("[ dict already built .]")
+        # Dictionary already built
+        print("[ dictionary already built .]")
         return
     if 'dict_class' in opt:
         # Custom dictionary class
@@ -47,6 +47,7 @@ def build_dict(opt):
         world_dict.parley()
     print('[ dictionary built. ]')
     dictionary.save(opt['dict_file'], sort=True)
+    # print('[ num words =  %d ]' % len(dictionary))
 
 def main():
     # Get command line arguments
--- a/examples/build_dict.py
+++ b/examples/build_dict.py
@@ -18,7 +18,7 @@ def build_dict(opt):
         return
     print('[ setting up dictionary. ]')
     if os.path.isfile(opt['dict_file']):
-        # dict already built
+        # Dict already built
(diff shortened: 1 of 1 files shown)


ajfisch:
test plan: 
python examples/train_model.py -dbf t -t squad -m drqa

examples/build_dict.py
```diff
@@ -32,24 +32,21 @@ def build_dict(opt):
...
+    for _ in world_dict:
+        cnt += 1
+        if cnt > opt['dict_maxexs'] and opt['dict_maxexs'] > 0:
+            print('Processed {} exs, moving on.'.format(opt['dict_maxexs']))
+            # don't wait too long...
+            break
+        world_dict.parley()
     print('[ dictionary built. ]')
     dictionary.save(opt['dict_file'], sort=True)
-    # print('[ num words =  %d ]' % len(dictionary))
```
jaseweston: wanted to keep this as I'm putting it back as soon as Alex fixes the \n bug

examples/build_dict.py
```diff
@@ -18,7 +18,7 @@ def build_dict(opt):
         return
     print('[ setting up dictionary. ]')
     if os.path.isfile(opt['dict_file']):
-        # dict already built
```
alexholdenmiller: maybe we should changing this to the full word since we are not referring to a python dict?

examples/build_dict.py
ajfisch: I actually was happy this was taking out. This means that the dict object has to define a __len__ attribute, which may not be natural. For example, I would like to use the dict as more of a "dataset observer" -- it counts the words, but also might count things like the most common words of some type (e.g. question words), the number of named entities types (if doing ner tagging), etc.

examples/build_dict.py
ajfisch: sure

examples/build_dict.py
jaseweston: yes, but it's incredibly useful debugging tool when it IS natural, so we should have something like it..  there's another instance of it during loading btw

ajfisch: Right, that's put in the Dictionary load function, not externally called though. We could do the same thing for save? Also fine with just requiring a __len__ attribute.

jaseweston:
ok -- just make it work plz! :) :)

ajfisch:
Added back in commen -- @jaseweston you can remove whenever you're ready

PR #143, opened 2017-06-15
Fixing train/val/test in visdial

belbs:
Visdial did not have a test set. Fixed so that 1000  examples are picked from what was previously training to create a validation set. What was previously validation is now test set.


jaseweston:
did you ask jiasen what the right thing to do is here?

belbs:
Yep, he's the one who said the usual split is this one (1k examples from training for validation)

jaseweston:
ok!

jaseweston:
note we have a versioning of builds now which you should use here for the update (so people who already downloaded get the update):
https://github.com/facebookresearch/ParlAI/commit/8298ce93b52e2b25f2b883332746116470333771

parlai/tasks/visdial/build.py
```diff
@@ -41,19 +43,45 @@ def build(opt):
...
         build_data.untar(dpath, fname1)
         build_data.untar(dpath, fname2)
+        
+        print('processing unpacked files')
+        # Use 1000 examples from training set as validation.
+        json1 = os.path.join(dpath, fname1.rsplit('.', 1)[0] + '.json')
+        with open(json1) as t_json:
+            train_data = json.load(t_json)
+
+        random.seed()
```
alexholdenmiller: actually can you fix this so it's the same valid set every time? e.g. `random.seed(42)`

jaseweston:
oh yes in fact maybe you should split the data first and save it  -- is it
possible even with the same random seed with some different library people
might get different sets? we want everyone to have the same set

On Thu, Jun 15, 2017 at 12:14 PM, Alexander Miller <notifications@github.com

alexholdenmiller:
ah I didn't think about platform-independence, there's a python library with guaranteed consistency but random does seem to guarantee it between python versions.

if we don't want to rehost, though, another option would be to just e.g. use the last 1000 instead of a random thousand (or a fixed sampling, like use each example where curr_idx % (total // 1000) == 0).

belbs:
yes, my original approach was just to get the last 1000. but I was unsure if there was some sort of structure to the dataset

belbs:
any consensus on which approach to take on selecting the validation examples?

jaseweston:
You could take every Nth example or something. Pretty good chance it will
be random?

PR #145, opened 2017-06-16
Predictor module

alexholdenmiller:
per @ajfisch's request, added a module for loading up a model via a library function instead of command line arguments, and an example of using it (more or less equivalent to interactive, we can get rid of one soon)

modifications to params to allow this to work as well


jaseweston:
could you add some comment to the predictor.py explaining the difference to
interactive
- it is stateless and doesnt use a world
- but it is good in the case of API calls from another system, where
stateless is typically better
we could also get rid of this example if it's confusing, but keep the thing
in core
also, the thing in core -- could it be in utils or some other dir slightly
less core?

PR #150, opened 2017-06-19
move predictor to utils, more documentation for predictor module

alexholdenmiller:


examples/predictor.py
```diff
@@ -3,19 +3,27 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree. An additional grant
 # of patent rights can be found in the PATENTS file in the same directory.
-"""Basic example which allows local human keyboard input to talk to a trained model.
-
-For example:
+"""Parallel of `interactive.py`, a basic example which allows local human
```
jaseweston: note that this (unlike interactive) only works in a stateless "world" (as it doesn't create a world)
i discussed with adam, and we also thought we could just remove this example and keep only the util, as it will be less confusing for people looking at interactive

alexholdenmiller: agree with the removal, I mostly made it for debugging to make sure everything was working

PR #153, opened 2017-06-20
train/dict updates and move seq2seq

alexholdenmiller:


parlai/core/dict.py
```diff
@@ -127,11 +127,12 @@ def __init__(self, opt, shared=None):
                 self.tok2ind[self.unk_token] = index
                 self.ind2tok[index] = self.unk_token
 
-            if 'dict_file' in opt and os.path.isfile(opt['dict_file']):
-                self.load(opt['dict_file'])
-            elif 'dict_initpath' in opt:
-                # load pre-existing dictionary
+            if opt.get('dict_initpath'):
```
jaseweston: this is different behavior to what i wanted, let us discuss

jaseweston: that is, you should only load the init if you don't have the actual real dict (same thing with models usually)

PR #157, opened 2017-06-21
updated dictionary_class(), added params argument groups, bug fixes

alexholdenmiller:
- Fixed some bugs with check parameters which used to be missing rather than none
- Updated dictionary_class to return a class, but allow users to specify a class on the command line and have the params class parse that into a specific class automatically
- Updated params to use argument groups per @ajfisch (you're beautiful), so now you get e.g.
```bash
$ py examples/train_model.py -m drqa --help
...
usage:
...
...
ParlAI parser.

optional arguments:
  -h, --help            show this help message and exit

Main ParlAI Arguments:
  -t TASK, --task TASK  ParlAI task(s), e.g. "babi:Task1" or "babi,cbt"
  --download-path DOWNLOAD_PATH
                        path for non-data dependencies to store any needed
                        files.defaults to {parlai_dir}/downloads
  -dt {train,train:ordered,valid,test}, --datatype {train,train:ordered,valid,test}
                        choose from: train, train:ordered, valid, test. by
                        default: train is random with replacement, valid is
                        ordered, test is ordered.
  -im IMAGE_MODE, --image-mode IMAGE_MODE
                        image preprocessor to use. default is "raw". set to
                        "none" to skip image loading.
  -nt NUMTHREADS, --numthreads NUMTHREADS
                        number of threads, e.g. for hogwild
  -bs BATCHSIZE, --batchsize BATCHSIZE
                        batch size for minibatch training schemes
  -dp DATAPATH, --datapath DATAPATH
                        path to datasets, defaults to {parlai_dir}/data

ParlAIModel Arguments:
  -m MODEL, --model MODEL
                        the model class name, should match
                        parlai/agents/<model>
  -mf MODEL_FILE, --model-file MODEL_FILE
                        model file name for loading and saving models
  --dict-class DICT_CLASS
                        the class of the dictionary agent used

DrQA Arguments:
  --no_cuda NO_CUDA
  --gpu GPU
...
...
```

### DIFF
--- a/parlai/agents/drqa/config.py
+++ b/parlai/agents/drqa/config.py
@@ -100,7 +100,7 @@ def set_defaults(opt):
 
     # Make sure fix_embeddings and embedding_file are consistent
     if opt['fix_embeddings']:
-        if opt.get('embedding_file') and not opt.get('pretrained_model'):
+        if not opt.get('embedding_file') and not opt.get('pretrained_model'):
             print('Setting fix_embeddings to False as embeddings are random.')
             opt['fix_embeddings'] = False
 
--- a/parlai/agents/drqa/config.py
+++ b/parlai/agents/drqa/config.py
@@ -13,71 +13,72 @@ def str2bool(v):
 
 def add_cmdline_args(parser):
     # Runtime environment
-    parser.add_argument('--no_cuda', type='bool', default=False)
-    parser.add_argument('--gpu', type=int, default=-1)
-    parser.add_argument('--random_seed', type=int, default=1013)
+    agent = parser.add_argument_group('DrQA Arguments')
+    agent.add_argument('--no_cuda', type='bool', default=False)
+    agent.add_argument('--gpu', type=int, default=-1)
+    agent.add_argument('--random_seed', type=int, default=1013)
 
     # Basics
-    parser.add_argument('--embedding_file', type=str, default=None,
+    agent.add_argument('--embedding_file', type=str, default=None,
                         help='File of space separated embeddings: w e1 ... ed')
-    parser.add_argument('--pretrained_model', type=str, default=None,
+    agent.add_argument('--pretrained_model', type=str, default=None,
                         help='Load dict/features/weights/opts from this file')
-    parser.add_argument('--log_file', type=str, default=None)
+    agent.add_argument('--log_file', type=str, default=None)
 
     # Model details
-    parser.add_argument('--fix_embeddings', type='bool', default=True)
-    parser.add_argument('--tune_partial', type=int, default=0,
+    agent.add_argument('--fix_embeddings', type='bool', default=True)
+    agent.add_argument('--tune_partial', type=int, default=0,
                         help='Train the K most frequent word embeddings')
-    parser.add_argument('--embedding_dim', type=int, default=300,
+    agent.add_argument('--embedding_dim', type=int, default=300,
                         help=('Default embedding size if '
                               'embedding_file is not given'))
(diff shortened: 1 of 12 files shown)


parlai/agents/drqa/config.py
```diff
@@ -99,10 +100,8 @@ def set_defaults(opt):
 
     # Make sure fix_embeddings and embedding_file are consistent
     if opt['fix_embeddings']:
-        if not 'embedding_file' in opt and not 'pretrained_model' in opt:
-            print(
-                'Setting fix_embeddings to False as embeddings are random.'
-                )
+        if opt.get('embedding_file') and not opt.get('pretrained_model'):
```
jaseweston: you flipped the not here on the first arg ? is that right?

parlai/core/params.py
```diff
@@ -127,22 +150,24 @@ def add_model_args(self, args=None):
             if hasattr(agent, 'add_cmdline_args'):
                 agent.add_cmdline_args(self)
             if hasattr(agent, 'dictionary_class'):
-                self.add_argument(
-                    '--dict-class', default=agent.dictionary_class(), type=str,
-                    help='the class of the dictionary agent used')
+                model_args.set_defaults(dict_class=agent.dictionary_class())
```
jaseweston: you actually put a whole class in the opt?  this seems really really weird. . can't it just be the name?

parlai/agents/drqa/config.py
alexholdenmiller: shoot, thanks

parlai/core/params.py
alexholdenmiller: yes the whole class--I thought it was better because when you are using this opt, you're likely to be initializing an instance of it, not just inspecting the name, so then you can just do that instead of also having to figure out how to import from a string when you want to use it. (see the change to examples/build_dict.py: that seems much more natural to me)

jaseweston: I don't like it because I want to keep the options simple and save them separately, not include a class in there.
(also what happens when you deep copy the options)
Can we PLEASE change it back?

alexholdenmiller: or we can talk in person tomorrow

parlai/core/params.py
```diff
@@ -46,75 +72,84 @@ def __init__(self, add_parlai_args=True, add_model_args=False, model_argv=None):
...
-        self.add_argument(
+        parlai.add_argument(
             '-bs', '--batchsize', default=1, type=int,
             help='batch size for minibatch training schemes')
-        self.add_parlai_data_path()
+        self.add_parlai_data_path(parlai)
 
     def add_model_args(self, args=None):
-        self.add_argument(
+        model_args = self.add_argument_group('ParlAIModel Arguments')
```
ajfisch: Missing a space

parlai/core/dict.py
```diff
@@ -62,37 +62,39 @@ class DictionaryAgent(Agent):
...
             '--dict-initpath',
             help='path to a saved dictionary to load tokens / counts from to ' +
                  'seed the dictionary with initial tokens and/or frequencies')
-        argparser.add_arg(
+        dictionary.add_argument(
             '--dict-language', default=DictionaryAgent.default_lang,
             help='sets language for the punkt sentence tokenizer')
-        argparser.add_arg(
+        dictionary.add_argument(
             '--dict-max-ngram-size', default=DictionaryAgent.default_maxngram,
```
belbs: the type=int bug I had mentioned to you before could be corrected here

PR #163, opened 2017-06-26
Added Personalized Dialog dataset

chaitjo:
Added ability to download and use data from the paper "Personalization in Goal-Oriented Dialog" by Joshi et al. '17 (https://arxiv.org/abs/1706.07503).


parlai/tasks/task_list.py
```diff
@@ -116,6 +116,20 @@
         "description": "Dataset of dialogs from movie scripts: http://opus.lingfil.uu.se/OpenSubtitles.php. A variant of the dataset used in Vinyals & Le '15, https://arxiv.org/abs/1506.05869."
     },
     {
+        "id": "personalized-dialog-full",
+        "display_name": "Personalized Dialog Full Set",
+        "task": "personalized-dialog:full",
+        "tags": [ "all",  "Goal", "Personalization" ],
```
alexholdenmiller: @jaseweston cool with these tags?

jaseweston: looks ok to me

PR #165, opened 2017-06-26
Change MTurk local db to be in-memory

yf225:
Local db file creation seems to have issues on the FAIR cluster as discovered by Harm. Moving it to be in-memory to avoid such issue.


alexholdenmiller:
sneaky

PR #169, opened 2017-06-26
Add image feature extraction modules and fix minor bugs. 

jiasenlu:
Add `image_featurizers.py.py` in `parlay/core/` to handle image loading and feature extraction. 
Add `extract_image_feature.py` in `parlay/examples` to show how to extract the image feature using the new module. 
Change `dialog_teacher.py, vqa_v1.py and vqa_v2.py` to adapt the new module. 
Minor bug fixed (VQA test image loading).


examples/extract_image_feature.py
```diff
@@ -0,0 +1,50 @@
...
+
+import random
+
+def main():
+    random.seed(42)
+
+    # Get command line arguments
+    parser = ParlaiParser()
+    parser.add_argument('-n', '--num-examples', default=10)
+
```
alexholdenmiller: (should help with hitting each image only once)
`parser.set_defaults(datatype='train:ordered')`

parlai/core/dialog_teacher.py
```diff
@@ -240,6 +242,7 @@ def _load(self, data_loader):
                                 new_entry.append(None)
                             if len(entry) > 4 and entry[4] is not None:
                                 new_entry.append(sys.intern(entry[4]))
+                                self.image_loader = ImageLoader(opt)
```
alexholdenmiller: this will create a new image loader on every example with an image--not good. since the image initializer is lazy anyways, let's move this to `__init__`

parlai/core/image_featurizers.py
```diff
@@ -0,0 +1,145 @@
...
+from PIL import Image
+
+_greyscale = '  .,:;crsA23hHG#98&@'
+
+class ImageLoader():
+	"""Extract image feature using pretrained CNN network.
+	"""	
+	@staticmethod
+	def add_cmdline_args(argparser):
+		argparser.add_arg('--image-cnntype', type=str, default='resnet152',
```
alexholdenmiller: actually: `--image-mode` can cover this, unless you can think of a case where image-mode might be set to something that would clash? otherwise let's delete this param

parlai/core/image_featurizers.py
```diff
@@ -0,0 +1,145 @@
...
+			help='')
+		argparser.add_arg('--image-cropsize', type=int, default=224,
+			help='')
+
+	def __init__(self, opt):
+		
+		self.opt = copy.deepcopy(opt)
+		self.netCNN = None
+		self.transform = None
+		self.xs = None
```
alexholdenmiller: should we move these three into `init_cnn`?

parlai/core/dialog_teacher.py
```diff
@@ -297,42 +300,3 @@ def get(self, episode_idx, entry_idx=0):
         table['episode_done'] = episode_done
         return table, end_of_data
 
-
-_greyscale = '  .,:;crsA23hHG#98&@'
-
-
-def img_to_ascii(path):
```
jaseweston: move this stuff to utils or something?

parlai/core/dialog_teacher.py
jiasenlu: Ahh, right.

parlai/core/image_featurizers.py
jiasenlu: there are two parameters used for image feature extraction. --image_cnntype (which specify the CNN type), and --image_layernum (which specify which layer of feature we need). Usually, we use output of fully connected layer as global image feature, or output of last convolutional layer as spatial image feature or others. Previously, I'm thinking we can let user to set the --image_cnntype or --image_layernum, and --image-mode just specify which folder to store the feature.

jiasenlu: Maybe we can limit this option --image-mode (Different ResNet architecture + Global or Spatial). Then we can delete the --image-layernum and --image--cnntype option.

parlai/core/image_featurizers.py
jiasenlu: yeah, let's do this.

parlai/core/dialog_teacher.py
jiasenlu: those stuff are move to image_featurizers.py

PR #170, opened 2017-06-26
ExecutableWorld,  designed to also work with BatchWorld

jaseweston:
A world where messages from agents can be interpreted as _actions_ in the world which result in changes in the environment (are executed). Hence a grounded simulation can be implemented rather than just dialogue between agents.

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -562,7 +562,7 @@ class BatchWorld(World):
         if self.epoch_done():
             raise StopIteration()
 
-    def batch_observe(self, index, batch_actions, index_acting)
+    def batch_observe(self, index, batch_actions, index_acting):
         batch_observations = []
         for i, w in enumerate(self.worlds):
             agents = w.get_agents()
@@ -617,7 +617,7 @@ class BatchWorld(World):
                     w.execute(batch_actions[i])
             # All agents (might) observe the results.
             for other_index in range(num_agents):
-                obs = self.batch_observe(other_index, batch_act, index))
+                obs = self.batch_observe(other_index, batch_act, index)
                 if obs is not None:
                     batch_observations[other_index] = obs
 
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -268,7 +268,7 @@ class DialogPartnerWorld(World):
 
 class MultiAgentDialogWorld(World):
     """Basic world where each agent gets a turn in a round-robin fashion,
-    recieving as input the actions of all other agents since that agent last
+    receiving as input the actions of all other agents since that agent last
     acted.
     """
     def __init__(self, opt, agents=None, shared=None):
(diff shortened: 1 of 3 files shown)


parlai/core/worlds.py
```diff
@@ -315,6 +315,53 @@ def shutdown(self):
...
+        """
+        pass
+
+    def observe(self, agent, act):
+        """An executable world class should implement this function, otherwise
+        the observations for each agent are just the messages from other agents
+        and not confitioned on the world at all (and it is thus the same as
+        MultiAgentDialogWorld). """
+        if agent.id == act['id']:
+            return None
```
alexholdenmiller: None or {}?

parlai/core/worlds.py
```diff
@@ -315,6 +315,53 @@ def shutdown(self):
...
+        acts = self.acts
+        for index, agent in enumerate(self.agents):
+            # The agent acts.
+            acts[index] = agent.act()
+            # We execute this action in the world. 
+            self.execute(agent, acts[index])
+            # All agents (might) observe the results.
+            for other_agent in self.agents:
+                obs = self.observe(other_agent, acts[index])
+                if obs is not None:
```
alexholdenmiller: `if obs` instead of you change `None` to `{}`, but either way is fine with me

parlai/core/worlds.py
```diff
@@ -481,11 +528,20 @@ def __next__(self):
...
         batch_observations = []
         for i, w in enumerate(self.worlds):
             agents = w.get_agents()
-            observation = agents[index].observe(validate(batch_actions[i]))
+            observation = None
+            if hasattr(w, 'observe'):
+                # The world has its own observe function, which the action
+                # first goes through (agents receive messages via the world,
+                # not from each other).
+                observation = w.observe(agents[index], validate(batch_actions[i]))
```
alexholdenmiller: does this have a chance of returning `None`?

parlai/core/worlds.py
```diff
@@ -523,11 +579,17 @@ def parley(self):
...
+                if hasattr(w, 'execute'):
+                    w.execute(w.agents[i], batch_act[i])
+            # All agents (might) observe the results.
             for other_index in range(num_agents):
-                if index != other_index:
-                    batch_observations[other_index] = (
-                        self.batch_observe(other_index, batch_act))
+                obs = self.batch_observe(other_index, batch_act, index)
+                if obs is not None:
+                    batch_observations[other_index] = obs
```
alexholdenmiller: I wonder if you need to override this always to make sure that you don't accidentally view a stale message? maybe better to fill it with `{}`

jaseweston:
I just didn't want to break existing code that doesn't set anything. I think all models should be able to cope with an empty observation though, ideally. So, in a batch there could be empty ones. But right now we don't allow for that.. should I change it?

alexholdenmiller:
we can follow up later with the empty table changes

Issue #168, opened 2017-06-26
"true label missing from candidate labels" in QACNN and QADailyMail


belbs:
```[task:qacnn]
[download_path:/Users/filipeabperes/ParlAI/downloads]
[datatype:train]
[image_mode:raw]
[numthreads:1]
[batchsize:1]
[datapath:/Users/filipeabperes/ParlAI/data]
[model:repeat_label]
[model_file:None]
[dict_class:None]
[evaltask:None]
[display_examples:False]
[num_epochs:1]
[max_train_time:inf]
[log_every_n_secs:1]
[validation_every_n_secs:0]
[validation_patience:5]
[dict_build_first:True]
[parlai_home:/Users/filipeabperes/ParlAI]
[creating task(s): qacnn]
[loading fbdialog data:/Users/filipeabperes/ParlAI/data/QACNN/train.txt]
[ training... ]
Traceback (most recent call last):
  File "train_model.py", line 161, in <module>
    main()
  File "train_model.py", line 106, in main
    world.parley()
  File "/Users/filipeabperes/ParlAI/parlai/core/worlds.py", line 236, in parley
    acts[0] = agents[0].act()
  File "/Users/filipeabperes/ParlAI/parlai/core/dialog_teacher.py", line 130, in act
    action, self.epochDone = self.next_example()
  File "/Users/filipeabperes/ParlAI/parlai/core/dialog_teacher.py", line 115, in next_example
    action, epoch_done = self.data.get(self.episode_idx, self.entry_idx)
  File "/Users/filipeabperes/ParlAI/parlai/core/dialog_teacher.py", line 294, in get
    raise RuntimeError('true label missing from candidate labels')
RuntimeError: true label missing from candidate labels```

jaseweston:
what was the command?

belbs:
python train_model.py -m repeat_label -t qacnn

I'm fixing into this now, apparently its a problem with how the fb format is built from the dataset

belbs:
The data builder was ignoring last line which sometimes included correct label.

This was happening both on QACNN and QADailyMail

PR #171, opened 2017-06-26
Fixed bug on building fb data

belbs:
Fix for issue Issue #168 

Data builder was ignoring last line which sometimes included correct label.


alexholdenmiller:
nice

PR #172, opened 2017-06-27
updates to the training loop / logging

alexholdenmiller:
added reset_metrics, metrics.report() rounds floats, default params for train_loop use mostly -1 now (params are ignored when set to -1)

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -41,14 +41,17 @@ def run_eval(agent, opt, datatype, still_training=False, max_exs=-1):
         opt['task'] = opt['evaltask']
 
     valid_world = create_task(opt, agent)
-    num_exs = max_exs if max_exs > 0 else len(valid_world)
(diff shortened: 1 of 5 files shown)


examples/train_model.py
```diff
@@ -33,14 +33,17 @@
 import math
 import os
 
-def run_eval(agent, opt, datatype, still_training=False):
+def run_eval(agent, opt, datatype, still_training=False, max_exs=-1):
```
jaseweston: can't that be in the opt?

alexholdenmiller: oh I put it there to make it explicit, since even if you set max-valid on the command line you probably don't want to use it at the very end when you're doing the final validation/testing (you can see below it's only used for one `run_eval` but not the other two).

I guess that coincides with the use of `still_training`, so I could interpret from that instead, this is just more explicit

alexholdenmiller:
just making everything better!

PR #176, opened 2017-06-29
bug fixes in drqa

hitvoice:
The default key for the NULL token is `'__NULL__'` (See line 59 in
parlai/core/dict.py), but `'<NULL>'` is used in line 193 and line 226
of parlai/agents/drqa/drqa.py. As a result, `'<NULL>'` is treated as an
unknown word and the index of `'__UNK__'` is actually used for padding,
which is not expected.

The string 'merge\_mode' only appears in line 80 of
parlai/agents/drqa/rnn\_reader.py, which seems to be a typo. It should
be 'question\_mode' according to the context.

### DIFF
--- a/parlai/agents/drqa/drqa.py
+++ b/parlai/agents/drqa/drqa.py
@@ -190,7 +190,7 @@ class DrqaAgent(Agent):
         if ex is None:
             return reply
         batch = batchify(
-            [ex], null=self.word_dict['__NULL__'], cuda=self.opt['cuda']
+            [ex], null=self.word_dict[self.word_dict.null_token], cuda=self.opt['cuda']
         )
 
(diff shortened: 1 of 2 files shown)


parlai/agents/drqa/drqa.py
```diff
@@ -190,7 +190,7 @@ def act(self):
         if ex is None:
             return reply
         batch = batchify(
-            [ex], null=self.word_dict['<NULL>'], cuda=self.opt['cuda']
+            [ex], null=self.word_dict['__NULL__'], cuda=self.opt['cuda']
```
ajfisch: Instead of hard coding it again, let's just use self.word_dict[self.word_dict.null_token]

PR #188, opened 2017-07-05
add ner and pos feature to drqa

gfgkmn:
I found in http://arxiv.org/abs/1704.00051v1, there implement artificial feature "POS" and "NER", but in ParlAI, it seems we don't implement these function.

So i add code to reproduce Drqa like http://arxiv.org/abs/1704.00051v1.

After experiment, after use these feature, I got
test:{'accuracy': 0.6616840113528856, 'hits@k': {1: 0.6288552507095554, 10: 0.6288552507095554, 50: 0.6288552507095554, 100: 0.6288552507095554, 5: 0.6288552507095554}, 'total': 10570, 'f1': 0.761960913560991}

when i use these parameter:
python3 examples/train_model.py -m drqa -t squad -bs 32 -dbf True -mf 'drqa' -vtim 3600 -e 10 --embedding_file glove.840B.300d.txt --tune_partial 1000 --dropout_emb 0.3 --dropout_rnn 0.3

Compared without these features:
valid:{'accuracy': 0.6526963103122043, 'f1': 0.7543689069607271, 'total': 10570, 'hits@k': {1: 0.6203405865657521, 10: 0.6203405865657521, 50: 0.6203405865657521, 100: 0.6203405865657521, 5: 0.6203405865657521}}

python3 examples/train_model.py -m drqa -t squad -bs 32 -dbf True -mf 'drqa' -vtim 3600 -e 10 --embedding_file glove.840B.300d.txt --tune_partial 1000 --dropout_emb 0.3 --dropout_rnn 0.3

### DIFF
--- a/parlai/agents/drqa/utils.py
+++ b/parlai/agents/drqa/utils.py
@@ -7,6 +7,20 @@ import torch
 import time
 import unicodedata
 from collections import Counter
+import spacy
+
+NLP = spacy.load('en')
+pos_list = [
+    'DET', 'ADP', 'PART', 'ADJ', 'PUNCT', 'INTJ', 'NOUN', 'ADV', 'X', 'PRON',
+    'PROPN', 'VERB', 'CONJ', 'SPACE', 'NUM', 'SYM', 'CCONJ'
+]
+ner_list = [
+    'QUANTITY', 'PRODUCT', 'EVENT', 'FACILITY', 'NORP', 'TIME', 'LANGUAGE',
+    'ORG', 'DATE', 'CARDINAL', 'PERSON', 'ORDINAL', 'LOC', 'PERCENT', 'MONEY',
+    'WORK_OF_ART', 'GPE', 'FAC', 'LAW'
+]
+pos_dict = {i: pos_list.index(i)/len(pos_list) for i in pos_list}
+ner_dict = {i: ner_list.index(i)/len(ner_list) for i in ner_list}
 
 
 # ------------------------------------------------------------------------------
@@ -49,6 +63,10 @@ def build_feature_dict(opt):
         feature_dict['in_question_uncased'] = len(feature_dict)
     if opt['use_tf']:
         feature_dict['tf'] = len(feature_dict)
+    if opt['use_ner']:
+        feature_dict['ner_type'] = len(feature_dict)
+    if opt['use_pos']:
+        feature_dict['pos_type'] = len(feature_dict)
     if opt['use_time'] > 0:
         for i in range(opt['use_time'] - 1):
             feature_dict['time=T%d' % (i + 1)] = len(feature_dict)
@@ -70,6 +88,8 @@ def vectorize(opt, ex, word_dict, feature_dict):
     # Create extra features vector
     features = torch.zeros(len(ex['document']), len(feature_dict))
 
+    spacy_doc = NLP(' '.join(ex['document']))
+
     # f_{exact_match}
     if opt['use_in_question']:
         q_words_cased = set([w for w in ex['question']])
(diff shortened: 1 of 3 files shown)


parlai/agents/drqa/utils.py
```diff
@@ -7,6 +7,20 @@
 import time
 import unicodedata
 from collections import Counter
+import spacy
```
ajfisch: Since we already load a spacy model as part of the DrQA SimpleDictionary, I think we should just have these as an extra function -- e.g. `features(pos=False, ner=False)`. And those fields should be passed in as part of the `ex` dict to vectorize.

parlai/agents/drqa/utils.py
```diff
@@ -49,6 +63,10 @@ def build_feature_dict(opt):
         feature_dict['in_question_uncased'] = len(feature_dict)
     if opt['use_tf']:
         feature_dict['tf'] = len(feature_dict)
+    if opt['use_ner']:
```
ajfisch: This is a little weird. You are using one feature each for `pos` and `ner`, and encoding that later as a integer value {0, num_types} ? This should either be a one-hot encoding or used as a low-dimensional embedding.

jaseweston:
note, if we do changes like this we should also make sure the existing
saved model can still be loaded, is that the case?:

wget https://s3.amazonaws.com/fair-data/parlai/_models/drqa/squad.mdl
python eval_model.py -m drqa -t squad -mf squad.mdl -dt valid

ajfisch:
@jaseweston -- should be able to load and run that model. `use_pos` and `use_ner` are options. But yes, @gfgkmn, they should be false by default.

gfgkmn:
Very useful opinion, thanks for commenting, I will change this code and pull again.

jaseweston:
@ajfisch  should we just drop this or can you review it?

parlai/agents/drqa/utils.py
```diff
@@ -70,6 +88,8 @@ def vectorize(opt, ex, word_dict, feature_dict):
     # Create extra features vector
     features = torch.zeros(len(ex['document']), len(feature_dict))
 
+    spacy_doc = NLP(' '.join(ex['document']))
```
ajfisch: this should not be done unless the ner and pos tags are actually used (its expensive)

parlai/agents/drqa/utils.py
ajfisch: Sorry I misread this last time, but same idea remains. see line 23 comment

parlai/agents/drqa/utils.py
```diff
@@ -7,6 +7,20 @@
...
+    'DET', 'ADP', 'PART', 'ADJ', 'PUNCT', 'INTJ', 'NOUN', 'ADV', 'X', 'PRON',
+    'PROPN', 'VERB', 'CONJ', 'SPACE', 'NUM', 'SYM', 'CCONJ'
+]
+ner_list = [
+    'QUANTITY', 'PRODUCT', 'EVENT', 'FACILITY', 'NORP', 'TIME', 'LANGUAGE',
+    'ORG', 'DATE', 'CARDINAL', 'PERSON', 'ORDINAL', 'LOC', 'PERCENT', 'MONEY',
+    'WORK_OF_ART', 'GPE', 'FAC', 'LAW'
+]
+pos_dict = {i: pos_list.index(i)/len(pos_list) for i in pos_list}
+ner_dict = {i: ner_list.index(i)/len(ner_list) for i in ner_list}
```
ajfisch: It doesn't make much sense to me to encode the feature as a fixed scalar between 0 and 1. My guess is the benefit you got from this is really just somewhat binary NE not NE (no typing)...

jaseweston:
ok, i'm just going to close this for now as it has been a while, if the authors want to work on it they can reopen?

PR #192, opened 2017-07-06
Refactor MTurk html and server setup

yf225:
1. Setting up the AWS server no longer requires information about the HIT. Instead the HIT config would be loaded from a separate file called ``hit_config.json``.

2. We now support custom cover page and chat interface for MTurk agent, so that different types of MTurk agent can be shown a different UI.


parlai/mturk/core/agents.py
```diff
@@ -58,7 +60,14 @@ def init_aws(self, opt):
             return
 
         print('Setting up MTurk backend...')
-        html_api_endpoint_url, json_api_endpoint_url, requester_key_gt = setup_aws(task_description=opt['task_description'], num_hits=opt['num_hits'], num_assignments=opt['num_assignments'], is_sandbox=opt['is_sandbox'])
+        create_hit_config(task_description=opt['task_description'], num_hits=opt['num_hits'], num_assignments=opt['num_assignments'], is_sandbox=opt['is_sandbox'])
+        if not self.task_files_to_copy:
```
harmdevries89: @yf225 shouldn't you check here if each mturk_agent_id has its templates in task_files_to_copy?

parlai/mturk/core/setup_aws.py
```diff
@@ -233,7 +243,24 @@ def setup_rds():
 
     return host
 
-def setup_relay_server_api(mturk_submit_url, rds_host, task_description, is_sandbox, num_hits, num_assignments, requester_key_gt, should_clean_up_after_upload=True):
+def create_hit_config(task_description, num_hits, num_assignments, is_sandbox):
```
harmdevries89: This looks much better

parlai/mturk/tasks/model_evaluator/run.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-present, Facebook, Inc.
+1# Copyright (c) 2017-present, Facebook, Inc.
```
alexholdenmiller: oops

parlai/mturk/core/agents.py
yf225: @harmdevries89 I found that setup_aws.py:create_zip_file() would skip missing files, so I guess we don't have to check here

parlai/mturk/core/agents.py
```diff
@@ -31,6 +31,7 @@
 polling_interval = 1 # in seconds
 create_hit_type_lock = threading.Lock()
 local_db_lock = threading.Lock()
+task_dir = os.getcwd()
```
yf225: I agreed - os.getcwd() is not guaranteed to return the directory that the file lives in. Changing it to be using the method you suggested.

parlai/mturk/tasks/model_evaluator/run.py
yf225: oops my bad :/

yf225:
Zhilin's issue seems to be on the number of concurrent database queries during HIT creation which overwhelms the RDS instance. Currently making it so that we are doing bulk message inserts (1 SQL statement for whatever number of new messages we send, and doing it every 1 second). Still need to test for stability and other potential issues and I am planning to run it tonight.

Also added script for automatically completing HITs by sending dummy text. This can replace manual testing for all of our tasks, and it would be especially useful when we want to test for scalability issues.

yf225:
Some investigation results: currently we are using t2.micro for RDS instance, which can only support dozens of concurrent connections (i.e. there can only be dozens of Turkers who have the mturk page open at the same time). I am thinking of creating larger instances when the user runs a larger number of HITs, to avoid the database overload issue. I will create another pull request for it.

harmdevries89:
@yf225 when do you plan to merge this PR? Zhilin's issue seems orthogonal to the refactoring, so I hope we can merge this one soon.

parlai/mturk/core/auto_test.py
```diff
@@ -0,0 +1,78 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+'''
```
alexholdenmiller: Nit: ''' => """

PR #194, opened 2017-07-06
Added download resuming

belbs:


parlai/core/build_data.py
```diff
@@ -26,7 +26,7 @@ def built(path, version_string=None):
         else:
             with open(fname, 'r') as read:
                 text = read.read().split('\n')
-            return (len(text) == 2 and text[1] == version_string)
+            return (len(text) > 1 and text[1] == version_string)
```
belbs: Not completely related, but makes version-checking more robust

belbs:
Also I feel like the interface could be changed if this is accepted. Having only a redownload parameter seems not to make as much sense anymore, but I didnt have great ideas on how to change it, so suggestions accepted

parlai/core/build_data.py
```diff
@@ -56,13 +56,23 @@ def download(url, path, fname, redownload=True):
...
+    if redownload or not resume:
         with requests.Session() as session:
-            response = session.get(url, stream=True)
+            header = {'Range': 'bytes=%d-' % resume_pos}
+            response = session.get(url, stream=True, headers=header)
             CHUNK_SIZE = 32768
             total_size = int(response.headers.get('Content-Length', -1))
-            done = 0
-            with open(outfile, 'wb') as f:
+            total_size += resume_pos
```
alexholdenmiller: shouldn't the total size to download still be the same as the original size?

parlai/core/build_data.py
```diff
@@ -56,13 +56,23 @@ def download(url, path, fname, redownload=True):
...
+            done = resume_pos
+            with open(outfile, mode) as f:
                 for chunk in response.iter_content(CHUNK_SIZE):
                     if chunk:  # filter out keep-alive new chunks
                         f.write(chunk)
```
alexholdenmiller: I'm amazed that this works though, that's awesome!! can you try interrupting a few zip downloads (maybe even more than once during the download) and doing a checksum on them vs a fully downloaded one?

parlai/core/build_data.py
belbs: added comment explaining that

parlai/core/build_data.py
belbs: yep, tried that. it works

belbs:
reorganized a lot of the code for resuming. now it starts downloading into a .part file and creates the  actual file only when done. 

changed the redownload setting because I think it makes more sense not to redownload by default now that we can resume. 

also handling the case where server cant resume.

PR #196, opened 2017-07-07
Add command-line arguments for task agents as well

qipeng:


parlai/core/params.py
```diff
@@ -138,6 +138,19 @@ def add_parlai_args(self):
             '-bs', '--batchsize', default=1, type=int,
             help='batch size for minibatch training schemes')
         self.add_parlai_data_path(parlai)
+        # Find which task specified, and add its specific arguments.
```
jaseweston: can you make a separate function, "add_task_args"  (like we have a "add_model_args" for this)

parlai/core/params.py
```diff
@@ -138,6 +138,19 @@ def add_parlai_args(self):
...
+        args = sys.argv
+        task = None
+        for index, item in enumerate(args):
+            if item == '-t' or item == '--task':
+                task = args[index + 1]
+        if task:
+            agent = get_agent_module(task)
+            if hasattr(agent, 'add_cmdline_args'):
+                agent.add_cmdline_args(self)
+            if hasattr(agent, 'dictionary_class'):
```
jaseweston: you shouldnt need this. we dont have this for models?? that is done by the agent no? @alexholdenmiller  please fix this is another diff (not this one, peng)

qipeng: The task agent is not a "model" (whose cmdline args are added in `add_model_args`, so unfortunately its `add_cmdline_args` is never called for `Teacher` agents unless we add this.

jaseweston:
i get that. thats why i asked:

"can you make a separate function, "add_task_args" (like we have a
"add_model_args" for this)
"

parlai/core/params.py
qipeng: Does 5cbf1d6c67f28c4335d04c8398d34709a025418f seem to do the job?

jaseweston:
- it needs to deal with the case of multitask and trailing ":" args

- it needs an argument group

- not sure why you need to pass in args?

alexholdenmiller:
I kind of like the `args` as is--it allows for calling this function programmatically if we want to in the future, e.g. like we do from the Predictor module

We actually won't need an argument group after a few changes Peng and I discussed (all args will be added within the task so they can make their own arg group)

qipeng:
@jaseweston Short tests and check examples are passing. Will merge in a bit unless you have further comments!

jaseweston:
cool

PR #197, opened 2017-07-07
Clean up dir only on version update

belbs:


parlai/tasks/vqa_v1/build.py
```diff
@@ -11,10 +11,13 @@
...
+    version = '1'
 
-    if not build_data.built(dpath, version_string='1'):
+    if not build_data.built(dpath, version_string=version):
         print('[building image data: ' + dpath + ']')
-        build_data.remove_dir(dpath)
+        # if already built but updating version, remove outdated files
+        if build_data.built(dpath):
+            build_data.remove_dir(dpath)
         build_data.make_dir(dpath)
```
belbs: maybe all these three lines could be packed into a clean_dir function of sorts

alexholdenmiller: yeah I think let's leave for now so it's clear what steps it's doing, people will just copy & paste it anyways

parlai/tasks/vqa_v1/build.py
```diff
@@ -11,10 +11,13 @@
 
 def buildImage(opt):
     dpath = os.path.join(opt['datapath'], 'COCO-IMG')
+    version = '1'
 
-    if not build_data.built(dpath, version_string='1'):
+    if not build_data.built(dpath, version_string=version):
         print('[building image data: ' + dpath + ']')
-        build_data.remove_dir(dpath)
+        # if already built but updating version, remove outdated files
```
alexholdenmiller: let's actually move this comment into the if block (before calling remove_dir), I think it's standard python style to do it that way (weirdly)

maybe change it to
`# an older version exists, so remove these outdated files`

belbs:
Should I apply the same to all tasks?

alexholdenmiller:
Looks great, yeah I'd be happy to roll it out to the other tasks now!

belbs:
Done, added modifications to all tasks

PR #204, opened 2017-07-12
Add TriviaQA task

SanyamSatia:
Add TriviaQA task with the RC version of the dataset: http://nlp.cs.washington.edu/triviaqa/

### DIFF
--- a/parlai/tasks/triviaqa/agents.py
+++ b/parlai/tasks/triviaqa/agents.py
@@ -19,62 +19,114 @@ def _path(opt):
     return (os.path.join(opt['datapath'], 'TriviaQA', 'qa'),
             os.path.join(opt['datapath'], 'TriviaQA', 'evidence'))
 
-def setup_data_common(dataset, path, suffix, evidence_dir):
-    dataset_path = os.path.join(path,
-                                dataset + '-' + suffix + '.json')
-    print('loading: ' + dataset_path)
-    with open(dataset_path) as data_file:
-        data = json.load(data_file)['Data']
-    for datapoint in data:
-        question = datapoint['Question']
-        answers = datapoint['Answer']['Aliases']
-        if dataset == 'web':
-            evidence_list = datapoint['SearchResults']
-        else:
-            evidence_list = datapoint['EntityPages']
-
-        if len(evidence_list) == 0:
-            continue
-
-        evidence_num = random.randrange(len(evidence_list))
-        evidence_filename = evidence_list[evidence_num]['Filename']
-        evidence_file_path = os.path.join(evidence_dir, dataset,
-                                          evidence_filename)
-        with open(evidence_file_path) as evidence_file:
-            evidence = evidence_file.read()
-            yield (evidence + '\n' + question, answers), True
-
 
 class WebTeacher(DialogTeacher):
     def __init__(self, opt, shared=None):
-        if opt['datatype'].startswith('train'):
-            self.suffix = 'train'
-        else:
-            self.suffix = 'dev'
+        if not hasattr(self, 'prefix'):
+            self.prefix = ''
+            if opt['datatype'].startswith('train'):
+                self.suffix = 'train'
+            else:
+                self.suffix = 'dev'
 
         opt['datafile'], self.evidence_dir = _path(opt)
         self.id = 'triviaqa'
         super().__init__(opt, shared)
 
     def setup_data(self, path):
-        return setup_data_common('web', path, self.suffix, self.evidence_dir)
+        dataset = 'web'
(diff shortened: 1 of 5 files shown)


parlai/tasks/triviaqa/agents.py
```diff
@@ -0,0 +1,57 @@
...
+                if len(evidence_list) == 0:
+                    continue
+
+                evidence_num = random.randrange(len(evidence_list))
+                evidence_filename = evidence_list[evidence_num]['Filename']
+                evidence_file_path = os.path.join(self.evidence_dir, dataset,
+                                                  evidence_filename)
+                with open(evidence_file_path) as evidence_file:
+                    evidence = evidence_file.read()
+                    yield (evidence + '\n' + question, answers), True
```
alexholdenmiller: hm I'm going to talk to Jason about how we want to set up the evidence, I think selecting one random one (fixed at launch time) isn't how we'll want to do this

parlai/tasks/triviaqa/agents.py
```diff
@@ -0,0 +1,57 @@
...
+        opt['datafile'], self.evidence_dir = _path(opt)
+        self.id = 'triviaqa'
+        super().__init__(opt, shared)
+
+    def setup_data(self, path):
+        datasets = ['web', 'wikipedia']
+
+        for dataset in datasets:
+            dataset_path = os.path.join(path,
+                                        dataset + '-' + self.suffix + '.json')
```
alexholdenmiller: I may want to split these into different teachers (ie. a WikiTeacher and a WebTeacher)

SanyamSatia: Split into different teachers: WikipediaTeacher and WebTeacher. The DefaultTeacher now inherits from MultiTaskTeacher.

SanyamSatia:
* Split existing teacher into different teachers and changed DefaultTeacher to inherit from MultiTaskTeacher in order to allow training on both datasets simultaneously. 
* Added download test

parlai/tasks/triviaqa/agents.py
```diff
@@ -0,0 +1,83 @@
...
+            self.suffix = 'dev'
+
+        opt['datafile'], self.evidence_dir = _path(opt)
+        self.id = 'triviaqa'
+        super().__init__(opt, shared)
+
+    def setup_data(self, path):
+        return setup_data_common('wikipedia', path, self.suffix,
+                                 self.evidence_dir)
+
```
alexholdenmiller: can you also add a "verified teacher" for the verified data? it will just always use the dev datatype (say, print a warning but don't error if datatype != 'valid'). this won't affect the default teacher

parlai/tasks/triviaqa/agents.py
```diff
@@ -0,0 +1,83 @@
...
+        answers = datapoint['Answer']['Aliases']
+        if dataset == 'web':
+            evidence_list = datapoint['SearchResults']
+        else:
+            evidence_list = datapoint['EntityPages']
+
+        if len(evidence_list) == 0:
+            continue
+
+        evidence_num = random.randrange(len(evidence_list))
```
alexholdenmiller: adjust this as we discussed on messenger, thanks! (separate ex for each web evidence, concat all wiki evidence)

parlai/tasks/triviaqa/agents.py
SanyamSatia: Added a VerifiedWikipediaTeacher and VerifiedWebTeacher inheriting from WikipediaTeacher and WebTeacher, respectively. Also added a VerifiedTeacher inheriting from MultiTaskTeacher.

SanyamSatia:
Changed evidence handling as discussed and added verified teachers

parlai/tasks/triviaqa/agents.py
```diff
@@ -0,0 +1,135 @@
...
+
+        opt['datafile'], self.evidence_dir = _path(opt)
+        self.id = 'triviaqa'
+        super().__init__(opt, shared)
+
+    def setup_data(self, path):
+        dataset = 'web'
+        dataset_path = os.path.join(path,
+                                    self.prefix + dataset + '-' + self.suffix +
+                                    '.json')
```
alexholdenmiller: actually, can you move this to the init, so that `opt['datafile']` points to this? then the user can inspect the exact datafile being loaded by looking at the opt file

SanyamSatia: Moved the initialization of the datafile option to the constructor.

alexholdenmiller:
Awesome, thank you!!

PR #205, opened 2017-07-12
New MTurkAgent APIs, performance improvement

yf225:
1. We can now call approve_work() / reject_work() / pay_bonus() / block_worker() on individual MTurkAgents to take actions on their work. 

2. Performance improvement for concurrent client requests. Now we can support 100 workers working on the HITs at the same time (database connections will not be used up anymore). Plans for supporting even more concurrent workers are underway.


parlai/mturk/core/data_model.py
```diff
@@ -12,87 +12,82 @@
...
-    __tablename__ = 'mturk_hit_assignment_info'
+class MTurkHITAgentAllocation(Base):
+    __tablename__ = 'mturk_hit_agent_allocation'
     id = Column(Integer, primary_key=True)
     task_group_id = Column(String(255), index=True)
     agent_id = Column(String(255), index=True)
 
 
-def is_database_schema_consistent(Base, engine):
+def check_database_health():
```
alexholdenmiller: [I need to continue from here tomorrow]

parlai/mturk/tasks/model_evaluator/worlds.py
```diff
@@ -48,3 +48,10 @@ def report(self):
     def shutdown(self):
         self.task_world.shutdown()
         self.mturk_agent.shutdown()
+        
+    def review_work(self):
```
alexholdenmiller: maybe add something like
```
"""Programmatically approve/reject the turker's work.
For example:
.. code-block:: python
    if self.turkerResponse == '0':
        self.mturk_agent.reject_work('You rated our model's response as a 0/10 but we know we're better than that')
    else:
        if self.turkerResponse == '10':
            self.mturk_agent.payBonus(1, 'Thanks for the great rating!')
        self.mturk_agent.approve_work()
"""
```

maybe we should have a parent class for these mturk worlds so that we have a better place to put documentation that applies to all tasks, like this

yf225:
gonna add more documentation these couple days :)

parlai/mturk/core/setup_aws.py
```diff
@@ -545,7 +545,7 @@ def check_mturk_balance(balance_needed, is_sandbox):
         else:
             raise
 
-    balance_needed = balance_needed * 1.2
+    balance_needed = balance_needed * 1.2 # AWS charges 20% fee for both reward and bonus payment
```
alexholdenmiller: is it not more for bonuses?

parlai/mturk/core/worlds.py
```diff
@@ -0,0 +1,43 @@
...
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+from parlai.core.worlds import World, validate
+
+class MTurkWorld(World):
+    """
+    Generic world for MTurk
+    """
+    def __init__(self, mturk_agent):
```
alexholdenmiller: ideally this should look more like its parent's API, so maybe add opt back in at least?

parlai/mturk/core/setup_aws.py
```diff
@@ -523,6 +523,34 @@ def setup_relay_server_api(rds_host, task_files_to_copy, should_clean_up_after_u
 
     return html_api_endpoint_url, json_api_endpoint_url
 
+def calculate_mturk_cost(payment_opt):
+    """MTurk Pricing: https://requester.mturk.com/pricing
+    20% fee on the reward and bonus amount (if any) you pay Workers.
+    HITs with 10 or more assignments will be charged an additional 20% fee on the reward you pay Workers.
+
```
yf225: Just found that MTurk has a more complex payment structure... Adding a new method to handle it

alexholdenmiller:
will we have issues with the subworlds? they all have different `__init__` params...

yf225:
I feel like each subworld will need their own ``__init__`` params because the task configuration can be quite different from one another (e.g. model_evaluator task expects both a model and a task dataset, while qa_data_collection task expects only a task dataset), so it would probably be good to have users implement their own ``__init__`` methods

PR #208, opened 2017-07-12
add task directory to mturk manager

harmdevries89:
Pass a task-specific directory path from which agent-tailored HTML templates will be loaded. 

Use case: when you're not developing from parlai/mturk/tasks/..

CC: @yf225 @alexholdenmiller


alexholdenmiller:
simple enough!

PR #211, opened 2017-07-13
fix insuranceqa bug

alexholdenmiller:


parlai/tasks/insuranceqa/agents.py
```diff
@@ -10,31 +10,35 @@
...
     def __init__(self, opt, shared=None):
         opt = copy.deepcopy(opt)
-        task = opt.get('task', 'insuranceqa:V2:100')  # options are 100, 500, 1000, or 1500
-        opt['datafile'] = _path("V2", opt, task.split(':')[2])
+        task = opt.get('task', None)
+        if not task:
+            # options are 100, 500, 1000, or 1500
+            task = 'insuranceqa:V2:100'
+        split = task.split(':')
+        opt['datafile'] = _path('V2', opt, split[2])
```
alexholdenmiller: could also add a default, if len(split) < 3 then use "100"

PR #220, opened 2017-07-17
Batch observe

qipeng:
Call `batch_observe`, if available, before `observe` is called. The reason that this is not used **instead** of `observe` is for better backwards compatibility, as well as making it easy to reuse the nice things in the `observe` method in some super classes. 

(Maybe we should call it `batch_preobserve` instead?)

### DIFF
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -83,6 +83,10 @@ class DialogTeacher(Teacher):
         if self.epochDone:
             raise StopIteration()
 
+    @property
+    def internal_states(self):
+        return super().internal_states + ['lastY']
+
     def share(self):
         shared = super().share()
         shared['data'] = self.data
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -85,7 +85,7 @@ class DialogTeacher(Teacher):
 
     @staticmethod
     def internal_states():
-        return super(DialogTeacher, DialogTeacher).internal_states() + ['lastY']
+        return super().internal_states() + ['lastY']
 
     def share(self):
         shared = super().share()
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -83,9 +83,9 @@ class DialogTeacher(Teacher):
         if self.epochDone:
             raise StopIteration()
 
-    @property
-    def internal_states(self):
-        return super().internal_states + ['lastY']
+    @staticmethod
+    def internal_states():
+        return super(DialogTeacher, DialogTeacher).internal_states() + ['lastY']
 
     def share(self):
         shared = super().share()
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -83,8 +83,8 @@ class DialogTeacher(Teacher):
         if self.epochDone:
             raise StopIteration()
 
-    @staticmethod
-    def internal_states():
+    @classmethod
+    def internal_states(cls):
         return super().internal_states() + ['lastY']
 
     def share(self):
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -550,6 +550,13 @@ class BatchWorld(World):
 
     def batch_observe(self, index, batch_actions, index_acting):
         batch_observations = []
+        a = self.world.get_agents()[index]
+
+        if (batch_actions is not None and len(batch_actions) > 0 and
+                hasattr(a, 'batch_observe')) and index != index_acting:
+            # if there's anything that needs to be batched, do it here
+            a.batch_observe(batch_actions)
+
         for i, w in enumerate(self.worlds):
             agents = w.get_agents()
             observation = None
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -569,7 +569,17 @@ class BatchWorld(World):
                 hasattr(a, 'batch_observe')):
             batch_observations = a.batch_observe(batch_observations)
         else:
+            # Set internal states from batched agent
+            for attname in a.internal_states:
+                if not hasattr(a, 'batch_{}'.format(attname)): continue
+                for w, att in zip(self.worlds, getattr(a, 'batch_{}'.format(attname))):
+                    setattr(w.get_agents()[index], attname, att)
+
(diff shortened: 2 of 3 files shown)


parlai/core/worlds.py
```diff
@@ -550,6 +550,13 @@ def __next__(self):
 
     def batch_observe(self, index, batch_actions, index_acting):
         batch_observations = []
+        a = self.world.get_agents()[index]
+
+        if (batch_actions is not None and len(batch_actions) > 0 and
+                hasattr(a, 'batch_observe')) and index != index_acting:
+            # if there's anything that needs to be batched, do it here
+            a.batch_observe(batch_actions)
```
jaseweston: i don't think this will work in the case of hasattr(w, 'observe'), in that case the agents are not supposed to see the batch actions, only observe through the world in that case...
maybe it's better to put it on line 574.. after the batch_observations have been built??
would need a comment explaining. i think we should get alex and adam's view too

alexholdenmiller: the point of this function should be to give the agents the opportunity to implement a ``batch_observe`` if they prefer it over shared ``observe`` definitions, just as we do with act. in that case, we need to make sure line 571 is never called, so that needs to be contained in an else block (like we do in batch_act below--we have a "batch" case and a "call each" case).

however, as jason said, we may need to transform the observations via w.observe. Thus, we probably need to do that first to each observation, looping over the words and changing the contents of batch acts according to either calling w.observe or just validate.

Then, check for batch_observe. If it's available, call it as you do. Otherwise, call it on the particular agent as we're doing now on 571.

I think that's what we're looking for.

Here's a simple example of a batch_observe definition for the seq2seq agent, though I don't think it would help that agent since it doesn't need to pass any state around or do any batched computation in observe, so I'm not going to add it to that class for now. It might be good for us to see how you'd like to use it (since you have a use case of actual batched computation). the one advantage of this is that you wouldn't need any kind of shared definition parameters to be set up.

```
    def observe(self, observation):
        observation = copy.deepcopy(observation)
        if not self.episode_done:
            # if the last example wasn't the end of an episode, then we need to
            # recall what was said in that example
            prev_dialogue = self.observation['text']
            observation['text'] = prev_dialogue + '\n' + observation['text']
        self.observation = observation
        self.episode_done = observation['episode_done']
        return observation

    def batch_observe(self, observations):
        # observations is a list of observation dicts
        for i, obs in enumerate(observations):
            obs = copy.deepcopy(obs)
            if not self.episode_done_arr[i]:
                # if the last example wasn't the end of an episode, then we need
                # to recall what was said in that example
                prev_dialogue = self.observations[i]['text']
                obs['text'] = prev_dialogue + '\n' + obs['text']
            self.observations[i] = obs
            self.episode_done_arr[i] = obs['episode_done']
        return self.observations
```

qipeng: @jaseweston @alexholdenmiller Thanks for the comments! `w.observe` is an easy fix.

However, the point of keeping both `batch_observe` and `observe` (instead of making them mutually exclusive) is so that task agents (`DialogTeacher`s) can directly inherit the nice behavior of metric updates from the base class' `act` and `observe` methods. Would be happy to make them mutually exclusive if we find out a nice way of inheriting those...

alexholdenmiller: you could still call `super().observe(obs)` or `self.observe(obs)`

qipeng: @alexholdenmiller Not exactly. The instance that `batch_observe` is being called on is different from the ones that `observe` would have been called on, which relies on their internal states to update the metrics (`self.lastY`). If `batch_observe` is mutually exclusive with `observe` for agents, all `Teacher`s would have to implement `batch_act` as well and update some batch-shared states for metric update to work?

qipeng:
Making `batch_observe` and `observe` mutually exclusive after in-person discussion with @alexholdenmiller

parlai/core/dialog_teacher.py
```diff
@@ -83,6 +83,10 @@ def __next__(self):
         if self.epochDone:
             raise StopIteration()
 
+    @property
+    def internal_states(self):
+        return super().internal_states + ['lastY']
```
alexholdenmiller: TIL about Python properties--cool! But it doesn't look like we're taking advantage of them at all, right? This is just removing `()` from after `a.internal_states`

parlai/core/worlds.py
```diff
@@ -569,8 +569,18 @@ def batch_observe(self, index, batch_actions, index_acting):
                 hasattr(a, 'batch_observe')):
             batch_observations = a.batch_observe(batch_observations)
         else:
+            # Set internal states from batched agent
+            for attname in a.internal_states:
+                if not hasattr(a, 'batch_{}'.format(attname)): continue
```
alexholdenmiller: I'd prefer `if hasattr` to `if not hasattr... continue`

parlai/core/worlds.py
```diff
@@ -569,8 +569,18 @@ def batch_observe(self, index, batch_actions, index_acting):
...
+                for w, att in zip(self.worlds, getattr(a, 'batch_{}'.format(attname))):
+                    setattr(w.get_agents()[index], attname, att)
+
             batch_observations = [w.get_agents()[index].observe(o) for w, o in zip(self.worlds, batch_observations)]
 
+            # Get internal states and save in batched agent
+            for attname in a.internal_states:
+                setattr(a, 'batch_{}'.format(attname), [getattr(w.get_agents()[index], attname) for w in self.worlds])
+
```
alexholdenmiller: should we skip this if the agent doesn't have batch_act or batch_observe? maybe that extra code isn't worth the speedup

qipeng: Skipping this when unnecessary does seem to give a small performance boost, but I didn't profile too carefully.

qipeng:
@jaseweston My code is fully compatible with the latest commit here at the moment, so I'm not blocked by this at all.

jaseweston:
ok, cool

alexholdenmiller:
ha, the github bot is watching you @qipeng

qipeng:
@alexholdenmiller haha just signed the CLA (not sure if the bot is all caught up though, it didn't remove the CLA signed tag in the first place, which I think might be a corner case for it?)

jaseweston:
We're going to drop this for now.

PR #221, opened 2017-07-18
Add start of sentence token to dict.py

Henry-E:
Added a start of sentence token. And reordered the tokens from 
NULL -> EOS -> UNK 
to 
NULL -> UNK -> EOS -> GO

This could break code that uses hardcoded eos_idx = 1

Also sorry about the extraneous commits, not sure how to cherry pick them

### DIFF
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -71,8 +71,9 @@ class DictionaryAgent(Agent):
     default_maxngram = -1
     default_minfreq = 0
     default_null = '__NULL__'
-    default_eos = '__EOS__'
     default_unk = '__UNK__'
+    default_eos = '__EOS__'
+    default_go = '__GO__'
 
     @staticmethod
     def add_cmdline_args(argparser):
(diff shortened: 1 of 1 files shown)


parlai/core/dict.py
```diff
@@ -71,8 +71,9 @@ class DictionaryAgent(Agent):
     default_maxngram = -1
     default_minfreq = 0
     default_null = '__NULL__'
-    default_eos = '__EOS__'
     default_unk = '__UNK__'
+    default_eos = '__EOS__'
+    default_go = '__GO__'
```
jaseweston: maybe having a _START_ and an _END_ is better?

parlai/core/dict.py
```diff
@@ -71,8 +71,9 @@ class DictionaryAgent(Agent):
     default_maxngram = -1
     default_minfreq = 0
     default_null = '__NULL__'
-    default_eos = '__EOS__'
     default_unk = '__UNK__'
+    default_eos = '__EOS__'
```
jaseweston: don't think it's a good idea to change the existing order..

parlai/core/dict.py
Henry-E: So SOS instead of GO throughout? SOS just looks kind of odd but I'm happy to go with it

parlai/core/dict.py
Henry-E: ok will undo that

parlai/core/dict.py
alexholdenmiller: Thanks for the addition Henry! Can you change __GO__ to __START__ and __EOS__ to __END__? slightly more generic names I guess / they match each other better

Henry-E:
Changes have been made

alexholdenmiller:
looks good to me, thanks Henry! I'll merge now

alexholdenmiller:
this will break seq2seq for a sec but I'll fix that now too

PR #222, opened 2017-07-18
Tutorial for seq2seq agent, updated agent

alexholdenmiller:
Long tutorial explaining a (slightly modified) version of the seq2seq agent, and updated that agent to have more documentation / be simpler / fixed utility things like loading from file.

Also removed an observation table validation check I don't think we need from the worlds file (text_candidate[0] =! text is okay, as you may want to do ranking and generation separately).


parlai/core/agents.py
```diff
@@ -149,9 +149,7 @@ def reset_metrics(self):
         self.metrics.clear()
 
     def share(self):
-        """If applicable, share any parameters needed to create a shared version
```
jaseweston: why is this gone?

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -37,70 +44,105 @@ def add_cmdline_args(argparser):
...
+                # load model parameters if available
+                print('Loading existing model params from ' + opt['model_file'])
+                new_opt, self.states = self.load(opt['model_file'])
+                # override options with stored ones
+                opt = self.override_opt(new_opt)
+
             self.dict = DictionaryAgent(opt)
             self.id = 'Seq2Seq'
-            hsz = opt['hiddensize']
+            # we use EOS markers to break input and output and end our output
```
jaseweston: maybe 'END' would be a better name than 'EOS'?  (it might not be a sentence?)

docs/source/basic_tutorial.rst
```diff
@@ -24,19 +24,20 @@ Follow the step by step guide on how to download and install ParlAI.
...
 .. code-block:: bash
 
-        cd ~/ParlAI; python setup.py develop
+    cd ~/ParlAI; python setup.py develop
 
-3. Several models have additional requirements
+3. Several models have additional requirements:
 
-  a. DrQA requires installing `PyTorch <http://pytorch.org/>`
+- DrQA requires installing `PyTorch <http://pytorch.org/>`_.
```
jaseweston: and seq2seq.. be good to put it in general examples!

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -17,10 +17,17 @@
 
 
 class Seq2seqAgent(Agent):
-    """Simple agent which uses an LSTM to process incoming text observations."""
+    """Simple agent which uses an RNN to process incoming text observations.
```
jaseweston: can you give me a summary of what all the updates do?
does it support ranking now? or anything else new?

docs/source/seq2seq_tutorial.rst
```diff
@@ -0,0 +1,472 @@
...
+        # valid_inds tells us the indices of all valid examples
+        # e.g. for input [{}, {'text': 'hello'}, {}, {}], valid_inds is [1]
+        # since the other three elements had no 'text' field
+        xs, ys, valid_inds = self.batchify(observations)
+
+        if len(xs) == 0:
+            # no valid examples, just return the empty responses we set up
+            return batch_reply
+
+        # produce prodictions either way, but use the targets if available
```
jaseweston: predictions

parlai/agents/seq2seq/seq2seq.py
alexholdenmiller: it looks like a lot of changes but they're almost all aesthetic / refactoring / commenting. only real change is that I now support loading models which had different params (the current params are overridden, with a messaged displayed for each)

parlai/agents/seq2seq/seq2seq.py
alexholdenmiller: sure we could switch, the mcgill folks used eos and eou (end of utterance) in the ubuntu dataset, and douwe was saying we should have a "start of sentence" token as well, so I'm open to changing this to whatever is most useful / clear

parlai/core/agents.py
alexholdenmiller: just changed it to a more accurate description, this comment was copypasta from Agents.share()

parlai/agents/seq2seq/seq2seq.py
jaseweston: ok, whatever you think, but sync with https://github.com/facebookresearch/ParlAI/pull/221 please

parlai/core/worlds.py
```diff
@@ -54,12 +54,6 @@
 def validate(observation):
     """Make sure the observation table is valid, or raise an error."""
     if observation is not None and type(observation) == dict:
-        if ('text_candidates' in observation and
```
jaseweston: why is this gone?

alexholdenmiller: I was thinking I would like to see both my generation and ranking metrics separately once I add ranking. That is, accuracy would be based on generation and ranking based on my candidates. This check in validation forces them to be synced (aka overrides generation if we do ranking), which didn't seem necessary.

alexholdenmiller:
removing save descripton

parlai/agents/seq2seq/seq2seq.py
alexholdenmiller: will do, I'll let that one settle first

PR #223, opened 2017-07-18
Tensorboard logging/visualization

filipeabperes:
This is something I've been using, but I'm not sure if it should be merged to master. Since it could be useful to some, I'm placing it here. It's a modification to the train script that sets up a flag for allowing tensorboard logging, which can then be visualized (even remotely) with a tensorboard server . The train script only sets up the accuracy, but the agent can set up other values for logging by placing them in the opt['tensorboard'] dict, eg below.

image

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -32,6 +32,8 @@ import copy
 import importlib
 import math
 import os
+import tensorboard_logger
+
 
 def run_eval(agent, opt, datatype, still_training=False, max_exs=-1):
     ''' Eval on validation/test data. '''
@@ -93,6 +95,8 @@ def main():
     train.add_argument('-dbf', '--dict-build-first',
                         type='bool', default=True,
                         help='build dictionary first before training agent')
+    train.add_argument('--tensorboard-dir', default='',
+                        help='directory for tensorboard logging')
     opt = parser.parse_args()
     # Possibly build a dictionary (not all models do this).
     if opt['dict_build_first'] and 'dict_file' in opt:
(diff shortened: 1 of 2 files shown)


belbs:
yeah, I thought it seemed too hacky (?) to merge into the current script as well, which is pretty clean. not sure where to put it though, but thought someone might benefit from it just by putting it over here as a separate branch

alexholdenmiller:
we could add a logger to the utils which handles printing / logging to tensorboard / w.e depending on params?

alexholdenmiller:
then just simple log command in the train_model loop and the logger will do all the fancy stuff

jaseweston:
that would be cool

belbs:
that could work, but I'm using a third party library to log in tensorboard format. the reference to it could just be moved into the logger, I suppose (?)

alexholdenmiller:
right!

belbs:
I moved the logger into a class in utils. Once it is created I keep a reference to it in opt, so that models can log data as well, eg below with the loss. (But the logger is not copied on deepcopy of opt, in order not to waste space/time.) 

```
if 'logger' in self.opt:
            self.opt['logger'].log_train({'loss': loss.data[0]})
```

examples/train_model.py
```diff
@@ -93,13 +94,20 @@ def main():
...
     # Possibly build a dictionary (not all models do this).
     if opt['dict_build_first'] and 'dict_file' in opt:
         if opt['dict_file'] is None and opt.get('model_file'):
             opt['dict_file'] = opt['model_file'] + '.dict'
         print("[ building dictionary first... ]")
         build_dict.build_dict(opt)
+
+    # Set up logger if directory provided
+    if opt['tensorboard_dir']:
+        opt['logger'] = TensorboardLogger(opt['tensorboard_dir'])
```
jaseweston: we are trying to keep opt as only strings, float, bools etc,  not whole classes.

belbs:
Though it doesn't seem very clean, the only other way I can think of to pass values between the training script and the actual model is via a dictionary in opt (the way I put it now).

alexholdenmiller:
so does the agent need to fill that log dict then--that info isn't available at this level of the loop? (we don't have your agent's side of it in this diff--what is it putting in that dict exactly?)

one other option other than using "opt" would be to pass it in the "shared" table to the agent but maybe that's too hacky too

belbs:
ah, yes, forgot to provide an example for that. eg, the agent would do:

```
if 'train_log' in self.opt:
            self.opt['train_log']['loss'] = loss.data[0]
```

And yes the shared is another possibility. In general I think it is good to think about this because it might be useful to have a defined way for the train loop to get information from the agent. Even if the type of logger is swapped later, this would still be useful.

jaseweston:
opt is meant to be for options..

jaseweston:
closing this task as we have a new, fresh one coming!

PR #230, opened 2017-07-21
MTurk improvements

yf225:
- Add email_worker() API to MTurkAgent
- Add start_new_run() method to allow creating HITs for new round without re-initializing AWS
- Allow setting `assignment_duration_in_seconds` (HIT expiration time) in opt dict
- Busy-wait for worker to accept HIT in MTurkAgent.__init__ method
- Hide implementation detail of conversation_id to prevent change, since the format for conversation_id is also hard-coded in other places
- Remove lock for create_hit_type() since it will only be called from one thread right now

### DIFF
--- a/parlai/mturk/core/agents.py
+++ b/parlai/mturk/core/agents.py
@@ -301,6 +301,11 @@ class MTurkAgent(Agent):
         self.hit_id = None
         self.worker_id = None
 
+        # Wait for MTurk-specific info
+        while not (self.assignment_id and self.hit_id and self.worker_id):
+            self.assignment_id, self.hit_id, self.worker_id = self.manager.get_hit_assignment_info(self.manager.task_group_id, self.conversation_id, self.id)
+            time.sleep(polling_interval)
+
     def observe(self, msg):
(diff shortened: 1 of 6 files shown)


parlai/mturk/core/agents.py
```diff
@@ -345,6 +363,13 @@ def pay_bonus(self, bonus_amount, reason='unspecified'):
         if self.manager.pay_bonus(worker_id=self.worker_id, bonus_amount=bonus_amount, assignment_id=self.assignment_id, reason=reason, unique_request_token=unique_request_token):
             print("Paid $" + str(bonus_amount) + " bonus to WorkerId: " + self.worker_id)
 
+    def email_worker(self, subject, message_text):
+        response = self.manager.email_worker(worker_id=self.worker_id, subject=subject, message_text=message_text)
+        if 'success' in response:
+            print("Email sent to worker ID: "+str(self.worker_id)+": Subject: "+str(subject)+": Text: "+str(message_text))
+        elif 'failure' in response:
+            print("Unable to send email to worker ID: "+str(self.worker_id)+". Error: "+str(response['failure']))
+
```
alexholdenmiller: would a runtime error make more sense than just a print, do you think? your call on how robustness would affect the usage

yf225: great catch - I added a return value to the method and leave it to the user to decide how to handle the error

PR #232, opened 2017-07-21
Exception to alert user of possible mistake when providing label and label candidate

belbs:
Currently in DialogData if you make a mistake and a pass a string as a label instead of a list it iterates over the string and provides a list of labels that is, for example ['y','e','s'], instead of ['yes']. Same applies to label candidates. Added an exception to alert the user and prevent unnoticed mistakes.

### DIFF
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -228,10 +228,12 @@ class DialogData(object):
                     new_entry.append(None)
                 if len(entry) > 1:
                     # process labels if available
-                    if entry[1] is not None:
+                    if entry[1] is None:
+                        new_entry.append(None)
+                    elif hasattr(entry[1], '__iter__') and type(entry[1]) is not str:
                         new_entry.append(tuple(sys.intern(e) for e in entry[1]))
                     else:
-                        new_entry.append(None)
+                        raise TypeError('Must provide iterable over labels, not a single string.')
                     if len(entry) > 2:
                         # process reward if available
                         if entry[2] is not None:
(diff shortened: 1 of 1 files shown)


parlai/core/dialog_teacher.py
```diff
@@ -228,30 +228,33 @@ def _load(self, data_loader):
...
-                            else:
+                            # process label candidates if available
+                            if entry[3] is None:
                                 new_entry.append(None)
+                            elif last_cands and entry[3] is last_cands:
+                                # if cands are shared, say "same" so we
+                                # don't store them again
+                                new_entry.append(
+                                    sys.intern('same as last time'))
+                            elif hasattr(entry[3], '__iter__') and type(entry[3]) is not str:
```
jaseweston: explain in comment what this one is?

parlai/core/dialog_teacher.py
```diff
@@ -228,30 +228,33 @@ def _load(self, data_loader):
...
-                                        sys.intern('same as last time'))
-                                else:
-                                    last_cands = entry[3]
-                                    new_entry.append(tuple(
-                                        sys.intern(e) for e in entry[3]))
-                            else:
+                            # process label candidates if available
+                            if entry[3] is None:
                                 new_entry.append(None)
+                            elif last_cands and entry[3] is last_cands:
```
jaseweston: seems you are doing more in this diff than the description, or?

parlai/core/dialog_teacher.py
belbs: added

parlai/core/dialog_teacher.py
belbs: its just what's in the description, I just changed the order of the if-elses in order to avoid unnecessary nesting

PR #235, opened 2017-07-24
weak ranking system for seq2seq

alexholdenmiller:
Works great for single-token candidates, questionable method for multi-token ranking (as it generates word i, it increases the score of the candidates that had the most likely word in the ith position)

plenty of room for improvement:
- independently ranking the candidates separate from the generation
- improving GPU performance of ranking by switching to batched tensor operations

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -3,7 +3,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree. An additional grant
 # of patent rights can be found in the PATENTS file in the same directory.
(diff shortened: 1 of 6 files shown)


alexholdenmiller:
pending: disable ranking during training, do full decoding on each candidate during eval

jaseweston:
On Tue, Jul 25, 2017 at 4:00 PM, Alexander Miller <notifications@github.com>
wrote:

> pending: disable ranking during training, do full decoding on each
> candidate during eval
>

yes on eval, if switched on (otherwise only do generation)?

PR #243, opened 2017-07-27
Added candidates to clevr

belbs:


parlai/tasks/clevr/agents.py
```diff
@@ -28,7 +28,17 @@ def _path(opt):
...
+counts = [str(i) for i in range(11)]
+materials = ['metal', 'rubber']
+sizes = ['small', 'large']
+shapes = ['cube', 'sphere', 'cylinder']
+colors = ['gray', 'blue', 'brown', 'yellow', 'red', 'green', 'purple', 'cyan']
+
+
 class DefaultTeacher(DialogTeacher):
+    # all possile answers for the questions
+    cands = ['yes', 'no'] + counts + materials + sizes + shapes + colors
```
alexholdenmiller: won't some answers be combinations of these? like "metal cube" or "gray sphere"?

belbs: it seems that the questions are always related only to one aspect. I tested the current candidates set and all the answers are in it.

PR #245, opened 2017-08-01
Simple task that just loads the specified FbDialogData file

jaseweston:


parlai/tasks/fromfile/agents.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+# This task simply loads the specified file: useful for quick tests without
```
alexholdenmiller: actually, can you change this to a """docstring"""? that way it will get loaded in the documentation

PR #249, opened 2017-08-01
Added image args to parlai args

belbs:
Moved the image args into the main args. They are only used if image mode is set, though.

(Diff looks huge on image featurizers, but only change is removing the args method)

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -70,9 +70,9 @@ class ParlaiParser(argparse.ArgumentParser):
 
         if add_parlai_args:
             self.add_parlai_args()
+            self.add_image_args()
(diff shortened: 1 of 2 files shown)


parlai/core/params.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, add_parlai_args=True, add_model_args=False, model_argv=None):
             self.add_parlai_args()
         if add_model_args:
             self.add_model_args(model_argv)
+        self.add_image_args()
```
jaseweston: should be under add_parlai_args on line 71? or no?

belbs: yeah, I guess it's cleaner like that

PR #250, opened 2017-08-02
misc fixes, removed display_data text_candidates code to speed it up

alexholdenmiller:


examples/display_data.py
```diff
@@ -31,12 +32,17 @@ def main():
...
-        for k in range(int(opt['num_examples'])):
+        cnt = 0
+        for _ in world:
             world.parley()
             print(world.display() + '\n~~')
+            cnt += 1
             if world.epoch_done():
                 print('EPOCH DONE')
                 break
+            elif opt['num_examples'] > 0 and cnt >= opt['num_examples']:
```
jaseweston: why does it need to be this way?

parlai/agents/repeat_label/repeat_label.py
```diff
@@ -52,9 +51,4 @@ def act(self):
         else:
             reply['text'] = "I don't know."
 
-        if 'label_candidates' in obs and len(obs['label_candidates']) > 0:
```
jaseweston: why not do this any more?   it seemed to make sense to me, and repeat_label is only a debugging thing anyway? is there a speed problem or what?

jaseweston: this shouldnt be slow

parlai/core/agents.py
```diff
@@ -237,8 +237,11 @@ def report(self):
         num_tasks = 0
         total = 0
         for i in range(len(self.tasks)):
+            tid = self.tasks[i].getID()
             mt = self.tasks[i].report()
-            m['tasks'][self.tasks[i].getID()] = mt
+            while tid in m['tasks']:
+                tid += '_'
```
jaseweston: what is this doing? comment?

parlai/agents/repeat_label/repeat_label.py
alexholdenmiller: this was slow because label_candidates might not allow indexing (might be a set). we keep it in though.

parlai/core/agents.py
alexholdenmiller: this prevents id cloberring, so if you have e.g.
"-t babi:task1k:1,babi:task1k:2,babi:task1k:2" then your report will show the results from all three, rather than hiding the second instance of task1k:2

examples/display_data.py
alexholdenmiller: I did this so that num_examples can be -1 to show all examples

jaseweston: code is uglier though like that

parlai/agents/repeat_label/repeat_label.py
jaseweston: what do you mean we keep it in though, you seem to have removed it?

parlai/core/agents.py
jaseweston: ah cool

parlai/agents/repeat_label/repeat_label.py
alexholdenmiller: oops meant "we can keep it in", added back

examples/display_data.py
alexholdenmiller: reverting

PR #252, opened 2017-08-04
[WIP] Handle disconnecting mturk agents

harmdevries89:
- Added heartbeat on top of the sockets
- Fires a disconnect event when client doesn't send a heartbeat for more than 4 seconds
- On a disconnect event, the dialog ends and all other agents are informed about the disconnect
- TODO: also catch disconnect event in MTurkAgent.act() 

cc: @yf225


JackUrb:
Functionality from this branch was introduced into #277

PR #258, opened 2017-08-11
Reliable socket messaging

harmdevries89:
Refactor for fast and reliable message passing over socketIO. 

Features:
- Message queue on both World and Client side to send messages in the order as they are put in the queue. 
- NodeJs server act as router and does not implement any retry logic i.e. the sending client is responsible for making sure the other side actually received the message. 
- Automatic resends when receiver does not ack the message
- Supports blocking and non-blocking messages. When sending blocking messages we ensure that the messages are received in the order as they have been sent. 
- Much faster than polling. 

Todo:
- Proper handling of disconnects i.e. when no heartbeat is receiver for X amount of time.

@yf225  @jaseweston @SaizhengZhang @kimiyoung


alexholdenmiller:
41 new commits? :O

alexholdenmiller:
just confirming: are these all new commits?

yf225:
There are quite a few incremental commits that are small in size, maybe we can squash them into just a couple of big ones?

alexholdenmiller:
Oh no that's not a problem--when we merge we'll squash them. I was just wondering if they were all actually new or if the diff head got screwed up and it was including old commits (has happened on some of our other PRs)

harmdevries89:
Just want to add that Jack is working on top of this branch here:
https://github.com/JackUrb/ParlAI/tree/unique_users

He will probably have a stable and tested version in a few days.

JackUrb:
Features of this branch were included in #277

PR #259, opened 2017-08-15
Added clevr to task list and item on news

belbs:


parlai/tasks/task_list.py
```diff
@@ -234,4 +234,11 @@
         "tags": [ "All",  "QA" ],
         "description": "A large scale Machine Reading Comprehension Dataset with questions sampled from real anonymized user queries and contexts from web documents. From Nguyen et al. '16. Link: https://arxiv.org/abs/1611.09268"
     }
```
yf225: @belbs Just found that we are probably missing a comma at the end of this line (tried to use this file to automatically create tasks on the leaderboard but failed to import the file)

belbs: By the time I noticed this it had already been fixed, I think its already corrected in the current master

jaseweston:
yes, someone submitted a fix

PR #263, opened 2017-08-16
Initial commit of FVQA task.

sxjzwq:
Added our FVQA dataset (https://arxiv.org/abs/1606.05433).

Includes some COCO images and some imagenet images.  I understand you probably want to avoid overlap with the existing COCO image data, but we only use a small subset of them so it may be painful for users to download the full set if they're not using it otherwise.

I've also done some adventurous things with metrics, maybe there was a better way to handle that?


jaseweston:
Cooool! I guess Alex will review this when he gets back from vacation (Mon+
I think)..

sxjzwq:
That's great! Thanks Alex.

alexholdenmiller:
@sxjzwq delaying this for the moment--we'd like the 'text' field of the tasks to only contain human-readable data.

Obviously, for the facts that are read as "conceptnet/e/3d86ac72e2b2ff15e89eec2303f8c14a1bbefa80" this doesn't apply. We want to still be able to support reproducing paper results within parlai, so I'm fine if this reference was moved to a field "custom" within the observation table rather than "labels". But can we modify the labels field to contain the textual content of the fact rather than its address?

sxjzwq:
OK in this case maybe we remove the custom metric and do natural language matching on the fact text?  Do you want me to make these changes and update the pull request?

alexholdenmiller:
That would be stellar. This makes the task more in line with the others, in that a "skilled" human could have a chance at doing it.

In that case yeah, the normal metrics class should be able to cover it for you.

Thanks!

sxjzwq:
OK, pushed an update.  Let me know of any outstanding issues.  Cheers!

alexholdenmiller:
Okay looks great now, thank you! Do you have a preference for the fact questions and the image questions being tallied separately in the metrics rather than using a single metrics class? Committing for now but maybe better to just combine them back to one metrics class.

alexholdenmiller:
by the way, before I add I'm going to send a commit to your repo with the "PATENT" info as well as a few smaller changes

sxjzwq:
We think it's better to separate the fact question and image question metrics at this stage. Thanks! Let's move on.

alexholdenmiller:
https://github.com/sxjzwq/ParlAI/pull/1

alexholdenmiller:
Thanks!

PR #264, opened 2017-08-16
xrange() was removed in Python 3, use range() instead

cclauss:


cclauss:
@jaseweston I am not clear on which part of the code is not yet being used.

PR #267, opened 2017-08-21
use START token between input and label instead of END

alexholdenmiller:
this helps learning with padding, because when there's padding the model is observing...

`Where is John ? END bathroom END NULL NULL NULL`

...and is learning that NULL follows END but also that end follows the question and prompts an answer. better to see...

`Where is John ? START bathroom END NULL NULL NULL`

...and just learn to always say NULL after END but to answer the question after START

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -37,13 +37,15 @@ class Seq2seqAgent(Agent):
             help='learning rate')
         agent.add_argument('-dr', '--dropout', type=float, default=0.1,
             help='dropout rate')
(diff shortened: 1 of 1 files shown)


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -447,6 +460,13 @@ def save(self, path=None):
             with open(path, 'wb') as write:
                 torch.save(model, write)
 
+    def shutdown(self):
```
jaseweston: when is this used?

alexholdenmiller: if you hit ctrl-c it saves the current model :)

PR #268, opened 2017-08-21
add ability to use f1 for validation instead of accuracy

alexholdenmiller:
for some of the more challenging datasets (e.g. opensubtitles), the seq2seq model struggles to get accuracy above zero, so using F1 can help it to make better partial progress and save incremental models during validation


jaseweston:
yes or optimizing the rank loss might be good too.. will that work too?

alexholdenmiller:
yes that would probably work too though the code doesn't support it yet! (can only do aggregated metrics, which is only accuracy and f1 in batchworld/multitask)

Issue #270, opened 2017-08-22
Runtime error in Seq2Seq agent caused by changes to .t() in PyTorch v0.2


kearnsw:
As of PyTorch v0.2, the .t() function is only applicable to 2D Variables: https://github.com/pytorch/pytorch/commit/8d33603901100dbdff43b1947867430a93432999. This causes a runtime error when calling the Seq2Seq agent as in the following documentation example:
```
user:~/projects/ParlAI λ python examples/train_model.py -m seq2seq -t babi:task1k:1 -bs 8 -e 1 -mf /tmp/model_s2s 
Traceback (most recent call last):
  File "examples/train_model.py", line 216, in <module>
    main()
  File "examples/train_model.py", line 130, in main
    world.parley()
  File "/home/user/projects/ParlAI/parlai/core/worlds.py", line 601, in parley
    batch_act = self.batch_act(index, batch_observations[index])
  File "/home/user/projects/ParlAI/parlai/core/worlds.py", line 574, in batch_act
    batch_actions = a.batch_act(batch_observation)
  File "/home/user/projects/ParlAI/parlai/agents/seq2seq/seq2seq.py", line 425, in batch_act
    predictions, text_cand_inds = self.predict(xs, ys, cands)
  File "/home/user/projects/ParlAI/parlai/agents/seq2seq/seq2seq.py", line 218, in predict
    xes = self.lt(xs).t()
  File "/home/user/packages/anaconda3/lib/python3.6/site-packages/torch/autograd/variable.py", line 729, in t
    raise RuntimeError("t() expects a 2D Variable, but self is {}D".format(self.dim()))
RuntimeError: t() expects a 2D Variable, but self is 3D 
```

alexholdenmiller:
thanks for flagging, I'll update

PR #272, opened 2017-08-22
Make truncation an option, fix runtime error from updated pytorch

alexholdenmiller:
fix pytorch bug reported in #270
default to truncating long inputs, but can turn this off with -tr false

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -49,6 +49,9 @@ class Seq2seqAgent(Agent):
             help='rank candidates if available. this is done by computing the' +
                  ' mean score per token for each candidate and selecting the ' +
                  'highest scoring one.')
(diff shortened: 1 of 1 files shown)


alexholdenmiller:
(is truncating on a good default? or should we default off? truncation values currently need to be modified manually)

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -49,6 +49,9 @@ def add_cmdline_args(argparser):
             help='rank candidates if available. this is done by computing the' +
                  ' mean score per token for each candidate and selecting the ' +
                  'highest scoring one.')
+        agent.add_argument('-tr', '--truncate', type='bool', default=True,
```
jaseweston: explain what this means in the comment please

jaseweston: like...remove padding or?

alexholdenmiller:
this look okay now, @jaseweston?

PR #273, opened 2017-08-24
Add end-to-end negotiation task

swkrueger:
Add end-to-end negotiation task that loads data from https://github.com/facebookresearch/end-to-end-negotiator.

### DIFF
--- a/parlai/tasks/negotiation/__init__.py
+++ b/parlai/tasks/negotiation/__init__.py
@@ -1,0 +1,5 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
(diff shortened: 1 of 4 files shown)


1nadequacy:
Hey @swkrueger, thanks for doing this!

I took a look at your diff and it seems that you are only providing a way to perform the dialogue part of negotiation, since I wasn't able to find the selection part (when the agents make a selection at the end of the dialogue). Is this expected or I'm missing something? 

Ideally we want it to be similar to openai gym, where at there very end of an episode you are getting a reward.

swkrueger:
This is only the teacher. The learner as well as the functionality for reinforced learning (e.g. rewards) will follow later. The teacher will train the agent to output their selection at the end of an episode after the selection token. For the simulated version, a special World will ensure that this selection is not communicated to the other party.

1nadequacy:
Makes sense, thanks for the explanation!

alexholdenmiller:
Thanks for checking @1nadequacy! We'll work on the simulated version after adding the model.

PR #274, opened 2017-08-25
Fixed apostrophe and underscore normalizing (I'm vs I 'm)

belbs:


alexholdenmiller:
(This fixes an issue where the label contains "I'm" but the model produces "I 'm" and is marked incorrect. We couldn't think of a case where removing apostraphes or underscores was better than replacing them with a space.)

PR #275, opened 2017-08-25
Added generative output to memnn agent

belbs:
Added possibility to generate outputs from the MemNN model (currently using a GRU). The --output parameter controls if 'rank' or 'generate' is used. --dropout controls dropout rate for the RNN training and --rnn-layers controls the numlayers in the GRU.


belbs:
@jaseweston I just put the bAbI results on the baselines spreadsheet, but in sum, yes it's performing the same expect much better on task 19 (100%)

PR #278, opened 2017-08-29
StringMatchRetriever

phecdarain:
Adding the string matching retriever.

Test: python tests/test_retrieve.py TestStringMatchRetriever.test_retriever


alexholdenmiller:
Good start!

For the frequency weighting, use the frequencies from the dictionary (DictionaryAgent.freqs()) instead of from the tokens the agent sees. Let's set the default to using this to sort results, rather than defaulting to random.

Throw out stopword tokens and punctuation as well--you can see something similar in the Metrics class in parlai/core/metrics.py in the function `_normalize_answer`, but you'll want to include more stopwords than just those articles as well. For example, you could use `nltk.corpus.stopwords.words('english')` which contains 153 english stopword tokens.

And as Jason reminded, you'll need to add saving/loading as well, like the dictionary has.

parlai/agents/ir_baseline/ir_retrieve.py
```diff
@@ -0,0 +1,123 @@
...
+        print('StringMatchRetriever: %d facts with %d tokens loaded.'
+              % (len(self.facts), len(self.token2facts)))
+
+    def save(self, filename=None):
+        """Save StringMatchRetriever to file.
+        Format is:
+            the first line: 'fact1<TAB>fact2<TAB>...'
+            starting from second line: 'token<TAB>fact_index1<TAB>fact_index2...'
+        """
+        filename = self.opt['model_file'] if filename is None else filename
```
alexholdenmiller: oops. this needs its own command line arg to specify the file name.

alexholdenmiller:
@jaseweston we realized a bunch of the functionality was shared with the ir_baseline model, so put it in the same folder for now and moved shared functions out to a new file

jaseweston:
shall we close this one @alexholdenmiller ?

alexholdenmiller:
Yes, thanks! replacing with #477

PR #281, opened 2017-08-31
Add deal-no-deal selection agent.

swkrueger:
The agent will, given a dialogue of a negotiation between himself and an opponent, output what it thinks the agreed decision was.


alexholdenmiller:
sorry for the delay, looking now!

alexholdenmiller:
overall looks good! I got a few runtime errors due to pytorch updates (I think), committed a few of them locally and I'm testing out training on my GPUs

alexholdenmiller:
I realized gpu training is disabled, merging for now and we can update it when we get the chance with the authors

PR #288, opened 2017-09-07
Creating a script to delete orphan HITs

JackUrb:
When the MTurk run script dies in a fatal error and the HITs are not deleted, this cleanup script can be run to find and remove orphan scripts without having to manually do so.


JackUrb:
Tested by running in sandbox and live and killing HITs that were orphans. Potential advancement would be a version that determines if a HIT is an orphan automatically instead of requiring user approval to delete (to ensure you're not killing an active task that you don't want to kill).

PR #309, opened 2017-09-19
squad indices / handwritten teacher update

alexholdenmiller:
Add support for batching and metrics to squad handwritten teacher, and add answer indices

### DIFF
--- a/parlai/tasks/squad/agents.py
+++ b/parlai/tasks/squad/agents.py
@@ -15,62 +15,116 @@ import os
 
 class HandwrittenTeacher(Teacher):
     """Hand-written SQuAD teacher, which loads the json squad data and
-    implements its own `act()` method for interacting with student agent, rather
-    than inheriting from the core Dialog Teacher. This code is here as an
-    example of rolling your own without inheritance.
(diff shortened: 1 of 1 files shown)


parlai/tasks/squad/agents.py
```diff
@@ -15,62 +15,116 @@
...
             self.epochDone = True
 
-        return {
+        action = {
+            'id': 'squad',
             'text': context + '\n' + question,
-            'labels': answers,
-            'episode_done': True
+            'episode_done': True,
+            'answer_starts': answer_starts
```
jaseweston: hmm..slightly weird name but i guess that's what they called it... why plural though?

alexholdenmiller: one start for each label :) (there are several labels for every question)

alexholdenmiller:
some of this functionality should maybe be moved to a "SupervisedTeacher" or "IndexedTeacher" which could be a parent of DialogTeacher--then people could inherit the batch indexing

alexholdenmiller:
at least now this class is a more detailed example

PR #314, opened 2017-09-20
Add 'same' as a dec option for s2s, fix runtime error on GPUs

alexholdenmiller:


alexholdenmiller:
Merged to fix runtime error

PR #317, opened 2017-09-21
add optim options to seq2seq and allow different embdim vs hsz

alexholdenmiller:
adds linear layers (one before enc, one before dec) between the lookup table and the encoding / decoding layers


alexholdenmiller:
also adds dropout on embeddings

PR #319, opened 2017-09-22
bAbI+ dataset integration

aleSuglia:
# bAbI+ integration

## Description
The aim of the present pull-request is to add a new world in ParlAI based on the recently proposed dataset *bAbI+*. *bAbI+* is an extension of the bAbI Task 1 dialogues with everyday incremental dialogue phenomena (hesitations, restarts, and corrections) which model the disfluencies and communication problems in everyday spoken interaction in real-world environments. See [[1]](https://www.researchgate.net/publication/319128941_Challenging_Neural_Dialogue_Models_with_Natural_Data_Memory_Networks_Fail_on_Incremental_Phenomena), [[2]](http://aclweb.org/anthology/D17-1235) for additional information about the dataset.

## Contribution
In the following we report the main changes to the ParlAI codebase:
1. Added to the .gitignore Intellij IDEA related files;
2. Created dialog bAbI+ world which supports **only** the task API calls;
3. Added reference to dialog bAbI+ to the tasks list.

## Authors
- Prof. Oliver Lemon ([https://sites.google.com/site/olemon/](https://sites.google.com/site/olemon/))
- Igor Shalyminov (@ishalyminov)
- Alessandro Suglia (@aleSuglia)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,5 @@ tmp/
 *.pickle
 # Heroku things
 heroku-cli-*
+
+.idea
--- a/parlai/tasks/dialog_babi_plus/__init__.py
+++ b/parlai/tasks/dialog_babi_plus/__init__.py
@@ -1,0 +1,3 @@
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
--- a/parlai/tasks/dialog_babi_plus/agents.py
+++ b/parlai/tasks/dialog_babi_plus/agents.py
@@ -1,0 +1,36 @@
+import os
+
+from parlai.core.fbdialog_teacher import FbDialogTeacher
+from parlai.tasks.dialog_babi_plus.build import build
+
+tasks = {}
+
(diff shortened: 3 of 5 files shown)


aleSuglia:
Thank you for your reviews. I'm glad that it works well and that it is nicely integrated in ParlAI.

Do you think that we should provide an implementation for the TaskTeacher as well? Ideally, by using the tool written by @ishalyminov [bAbI tools](https://github.com/ishalyminov/babi_tools), it is possible to generate dialogues for each task in the original dialog bAbI. So in the future, it might be useful to support this use case. What do you think?

alexholdenmiller:
Ah I see! Then for now let's at least add a DefaultTeacher that defaults to task 1--this way someone can do `dialog_babi_plus` in the meantime without needing to spell out the whole task. (It could be written as a child of TaskTeacher, which sets the task to 1)

aleSuglia:
In my latest commits I have implemented the *DefaultTeacher* which defaults to task 1. Is it ok?

alexholdenmiller:
Pardon me -- I didn't notice the commits. Looks good, thanks!

PR #320, opened 2017-09-23
Bidirectional encoding seq2seq

klshuster:
Adds functionality for bidirectional encoding in the seq2seq model. In addition, builds out the attention option to include "max-pool," which does an element-wise max of the final encoder output before applying attention.  Also fixes an issue with LSTMs and attention, via unwrapping the "hidden" tuple. 

Ensured that the model can still be trained via running train_model.py with the seq2seq model and the appropriate command line options set


alexholdenmiller:
(so none of your code will change, just removing some currently-unchanged code)

additionally, we should modifying the way the "attention" on the max-pool state is used. I guess it's not really attention--it would be more accurate to say it's "context" that you're paying attention to at every state. so maybe just changing the name is right.

klshuster:
After removing 'same' from the decoder options, should the default option then be 'shared,' or 'gru' like the encoder?

When you say changing the name, do you mean simply changing 'max-pool' to something like 'max-pool-context', or should the whole 'attention' name should change to something like 'context'?

alexholdenmiller:
Merged some of these manually into my new seq2seq branch, I'll merge more after that one has landed--a lot has changed in the seq2seq file...

alexholdenmiller:
changes have been incorporated directly in another branch, so closing this out

PR #321, opened 2017-09-23
Fixed that the episode_done always be False in task VisDial

wolegechu:


alexholdenmiller:
Thanks for flagging! I wasn't sure exactly what was wrong so I was experimenting, I think #327 is a little cleaner so I'm going to land that one in lieu of this.

PR #322, opened 2017-09-25
Add MTurk interface to negotiation task

mikelewis0:


alexholdenmiller:
@JackUrb can you take a look at this as well?

alexholdenmiller:
perused the code, looked good to me.

then I ran
```sh
git fetch origin
git checkout -b negotiation origin/negotiation
git merge master
cd parlai/mturk/tasks/dealnodeal/
python run.py
```

Here it is:
screenshot 2017-09-26 15 15 11


Looks good, except it looks like the images didn't load for the items...

alexholdenmiller:
If you want, you could change run.py to allow the user to specify the model they want to use from the commandline, e.g. "-m seq2seq -mf /tmp/model" would launch the seq2seq model and tell it to load the model file at /tmp/model.

The `create_agent` function (e.g. https://github.com/facebookresearch/ParlAI/blob/master/examples/train_model.py#L115) checks this `--model` param. You could use LocalHumanAgent `if 'model' not in opt` and `create_agent(opt)` otherwise.

parlai/mturk/tasks/dealnodeal/run.py
```diff
@@ -0,0 +1,85 @@
...
+    argparser.add_mturk_args()
+    opt = argparser.parse_args()
+    opt['task'] = 'dealnodeal'
+    opt['datatype'] = 'valid'
+    opt.update(task_config)
+
+    mturk_agent_1_id = 'mturk_agent_1'
+    mturk_agent_2_id = 'mturk_agent_2'
+    human_agent_1_id = 'human_1'
+    mturk_agent_ids = [mturk_agent_1_id] #, mturk_agent_2_id]
```
JackUrb: The hardcoded/commented option of using human agents vs using two mturk agents isn't as good form as it could be. Any chance this can be done programmatically and an additional flag can be added to the parser?

Can use `parser.add_argument('--use_human', dest='use_human', type=bool, default=False, action='store_true')` and then select the ids based on whether or not `opt['use_human']` is set.

alexholdenmiller: @mikelewis0 assuming this is just for other people to evaluate their models, you could do what I commented at the top-level.

JackUrb: In which case I'd remove the unused second turk agent and comment. If you want them to be able to collect more data if they'd want to, you can do both as well (as they don't conflict)

parlai/mturk/tasks/dealnodeal/worlds.py
```diff
@@ -0,0 +1,82 @@
...
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+from parlai.mturk.core.worlds import MTurkTaskWorld
+from parlai.core.worlds import validate
+from joblib import Parallel, delayed
+from parlai.tasks.dealnodeal.agents import NegotiationTeacher
+import random
+
+class MTurkDealNoDealDialogWorld(MTurkTaskWorld):
+    """Basic world where each agent gets a turn in a round-robin fashion,
```
JackUrb: Can this comment (and the comment for parley) be updated to describe what DealNoDeal is doing?

JackUrb:
Other than the image issue and my 2 comments this is looking quite good to me!

mikelewis0:
Thanks a lot for the feedback! I've pushed some fixes.

parlai/mturk/tasks/dealnodeal/worlds.py
```diff
@@ -0,0 +1,79 @@
...
+        """Alternate taking turns, until both agents have made a choice (indicated by a turn starting with <selection>)
+        """
+        if self.first_turn:
+            self.num_negotiations += 1
+
+            # Use NegotiationTeacher to load data for us
+            self.task.dialogue_idx = None
+            act = self.task.act()
+            for other_agent in self.agents:
+                other_agent.observe(validate(act))
```
JackUrb: It would appear that the NegotiationTeacher is giving both of the agents the same values for both items here. Shouldn't there be something that provides different values to each of the agents while maintaining the same counts?

mikelewis0:
Thanks Jack - great catch! I've added a fix.

mikelewis0:
Awesome, thanks both of you for the reviews!

PR #323, opened 2017-09-26
Add dict-build-first option to eval_model.py

jojonki:
I added an option for using trained `model file`.

### Test 

```
// train and save memnn model with babi task
$ python examples/train_model.py -m "memnn" -t "babi:task1k:1" -e 10 -bs 32 --gpu 0 -mf /tmp/model_babi_task1k_1


// eval memnn without the model_file
$ python examples/eval_model.py -m "memnn" -t "babi:task1k:1"
...
{'total': 100, 'accuracy': 0.14, 'f1': 0.14, 'hits@k': {1: 0.14, 5: 0.78, 10: 1.0, 50: 1.0, 100: 1.0}}


// eval memnn with the model_file
$ python examples/eval_model.py -m "memnn" -t "babi:task1k:1" -mf /tmp/model_memnn_babi_task1k_1
...
{'total': 100, 'accuracy': 1.0, 'f1': 1.0, 'hits@k': {1: 1.0, 5: 1.0, 10: 1.0, 50: 1.0, 100: 1.0}}
```

### DIFF
--- a/examples/eval_model.py
+++ b/examples/eval_model.py
@@ -14,6 +14,7 @@ or
 from parlai.core.params import ParlaiParser
 from parlai.core.agents import create_agent
 from parlai.core.worlds import create_task
+import build_dict
 
 import random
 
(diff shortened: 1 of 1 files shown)


examples/eval_model.py
```diff
@@ -24,8 +25,16 @@ def main():
     parser = ParlaiParser(True, True)
     parser.add_argument('-n', '--num-examples', default=100000000)
     parser.add_argument('-d', '--display-examples', type='bool', default=False)
+    parser.add_argument('-dbf', '--dict-build-first', type='bool', default=True)
```
jaseweston: a dictionary should already be built for a trained model?

jojonki: Yes, I thought it was useful. And the code was copied from `train_model.py`.

alexholdenmiller: I think Jason was referring to the fact that a pre-trained model should have already built a dictionary, which was used to train it. You should be loading that dictionary instead of building a new one, because if the new one isn't exactly identical, then the word embeddings stored by the model become meaningless. Thus, we don't necessarily want to build a new one for you here--we should make you go find the one that is saved already.

alexholdenmiller:
I think we don't want to include this--see our inline comments. I see where you were coming from--thanks for making the pull request.

PR #324, opened 2017-09-26
Adds VCRNN support to seq2seq

SsnL:
Commit 1: Adds function call on every new training epoch to support increasing lambda per epoch for VCRNN. This can also be used for updating lr.

Commit 2: Adds enc and dec train/eval mode transition in `seq2seq.py` so that 

1. in `_decode_and_train`, enc and dec are in train mode.
2. in `_score_candidates` and `_decode_only`, enc and dec are in eval mode.

Commit 3: Adds single-layer VCRNN class implemented in PyTorch autograd.

Commit 4: Integrated VCRNN in seq2seq:

1. For VCRNN, the parameters for `m` regularization loss is arbitrarily set as 0.6*|m -0.4|. This is enforced in `_decode_and_train` of `seq2seq.py`.
2. Adds checks in multiple places of `seq2seq.py` to account for `m` values returned as part of hidden state.

Test:
`python3 examples/train_model.py -m seq2seq -mf /tmp/mf -t babi:task10k:1 -enc * -dec * -nl 1` for 
1. both `vcrnn`
2. enc `vcrnn`, dec `lstm`
3. enc `lstm`, dec `vcrnn`

Also tested eval mode with same three cases.

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -127,6 +127,7 @@ def main():
     impatience = 0
     saved = False
     valid_world = None
+    epoch = 0
     while True:
         world.parley()
         parleys += 1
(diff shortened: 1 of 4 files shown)


SsnL:
Minimized VCRNN-specific changes to seq2seq while being as general as I can. The resulting compromise is the `get_last_forward_loss` function.

alexholdenmiller:
still might add this at some point :)

SsnL:
That would be great. :) I think we are figuring out better ways to write custom RNN. (Currently it is a mess as you can see in this PR.) I'll let you know if there is considerble progress.

PR #329, opened 2017-09-27
updates to seq2seq model

alexholdenmiller:
1) update start and end tokens, so every utterance starts with __START__ and ends with __END__
2) fix context history, remembering past answers / labels
3) remove linear layer in between emb and enc/dec, instead just change enc/dec input dim
4) during training & with no attention, do teacher forcing on the entire sequence at once instead of one token at a time
5) when ranking candidates & with no attention, process the whole sequence at once instead of one token at a time

still todo: allow different LT for encoding and decoding, and allow sharing the decoding LT with the linear output layer


alexholdenmiller:
made new branch since this one got messy

PR #335, opened 2017-10-04
Slightly better message handling

JackUrb:
Some error states would pollute the logs in the past, this reduces the rate at which those errors happen and properly silences them when they do


JackUrb:
Ran and passed MTurk tests

PR #340, opened 2017-10-09
fix #339: resolve return ambiguity

### LINKED ISSUE #339: Return type ambiguous when extracting image features
The return type of the extracted image features (if the features file is present or not) is different. If the file is present then it returns a numpy.ndarray type object otherwise it returns a torch.autograd.variable.Variable object.
( https://github.com/facebookresearch/ParlAI/blob/3d86ccdbb4d87002cc6c4782afd0ee5277e742f1/parlai/core/image_featurizers.py#L149 )

backpropper:
fix #339


alexholdenmiller:
thanks!

PR #341, opened 2017-10-11
Adding new teacher for VQA task

backpropper:
Adding AllTeacher which gives access to all the information in the data.


jaseweston:
"Adding AllTeacher which gives access to all the information in the data."
-- can you clarify what the other teachers do in contrast  in the comment? thanks

jaseweston:
seems fine to me, @alexholdenmiller ?

PR #349, opened 2017-10-16
Small fix to Fairseq

hadasah:
1. Added check in save() to see it's a shared instance
2. Changed paper reference at the top to Gehring, et al.


alexholdenmiller:
thanks!

PR #356, opened 2017-10-18
update truncation and remove loss for NULL, clean up dictionary, switch to Xentropy

alexholdenmiller:


alexholdenmiller:
merging for saizheng

PR #360, opened 2017-10-19
allowed fixed glove emb, never predict NULL, option to truncate targets again

alexholdenmiller:


alexholdenmiller:
merging for saizheng

PR #364, opened 2017-10-23
Add parameter sweep for bAbI 10k task

klshuster:
Script for running a parameter sweep for bAbI 10k task that reproduces results from https://arxiv.org/abs/1503.08895 for tasks 1-3

### DIFF
--- a/projects/babi/memnn/memnn_bAbI_10k_task_parameter_sweep.sh
+++ b/projects/babi/memnn/memnn_bAbI_10k_task_parameter_sweep.sh
@@ -3,15 +3,14 @@
 # Reproduce results from https://arxiv.org/abs/1503.08895 bAbI tasks 1-3 with
 # 10k training examples 
 
-PARLAI=/path/to/ParlAI                 # Where ParlAI exists in filesystem
+PARLAI=../../..                # Where ParlAI exists in filesystem
 
 DATE=`date +"%Y%m%d"`
 SWEEP_NAME=memnn_bAbI_task10k_t1to3_parameter_sweep  # Sweep name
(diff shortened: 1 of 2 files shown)


projects/babi/memnn/memnn_bAbI_10k_task_parameter_sweep.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/sh
+
+# Reproduce results from https://arxiv.org/abs/1503.08895 bAbI tasks 1-3 with
+# 10k training examples 
+
+PARLAI=/path/to/ParlAI                 # Where ParlAI exists in filesystem
```
jaseweston: is there a way this can default to knowing where it is? we are in the directory after all?

klshuster:
Yes I believe we can - we could either specify a SAVE_ROOT ourselves (e.g. /checkpoint/$USER/${DATE}/${SWEEP_NAME}) or let someone specify it via a command-line arg to the script.

In terms of the ParlAI path, if one were to run the script as is within this repo, then we could simply specify the path to be "PARLAI=../../..", if that's what you mean; alternatively, if one wanted to copy the script and run it from somewhere else, that is where they would specify the ParlAI path

jaseweston:
/tmp for saving as default or something? something anyone would have..?

yes "PARLAI=../../.." or default or something

i want when someone runs it there's chances it will "just work", no messing
about involved

klshuster:
Ok, /tmp should work. If we do that, then this should in theory "just work", as long as someone has the right dependencies installed and modules loaded for running pytorch

PR #368, opened 2017-10-25
Add ConvAI_ChitChat task support

bavadim:
This is the data collected in Moscow Institute of Physics and Technology as a part of the ConvAI (competition of chatbots). This task enables to learn chitchat models on this data. For more information about the dataset see  http://convai.io/data/.


alexholdenmiller:
Hi @bavadim,

Thanks for the PR! Question for you: the way we've been formatting conversational datasets has been to place the agent on one "side" of the conversation and then the model on the other.

That is, in the conversation:
Alice: Hello there!
Bob: Hey, what's up?
Alice: Nothing much. You?
Bob: Same.

we would create two examples:
```
'text': 'hello there!', 'labels': ['Hey, what's up?']
'text': 'Nothing much. You?', 'labels': ['Same.']
```

This way, the agent might remember its history, and all incoming messages are always from the other party.

Your dataset is currently set up as follows:
```
'text': None, 'labels': ['hello there!']
'text': 'hello there!', 'labels': ['Hey, what's up?']
'text': 'Hey, what's up?', 'labels': ['Nothing much. You?']
'text': 'Nothing much. You?', 'labels': ['Same.']
```
So the agent here is learning both to respond to the other agent and to itself in the same episode.

The way we've captured this in other datasets is to actually create two different episodes instead: one from Alice's perspective, and a separate one from Bob's perspective. Can we do that here? I can help to set that up.

As an aside, there's also a small bug here--I can send you a fix for that.

PR #380, opened 2017-10-31
dialogue_qe task support

bavadim:
This is the data collected in Moscow Institute of Physics and Technology as a part of the ConvAI (competition of chatbots). This task enables to learn dialogue-level metric for dialogue systems . For more information about the dataset see http://convai.io/data/.


alexholdenmiller:
would it make more sense for this data to be at convai_chitchat:qe--that is, to be another teacher within that some task?

bavadim:
Hi! We think of this as another task, since this task is about estimation of dialogue as a whole similarity to natural dialogues between humans. A bit later we'll make another pull-request with a task to estimate utterance similarity to human utterances. But we can merge these tasks. It's up to you.

alexholdenmiller:
Hi sorry for the delay here, I'll merge this now.

We'd like for tasks to have instructions such that a human could do them if given the task independently, e.g. prepending something similar to "Please provide a score for Alice and Bob in the following conversation:"

However, we can do that in a follow-up. We have similar messages in other tasks which we would like to migrate to a separate "Teacher" message so that it's easy to strip away before passing the remaining text to the model, so that the model doesn't get "distracted" during training if you don't want to keep that text.

PR #381, opened 2017-10-31
Move seq2seq to use a modular format, other updates

alexholdenmiller:


alexholdenmiller:
also add reverse-direction conversations from opensubtitles and add preprocessing to OS dataset

PR #383, opened 2017-11-01
Threadpool for image loading in vqa v1

klshuster:
1. Adds a thread pool for loading images for VQA v1.
2. Switches saving resnet features from numpy pickles to hdf5 files
3. Converts lru_cache in image_featurizers to a first_n cache of size 84k (around 32gb for hdf5 features)
4. Allows numthreads and batchsize to both be specified and lead to creation of a BatchWorld rather than HogwildWorld.

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -827,7 +827,7 @@ def create_task(opt, user_agents):
     # Single threaded or hogwild task creation (the latter creates multiple threads).
     # Check datatype for train, because we need to do single-threaded for
     # valid and test in order to guarantee exactly one epoch of training.
-    if opt.get('numthreads', 1) == 1 or 'train' not in opt['datatype']:
+    if opt.get('numthreads', 1) == 1 or 'train' not in opt['datatype'] or opt.get('batchsize', 1) > 1:
         if ',' not in opt['task']:
             # Single task
             world = create_task_world(opt, user_agents)
--- a/parlai/tasks/vqa_v1/agents.py
+++ b/parlai/tasks/vqa_v1/agents.py
@@ -11,6 +11,9 @@ from .build import build, buildImage
 import json
 import random
 import os
+from threading import Thread
+import queue
+import concurrent.futures
 
 
 def _path(opt):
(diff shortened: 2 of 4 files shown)


parlai/core/worlds.py
```diff
@@ -827,7 +827,7 @@ def create_task(opt, user_agents):
     # Single threaded or hogwild task creation (the latter creates multiple threads).
     # Check datatype for train, because we need to do single-threaded for
     # valid and test in order to guarantee exactly one epoch of training.
-    if opt.get('numthreads', 1) == 1 or 'train' not in opt['datatype']:
+    if opt.get('numthreads', 1) == 1 or 'train' not in opt['datatype'] or opt.get('batchsize', 1) > 1:
```
jaseweston: maybe we need to explain the interplay of numthreads and batchsize here (as i dont understand myself)

parlai/tasks/vqa_v1/agents.py
```diff
@@ -44,6 +47,24 @@ def _path(opt):
     return data_path, annotation_path, image_path
 
 
+class MasterLoader(Thread):
```
jaseweston: so if this stuff works, we move it to other tasks as well, or what?

parlai/core/worlds.py
klshuster: Basically, one can specify the amount of threads in the threadpool with numthreads; if they set a batchsize greater than 1, we should construct a BatchWorld rather than a HogwildWorld

parlai/tasks/vqa_v1/agents.py
klshuster: Yes, Alex and I spoke about extending this to a new parent teacher that would allow a subclass to make use of this

parlai/core/worlds.py
jaseweston: make comment a bit more explicit? also is it clear in parlai params?

klshuster: I'll make sure the comment is more explicit and that the parlai params are more clear in another PR

jaseweston:
thanks

PR #387, opened 2017-11-02
Clarify batchworld and params comments

klshuster:


parlai/core/worlds.py
```diff
@@ -827,6 +827,8 @@ def create_task(opt, user_agents):
     # Single threaded or hogwild task creation (the latter creates multiple threads).
     # Check datatype for train, because we need to do single-threaded for
     # valid and test in order to guarantee exactly one epoch of training.
+    # If batchsize is > 1, default to BatchWorld, regardless of numthreads, as
+    # numthreads may be used for a threadpool in BatchWorld.
```
alexholdenmiller: "used for multithreaded data loading" is mos re clear I think

jaseweston:
but couldn't it be used for other multithreaded stuff in the long-term even
during batch mode, or no?
it might be better to write for general case (and give example)

klshuster:
so something like "If batchsize > 1, default to BatchWorld, as numthreads can be used for multithreading in batch mode, e.g. multithreaded data loading"?

jaseweston:
sounds good to me..alex?

alexholdenmiller:
we may want to also just add another parameter--'hogwild' :)

PR #388, opened 2017-11-06
Adding nlvr

debajyotidatta:
Hi, I am collaborating with @yoavartzi (Dr. Yoav Artzi) to add the NLVR dataset to ParlAI. 

I ran `python examples/display_data.py -t nlvr ` and that works fine. Are there specific tests that one can run to ensure that a task has been added correctly?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,3 +114,7 @@ heroku-cli-*
 # IntelliJ IDEA
 .idea
 *.iml
+
+# VSCode
+
+*.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ __pycache__/
 
 # Distribution / packaging
 .Python
+.vector_cache/
 env/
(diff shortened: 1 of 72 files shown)


alexholdenmiller:
Overall looks good, except that using `-dt valid` gives an "index out of range" error: your `image_file_names` list is empty.

Also, can we add a prefix to the questions: "True or False: "
We like tasks to be clear to a human doing the tasks as much as possible, and a model could safely ignore that prefix.

You said "WIP"--are there any other changes planned?

debajyotidatta:
No other changes are planned as part of this pull request. I added removed the WIP label and fixed the dt valid that you mentioned.

debajyotidatta:
@alexholdenmiller : Any updates on this?

alexholdenmiller:
pardon me--I had missed the update notification. looks great!

PR #390, opened 2017-11-07
Fixed index teacher

klshuster:
Eliminate redundancy in VQA, SQuAD, and Dialog teachers by consolidating into one superclass, "FixedDataTeacher", which captures shared functionality amongst teachers that present "fixed data," i.e. data that is not dynamically generated during training.

### DIFF
--- a/parlai/core/dialog_teacher.py
+++ b/parlai/core/dialog_teacher.py
@@ -85,7 +85,7 @@ class DialogTeacher(FixedDataTeacher):
         num_eps = self.data.num_episodes()
         if not self.stream:
             if self.episode_done:
-                self.episode_idx, self.epochDone = self.next_episode_idx(num_eps=num_eps)
+                self.episode_idx, epoch_done = self.next_episode_idx(num_eps=num_eps)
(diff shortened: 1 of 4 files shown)


parlai/core/dialog_teacher.py
```diff
@@ -96,37 +81,21 @@ def label_candidates(self):
...
-
-        if self.random:
-            epoch_done = False
-        elif (self.episode_idx + self.step_size >= num_eps and
-                action['episode_done'] and not self.stream):
+        if not self.random and action['episode_done'] and not self.stream:
             # this is used for ordered data to check whether there's more data
             epoch_done = True
-        return action, epoch_done
+        return action, self.epochDone or epoch_done
```
alexholdenmiller: is this right? maybe the `self.epochDone =` should be `epoch_done =`

klshuster: Yeah you're right, that's what it should be

alexholdenmiller:
@klshuster feel free to merge when you're ready!

PR #392, opened 2017-11-09
add iwslt14, various performance improvements

alexholdenmiller:


parlai/tasks/task_list.py
```diff
@@ -277,6 +277,13 @@
         "description": "Workshop on Machine Translation task, currently only includes en_de."
     },
     {
+        "id": "IWSLT14",
+        "display_name": "IWSLT14",
+        "task": "iwslt14",
+        "tags": ["All", "MT"],
+        "description": "2014 International Workshop on Spoken Language task, currently only includes en_de and de_en. From Cettolo et al. '12. Link: wit3.fbk.eu"
```
jaseweston: is it ...an MT task? can you try to explain it a little more?

PR #395, opened 2017-11-15
Faster image cpu loading

klshuster:
If cuda is available, load extracted image features into the gpu before returning them. This improves training time in models using extracted image features

### DIFF
--- a/parlai/core/image_featurizers.py
+++ b/parlai/core/image_featurizers.py
@@ -203,5 +203,5 @@ class ImageLoader():
                     hdf5_file = h5py.File(new_path, 'r')
                     feature = hdf5_file['feature'].value
                 if self.use_cuda:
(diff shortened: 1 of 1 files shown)


parlai/core/image_featurizers.py
```diff
@@ -167,7 +170,7 @@ def img_to_ascii(self, path):
             asc.append('\n')
         return ''.join(asc)
 
-    @first_n_cache
+    # @first_n_cache
```
alexholdenmiller: did you mean to disable the cache?

klshuster: yes, as it was caching the images in GPU which led to cuda OOM errors; though now that I think about it, I could just cache the cpu version, as loading those from cache is still faster than loading from disk. But this would probably involve moving the cuda logic to the cache, rather than the load/extract methods

PR #396, opened 2017-11-15
move convai project folder

alexholdenmiller:


alexholdenmiller:
@bavadim heads up --we're moving the folder

PR #397, opened 2017-11-15
Fixed data teacher reset

klshuster:


dehuachen:
This change makes `python examples/display_data.py -t babi:task1k:1` failed,
causing: AttributeError Exception, at "ParlAI/parlai/core/dialog_teacher.py", line 60, in reset
    if self.stream:
AttributeError: 'Task1kTeacher' object has no attribute 'stream'

alexholdenmiller:
Thanks for the heads up @dehuachen I'll push a fix now.

PR #399, opened 2017-11-16
move self.stream to fixed data teacher

klshuster:
fix bug introduced in #397 where `python examples/display_data.py -t babi:task1k:1` failed


alexholdenmiller:
closing since I pushed a different change

Issue #181, opened 2017-06-30
No module named 'torch'


JackZhang1988:
i try to run ParlAi example `python examples/display_data.py -t babi:task1k:1` but i got belwo error:

```
[task:babi:task1k:1]
[download_path:/Applications/My-Project/ParlAI/downloads]
[datatype:train]
[image_mode:raw]
[numthreads:1]
[batchsize:1]
[datapath:/Applications/My-Project/ParlAI/data]
[num_examples:10]
[parlai_home:/Applications/My-Project/ParlAI]
[creating task(s): babi:task1k:1]
Traceback (most recent call last):
  File "examples/display_data.py", line 42, in <module>
    main()
  File "examples/display_data.py", line 30, in main
    world = create_task(opt, agent)
  File "/Applications/My-Project/ParlAI/parlai/core/worlds.py", line 806, in create_task
    world = create_task_world(opt, user_agents)
  File "/Applications/My-Project/ParlAI/parlai/core/worlds.py", line 778, in create_task_world
    world_class, task_agents = _get_task_world(opt)
  File "/Applications/My-Project/ParlAI/parlai/core/worlds.py", line 773, in _get_task_world
    task_agents = _create_task_agents(opt)
  File "/Applications/My-Project/ParlAI/parlai/core/agents.py", line 371, in _create_task_agents
    my_module = importlib.import_module(module_name)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 978, in _gcd_import
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load
  File "<frozen importlib._bootstrap>", line 950, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 655, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 205, in _call_with_frames_removed
  File "/Applications/My-Project/ParlAI/parlai/tasks/babi/agents.py", line 7, in <module>
    from parlai.core.fbdialog_teacher import FbDialogTeacher
  File "/Applications/My-Project/ParlAI/parlai/core/fbdialog_teacher.py", line 42, in <module>
    from .dialog_teacher import DialogTeacher
[...]

PR #402, opened 2017-11-17
remove torch import when not featurizing images

alexholdenmiller:
@klshuster can you double check this doesn't cause any errors in your pipeline?

### DIFF
--- a/parlai/core/image_featurizers.py
+++ b/parlai/core/image_featurizers.py
@@ -11,7 +11,6 @@ import h5py
 from PIL import Image
 from functools import wraps
 from threading import Lock, Condition
-import torch
 
 _greyscale = '  .,:;crsA23hHG#98&@'
 _cache_size = 84000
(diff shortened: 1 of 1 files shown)


alexholdenmiller:
re-fix for #181

parlai/core/image_featurizers.py
```diff
@@ -71,6 +70,7 @@ def init_cnn(self):
         """Lazy initialization of preprocessor model in case we don't need any image preprocessing."""
         try:
             import torch
+            self.torch = torch
```
klshuster: We would need to change line 63 to be `self.use_cuda = False` and then move `self.use_cuda = not opt.get('no_cuda', False) and torch.cuda.is_available()` to this function (or at least after this import) but otherwise looks good

alexholdenmiller: oops thanks good  catch

PR #403, opened 2017-11-17
Move threadpool to fixed data teacher

klshuster:
Moves the threadpool to fixed data teacher to provide threadpool data loading to any subclass of the fixed data teacher

### DIFF
--- a/parlai/core/fixed_data_teacher.py
+++ b/parlai/core/fixed_data_teacher.py
@@ -17,10 +17,11 @@ class DataLoader(Thread):
     A teacher may submit a request to the loader, which will return the
     appropriate data.
 
-    To submit a request, a teacher must provide the following:
-        - a receive function (for receiving the data)
-        - a load function (for loading the data), e.g. ImageLoader.load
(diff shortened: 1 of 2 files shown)


alexholdenmiller:
cool!

so I notice that it's still not very clear how to take advantage of this apart from copying code from VQA--it's not actually used by default from code within FixedData, right?

could we add in a basic example of how to use it in the docstring, in addition to the descriptions that you wrote?

alexholdenmiller:
maybe we could include a parent implementation of the VQA image request?

klshuster:
I can add an example to the teacher's docstring; by parent implementation, do you mean adding an implementation of the image request in VQA to the FixedDataTeacher

jaseweston:
is this stuff going to work seamlessly  in the multitask case?

klshuster:
from my initial testing, it seems to work in the multitask case

jaseweston:
sweet!

alexholdenmiller:
it will currently spawn n_threads per task--@klshuster we should try to guarantee only one threadpool is instantiated

PR #404, opened 2017-11-17
 move more functionality to fixed data teacher

alexholdenmiller:
this way subclasses can essentially just inherit from fixed data, define `get` function, and start rolling--they can inherit `act` and `next_example`

### DIFF
--- a/parlai/core/fixed_data_teacher.py
+++ b/parlai/core/fixed_data_teacher.py
@@ -74,7 +74,7 @@ class FixedDataTeacher(Teacher):
            and computes the image path for the load request.
         3. At the end of ``submit_load_request``, the teacher calls
            ``self.data_loader.request_load`` with three args:
-           - ``self.receive`` - the function that the DataLoader calls to
+           - ``self.receive_data`` - the function that the DataLoader calls to
                return the the loaded object
            - ``self.image_loader.load`` - the function used to load the image
                from the image path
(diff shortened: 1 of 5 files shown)


alexholdenmiller:
also fixed some bugs in the epochDone marking

parlai/core/fixed_data_teacher.py
```diff
@@ -17,8 +56,35 @@ class FixedDataTeacher(Teacher):
...
+
+    The following is an example of the DataLoader usage in the VQA-V1 teacher.
+
+        1. In the teacher's ``init`` function, the teacher calls its
+           ``submit_load_request`` function to preload an image.
+        2. The ``submit_load_request`` function gets the next ``episode_idx``,
+           and computes the image path for the load request.
+        3. At the end of ``submit_load_request``, the teacher calls
+           ``self.data_loader.request_load`` with three args:
+           - ``self.receive`` - the function that the DataLoader calls to
```
klshuster: this should be ``self.receive_data`` now I believe

parlai/core/image_featurizers.py
```diff
@@ -60,33 +58,35 @@ class ImageLoader():
...
-        opt = self.opt
         self.image_size = opt['image_size']
         self.crop_size = opt['image_cropsize']
         self.datatype = opt['datatype']
         self.image_mode = opt['image_mode']
 
         if self.use_cuda:
             print('[ Using CUDA ]')
-            torch.cuda.set_device(opt.get('gpu', 0))
+            torch.cuda.set_device(opt.get('gpu', opt.get('gpu', -1)))
```
klshuster: does this need to be nested like this? or can it just be ``opt.get('gpu', -1)``?

PR #406, opened 2017-11-22
Fix train loop for streaming data

klshuster:
Fixes training loop so that when streaming data (for more than 1 epoch), the training loop functions correctly. Also, fixes minor bugs in wmt and mutualfriends so that the agents run correctly.


examples/train_model.py
```diff
@@ -128,12 +128,20 @@ def main(parser):
...
+    random = opt.get('datatype', 'train') == 'train'
     with world:
         while True:
             world.parley()
             parleys += 1
 
-            if opt['num_epochs'] > 0 and parleys >= max_parleys:
+            if world.epoch_done():
+                num_epochs += 1
+            elif type(world) is BatchWorld:
```
jaseweston: we shouldnt need a batchworld check inside this loop, how can we avoid this?
this looks pretty horrible and points to something being wrong in our design if required :(

jaseweston: this script should be a simple entry point into parlai, not ridiculously complicated..

examples/train_model.py
```diff
@@ -128,12 +128,20 @@ def main(parser):
     impatience = 0
     saved = False
     valid_world = None
+    num_epochs = 0
+    random = opt.get('datatype', 'train') == 'train'
```
jaseweston: bad variable name

examples/train_model.py
klshuster: I'll move the logic out of here into BatchWorld - sorry about that

examples/train_model.py
```diff
@@ -128,12 +128,17 @@ def main(parser):
...
+    sampling_data = opt.get('datatype', 'train') == 'train'
     with world:
         while True:
             world.parley()
             parleys += 1
 
-            if opt['num_epochs'] > 0 and parleys >= max_parleys:
+            if world.epoch_done():
+                num_epochs += 1
+                world.reset()
```
jaseweston: do you have to reset when an epoch is done? it doesnt keep going round? i cant remember what we settled on..

klshuster: if the world doesn't reset, then the world in the batch that finished the epoch will not see any more examples (it will continue to return epoch_done as true). In addition, resetting puts the batches back at the beginning of the data, so that they may go through it again

examples/train_model.py
```diff
@@ -172,9 +177,11 @@ def main(parser):
                         time_left = min(time_left, other_time_left)
                     else:
                         time_left = other_time_left
-                if time_left is not None:
+                if time_left is not None and max_exs != 0:
                     logs.append('time_left:{}s'.format(math.floor(time_left)))
-
+                if opt['num_epochs'] > 0:
+                    display_epochs = int(total_exs / len(world)) if sampling_data else num_epochs
```
jaseweston: thought we were trying to get rid of using len ?

examples/train_model.py
```diff
@@ -24,7 +24,7 @@
 """
 
 from parlai.core.agents import create_agent
-from parlai.core.worlds import create_task
+from parlai.core.worlds import create_task, BatchWorld
```
jaseweston: we shouldnt need batchworld?

examples/train_model.py
alexholdenmiller: we've always had in in train_model.py on line 125

parlai/core/worlds.py
```diff
@@ -631,6 +631,8 @@ def episode_done(self):
         return False
 
     def epoch_done(self):
+        if not self.random:
+            return any([world.epoch_done() for world in self.worlds])
```
alexholdenmiller: I'm not sure about this.

Before: "return done if all worlds are done"
now: "if we aren't doing random sampling, return done if any world is done"

however, there was an implicit "if we aren't doing random sampling" already, since random sampling never returns epoch_done. so the logic here has been flipped from "all" to "any". I'm worried this will prevent all examples from being covered.

examples/train_model.py
alexholdenmiller: reset can be good for resetting the data, but it also resets the metrics, so we need to make sure we get the train metrics first (currently happening on line 157)

klshuster:
Per discussion with @alexholdenmiller, we now compute the length of the dataset for streaming data exactly once and save in a file for later access; this now works with the current training loop

parlai/core/dialog_teacher.py
```diff
@@ -319,8 +319,17 @@ def share(self):
...
+        if not os.path.isfile(length_file):
+            length = 0
+            for episode in self._read_episode(self.data_loader(self.datafile)):
+                length += len(episode)
+            with open(length_file, 'w') as f:
+                f.write(str(length))
+        else:
+            with open(length_file, 'r') as f:
+                length = int(f.read())
+        return length
```
alexholdenmiller: if we've already loaded the data, then we can also get it just from there first rather than iterating over the data again

alexholdenmiller: nvm

parlai/core/dialog_teacher.py
```diff
@@ -322,7 +322,9 @@ def __len__(self):
         length_file = self.datafile + ".length"
         if not os.path.isfile(length_file):
             length = 0
-            for episode in self._read_episode(self.data_loader(self.datafile)):
+            for episode in self.data:
+                if episode == -1:
+                    break
```
alexholdenmiller: wait--will this prevent it from iterating again? try this with validation data or something. may need to reset after this.

alexholdenmiller:
sure, I hear you.

currently train_model does not support infinite len. this patch is converting stream data from failing to working by looping over the data once only the first time, on data that we know has a length (because it's a child of FixedDataTeacher). this seems like a reasonable price to pay (only once).

separate from this patch to making streaming working immediately, we should upgrade our train_model code to properly support infinite data (of which we don't have any yet).

jaseweston:
On Tue, Nov 28, 2017 at 11:36 AM, Alexander Miller <notifications@github.com
> wrote:

> sure, I hear you.
>
> currently train_model does not support infinite len. this patch is
> converting stream data from failing to working by looping over the data
> once only the first time,
>

I think it used to support until all these revisions using lens, iterators,
etc!

alexholdenmiller:
we've used `len` for num_epochs for as long as we've had train_model though, right? so it should have always broken if num_epochs was set when streaming.

parlai/tasks/dealnodeal/agents.py
```diff
@@ -64,6 +64,9 @@ def __init__(self, opt, shared=None):
 
         self.reset()
 
+    def __len__(self):
+        return len(self.episodes)
+
```
jaseweston: is len episodes or examples?

klshuster: Because each episode has one example, the len for this teacher is len episodes

jaseweston:
dealnodeal is a dialog about making it a deal, it has a bunch of turns in
it?

bash-3.2$ py examples/display_data.py -t dealnodeal

[ optional arguments: ]

[  num_examples: 10 ]

[ Main ParlAI Arguments: ]

[  task: dealnodeal ]

[  download_path: /Users/jase/src/ParlAI/downloads ]

[  datatype: train ]

[  image_mode: raw ]

[  numthreads: 1 ]

[  batchsize: 1 ]

[  datapath: /Users/jase/src/ParlAI/data ]

[creating task(s): dealnodeal]

loading:
/Users/jase/src/ParlAI/data/negotiation/end-to-end-negotiator-master/src/data/negotiate/train.txt

Negotiate with your opponent to decide who gets how many of each item. You
must both agree on the distribution of items (or you both get zero), but
try to get as much value as you can. There are 1 book(s), each with a value
of 4, 3 hat(s), each with a value of 0, and 1 ball(s), each with a value of
6.

i need the ball and 1 hat please .

[labels: sorry i really need the ball and book . you can have the hat]

   [RepeatLabelAgent]: sorry i really need the ball and book . you can have
the hat

~~

i really need the ball . you can have the book and the hats if i can have
the ball .

[labels: sorry i cant make a deal without the ball .]

   [RepeatLabelAgent]: sorry i cant make a deal without the ball .

~~

i can't make a deal without the ball either . i guess we have no deal .

[labels: ok , i understand]

   [RepeatLabelAgent]: ok , i understand

~~

no deal .

[labels: i agree]

   [RepeatLabelAgent]: i agree

~~

that button should come up soon . no deal .

[labels: yeah just waiting on it]

   [RepeatLabelAgent]: yeah just waiting on it

~~

<selection>

[labels: <no_agreement> <no_agreement> <no_agreement> <no_agreement>
<no_agreement> <no_agreement>]

   [RepeatLabelAgent]: <no_agreement> <no_agreement> <no_agreement>
<no_agreement> <no_agreement> <no_agreement>

~~

Negotiate with your opponent to decide who gets how many of each item. You
must both agree on the distribution of items (or you both get zero), but
try to get as much value as you can. There are 1 book(s), each with a value
of 6, 4 hat(s), each with a value of 1, and 1 ball(s), each with a value of
0.

[labels: you can have the ball if i can have the book and the hats]

   [RepeatLabelAgent]: you can have the ball if i can have the book and the
hats

~~

i need the book and the ball if you want you can have all of the balls

[labels: i cannot give up the book . i need the book and at least 2 hats]

   [RepeatLabelAgent]: i cannot give up the book . i need the book and at
least 2 hats

~~

i can not give up the book at all either . it looks like we can not make a
deal .

[labels: even if you take all of the hats and the ball ?]

   [RepeatLabelAgent]: even if you take all of the hats and the ball ?

~~

yes , i cannot make a deal without the book

[labels: ok no deal]

   [RepeatLabelAgent]: ok no deal

~~

klshuster:
ahh yes sorry I misspoke, I was simply looking at the fact that each episode was only one line in the data file; this should represent the number of examples, not episodes, though the agent does not programmatically (currently) have a concept of 'examples'; i will go in and fix this to ensure len returns the correct value

jaseweston:
I think we should rename len() -> num_examples()
but we should try to use it as little as possible anyway, e.g. train
doesn't require it

klshuster:
yeah that makes sense; the len() is only necessary for our current implementation of train_model, which uses len(world). I'll implement both for now (num_examples() and len()) and then we can deal with len() when we deal with len(world) in train_model

Issue #409, opened 2017-11-27
[BUG] QA Data collection Mturk task fails with KeyError: 'image_mode'


koustuvsinha:
On runnning the provided `qa_data_collection` Mturk task, the taks fails after first HIT is accepted, with the following stack trace:

```
[building data: /root/ParlAI_Mturk/data/SQuAD]
[||||||||||||||||||||||||||||||||||||||||] 30288272 / 30288272
[||||||||||||||||||||||||||||||||||||||||] 4854279 / 4854279
Exception in thread task-t_1:
Traceback (most recent call last):
  File "/root/project/miniconda2/envs/parlai/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/root/project/miniconda2/envs/parlai/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/root/ParlAI/parlai/mturk/core/mturk_manager.py", line 776, in _task_function
    task_function(mturk_manager=self, opt=opt, workers=workers)
  File "run.py", line 62, in run_conversation
    task = task_class(task_opt)
  File "/root/ParlAI/parlai/tasks/squad/agents.py", line 98, in __init__
    super().__init__(opt, shared)
  File "/root/ParlAI/parlai/core/dialog_teacher.py", line 51, in __init__
    cands=self.label_candidates(), **kwargs)
  File "/root/ParlAI/parlai/core/dialog_teacher.py", line 144, in __init__
    self.image_loader = ImageLoader(opt)
  File "/root/ParlAI/parlai/core/image_featurizers.py", line 64, in __init__
    self.im = opt['image_mode']
KeyError: 'image_mode'
```

alexholdenmiller:
thanks for flagging!

alexholdenmiller:
should be fixed now!

PR #412, opened 2017-11-27
set default im in image featurizer to none

alexholdenmiller:
also, fix naming of wmt's init.py


alexholdenmiller:
fix for #409

alexholdenmiller:
good catch, added an error message so we can set the defaults only in one place (the params file) or could add more args in the future

PR #415, opened 2017-11-27
mastering the dungeon

kimiyoung:


jaseweston:
I think we are just checking in without MTurk code rn because he is not
sure it works with the new system. Perhaps you could help Jack to update
that later.

JackUrb:
@jaseweston That's totally fair on the side of the mturk collection code (though I don't foresee any issues with it working on the new system), but it's even more reason I feel these run files for training and evaluating should be moved out of the mturk directory and into the projects directory.

jaseweston:
oh i thought everything is a subfolder of projects right now isn't it?
which file are you talking about? you mean, there shouldnt be an mturk dir
inside projects?

JackUrb:
@jaseweston Yeah right now everything is in `/projects/...` however the training and evaluation code is under a directory `/projects/.../mturk/` and for the rest of the ParlAI codebase things in `/mturk/` directories refer to runable tasks to collect data from HITs. I feel like these would be more appropriate in the existant `/projects/mastering_the_dungeon/projects/` though I feel like that directory should be renamed something else too (maybe `/projects/mastering_the_dungeon/experiment`?).

jaseweston:
ok this is a little pedantic i think, but if zhilin is ok with it

kimiyoung:
@JackUrb @jaseweston Thanks for the comments! A lot of training code in mturk/tasks/MTD/run.py is also used during MTD. Later when we add the MTD part we can still reuse the code. And the current committed runs can be viewed as evaluation of MTD, so it might make sense semantically to just put it under mturk.

In terms of checking in the MTD part (with HITs), the parlai version I used is quite different from the current master branch, and I changed many APIs and such to deal with the corner cases. We will probably need to tweak it later, which could take some time. Thus I thought it might be good to just check in the offline experiments and the environments first.

kimiyoung:
@alexholdenmiller I have added the copyright.

PR #416, opened 2017-11-28
Consolidate dialog teachers

klshuster:
This PR does the following:

1. Renames FixedDataTeacher to FixedDialogTeacher.
2. Consolidates all dialog teachers (FixedDialog, FbDialog, Dialog) into one file, `teachers.py`, in parlai/core
3. Updates docs to reflect this change
4. Updates README to reflect this change
5. Fixes a few minor bugs dealing with metrics:
    - bug in `seq2seq` in returning the loss as a metric when metrics are using a SharedTable.
    - bug in `thread_utils` where a type was not converted to a str before printing


alexholdenmiller:
related, but haven't looked at more than just your description yet: we should probably add "hogwild" as a separate param to params.py, since using the sharedtable in metrics only is necessary during hogwild not just when you're using extra threads for data loading.

alexholdenmiller:
"files changed: 46" ha nice

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -357,7 +357,7 @@ def predict(self, xs, ys=None, cands=None, valid_cands=None, lm=False):
             loss.backward()
             self.update_params()
             losskey = 'loss' if not lm else 'lmloss'
-            loss_dict = {losskey: loss.mul_(len(xs)).data}
+            loss_dict = {losskey: loss.mul_(len(xs)).data[0]}
```
alexholdenmiller: actually this was intentionally returning the cuda tensor so that we didn't have to pull the tensor back from the GPU, it's possible it's not affecting the performance much though

klshuster: That makes sense, however I think there is an issue with some python built-ins when using cuda tensors; here's the error I get if I revert this change:

File "examples/train_model.py", line 220, in <module>
main(setup_args())
File "examples/train_model.py", line 157, in main
train_report = world.report()
File "/data/users/kshuster/ParlAI/parlai/core/worlds.py", line 645, in report
return self.world.report()
File "/data/users/kshuster/ParlAI/parlai/core/worlds.py", line 263, in report
return self.agents[0].report()
File "/data/users/kshuster/ParlAI/parlai/core/agents.py", line 140, in report                                                                            
return self.metrics.report()
File "/data/users/kshuster/ParlAI/parlai/core/metrics.py", line 181, in report                                                                              
m[k] = round_sigfigs(self.metrics[k] / total, 3)
File "/data/users/kshuster/ParlAI/parlai/core/utils.py", line 91, in round_sigfigs
return round(x, -math.floor(math.log10(abs(x)) - sigfigs + 1))

TypeError: bad operand type for abs(): 'torch.cuda.FloatTensor'

README.md
```diff
@@ -192,12 +192,14 @@ The core library contains the following files:
...
 - **build_data.py**: basic utilities for setting up data for tasks. you can override if your filesystem needs different functionality.
-- **dialog_teacher.py**: contains a base teacher class for doing dialog with fixed chat logs, along with a data class for storing the data
 - **dict.py**: contains code for building general NLP-style dictionaries from observations
   - DictionaryAgent: agent which tracks the index and frequency of words in a dictionary, and can parse a sentence into indices into its dictionary or back
-- **fbdialog_teacher.py**: contains a teacher class which implements a function setup_data which parses data in the FB Dialog data format
 - **metrics.py**: computes evaluation metrics for dialog, e.g. ranking metrics, etc.
 - **params.py**: uses argparse to interpret command line arguments for ParlAI
+- **teachers.py**: contains teachers that deal with dialog-based tasks, as well as data classes for storing data
+  - **_FixedDialogTeacher_**: base class for a teacher that utilizes fixed data
+  - **_DialogTeacher_**: base class for a teacher doing dialog with fixed chat logs
```
jaseweston: difference between _FixedDialogTeacher_ and DialogTeacher unclear here?

parlai/agents/seq2seq/seq2seq.py
alexholdenmiller: yeah actually it looks like it has a pretty minor effect so let's land it

PR #418, opened 2017-11-29
Speed up batched training by sorting examples by length

alexholdenmiller:
Dramatically reduces padding in batched training by allowing fixed data to have batches of similar-length data.

(I'll update conflicts after #416 lands)


examples/train_model.py
```diff
@@ -56,6 +56,10 @@ def setup_args():
     train.add_argument('-vmt', '--validation-metric', default='accuracy',
                        help='key into report table for selecting best '
                             'validation')
+    train.add_argument('-vcut', '--validation-cutoff',
```
jaseweston: default should be for >=1?

parlai/core/fixed_data_teacher.py
```diff
@@ -4,14 +4,88 @@
...
 
+import argparse
+from collections import deque
 import concurrent.futures
 from threading import Thread
 import queue
 import random
 
 
+def flatten(teacher, context_length=None, include_label=True):
```
jaseweston: i feel like these should be in some utils file or our teachers file is going to be hard to read and full of stuff? (also kurt is moving this file)

parlai/core/fixed_data_teacher.py
```diff
@@ -87,6 +161,30 @@ class FixedDataTeacher(Teacher):
 
 
     """
+    @staticmethod
+    def add_cmdline_args(argparser):
+        try:
+            args = argparser.add_argument_group('Batching Arguments')
+            args.add_argument('-bsrt', '--batch-sort', default=True, type='bool',
+                              help='If enabled (default True), create batches by '
+                                   'flattening all episodes to have length one '
```
jaseweston: what does "length one" mean?

parlai/core/worlds.py
```diff
@@ -622,7 +625,7 @@ def display(self):
         return s
 
     def __len__(self):
-        return math.ceil(sum(len(w) for w in self.worlds) / len(self.worlds))
+        return len(self.world)
```
jaseweston: how did that change from worlds to world? -- what's going on here?

parlai/core/fixed_data_teacher.py
```diff
@@ -87,6 +161,30 @@ class FixedDataTeacher(Teacher):
 
 
     """
+    @staticmethod
+    def add_cmdline_args(argparser):
```
jaseweston: can we only add these if bs>1 ?  hmm, doesnt look easy does it?

examples/train_model.py
alexholdenmiller: sure if you prefer! this was 0.995 below before ("rounds up to 1") but I'll change to 1

parlai/core/fixed_data_teacher.py
alexholdenmiller: sure!

parlai/core/fixed_data_teacher.py
alexholdenmiller: we can pop then from the opt dict in params.py if batchsize is not greater than one, so they will never be seen by the user.

jaseweston: oh sweet!

parlai/core/worlds.py
alexholdenmiller: The length of a batch world was here defined as the ceiling of the average length of all the subworlds. However, subworlds are just shared versions of the seed world (self.world), so they should all be the same. (They are all the same in all of our worlds.). However, if you don't actually set up the data in the subworlds, they break when they try to return the length, unless you have the seed world share the length with all of them. (This only now just came up because of the new batching system.)

parlai/core/fixed_data_teacher.py
alexholdenmiller: One example per episode.

alexholdenmiller:
hm I rebased on kurt's changes and now it includes all of kurt's changes in the diff...

parlai/core/worlds.py
jaseweston: ok

alexholdenmiller:
sure--should we just move the batching options to params.py too?

jaseweston:
hmm maybe, up to you

On Thu, Nov 30, 2017 at 11:52 AM, Alexander Miller <notifications@github.com

PR #420, opened 2017-11-30
hacky fix to batched streaming

alexholdenmiller:
This solution is pretty hacky but currently batched streaming is broken and this will fix it until a cleaner solution can be identified.


alexholdenmiller:
ha, sure

PR #422, opened 2017-11-30
Remove world iters, get rid of len, etc.

klshuster:
This PR includes the following:

- remove `__iter__`, `__next__`, and `__len__` from worlds and agents
- adds `num_examples` and `num_episodes` to all worlds, as well as agents in `teachers.py`
- converts vqa_v2 teacher to a `FixedDialogTeacher`
- misc changes to other agents to deal with removing `__len__`


parlai/core/agents.py
```diff
@@ -183,20 +171,17 @@ def __init__(self, opt, shared=None):
...
-            raise StopIteration()
+    def num_examples(self):
+        if not hasattr(self, 'num_exs'):
+            # num_examples is sum of all examples in all tasks
+            self.num_exs = sum(t.num_examples() for t in self.tasks)
+        return self.num_exs
+
+    def num_episodes(self):
+        if not hasattr(self, 'num_eps'):
+            # num_examples is sum of all examples in all tasks
```
jaseweston: episodes

parlai/core/teachers.py
```diff
@@ -242,6 +239,13 @@ def num_episodes(self):
         except Exception:
             raise RuntimeError('"num_episodes" must be overriden by children.')
 
+    def num_examples(self):
```
jaseweston: so what about if num_episodes (or num_examples) isnt defined -- what happens?

parlai/core/teachers.py
```diff
@@ -60,9 +60,6 @@ def __init__(self, opt):
         self.num_workers = opt.get('numthreads', 1)
         self.request_queue = queue.Queue()
 
-    def __len__(self):
```
jaseweston: was this wrong i guess?

parlai/core/worlds.py
```diff
@@ -623,8 +611,11 @@ def display(self):
         s += ("[--end of batch--]")
         return s
 
-    def __len__(self):
-        return len(self.world)
+    def num_examples(self):
+        return math.ceil(sum(w.num_examples() for w in self.worlds) / len(self.worlds))
```
jaseweston: hmm, what about it if it isnt defined for one of them?

parlai/core/agents.py
klshuster: good catch, thanks

parlai/core/teachers.py
klshuster: yeah, this was old code in the `DataLoader` that is useless now

parlai/core/worlds.py
alexholdenmiller: note that this should be `self.world.num_examples()` but jason's question still applies.

currently we'll just get some kind of error about num_examples not being defined, but we could add a new `try/except` saying specifically that it's either not implemented or not available for some of the tasks.

alexholdenmiller:
nice!

parlai/core/teachers.py
klshuster: Currently, any subclass of `FixedDialogTeacher` that does not have `self.episodes` and `self.examples`, and does not override `num_episodes` or `num_examples`, will get a runtime error. However, `DialogTeacher` has these functions implemented, and all of the teachers that subclass `FixedDialogTeacher` (in this repo) have these functions implemented as well

parlai/core/worlds.py
jaseweston: one solution is that if one of them returns something that means 'unknown' (via None? 0?) hen the whole thing becomes equal to that value?

parlai/core/teachers.py
jaseweston: i think the base class teacher should have them defined with returning whatever means 'unknown'  (for e.g. streaming or infinite data where we dont want to specify)

klshuster: Ok, do you mean like the base `Teacher` class? Before, we had a return of 0 be the default for streaming data in the `StreamDialogData`, but now we compute the length once and then save that to file

jaseweston: i mean for tasks where the data is infinite, or where they can't/don't declare it.  we probably shouldn't use 0 because it's numerical -- somehow None or inf seem to make more sense?

klshuster: I think None makes sense; I can add that to the base `Teacher` class, if you think that's the right call

jaseweston: yep cool,  then anything that sums over num_*  that sees a None in one of the worlds, should also return None, right?

klshuster: Yeah, I'll make that change right now

alexholdenmiller: (also--the reason we used 0 as the default is that __len__ is only allowed to return positive numbers or zero, can't do -1 or None or anything. None is good!)

PR #423, opened 2017-12-01
update drqa and fix batching bug

alexholdenmiller:


alexholdenmiller:
approved by jason over messenger

PR #428, opened 2017-12-04
change tokenizer to parlai in fairseq to improve performance

klshuster:
changes the tokenization in the fairseq agent to use the tokenizers provided by the parlai `DictionaryAgent`; this improves performance of fairseq on bAbI


alexholdenmiller:
yes it defaults to spacy because it gets higher accuracy with the pretrained model but you can just switch the tokenizer via a commandline arg to the faster-running one

PR #429, opened 2017-12-05
Move computation to metrics

klshuster:
This PR shifts the computation of `time_left` and `num_epochs` to `metrics.py`, and also moves metric aggregation functionality for multiple tasks/worlds to `aggregate_metrics` in `metrics.py`.

Also, a few minor fixes in some tasks for when `datatype` is, e.g., `test:stream`, rather than `test`


examples/train_model.py
```diff
@@ -176,42 +177,35 @@ def log(self):
         logs.append('time:{}s'.format(math.floor(self.train_time.time())))
         logs.append('parleys:{}'.format(self.parleys))
         # get report and update total examples seen so far
+        report_opts = {
```
jaseweston: these should be in the world or the agent, no?

examples/train_model.py
```diff
@@ -121,20 +122,15 @@ def __init__(self, parser):
...
-        self.total_exs = 0
         self.total_episodes = 0
         self.total_epochs = 0
-        self.max_exs = None
-        self.max_parleys = None
-        self.world_num_exs = self.world.num_examples()
-        if self.world_num_exs is not None:
-            self.max_exs = opt['num_epochs'] * self.world_num_exs
-            self.max_parleys = math.ceil(self.max_exs / opt['batchsize'])
+        self.num_epochs = opt['num_epochs'] if opt['num_epochs'] > 0 else float('inf')
```
jaseweston: max?

examples/train_model.py
```diff
@@ -172,72 +168,52 @@ def log(self):
...
-                    (self.world_num_exs is not None and self.world_num_exs > 0)):
-                display_epochs = int(self.total_exs / self.world_num_exs)
-            else:
-                display_epochs = self.total_epochs
-                logs.append('num_epochs:{}'.format(display_epochs))
-        # join log string and add full metrics report to end of log
+
+        # time elapsed
+        logs.append('time:{}s'.format(math.floor(self.world.get_train_time())))
+        logs.append('parleys:{}'.format(self.world.get_total_parleys()))
```
jaseweston: to report?

examples/train_model.py
```diff
@@ -121,20 +122,15 @@ def __init__(self, parser):
         # Create model and assign it to the specified task
         self.agent = create_agent(opt)
         self.world = create_task(opt, self.agent)
-        self.train_time = Timer()
```
jaseweston: keep this one?

examples/train_model.py
```diff
@@ -172,72 +168,52 @@ def log(self):
...
     def train(self):
         opt = self.opt
         world = self.world
         with world:
             while True:
                 world.parley()
-                self.parleys += 1
-                if world.epoch_done():
-                    self.total_epochs += 1
+                world_done = ((world.get_total_parleys() >= world.get_max_parleys())
```
jaseweston: no more parley counts

parlai/core/metrics.py
```diff
@@ -63,6 +64,70 @@ def _score(g_tokens, a_tokens):
...
+    m['total'] = total
+    m['accuracy'] = 0
+    if num_tasks > 0:
+        m['accuracy'] = sum_accuracy / num_tasks
+        if sum_f1 > 0:
+            m['f1'] = sum_f1 / num_tasks
+    return m
+
+
+def compute_time_metrics(world, opt):
```
jaseweston: this isnt really to do with the metrics class but i see why you've put it here.

it has an unhealthy use of options that only depend on train_model main (max_train_time), and requires the world too..

but i dont see how to do it better right now, other than making those all arguments instead?

PR #440, opened 2017-12-08
Simplify total epochs call

klshuster:
Move the `epoch_done` check to `world.get_total_epochs` to simplify trainloop code and increase clarity


parlai/core/worlds.py
```diff
@@ -192,9 +192,13 @@ def get_total_epochs(self):
         """Return total amount of epochs on which the world has trained."""
         if not self.max_exs:
             self.max_exs = self.num_examples() * self.opt['num_epochs'] if self.num_examples() else -1
+        # when we know the size of the data
         if self.max_exs > 0:
             return int(self.total_parleys * self.opt['batchsize'] / self.num_examples())
+        # when we do not know the size of the data
         else:
+            if self.epoch_done():
+                self.total_epochs += 1
```
alexholdenmiller: 👎

alexholdenmiller:
sorry another request--can we change update total epochs to `update_counters` and then you can move the `self.total_parleys += 1` code into there as well?

PR #441, opened 2017-12-08
[bugfix ] MTurkDealNoDealDialogWorld termination condition

sivareddyg:
Without this termination condition, world doesn't end. Thereby, memory keeps leaking until there is no memory left. This should fix it.


sivareddyg:
CLA is signed now.

Issue #446, opened 2017-12-12
Add 'taskntalk' dataset of synthetic shapes to ParlAI. 


kdexd:
Task N' Talk is a goal-based cooperative conversation game between two agents - a questioner and answerer. It was described in _Kottur et al. '17 "Natural Language Does Not Emerge 'Naturally' in Multi-Agent Dialog"_ ([arxiv](https://arxiv.org/abs/1706.08502)). 

I implemented this paper using ParlAI ([lang-emerge-parlai](https://www.github.com/karandesai-96/lang-emerge-parlai)). As it works fine, this issue is created to group a bunch of incoming PRs related to the task.

karandesai-96:
Next up: Our QBot and ABot in `parlai/agents`

jaseweston:
Cool.. are they general agents that might work on some of the other parlai
tasks? (
https://github.com/facebookresearch/ParlAI/blob/master/docs/source/_static/img/tasks2.png?raw=true
)

If yes, awesome! If no, they should go in a projects/ folder instead

karandesai-96:
I think they should be re-usable for the VisDial task. If I avoid hardcoding the dimensions of internal modules. I am speaking in context of this figure:

![image](https://user-images.githubusercontent.com/10494087/33867600-cfdd3c2a-df23-11e7-872a-b09585c6d265.png)

Taken from Das et al. 2017a _Learning Cooperative Visual Dialog Agents with Deep Reinforcement Learning_. Even this paper conducts experiments on both, Task N' Talk and VisDial.

Of course, the architecture hyperparameters are different but it has a similar structure. I will try to generalize well enough so that my agents can be used for both tasks. :)

karandesai-96:
For completeness, this figure is from _Natural Language Does Not Emerge 'Naturally' in Multi-Agent Dialog_.
![image](https://user-images.githubusercontent.com/10494087/33867687-5f0dff1a-df24-11e7-9e68-1d629c74d8d0.png)

ListenNet works like Encoder and SpeakNet works like Decoder.

karandesai-96:
@alexholdenmiller @jaseweston Hi, I created the agent classes and made a PR. Now in the `parlai/projects`, I can subclass the dictionary agent and provide task specific mappings (for example, `X` - `1`). Also I shall subclass my agents and simply override `dictionary_agent()` staticmethod. This should serve as an example to show easy subclassing of these agents as per requirement.

karandesai-96:
[...]

PR #447, opened 2017-12-12
Add 'taskntalk' to task list and create teachers for accessing it.

karandesai-96:
### Reference Issue #446

This PR adds `taskntalk` task to the current list of tasks in ParlAI. There are two types of datasets - small and large. Also, `TaskNTalkTeacher` is created for fetching examples from this dataset, although it lacks metric handling functionality (to be introduced later).

Here are few examples of `display_data.py` example:

```sh
$ python3 display_data.py -t taskntalk:small --num-examples 1
[ optional arguments: ] 
[  num_examples: 1 ]
[ Main ParlAI Arguments: ] 
[  image_mode: raw ]
[  datatype: train ]
[  numthreads: 1 ]
[  task: taskntalk ]
[  datapath: /home/kd/Documents/gt/parlai/data ]
[  download_path: /home/kd/Documents/gt/parlai/downloads ]
[ Batching Arguments: ] 
[  batchsize: 1 ]
[creating task(s): taskntalk]
loading: /home/kd/Documents/gt/parlai/data/taskntalk/small/train.json
blue square solid
color shape
[labels: blue square]
   [RepeatLabelAgent]: blue square
~~
```
Similar results for `-t taskntalk:large` and `-t taskntalk`. And for batchsize > 1:
```sh
$ python3 display_data.py -t taskntalk:small --num-examples 1 --batchsize 2 -dt valid
     --- similar as above, excluded for sanity ---
loading: /home/kd/Documents/gt/parlai/data/taskntalk/small/valid.json
[--batchsize 2--]
[batch world 0:]
blue triangle solid
color shape
[labels: blue triangle]
   [RepeatLabelAgent]: blue triangle
[batch world 1:]
green triangle dashed
color style
[labels: blue triangle]
   [RepeatLabelAgent]: green dashed
[--end of batch--]
```

### DIFF
--- a/parlai/tasks/taskntalk/agents.py
+++ b/parlai/tasks/taskntalk/agents.py
@@ -12,10 +12,13 @@ import random
 
 
 def _path(opt, task_size='small'):
+    """Return path to json file of dataset - it can be train/valid file
+    of small/large dataset. Validation data is used for test as well,
+    because labels are inferred from the image and task itself.
+    """
+    dt = opt['datatype'].split(':')[0]
     # ensure data is built
     build(opt)
-    dt = opt['datatype'].split(':')[0]
-
     if dt == 'train':
         file_name = 'train.json'
     elif dt == 'valid' or dt == 'test':
@@ -28,10 +31,12 @@ def _path(opt, task_size='small'):
 
 
 class TaskNTalkTeacher(Teacher):
+    """TaskNTalk basic teacher, it picks a random image and associates
+    a random task with it. Metric updates and observation are to be
+    implemented.
+    """
     def __init__(self, opt, shared=None):
         super().__init__(opt, shared)
-        # store datatype
-        self.datatype = opt['datatype']
         if not shared:
             self._setup_data(self.opt['datafile'])
         else:
@@ -40,7 +45,7 @@ class TaskNTalkTeacher(Teacher):
             self.task_index = shared['task_index']
 
     def _setup_data(self, data_path):
-        # loads data
+        """Read the json file and store images and task definitions."""
         print('loading: ' + data_path)
         with open(data_path) as data_file:
             json_data = json.load(data_file)
@@ -51,6 +56,7 @@ class TaskNTalkTeacher(Teacher):
         random.shuffle(self.data)
 
     def share(self):
+        """Share images and task definitions with other teachers."""
         shared = super().share()
         shared['data'] = self.data
         shared['task_defn'] = self.task_defn
@@ -67,10 +73,9 @@ class TaskNTalkTeacher(Teacher):
         return observation
 
     def act(self):
-        # TODO(kd): fetch all data for valid/test
-        # select random image and task
+        """Select random image and associate random task with it."""
         image = random.choice(self.data)
-        task  = random.choice(self.task_defn)
+        task = random.choice(self.task_defn)
         labels = [image[self.task_index[attr]] for attr in task]
         action = {
             'image': ' '.join(image),
@@ -78,20 +83,24 @@ class TaskNTalkTeacher(Teacher):
             'labels': [' '.join(labels)],
             'episode_done': True
         }
+        # TODO(kd): fetch all data for valid/test
         return action
 
 
 class SmallTeacher(TaskNTalkTeacher):
+    """Teacher for small dataset, invoked by ``taskntalk:small``."""
     def __init__(self, opt, shared=None):
         opt['datafile'] = _path(opt, 'small')
         super().__init__(opt, shared)
 
 
 class LargeTeacher(TaskNTalkTeacher):
+    """Teacher for large dataset, invoked by ``taskntalk:large``."""
     def __init__(self, opt, shared=None):
         opt['datafile'] = _path(opt, 'large')
         super().__init__(opt, shared)
 
 
 class DefaultTeacher(SmallTeacher):
+    """Default teacher for small dataset, invoked by ``taskntalk``."""
     pass
--- a/parlai/tasks/taskntalk/agents.py
+++ b/parlai/tasks/taskntalk/agents.py
@@ -3,14 +3,12 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree. An additional grant
 # of patent rights can be found in the PATENTS file in the same directory.
+from parlai.core.agents import Teacher
+from .build import build
 
-import itertools
 import json
 import os
 import random
-
-from parlai.core.agents import Teacher
-from .build import build
 
 
 def _path(opt, task_size='small'):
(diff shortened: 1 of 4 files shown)


karandesai-96:
@jaseweston, @alexholdenmiller: Hi, I'm starting to make a couple of PRs to integrate my implementation. I have a small suggestion - why is `observation['text']` assumed to be a string? (It gave an error during executing `display_data.py` when it was a list. Inline comment follows:

parlai/tasks/taskntalk/agents.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def act(self):
+        """Select random image and associate random task with it."""
+        image = random.choice(self.data)
+        task = random.choice(self.task_defn)
+        labels = [image[self.task_index[attr]] for attr in task]
+        action = {
+            'image': ' '.join(image),
+            'text': ' '.join(task),
+            'labels': [' '.join(labels)],
```
karandesai-96: I had to convert my task '['shape', 'color']' to 'shape color' just to make `display_data.py` work. Is it standard or should there be a relaxation on the datatype of 'text'?

jaseweston: indeed, agents are supposed to be speaking text, so the 'text' field is meant to contain text.

karandesai-96: Here, the `'task'` is supposed to be the goal of Questioner Agent, something which it has to perform at the end of dialog episode. Now it is not something which the agent would speak - but in the paper, it is treated just as text (as future conversations are embedded, passed through LSTM etc.).

While it is logically not 'text', and hold a different only at the start of dialog, I think it's fine to make an exception and do what I did here.

jaseweston:
you can fill in another field instead of text, specific to your task, but
it is not recommended unless strictly necessary as no other models will be
able to train on your task, because it is so specific

karandesai-96:
Okay.. well then `text` works fine... I wish to keep it generic and in fact analogous to VisDial task, this `text` describing goal (or task) of Questioner Agent will become the caption of the unseen image in VisDial (or GuessWhich game).

I hope haven't interchanged - the goal of QBot is described as `task` in the paper, while task in ParlAI has a totally different meaning!

alexholdenmiller:
@karandesai-96 basically you should think of tasks in ParlAI as something that a human would be able to do (albeit maybe a human expert, as in the case of translation). This is part of why we emphasize that the text field should be actually text, in addition to keeping the same format to ensure interoperability between models.

alexholdenmiller:
You can remove
```
# Copyright (c) 2017-present, Facebook, Inc.
# All rights reserved.
```
from your files if you wish (please leave the next three lines with the patent info).

Otherwise I'll merge this!

alexholdenmiller:
Actually I noticed something we might want to update quickly before merging, let me check with Jason

karandesai-96:
@alexholdenmiller Makes sense to me - why the `text` field is indeed kept to be a string.

I have one question here - this task is not about doing a dialog with a given vocabulary. This task is to view the emergence of grounded language between the bots. 

Other agents such as `memnn` and `seq2seq` have a DictionaryAgent which does the tokenization and de-tokenization. For that, the dict is built on the dataset itself.

But here, I cannot build the dict on words such as `red`, `triangle`, `dotted` etc. Instead, the bots have separate vocabulary which comprises of symbols - `X Y Z 1 2 3 4` or similar. At the end of training,  there has to be a correspondence between symbols and actual words of our synthetic dataset. So I am unsure whether to include the vocab symbols in this json file or not. 

Also, if such training is project specific and for example in future, there would be another game `guesswhich` (Das et al. 2017a) - then our agent should have a dictionary agent as described before, since it would be de-tokenizing to english words.

karandesai-96:
I mean, my _to-be-included-in-upcoming-PR_`CooperativeGameAgent` should have a DictionaryAgent to be generic enough. The question lies in inclusion of symbols in task json dataset for that agent to be able to tokenize.

Another alternative is to include docs on procedure to subclass this agent to `QBot` and `ABot` - according to the game, and we can define the tokenization and de-tokenization logic as per task. (Maybe even omit `DictionaryAgent` specifically for `taskntalk`).

parlai/tasks/taskntalk/agents.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    def __init__(self, opt, shared=None):
+        super().__init__(opt, shared)
+        if not shared:
+            self._setup_data(self.opt['datafile'])
+        else:
+            self.data = shared['data']
+            self.task_defn = shared['task_defn']
+            self.task_index = shared['task_index']
+
```
alexholdenmiller: can you add self.id='taskntalk'

alexholdenmiller:
@karandesai-96 I added one change request, and otherwise we're good!

As for the tokenization: one option is for you to extend the dictionary class with a subclass which maps the game tokens to different values (X <=> square, etc). This way you can inherit some of the existing parsing functionality, but you can maintain your own custom correspondence.

I'm not 100% sure if that will work for you in this exact setup--do you think that could work here?

parlai/tasks/taskntalk/agents.py
karandesai-96: Done :white_check_mark:

karandesai-96:
@alexholdenmiller I made the change you suggested.

> As for the tokenization: one option is for you to extend the dictionary class with a subclass which maps the game tokens to different values (X <=> square, etc). 

This mapping 'emerges' after training for a long time, 'X' can be mapped to any other word as well, and that is to be observed manually. But thanks for the subclassing idea, I can take it forward.

> This way you can inherit some of the existing parsing functionality, but you can maintain your own custom correspondence.

As far as I can see it, all the cooperative games of this style will be either on synthetic datasets, or have english dialog (`guesswhich` game uses `VisDial`). In `parlai/agents`, I shall keep the `DictionaryAgent` instance intact - generic enough for real datasets. 

In `parlai/projects/taskntalk`, I will extend it with a subclass and replace it in QBot and ABot. It will serve as a good example to deal with cooperative gaes involving synthetic datasets!

So my mapping would be `'X' - 1`, `'Y' - 2` and so on. Which would eventually become human interpretable after training these agents for a long time (`'X' - 1 - 'shape?'`, `'2' - 2 - 'star'` etc.)

karandesai-96:
That said, there must be a way to specify which dict class should be used for `DictionaryAgent` (default or custom). There should be an available example, I will search for it (`str2class` may come handy). If that doesn't work then I will extend both QBot and ABot and just make a `super().__init__()` call and later replace the dictionary instance.

alexholdenmiller:
Great! The dictionary class can be set on the command line `--dict-class`, or via the static `dictionary_class` method (or of course you could set your own flag that changes the logic as well).

Thanks for the contributions!

karandesai-96:
@alexholdenmiller Great, these are well-suited alternatives for the `DictionaryAgent`! I am closely following `Seq2seqAgent` as a reference since my agent will be behaving in a similar manner (tried to subclass `Seq2seqAgent` directly but it isn't _that_ similar, anyhow..)

Thanks for the merge, thrilled to get my contribution in ParlAI. I will make my next PR soon.

PR #448, opened 2017-12-13
Maintain dialogue history: as a utility function which any model can use

jaseweston:


parlai/core/utils.py
```diff
@@ -185,3 +185,50 @@ def sort_data(data, key='text_label', method='spaces'):
...
+                            historyLength=1, useReplies="labels",
+                            dict=None, useStartEndIndices=True):
+    """Keeps track of dialog history, up to a truncation length.
+    Either includes replies from the labels, model, or not all using param 'replies'."""
+
+    def parse(txt):
+        if dict is not None:
+            vec =  dict.txt2vec(txt)
+            if useStartEndIndices:
+                parsed_x = deque([dict[dict.start_token]])
```
alexholdenmiller: maxlen=?

alexholdenmiller: resolved, this is right!

parlai/core/utils.py
```diff
@@ -185,3 +185,50 @@ def sort_data(data, key='text_label', method='spaces'):
...
+
+    if history['episode_done']:
+        history['dialog'].clear()
+        history['episode_done'] = False
+
+    if useReplies != 'none':
+        if useReplies == 'model':
+            if reply != '':
+                history['dialog'].extend(parse(reply))
+            r = reply
```
alexholdenmiller: looks like this `r` isn't used

parlai/core/utils.py
```diff
@@ -185,3 +185,50 @@ def sort_data(data, key='text_label', method='spaces'):
...
+        else:
```
alexholdenmiller: elif

parlai/core/utils.py
```diff
@@ -185,3 +185,50 @@ def sort_data(data, key='text_label', method='spaces'):
...
+            if 'labels' in observation:
```
alexholdenmiller: elif

jaseweston:
ok, fixed

PR #449, opened 2017-12-14
Pytorch data loader

klshuster:
Provides a PytorchDataTeacher that utilizes a pytorch DataLoader for fast dataloading.
(I wrote the following for this PR and then subsequently added it to the PytorchDataTeacher class)
To use the PytorchDataTeacher, one must do the following:

1. Ensure that an appropriate teacher exists that can read the data currently saved and produce an action dict for an agent (this will be the `buildteacher`) 
2. Subclass PytorchDataTeacher in a task agents file (one does not necessarily need to implement any of the methods in PytorchDataTeacher, it just should exist in the `agents.py` file).
3. Build the data so that it can be used by the PytorchDataTeacher
    - This can be accomplished in 2 ways:
        1. Specify a `buildteacher`, `datafile` (where the data currently is, and what will be used to build the data), and `datatype` (e.g. train, valid, etc) when calling either `build_pytorch_data` or calling `train_model.py`. If one is training a model, the data will be built automatically in `train_model.py`. 
        2. Implement the `buildteacher` such that it saves the appropriate datafile in its `datafile` attribute (i.e. `self.datafile`) given the datatype, and then specify the `buildteacher` when calling either `build_pytorch_data.py` or `train_model.py`

Additionally, if `preprocessed` is set to `True`, then the model specified in the command line params will have its `observe` function called on the `buildteacher`'s action, and the data will be saved for that model specifically.

Here's an example of what would need to be done for `bAbI` 10k task 1, with preprocessed data from the `seq2seq` model.

1.  Implement a normal `bAbI` teacher that can read the data in its current format and create an action dict (this currently exists as the `Task1kTeacher`
2.  Implement a `bAbI` teacher that subclasses PytorchDataTeacher, e.g. `Task1kPyTeacher`
3. If `Task1kTeacher` saves the datafile in it's attributes, use one of the following 2 commands:
    - `python examples/train_model.py -t babi:task10kpy:1 --buildteacher babi:task10k:1 -m seq2seq -mf /tmp/pytorch_data_build --preprocess 1` - if one would like to train the model after building the data
    - `python examples/build_pytorch_data.py -m seq2seq --buildteacher babi:task10k:1 --preprocess 1`
4. If you would like to specify a specific datafile to build, e.g. a validation file, you could do either of the following:
    - `python examples/train_model.py -t babi:task10kpy:1 --buildteacher babi:task10k:1 --datafile data/bAbI/tasks_1-20_v1-2/en-valid-10k-nosf/qa1_valid.txt -dt valid -m seq2seq -mf /tmp/pytorch_data_build --preprocess 1`
    - `python examples/build_pytorch_data.py -m seq2seq --buildteacher babi:task10k:1 --datafile data/bAbI/tasks_1-20_v1-2/en-valid-10k-nosf/qa1_valid.txt --preprocess 1`


jaseweston:
new plan: move pytorchloader as a separate teacher so we do not have to modify any code of any existing datasets/tasks

klshuster:
It is now possible for one to simply have `python examples/train_model.py -t pytorch_data_teacher --buildteacher babi:task10k:1 -m seq2seq -mf /tmp/pytorch_data_build --preprocess 1`, i.e. specify the teacher to be the `PytorchDataTeacher`. Note that it is still necessary to either specify a `--datafile` or save a datafile in the `--buildteacher`, so that the `PytorchDataTeacher` knows where to find the data

jaseweston:
cool can we just change some of those flag names:

pytorch_data_teacher   -- something a little smaller maybe?
buildteacher --  something that references it is for pytorch loader? like a
prefix?
preprocess   -- something that references it is for pytorch loader? like a
prefix?

klshuster:
I changed it to 'pytorch_teacher', 'pytorch_buildteacher', and 'pytorch_preprocess' - does that work?

jaseweston:
sounds good to me

parlai/core/pytorch_data_teacher.py
```diff
@@ -0,0 +1,306 @@
...
+        return self.dataset.num_episodes()
+
+    def num_examples(self):
+        """Get the total number of examples in this dataset."""
+        return self.dataset.num_examples()
+
+    def observe(self, observation):
+        """Process observation for metrics."""
+        if self.bsz > 1:
+            self.lastY = self.lastYs[self.batchindex]
```
alexholdenmiller: also need `self.lastYs[self.batchindex] = None` (had a few changes in FixedDialogTeacher, this might not be the only one)

parlai/core/pytorch_data_teacher.py
```diff
@@ -0,0 +1,291 @@
...
+        batch, self.epochDone = self.next_batch()
+
+        # pad batch
+        if len(batch) < self.bsz:
+            batch += [{'episode_done': True, 'id': self.getID()}] * (self.bsz - len(batch))
+
+        for i, ex in enumerate(batch):
+            self.lastYs[i] = ex.get('labels', ex.get('eval_labels'))
+
+        return batch
```
alexholdenmiller: it looks like this function could be inherited if the parent class also implemented next_batch(), yeah?

parlai/core/pytorch_data_teacher.py
```diff
@@ -0,0 +1,291 @@
...
+    def act(self):
+        """Send new dialog message."""
+        if not hasattr(self, 'epochDone'):
+            # reset if haven't yet
+            self.reset()
+        if self.epochDone:
+            if not self.training:
+                return {'episode_done': True, 'id': self.getID()}
+            else:
+                self.reset_data()
```
alexholdenmiller: are we sure we want auto-reset?

PR #453, opened 2017-12-15
fix to seq2seq model

alexholdenmiller:
history was remembering the zero padding, so need to allocate new object when padding

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -391,9 +391,9 @@ class Seq2seqAgent(Agent):
             xs = Variable(self.xs)
         else:
             max_x_len = max([len(x) for x in parsed_x])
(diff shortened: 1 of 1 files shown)


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -421,8 +422,9 @@ def valid(obs):
                     y.extendleft(reversed(x))
 
             max_y_len = max(len(y) for y in parsed_y)
-            for y in parsed_y:
-                y += [self.NULL_IDX] * (max_y_len - len(y))
+            parsed_y = [y if len(y) == max_y_len else
```
jaseweston: this is padding things? should be a utility or something?

alexholdenmiller: you're right--this is padding. could be a utility but I'd like to land this first at least.

PR #455, opened 2017-12-17
Extended MemNN agent added that supports reward-based and forward prediction

ekmb:


examples/train_model.py
```diff
@@ -123,6 +124,11 @@ def __init__(self, parser):
             build_dict.build_dict(opt)
         # Create model and assign it to the specified task
         self.agent = create_agent(opt)
+        
+        # CUDA with multiprocessing requires 'spawn' start method
+        if opt.get('numthreads', 1) > 1 and opt.get('cuda', False):
+            set_start_method('spawn')
```
alexholdenmiller: I think we need to protect this call using `if __name__ == '__main__'`--so this should be in the training loop

parlai/tasks/dbll_babi/agents.py
```diff
@@ -55,8 +55,113 @@ def __init__(self, opt, shared=None):
...
+        self.opt = opt
         super().__init__(opt, shared)
 
+    def setup_data(self, path):
+        if len(self.opt['task'].split(':')) == 4:
+            return self.setup_data_with_feedback(path)
+        else:
+            return super().setup_data(path)
+
+    def setup_data_with_feedback(self, path):
```
alexholdenmiller: @jaseweston what do you think of this input setup?

examples/train_model.py
alexholdenmiller: So at the bottom of the train file under the name/main check we should set multiprocessing to spawn if cuda is enabled (we could wrap that in a try/except checking if `import torch; if torch.is_cuda_enabled()` is true as well?)

alexholdenmiller:
if Jason approves than I'm okay for now

examples/train_model.py
jaseweston: can't we move this to the world or something? it's annoying people have to put this in their train script isn't it?

parlai/agents/memnn_extended/memnn_extended.py
```diff
@@ -0,0 +1,492 @@
...
+
+class MemnnExtendedAgent(Agent):
+    """ Memory Network agent.
+    With setting IM acts as a regular Imitation learning agent
+
+    For more details on IM_feedback, RBI, FP, RBI+FP settings that support 
+    reward-based and forward prediction: https://arxiv.org/abs/1604.06045
+    These models assume that feedback and reward for the current example  
+    immediatly follow the query. 
+    py examples/train_model.py -m memnn_extended -t dbll_babi:task:3_p0.5:f
```
jaseweston: can ":f" be a little more descriptive?

parlai/tasks/dbll_babi/agents.py
jaseweston: len(self.opt['task'].split(':')) == 4 should more specifically search for a string name and have a comment at least

parlai/tasks/dbll_babi/agents.py
```diff
@@ -55,8 +55,113 @@ def __init__(self, opt, shared=None):
...
+        """Reads data in the fbdialog format.
```
jaseweston: this data doesnt exist (won't be downloaded, though right (right now)?)

parlai/agents/memnn_extended/memnn_extended.py
```diff
@@ -0,0 +1,492 @@
+# Copyright (c) 2017-present, Facebook, Inc.
```
jaseweston: maybe call it "feedback" or "fp" or something more specific than "extended" ?

parlai/agents/memnn_extended/memnn_extended.py
```diff
@@ -0,0 +1,492 @@
...
+
+import os
+import copy
+import random
+
+from .modules import MemNN, Decoder, to_tensors
+
+
+class MemnnExtendedAgent(Agent):
+    """ Memory Network agent.
```
jaseweston: slightly more descriptive comment please especially first line

examples/train_model.py
alexholdenmiller: the multiprocessing docs say:
```
Note that this should be called at most once, and it should be protected inside the if __name__ == '__main__' clause of the main module.
```
it might work in HogwildWorld as well but this motivated me to put it in the main module!

parlai/tasks/dbll_babi/agents.py
alexholdenmiller: I meant the actual data:
that is, you feel good about appending the next x to the current x?

alexholdenmiller: (I'm understanding that correctly, right?

that the data looks like:
```
Mary moved to the bedroom.
Mary travelled to the garden.
Where is John?
No, that's wrong.
[labels: garden]
```

jaseweston:
if you are understanding i agree it's not great -- but is ok for
experiments. for that reason maybe this should be moved to parlai/projects/
?

examples/train_model.py
ekmb: I tried putting it in HogwildWorld but was getting errors that the context was already created. Also valid world is also a HogWildWorld but the 'set_start_method('spawn')' call should appear only once in a program.

If in __main__ did you mean something like this:
 try:
        import torch
        from multiprocessing import set_start_method
        if torch.cuda.is_available():
            # CUDA with multiprocessing requires 'spawn' start method
            set_start_method('spawn')
    except:
        pass

parlai/tasks/dbll_babi/agents.py
ekmb: Your understanding is correct.

parlai/agents/memnn_extended/memnn_extended.py
ekmb: It supports fp/rbi/rbi+fp/im so I wanted something more general. I can rename it to 'fp' if you think it won't be confusing.

parlai/tasks/dbll_babi/agents.py
ekmb: No, this method only iterates over the dataset. I think the data is downloaded when a new task is created.

jaseweston: ok, can we just move this stuff to parlai/projects because it doesn't really fit what we see as a standard ParlAI dataset format. As I said I think it's ok for experiments, but our standard format has to be expecting responses in time sequence format (i.e. you wouldn't get the feedback before you've provided the label!)

jaseweston:
what's happening with this?

ekmb:
I'll fix it in a few weeks, sorry can't work on it now

parlai/tasks/dbll_babi/agents.py
```diff
@@ -55,113 +55,8 @@ def __init__(self, opt, shared=None):
         opt['datafile'] = _path(os.path.join('babi', 'babi1'), params, opt)
```
jaseweston: so dbll_babi hasnt changed?

ekmb: No, I reverted the changes I made before

PR #456, opened 2017-12-17
Label should be added to the dialog history instead of model's prediction

ekmb:


parlai/agents/memnn/memnn.py
```diff
@@ -119,10 +115,13 @@ def observe(self, observation):
...
-            if self.answers[batch_idx] is not None:
-                prev_dialogue += '\n' + self.answers[batch_idx]
-                self.answers[batch_idx] = None
+
+            # append answer given in the previous example to the dialog 
+            if 'eval_labels' in self.observation:
+                prev_dialogue += '\n' + random.choice(self.observation['eval_labels'])
+            elif 'labels' in self.observation:
+                prev_dialogue += '\n' + random.choice(self.observation['labels'])
+
```
alexholdenmiller: @jaseweston in the seq2seq model we were doing 'append label during training but append answer during valid' but maybe this makes more sense--append labels if available, else append eval_labels if available, else append self answer, even? (this will make it more robust during interactive modes as well)

jaseweston: we should be using my utility. yes, we could have an option where it uses label in train and pred when no label or eval_label: these should be added to the utility as extra options i think.

alexholdenmiller: do you want to land this or wait for a switch to your util?

jaseweston:
I would go for the util..

On Tue, Dec 19, 2017 at 11:44 AM, Alexander Miller <notifications@github.com

parlai/agents/memnn/memnn.py
ekmb: You're talking about maintain_dialog_history, right? It seems like it can use as reply either self.answer (predictions from the previous example) or the current label (the one we're trying to predict) or none. Am I missing something? Shouldn't it use labels from the previous example?

jaseweston: Hmm, that does look wrong.. needs to be fixed I think

jaseweston: should be like this i think: https://github.com/facebookresearch/ParlAI/pull/461

?

ekmb: Maybe also add a check if eval_labels become labels and vice versa? to fix that training data is appended to validation data when a model is evaluated in the middle of training

jaseweston: not sure that is the job of this function, but that should be definitely fixed -- are you aware of that problem @alexholdenmiller  et al?

parlai/agents/memnn/memnn.py
```diff
@@ -100,32 +108,38 @@ def __init__(self, opt, shared=None):
...
+        """Save observation for act.
+        If multiple observations are from the same episode, concatenate them.
+        """
         self.episode_done = observation['episode_done']
-        return observation
+        # shallow copy observation (deep copy can be expensive)
+        obs = observation.copy()
+        batch_idx = self.opt.get('batchindex', 0)
+        
+        obs['text'] = '\n'.join(maintain_dialog_history(
```
jaseweston: i think this can actually be much faster if you convert to vectors only once and keep the vector representation.. that's what seq2seq version does

ekmb: I agree, but it will require more changes throughout the file, for example, in the input parse(self, text) since the model uses '\n' to separate query from the context. Can we assume that every context sentence ends with '.' while query with '?'. Maybe leave it for a separate pull request?

jaseweston: no we can't assume that, this isn't just for question answering.. i think the \n split is good...

PR #458, opened 2017-12-18
Add CooperativeGameAgent, QuestionerAgent, AnswererAgent in parlai.agents

karandesai-96:
### Reference Issue #446

This PR adds three classes into `parlai.agents`: `CooperativeGameAgent`, `QuestionerAgent`, `AnswererAgent`. All these agents are made up of several pytorch modules declared in `parlai.agents.coopgame_agent.modules`

Releavnt documentation is included as well. Currently, this class doesn't handle batch training, my plan is to use these classes and write a training script according to [@karandesai-96/lang-emerge-parlai](https://www.github.com/karandesai-96/lang-emerge-parlai). It will go in `parlai/projects`. Once it works correctly, I shall make generalizations to accomodate batch training and validation.

Built documentation looks like this, followed by class APIdocs:

image


karandesai-96:
While working on this PR, I realized that a `CooperativeGameAgent` can use modules from `seq2seq` as its listen, speak, predict modules. Once the basic paper implementation works, I will try to make it work.

parlai/agents/coopgame_agent/coopgame_agent.py
```diff
@@ -0,0 +1,286 @@
...
+        return optim_class([module.parameters() for module in self.modules],
+                            **kwargs)
+
+    def tokenize(self, text):
+        """Convert text observaton (string) to a ``torch.autograd.Variable`` of
+        tokens using ``DictionaryAgent``.
+        """
+        text_tokens = self.dict.txt2vec(text)
+        # TODO(kd): handle requires_grad=False for valid/test
+        token_vars = Variable(torch.Tensor(text_tokens), requires_grad=True)
```
alexholdenmiller: do you need `requires_grad=True` here? maybe I'm missing something, but isn't that only true if you're passing gradients back into the dictionary?

karandesai-96: I think you're right. I am correcting this in upcoming commit. Also, I added an in-place operation on Variable - removing it.

karandesai-96:
@jaseweston I am thinking of giving a link of related example in `parlai/projects`, once it is added. Does that sound okay? 

Just to be sure, I will look into other `parlai/agents` as to how their instructions are presented in documentation.

alexholdenmiller:
great--merging now!

PR #461, opened 2017-12-20
label fix for maintain dialog history

jaseweston:


jaseweston:
will merge this as it's kinda broken rn

parlai/agents/seq2seq/.#seq2seq.py
```diff
@@ -0,0 +1 @@
+jase@jase-mbp.dhcp.thefacebook.com.89298
```
alexholdenmiller: lol

PR #464, opened 2017-12-21
custom metrics from agent +  log-every-n-secs in eval_model

jaseweston:


parlai/core/metrics.py
```diff
@@ -245,13 +254,19 @@ def report(self):
...
         return m
 
     def clear(self):
         with self._lock():
             self.metrics['cnt'] = 0
             self.metrics['correct'] = 0
             self.metrics['f1'] = 0.0
+            for k in self.custom_metrics:
+                self.metrics[k] = 0.0
+                self.metrics[k + '_cnt'] = 0.0
```
alexholdenmiller: maybe better to store `0` instead of `0.0` in _cnt

PR #466, opened 2017-12-22
Pytorch data fixes

klshuster:
A few minor fixes/improvements to the Pytorch Teacher

- fix preprocess command line arg so that it actually works (specify `--pytorch_preprocess` for preprocessing, `--no_pytorch_preprocess` for no preprocessing
- make sure seq2seq model doesnt preprocess twice
- fix StreamDataset `__getitem__` so that, when there are multiple workers (e.g. 4), they do not load the same example 4 times (as they each get their own copy of the StreamDataset).

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -318,12 +318,15 @@ class Seq2seqAgent(Agent):
         # shallow copy observation (deep copy can be expensive)
         obs = observation.copy()
         batch_idx = self.opt.get('batchindex', 0)
-        obs['text2vec'] = maintain_dialog_history(
-            self.history, obs,
-            reply=self.answers[batch_idx],
-            historyLength=self.opt['history_length'],
-            useReplies=self.opt['history_replies'],
-            dict=self.dict)
+        if not obs.get('preprocessed', False):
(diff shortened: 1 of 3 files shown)


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -318,12 +318,15 @@ def observe(self, observation):
...
-            dict=self.dict)
+        if not obs.get('preprocessed', False):
+            obs['text2vec'] = maintain_dialog_history(
+                self.history, obs,
+                reply=self.answers[batch_idx],
+                historyLength=self.opt['history_length'],
+                useReplies=self.opt['history_replies'],
+                dict=self.dict)
+        else:
+            obs['text2vec'] = deque(obs['text2vec'], self.opt['history_length'])
```
jaseweston: wouldn't that already be part of the preprocessing?

jaseweston: i guess it depends how much preprocessing is done.. there are levels...?

klshuster: A deque is not serializable for saving JSON, so when I save the data it needs to be converted to a list; this converts it back to a deque so it can be used accordingly later down the model's pipeline

jaseweston:
ah ok

PR #467, opened 2017-12-22
Debug seq2seq

emilydinan:
Fixed seq2seq bug in which correct hidden state is not passed into the decoder at eval time. Also fixed small bug in train_model which caused model to get stuck in validation loop even when accuracy is 1 (if previous best accuracy is 1).


examples/train_model.py
```diff
@@ -145,7 +145,8 @@ def validate(self):
         valid_report, self.valid_world = run_eval(
             self.agent, opt, 'valid', opt['validation_max_exs'],
             valid_world=self.valid_world)
-        if valid_report[opt['validation_metric']] > self.best_valid:
+        if valid_report[opt['validation_metric']] > self.best_valid or \
+                valid_report[opt['validation_metric']] == 1:
```
alexholdenmiller: hm this shouldn't be necessary, I think. The first time that it hits 1, it should stop on line 156 when the cutoff is achieved.

alexholdenmiller: did you observe different behavior?

PR #468, opened 2017-12-24
Fix create_task_world error for multiple agents.

karandesai-96:
Bug Description
============

Minimal snippet (import statements and other details omitted for brevity):
```python
agent1 = Agent(opt)
agent2 = Agent(opt)
world = create_task_world(opt, [agent1, agent2])
```

**Actual Behaviour:** `create_task_world` returns `DialogPartnerWorld` even if agents are more than 2.
```text
RuntimeError: There must be exactly two agents for this world.
```

**Desired Behaviour:** `world` should be an instance of `MultiAgentDialogWorld`.

Contribution
=========
- Fix this error by returning a `MultiAgentDialogWorld`.
- Fix `epoch_done` method of `MultiAgentDialogWorld` and make it similar to `DialogPartnerWorld`'s method of the same name.
- Modify ~`display_data.py`~ `base_train.py` example to allow specifying multiple agents, just to make sure this works. Old example usage works the same way (`python3 base_train.py -t taskname`)

Example Usage
============
```text
~$ python3 base_train.py -t taskntalk --num-iters 1 --num-agents 2
[ optional arguments: ] 
[  num_iters: 1 ]
[  num_agents: 2 ]
        ### other arguments omitted for brevity ###
[ training ]
agent received observation:
  {'episode_done': True, 'image': 'purple circle dashed', 'labels': ['purple circle'], 'text': 'color shape'}
agent sending message:
  {'text': 'hello, teacher!'}
agent received observation:
  {'text': 'hello, teacher!'}
agent sending message:
  {'text': 'hello, teacher!'}
[ training summary. ]
  {'total': 0}
[ validating ]
agent received observation:
  {'episode_done': True, 'image': 'red star dashed', 'labels': ['star red'], 'text': 'shape color'}
agent sending message:
  {'text': 'hello, teacher!'}
agent received observation:
  {'text': 'hello, teacher!'}
agent sending message:
  {'text': 'hello, teacher!'}
[ validation summary. ]
  {'total': 0}
finished in 0.0 s
```

### DIFF
--- a/examples/base_train.py
+++ b/examples/base_train.py
@@ -24,22 +24,25 @@ import time
 def main():
     # Get command line arguments
     parser = ParlaiParser()
-    parser.add_argument('-n', '--num-examples', default=10)
+    parser.add_argument('-n', '--num-iters', default=10, type=int)
+    parser.add_argument('-a', '--num-agents', default=1, type=int)
     opt = parser.parse_args()
 
-    agent = Agent(opt)
+    agents = []
+    for _ in range(opt['num_agents']):
+        agents.append(Agent(opt))
 
     opt['datatype'] = 'train'
-    world_train = create_task(opt, agent)
+    world_train = create_task(opt, agents)
 
     opt['datatype'] = 'valid'
-    world_valid = create_task(opt, agent)
+    world_valid = create_task(opt, agents)
 
     start = time.time()
     # train / valid loop
     for _ in range(1):
         print('[ training ]')
-        for _ in range(10):  # train for a bit
+        for _ in range(opt['num_iters']):  # train for a bit
             world_train.parley()
 
         print('[ training summary. ]')
--- a/examples/display_data.py
+++ b/examples/display_data.py
@@ -12,39 +12,15 @@ see a few of them:
 """
 
 from parlai.core.params import ParlaiParser
-from parlai.core.agents import Agent
 from parlai.agents.repeat_label.repeat_label import RepeatLabelAgent
 from parlai.core.worlds import create_task
 
 import random
 
-
-class RepeatTextAgent(Agent):
-    """Simple agent which repeats the text it receives."""
-    def __init__(self, opt, shared=None):
-        super().__init__(opt)
-        self.id = 'RepeatTextAgent'
-
-    def act(self):
-        obs = self.observation
-        if obs is None or 'text' not in obs:
-            return {'text': 'Nothing to repeat yet.'}
-        return {'id': self.getID(), 'text': obs['text']}
-
-
 def display_data(opt):
-    # create one repeat label agent and assign it to the specified task
-    agents = []
-    agents.append(RepeatLabelAgent(opt))
-
-    # since RepeatLabelAgent action has label in 'text', other agents
-    # should be `RepeatTextAgent`s
-    for _ in range(opt['num_agents'] - 1):
-        agents.append(RepeatTextAgent(opt))
-
-    # world will be `DialogPartnerWorld` for num_agents = 1
-    # (agent + task teacher), else it will be `MultiAgentDialogWorld`
-    world = create_task(opt, agents)
+    # create repeat label agent and assign it to the specified task
+    agent = RepeatLabelAgent(opt)
+    world = create_task(opt, agent)
 
     # Show some example dialogs.
     with world:
@@ -62,7 +38,6 @@ def main():
     # Get command line arguments
     parser = ParlaiParser()
     parser.add_argument('-n', '--num-examples', default=10, type=int)
-    parser.add_argument('-a', '--num-agents', default=1, type=int)
     opt = parser.parse_args()
 
     display_data(opt)
--- a/examples/display_data.py
+++ b/examples/display_data.py
@@ -12,15 +12,39 @@ see a few of them:
 """
 
 from parlai.core.params import ParlaiParser
+from parlai.core.agents import Agent
 from parlai.agents.repeat_label.repeat_label import RepeatLabelAgent
 from parlai.core.worlds import create_task
 
 import random
 
+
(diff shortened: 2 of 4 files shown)


karandesai-96:
Spotted this bug while working on #446 (worlds for playing cooperative games will have at least three agents - task teacher, questioner, and answerer).

parlai/core/worlds.py
```diff
@@ -356,8 +356,9 @@ def parley(self):
     def epoch_done(self):
         done = False
         for a in self.agents:
-            if a.epoch_done():
-                done = True
+            if hasattr(a, 'epoch_done'):
```
jaseweston: we could also put this function in the base agent class, returning false so it exists?
https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/agents.py#L8

examples/display_data.py
```diff
@@ -12,15 +12,39 @@
...
+
 def display_data(opt):
-    # create repeat label agent and assign it to the specified task
-    agent = RepeatLabelAgent(opt)
-    world = create_task(opt, agent)
+    # create one repeat label agent and assign it to the specified task
+    agents = []
+    agents.append(RepeatLabelAgent(opt))
+
+    # since RepeatLabelAgent action has label in 'text', other agents
```
jaseweston: i don't really understand this, what text are they repeating exactly? the same that the first repeat label agent said? not sure i understand the use case exactly..

parlai/core/worlds.py
```diff
@@ -888,7 +889,8 @@ def shutdown(self):
 
 ### Functions for creating tasks/worlds given options.
 
-def _get_task_world(opt):
+def _get_task_world(opt, user_agents):
```
jaseweston: this looks reasonable to me..

examples/display_data.py
```diff
@@ -12,15 +12,39 @@
...
 
 from parlai.core.params import ParlaiParser
+from parlai.core.agents import Agent
 from parlai.agents.repeat_label.repeat_label import RepeatLabelAgent
 from parlai.core.worlds import create_task
 
 import random
 
+
+class RepeatTextAgent(Agent):
```
jaseweston: if we need another agent we'd probably put it elsewhere, or extend RepeatLabelAgent somehow ... still trying to understand exactly what you want it to do ..

examples/display_data.py
karandesai-96: @jaseweston yes you are right, they repeat what the `RepeatLabelAgent` said. Let me clarify it in a separate comment.

karandesai-96:
@jaseweston but if you say: 

> if we need another agent we'd probably put it elsewhere, or extend RepeatLabelAgent somehow

Then I can write a sub-class of `RepeatLabelAgent` which keeps the `label` field intact.

karandesai-96:
@jaseweston: Update, just tried it. (Haven't pushed the commit here yet)

#### Agent class:
```python
class CustomRepeatLabelAgent(RepeatLabelAgent):
    """Sub-class of RepeatLabelAgent which repeats the label it received,
    and also keeps `observation['label']` intact."""
        super().__init__(opt, shared)

    def act(self):
        action = super().act()
        action['labels'] = self.observation['labels']
        return action
```

#### Output
```text
~$ python3 display_data.py -t taskntalk -n 1 --num-agents 2
[ optional arguments: ] 
[  num_agents: 3 ]
[  num_examples: 1 ]
### other arguments omitted for brevity ###

blue square solid
color shape
[labels: blue square]
[RepeatLabelAgent]: blue square
[labels: blue square]
[RepeatLabelAgent]: blue square
[labels: blue square]
~~
```

A bit cluttered, what'd you say? I think I can leave this example unchanged and find a way to add a unit test maybe?

karandesai-96:
I assume the thread is inactive due to holiday week?
@jaseweston: On to you whenever you are back after the celebrations :) (Happy New Year!)

karandesai-96:
I have changed my mind - 

`display_data.py` 's purpose it to display the examples from tasks, and not about using more agents - it is more of a teacher work rather than agent.

I am changing `base_train.py` example to hold more agents - efficiently showing a training loop where a world containing 2 or more agents is trained and validated. Upcoming commit...

examples/base_train.py
```diff
@@ -24,22 +24,25 @@
...
     parser = ParlaiParser()
-    parser.add_argument('-n', '--num-examples', default=10)
+    parser.add_argument('-n', '--num-iters', default=10, type=int)
+    parser.add_argument('-a', '--num-agents', default=1, type=int)
     opt = parser.parse_args()
 
-    agent = Agent(opt)
+    agents = []
+    for _ in range(opt['num_agents']):
+        agents.append(Agent(opt))
```
jaseweston: so it makes identical agents or?

karandesai-96: yes it makes identical agents. Just a simple example to demonstrate that a world can be created with any number of agents getting turns in a round robing fashion.

jaseweston:
ok cool

parlai/core/worlds.py
```diff
@@ -907,13 +907,15 @@ def _get_task_world(opt):
             world_class = getattr(my_module, world_name)
         except Exception:
             # Defaults to this if you did not specify a world for your task.
-            world_class = DialogPartnerWorld
-    task_agents = _create_task_agents(opt)
+            if len(task_agents + user_agents) == 2:
+                world_class = DialogPartnerWorld
+            else:
+                world_class = MultiAgentDialogWorld
```
alexholdenmiller: nice--while we're here, we should also do this when `'.' in sp[0]` (line 896).

karandesai-96: @alexholdenmiller Done!

alexholdenmiller:
thanks!!

PR #471, opened 2018-01-02
Update News

klshuster:
add pytorch data teacher to news


NEWS.md
```diff
@@ -1,5 +1,7 @@
 <h2>News</h2>
 
+2017-12-14: [New teacher added: PytorchDataTeacher](https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/pytorch_data_teacher.py)
```
jaseweston: I think this should explain what this is -- more like 
"Bigger, faster data loading supported with {PyTorch data loader}" or something?

klshuster: How about "Fast, multiprocessed data loading supported with [Pytorch data loader]"?

jaseweston:
yes, thnx!

PR #477, opened 2018-01-03
add retriever agent to parlai

alexholdenmiller:
The Retriever Agent is initialized with an existing ParlAI task, storing the 'text' - 'label' pairs as keys and values, and creating a sparse tfidf matrix of hashed word counts from the keys.

When the retriever agent acts, it retrieves the values associated with the keys most similar to the incoming 'text'. If the incoming observation includes label candidates, it ranks them against the incoming text and returns that in the 'text–candidates' field--otherwise, it returns the list of values it already retrieved.

Still could use a little more cleanup including removing functions imported from drqa but not used, and more features including support for returning the keys instead of the values if desired. This will allow it to be used by another agent as a tfidf document store.


alexholdenmiller:
posting first draft for now

alexholdenmiller:
@jaseweston 
1) not really, although it could be supported--but you need to rebuild the document matrix when you do that. what I do currently support is a cached on-the-fly building of a document matrix using all of the candidates, if candidates are provided, and use that instead of the saved documents if available.
2) I assumed that data to be added would be formatted as a task, so currently "no" not possible to store something that isn't a task (although original code read a directory of files and I changed that for this, so it wouldn't be that hard to add that kind of thing back in)

jaseweston:
ok, cool for now

jaseweston:
try with adam's data?

alexholdenmiller:
@jaseweston ready for re-review!

alexholdenmiller:
added a unit test on babi:task1k:1, testing retrieving one specific example (successfully), adding another entry and then retrieving it

parlai/agents/tfidf_retriever/tfidf_retriever.py
```diff
@@ -0,0 +1,168 @@
...
+        if 'text' in obs:
+            # we should reply or add this as a stored example
+            if 'labels' in obs:
+                # save this to our retrieval matrix
+                self.store(obs)
+            else:
+                self.rebuild()  # no-op if nothing has been queued to store
+                doc_ids, doc_scores = self.ranker.closest_docs(obs['text'], k=30)
+
+                if obs.get('label_candidates') and False:
```
jaseweston: and False?

parlai/agents/tfidf_retriever/build_db.py
```diff
@@ -0,0 +1,112 @@
...
+fmt = logging.Formatter('%(asctime)s: [ %(message)s ]', '%m/%d/%Y %I:%M:%S %p')
+console = logging.StreamHandler()
+console.setFormatter(fmt)
+logger.addHandler(console)
+
+# ------------------------------------------------------------------------------
+# Store corpus.
+# ------------------------------------------------------------------------------
+
+def store_contents(opt, task, save_path, context_length=-1, include_labels=True):
```
jaseweston: as i said before,  i thought it might be better to not do it this but to use the standard act() function like we do to build a dictionary, but maybe i'm wrong maybe this is simpler.
i think a workaround would be to allow both ways to do it? can that be done?

alexholdenmiller: if you don't specify a task, it doesn't do this--and it does allow adding them via the act function as well (that is, both ways are allowed!)

parlai/agents/tfidf_retriever/tfidf_retriever.py
alexholdenmiller: oops thanks

PR #480, opened 2018-01-04
Humanize and throttle download progress for better UX

### LINKED ISSUE #479: Humanize and throttle download progress output for better user experience
Currently everything is shown in bytes and doesn't really provide any meaningful readable information to user. Further, output to stdout at every chunk downloaded is harmful for overall performance. We should humanize and throttle download progress.

apsdehal:
- Resolves #479

Sample:
image

### DIFF
--- a/parlai/core/build_data.py
+++ b/parlai/core/build_data.py
@@ -14,6 +14,8 @@ import os
 import requests
 import shutil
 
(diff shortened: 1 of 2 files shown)


alexholdenmiller:
thanks!

PR #484, opened 2018-01-04
Opensubtitles teachers teacher name fix

emilydinan:
Also added 10k teacher

### DIFF
--- a/parlai/tasks/opensubtitles/agents.py
+++ b/parlai/tasks/opensubtitles/agents.py
@@ -57,6 +57,7 @@ class Task100kTeacher(HalfTeacher):
                 yield entry, new
             cnt += 1
 
+
 class Task10kTeacher(HalfTeacher):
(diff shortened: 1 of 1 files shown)


parlai/tasks/opensubtitles/agents.py
```diff
@@ -58,6 +58,16 @@ def setup_data(self, path):
             cnt += 1
 
 
+class Task10kTeacher(HalfTeacher):
+    """This version of opensubtitles only includes 10,000 dialogs."""
+    def setup_data(self, path):
+        cnt = 0
+        for entry, new in super().setup_data(path):
+            if cnt < 10000:
+                yield entry, new
```
jaseweston: this data isn't random though, right?

alexholdenmiller: right just the first 10k (from one perspective, no reversed perspectives)

PR #487, opened 2018-01-05
Add support for NarrativeQA

### LINKED ISSUE #475: NarrativeQA support
https://arxiv.org/abs/1712.07040 and https://github.com/deepmind/narrativeqa .
I can work on this one also if we want support.

apsdehal:
Resolves #475 

- Adds two tasks, one for summaries and one for stories, `narrative_qa` and `narrative_qa:summaries`
- Downloads all of the stories while building, so may take some time during downloading
- Provides both answer references as possible answer_labels
- In case a story is missing (unable to download), it is skipped.
- download_tests file has been update to reflect both stories and summaries download test
- For convenience, creates train, valid and test folders and creates separate CSVs for each of them
- Uses apsdehal's fork with some fixes for now, should be changed to deepmind's original repo after PR 1 on that repo is merged

### DIFF
--- a/parlai/tasks/narrative_qa/__init__.py
+++ b/parlai/tasks/narrative_qa/__init__.py
@@ -1,0 +1,5 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
--- a/parlai/tasks/narrative_qa/agents.py
+++ b/parlai/tasks/narrative_qa/agents.py
@@ -1,0 +1,135 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+from parlai.core.teachers import DialogTeacher
+from .build import build
+
+import os
+import copy
+import csv
+import glob
+
+
+def _path(opt):
+    build(opt)
+
+    dt = opt['datatype'].split(':')[0]
+
+    if not (dt == 'train' or dt == 'valid' or dt == 'test'):
+        raise RuntimeError('Not valid datatype.')
+
+    suffix = dt
+
+    data_path = os.path.join(opt['datapath'], 'NarrativeQA',
+                             'narrative_qa', suffix)
+
+    return data_path
+
+
+class SummariesTeacher(DialogTeacher):
+    def __init__(self, opt, shared=None):
+        opt = copy.deepcopy(opt)
+        data_path = _path(opt)
+        opt['datafile'] = data_path
+        self.id = 'NarrativeQA'
+
+        super().__init__(opt, shared)
+
+    def setup_data(self, path):
+        print('loading data from: ' + path)
+
+        qa_path = os.path.join(path, 'qaps.csv')
+        summaries_path = os.path.join(path, 'summaries.csv')
+
+        qa_pairs = dict()
+
+        with open(qa_path, 'r') as f:
+            reader = csv.DictReader(f)
+            for row in reader:
+                if row['document_id'] not in qa_pairs:
+                    qa_pairs[row['document_id']] = []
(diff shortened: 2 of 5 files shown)


jaseweston:
so, there is a (big) story for each question, is that it?
how does that work in your system -- does it dump the whole story into the observation, or you are expected to access that separately, or..?

apsdehal:
Yes, currently it dumps the whole story. There are two tasks overall, one is for story and one is for summaries. Then, it asks all of the questions related to that story as kind of a dialog interaction. 

Other option would be to provide user file path for the story and then they can read it themselves. I would be happy to change to that if it suits better to project's needs.

See https://gist.github.com/apsdehal/8bc6cf1469f5a014d735a70f503d4e42 for a sample output.

klshuster:
looks good to me, though I'm not able to run on my mac currently as I do not have `wget` installed, which is used in the download script for the stories (`ParlAI/data/NarrativeQA/narrativeqa-master/download_stories.sh`)

jaseweston:
hmm...i think it should be using wget from within python?

apsdehal:
@jaseweston It is possible, but I thought it might be best to use their own script as it downloads and modifies some stuff. 
I can look into creating a python version of the script using download tools but that may divert from original script in future if they change something. See the original version of bash script by deepmind: [here](https://github.com/apsdehal/narrativeqa/blob/master/download_stories.sh)

jaseweston:
I see. You can also check out a certain version of their github repo to
ensure nothing changes/breaks, the downside being you don't get any fixes
automatically either..

apsdehal:
Yes, that is what I am doing currently. I have forked their repo to mine and then made some small changes to bash script so that it works with `subprocess` execution. So, my fork won't be updated unless I make changes to it.

So, would you like me to convert their bash script to a pythonic version?

alexholdenmiller:
Thanks @apsdehal, looks good overall!

But yes it would be better for us to not refer to the external repo at all if possible--a small external change could break our own code then. The gutenberg downloading looks pretty straightforward, and we have equivalent download-with-retries and unzipping functions in our build_tools core file.

apsdehal:
Sure, give me some time, I will push a fix regarding that soon.

apsdehal:
@alexholdenmiller @jaseweston I have updated the build script to download all of the stories content directly instead of using the bash file.

alexholdenmiller:
Great, thanks @apsdehal!

Two points:
1) this is making two attempts to download full-size versions of each story, is that right?
```
download_stories(base_path)
# Try again for small stories
download_stories(base_path)
```

2) I'm getting tons of blank logging commands, I think due to when the download is requested for a story but fails, or when the file is very small. I'm pushing a fix to the download logger so that it is more clear.

However, I'm not sure if we fully want to accept that downloads may be incomplete without letting the user know (or try again later). It seems like we should at print out the stories that did not successfully download, and probably should keep a list of the ones that did not and perhaps retry downloading them on the next run (but okay maybe we should just go ahead and use what's there on this run?).
@jaseweston what do you think? the line here between convenience, reliability, and complexity is getting a little blurry.

jaseweston:
I think people will want the whole train/test set, no?

alexholdenmiller:
Seems like it to me--my impression here is that a "small" version of every story is included first, and then whichever long stories the downloader can get ahold of are downloaded, otherwise the short version is kept.

@apsdehal I guess DeepMind was doing the same thing? Making an attempt or two and downloading the full version and then just moving on? I see a two-second `sleep` in there, which looks like an attempt not to overwhelm the gutenberg servers.

apsdehal:
Yes, the second attempt tries to re-download the stories which were incompletely downloaded and the ones which were errored. I can add a check after second download to notify user of the stories which still have a small size.

alexholdenmiller:
@jaseweston do you want to just alert users, or keep retrying until every one has been downloaded?

apsdehal:
@alexholdenmiller I believe that retrying indefinitely would result in bad UX as a website might be permanently down. There should be a limit on max retries (like 10). We can notify user about the current download and if we are retrying (with the number of current retry).

jaseweston:
yes, lets just try a few times, and then notify. they can always try again manually, right?

apsdehal:
Yes, I will update the PR accordingly.

apsdehal:
@jaseweston @alexholdenmiller I have updated the PR.

PR #488, opened 2018-01-05
UnicodeDecodeError Fix

MannyKayy:
Fixes the following UnicodeDecodeError:
parlai-unicodeerror

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,8 +11,8 @@ import sys
 if sys.version_info < (3,):
     sys.exit('Sorry, Python3 is required for ParlAI.')
 
-with open('README.md', 'rb') as f:
-    readme = f.read().decode()
+with open('README.md', encoding="utf8") as f:
+    readme = f.read()
 
 with open('LICENSE') as f:
     license = f.read()
--- a/setup.py
+++ b/setup.py
@@ -11,8 +11,8 @@ import sys
 if sys.version_info < (3,):
     sys.exit('Sorry, Python3 is required for ParlAI.')
 
-with open('README.md') as f:
-    readme = f.read()
+with open('README.md', 'rb') as f:
+    readme = f.read().decode()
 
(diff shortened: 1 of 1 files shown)


JackUrb:
@MannyKayy thanks for flagging - I hadn't realized our setup script was running into encoding issues on windows. Can you try using `open(README.md, encoding="utf8")` without using the `decode` call and seeing if that works as well? That would be the preferred solution.

MannyKayy:
Hi @JackUrb , yep.. that works as well, I have updated the push request. I have to highlight that this encoding issue is being raised within an Ubuntu 14.04 docker image....

MannyKayy:
It also appears that this issue is more prevalent than i initially thought.

SQuAD dataset:
image

JackUrb:
Ah my bad on the windows assumption there - last time I saw this issue it was a discrepancy there. 

As far as the rest of the unicode issues go, they're definitely related to this one, though it's good to at least get this one in as a fix for the initial setup of ParlAI. We can address the rest in the followup issue and try to work through a fix with you there.

alexholdenmiller:
Hi @MannyKayy--what do you get if you run from your terminal `echo $LANG`? We have `LANG=en_US.UTF-8` set on our platform, which prevents this problems by having python set the default codec to 'utf-8' instead of 'ascii'. We could update the code to be more reliable, but this should get you back and working right away.

MannyKayy:
Hi @alexholdenmiller ,

here is a screenshot of all locale settings. I am working within [this docker container](https://gist.github.com/MannyKayy/82c6ae9875db7095ff443a699d4007ef):

image

This pull request, in conjunction with setting the PYTHONIOENCODING environment variable solves as mentioned in #491 these encoding issues (i.e. `export PYTHONIOENCODING=utf-8`).

I have included the updated setup in the dockerfile to automate this..

PR #496, opened 2018-01-10
Pytorch teacher batch cache

klshuster:
Adds a batch sort capability to the pytorch data teacher. Accomplishes this via building up a cache of episodes with a background process, and picking out batches from the cache. 

There are two caching methods:
- 'pop' - this pops batches out of the cache into a separate "batches" list, that can be polled later once all of the batches have been loaded
-'index' - this keeps episodes in the `{length : episode_list}` dictionary, and simply grabs batches from there

### DIFF
--- a/examples/build_pytorch_data.py
+++ b/examples/build_pytorch_data.py
@@ -50,7 +50,7 @@ def build_data(opt):
             df += '.pytorch' + (agent.getID() if opt.get('pytorch_preprocess', True) else '')
         if not os.path.isfile(df):
             raise Exception('Tried to find data but it is not built, please'
-                            'specify `--pytorch_buildteacher`')
+                            'specify `--pytorch-buildteacher`')
         else:
             return df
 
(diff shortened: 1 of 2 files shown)


klshuster:
I had some numbers but found a bug, fixing that and will have tests available after

klshuster:
Some general observations of the results of the speed tests:

- Compared to not using batch sorting, the speed is very similar, i.e. minimal difference depending on different runs
- Compared to not using a pytorch data loader, we achieve faster results when training on the same batch size (i.e. 64)
- However, using batch sorting uses significantly less (i.e. half) GPU memory compared to using the pytorch data loader without batch sorting; so, we can pump up the batch size a little bit without running out of GPU memory. This allows us to achieve even faster speed.

See below for speed test times. The green row at the top is without using the PytorchDataLoader, and then everything following that is in ascending order of speed (speed is the right most column, measured as exs/sec

image

jaseweston:
cool!

PR #522, opened 2018-01-26
Messenger Agent for ParlAI

JackUrb:
Introduces a messenger agent into ParlAI and provides supporting demos.

Adapts much of the flow from the MTurk management style but cuts out all of the fat that we don't need to deal with considering we can let FB messenger handle message management and we don't need to switch views or anything.

The two demos show tasks with and without overworlds. As expected, the task without an overworld only can perform one thing (QA data collection) while the task with an overworld can do multiple (echo bot, demo of using data provided in the overworld in the task world, chat world).

Leverages Heroku to act as a passthrough server for messages being sent from facebook, but uses the send API locally to pass messages through to Facebook.

Can handle attachments, but I don't have any demos that use anything other than quick replies yet.

Provides incomplete documentation. Will update that with more details before this is committed. (Otherwise this is code-complete for the baseline features. Agent active status remains as a TODO because I don't know how much people actually want that)


docs/source/messenger.rst
```diff
@@ -0,0 +1,113 @@
...
+
+
+Task 2: Exposing People to Multiple Tasks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO: Describe this.
+
+
+Creating Your Own Task
+----------------------
```
alexholdenmiller: ha maybe worth cutting all the mturk stuff out for now--can be pasted back in from the other doc when you have time

alexholdenmiller:
great work!

docs/source/messenger.rst
JackUrb: Updating it right now!

PR #526, opened 2018-01-29
Vqa speed improvements

klshuster:
1. Adds a VQA Dataset class for use in the PytorchDataLoader to improve VQA training times.
2. Generalizes PytorchDataTeacher to allow alternative datasets for tasks and different batch preparation functions (`collate_fn`) for models
3. Minor fixes to `build_pytorch_data`, and turns off cache in `image_featurizers`
4. Changes the way that the PytorchDataLoader handles epochs - now, only one iterator over the data is created, compared to a new `iter(self.data)` for every teacher


parlai/core/agents.py
```diff
@@ -284,7 +284,10 @@ def load_agent_module(opt):
         return None
 
 def get_agent_module(dir_name):
-    if ':' in dir_name:
+    if '.' in dir_name:
+        module_name = dir_name.split(':')[0]
+        class_name = dir_name.split(':')[-1]
+    elif ':' in dir_name:
```
alexholdenmiller: what broke here?

is `len(dir_name.split(':')) > 1` when `'.' in dir_name`?

parlai/core/pytorch_data_teacher.py
```diff
@@ -326,7 +325,7 @@ def __getitem__(self, index):
                 return (index, ep)
 
     def __len__(self):
-        return self.num_eps
+        return int(self.num_episodes() * max(self.opt.get('num_epochs', 1), 1))
```
alexholdenmiller: what's this for?

parlai/core/pytorch_data_teacher.py
```diff
@@ -385,31 +386,57 @@ def __init__(self, opt, shared=None):
...
             self.dataset = shared['dataset']
             self.pytorch_dataloader = shared['pytorch_dataloader']
             self.lastYs = shared['lastYs']
+            self.data = shared['data']
 
         self.num_batches = math.ceil(self.dataset.num_episodes()/self.bsz)
         self.reset()
 
+    def get_dataset_class(self, opt):
+        dataset_name = opt.get('dataset')
```
alexholdenmiller: Can you add a docstring here? Good to provide a few examples of how to specify the dataset name

alexholdenmiller:
Nice though! Can you post the speedup here?

parlai/core/agents.py
klshuster: I tried using an agent from parlai_external, which i had to specify as `-m "parlai_external.agents.mlb.mlb:MlbAgent"`

parlai/core/pytorch_data_teacher.py
klshuster: This is intended to fix the issue that a new data-loading iterator was constructed for each teacher in the `reset_data` function; now, only one iterator is created, and does not stop iterating until the appropriate amount of examples have been returned. I will change this so that it works for validation/test (as right now it looks like it would do too many iterations)

parlai/core/pytorch_data_teacher.py
klshuster: sure!

parlai/core/agents.py
alexholdenmiller: shouldn't the module/class-name code below work for that? it looks like in both cases you should get `module_name = parlai_external.agents.mlb.mlb` and `class_name = MlbAgent`, right?

parlai/core/pytorch_data_teacher.py
```diff
@@ -418,11 +424,25 @@ def __init__(self, opt, shared=None):
...
+        """ To use a custom dataset (as opposed to the StreamDataset above),
+            you can subclass the pytorch Dataset class in a task's ``agents.py``
+            file, and then specify it in the command line.
+
+            For example, the VQA v1 task provides a custom dataset, which can
+            be specified on the command line as follows:
+            ``--dataset parlai.tasks.vqa_v1.agents:VQADataset``
+
+            Where VQADataset is the dataset defined in the vqa_v1 teacher
+            file.
```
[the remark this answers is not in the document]
klshuster: Sure, I will change that

parlai/core/agents.py
klshuster: You know, I could've sworn this wasn't working for me before, but it does in fact work without my addition. Will remove in the next commit

alexholdenmiller:
let's get those speed differences explained next!

klshuster:
New commit adds additional speedup, to 1250 ex/sec (brings it to roughly 2x slower than pytorch)

jaseweston:
yay!

klshuster:
With this most recent commit, I can achieve ~2400 ex/sec, with only 1 GPU and around 9 GB of RAM (about 12% slower than pytorch)

jaseweston:
great!

jaseweston:
how is test acc?

klshuster:
there was an issue with my machine gpu, so had to switch machines/extract the features again; will have results once that's taken care of

klshuster:
Validation set accuracy for our implementation of the model is in line with (and at times superior to) validation accuracy achieved by the pytorch implementation of the model - after training for 100 epochs with the same model configurations, our model had 57.5% validation accuracy, while the pytorch implementation had 54.1% accuracy. I will be pushing a final commit to clean up some extraneous changes that made their way into this branch and will merge after

alexholdenmiller:
great!

PR #527, opened 2018-01-30
StarSpace [baseline model]

jaseweston:


parlai/agents/starspace/modules.py
```diff
@@ -0,0 +1,84 @@
...
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+import math
+import torch
+import torch.nn as nn
+from torch.nn.parameter import Parameter
+from torch.autograd import Variable,Function
+from torch.nn.utils.rnn import pad_packed_sequence, pack_padded_sequence
+import torch.nn.functional as F
```
alexholdenmiller: can cut many of these imports (all four of the last ones except Variable, looks like?)

parlai/agents/starspace/modules.py
```diff
@@ -0,0 +1,84 @@
...
+            # testing
+            xs_enc = []
+            ys_enc = []
+            xs_emb = self.encoder(xs)
+            c_scores = []
+            for c in cands:
+                xs_enc.append(xs_emb)
+                c_emb = self.encoder2(c)
+                ys_enc.append(c_emb)
+        return torch.cat(xs_enc), torch.cat(ys_enc)
```
alexholdenmiller: maybe good to merge this since most of it is shared?

```
xs_enc = []
ys_enc = []
xs_emb = self.encoder(xs)
if ys is not None:
    xs_enc.append(xs_emb)
    ys_enc.append(self.encoder2(ys))
for c in cands:
    xs_enc.append(xs_emb)
    c_emb = self.encoder2(c)
    ys_enc.append(c_emb)
return torch.cat(xs_enc), torch.cat(ys_enc)
```

parlai/agents/starspace/starspace.py
```diff
@@ -0,0 +1,459 @@
...
+            # load dictionary and basic tokens & vectors
+            self.dict = DictionaryAgent(opt)
+
+            self.model = Starspace(opt, len(self.dict), self.dict)
+            if opt.get('model_file') and os.path.isfile(opt['model_file']):
+                self.load(opt['model_file'])
+            self.model.share_memory()
+
+        # set up modules
+        self.criterion = torch.nn.CosineEmbeddingLoss(margin=opt['margin'], size_average=False)
```
alexholdenmiller: just curious--why `size_average=False`? I don't have any intuition either way

parlai/agents/starspace/starspace.py
```diff
@@ -0,0 +1,459 @@
...
+
+    def override_opt(self, new_opt):
+        """Set overridable opts from loaded opt file.
+
+        Print out each added key and each overriden key.
+        Only override args specific to the model.
+        """
+        model_args = {'hiddensize', 'embeddingsize', 'numlayers', 'optimizer',
+                      'encoder', 'decoder', 'lookuptable', 'attention',
+                      'attention_length'}
```
alexholdenmiller: if you're keeping this may want to update these. not sure if you need anything other than `embeddingsize` (and `tfidf`?)

parlai/agents/starspace/modules.py
jaseweston: done

parlai/agents/starspace/modules.py
jaseweston: ok

parlai/agents/starspace/starspace.py
jaseweston: i think it's just a scaling factor based on batch size? actually i only  batch size 1 right now anyway..

parlai/agents/starspace/starspace.py
jaseweston: we need to overhaul this whole system. will talk to you

jaseweston: (but not for this diff)

PR #532, opened 2018-01-31
Language model

emilydinan:
1) Adds language model to parlai agents. 
2) Deletes language modeling code from seq2seq
3) Updates metrics to only report "prediction metrics" (accuracy, etc.) when predictions are present. Additionally, all custom keys are added to custom metrics so that they are counted properly.


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -350,94 +346,35 @@ def predict(self, xs, ys=None, cands=None, valid_cands=None, lm=False):
...
+            loss_dict['ppl'] = (math.e**loss).mul(len(xs)).data[0]
         else:
             self.model.eval()
             predictions, scores, text_cand_inds = self.model(xs, ys, cands,
                                                              valid_cands)
 
         return predictions, text_cand_inds, loss_dict
 
-    def batchify(self, observations, lm=False):
+    def batchify(self, observations):
```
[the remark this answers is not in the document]
emilydinan: sure -- will change the name in seq2seq and the lm

parlai/core/metrics.py
```diff
@@ -136,7 +136,7 @@ def __init__(self, opt):
         self.metrics['cnt'] = 0
         self.metrics['correct'] = 0
         self.metrics['f1'] = 0.0
-        self.custom_metrics = ['mean_rank', 'loss']
+        self.custom_metrics = []
```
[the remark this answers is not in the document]
alexholdenmiller: shoot--you're right.

sorry @emilydinan: hogwild only allows proper multithreaded sharing of pre-declared metrics. so we should keep this in, actually--but should put a comment there saying "hogwild only allows for multithreaded sharing of metrics listed here"

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,480 @@
...
+                xs = Variable(xs)
+                ys = Variable(ys)
+            data_list = [xs]
+            targets_list = [ys]
+
+        return data_list, targets_list, labels, valid_inds, end_idxs, y_lens
+
+    def batch_act(self, observations):
+        batch_reply = [{'id': self.getID()} for _ in range(len(observations))]
+        if 'labels' in observations[0]:
```
alexholdenmiller: should probably use `any(('labels' in obs for obs in observations))` to be safe

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,480 @@
...
+                self.hidden = self.model.init_hidden(self.batchsize)
+                self.is_training = False
+            data_list, targets_list, labels, valid_inds, end_idxs, y_lens = self.batchify(observations, self.opt['seq_len'], self.is_training)
+
+        if data_list is None:
+            # not enough data to batch act yet, return empty responses
+            return batch_reply
+
+        batch_reply = []
+        for i in range(len(data_list)):
```
alexholdenmiller: can you add comment along the lines of: `# data_list is a list of batches: always len 1 during eval, but could be zero or more than 1 during train`

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,480 @@
...
+        """Convert a list of observations into input & target tensors."""
+        labels = None
+        valid_inds = None
+        end_idxs = None
+        y_lens = None
+        if is_training:
+            for obs in observations:
+                if obs:
+                    if 'text2vec' in obs:
+                        self.next_batch += obs['text2vec']
```
alexholdenmiller: this would be clearer if just a list of lists

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,480 @@
...
+                data_list = []
+                targets_list = []
+                total = len(self.next_batch)//((seq_len + 1) * self.batchsize)
+                for _ in range(total):
+                    batch = self.next_batch[:((seq_len + 1) * self.batchsize)]
+                    self.next_batch = self.next_batch[((seq_len + 1) * self.batchsize):]
+
+                    source = torch.LongTensor(batch).view(self.batchsize, -1).t().contiguous()
+                    data = Variable(source[:seq_len])
+                    targets = Variable(source[1:].view(-1))
```
alexholdenmiller: better to do view before passing into the criterion instead of here--easier to visualize when debugging

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,480 @@
...
+        if type(h) == Variable:
+            return Variable(h.data)
+        else:
+            return tuple(self.repackage_hidden(v) for v in h)
+
+    def get_target_loss(self, data, hidden, targets, y_lens):
+        """Calculates the loss with respect to the targets, token by token,
+           where each output token is conditioned on either the input or the
+           previous target token.
+        """
```
alexholdenmiller: should be handled by nn.CrossEntropyLoss when padding_idx=0 and size_average=True

(unless I misunderstand what the below could is supposed to do)

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,480 @@
...
+            value, word_idx = torch.max(word_weights, 2)
+            last_words = []
+            for j in range(len(end_idxs)):
+                last_words.append(word_idx[end_idxs[j]-1][j])
+            word_idx = torch.LongTensor(last_words)
+            if self.use_cuda:
+                word_idx = word_idx.cuda()
+        else:
+            word_weights = word_weights[-1]
+            value, word_idx = torch.max(word_weights, 0)
```
alexholdenmiller: this would be a little better if we just had one branch where input would be `1 x seqlen` rather than separate code for bsz > 1

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,480 @@
...
+        word_weights = output.squeeze().data.exp()
+        # get last word of output
+        if bsz > 1:
+            value, word_idx = torch.max(word_weights, 2)
+            last_words = []
+            for j in range(len(end_idxs)):
+                last_words.append(word_idx[end_idxs[j]-1][j])
+            word_idx = torch.LongTensor(last_words)
+            if self.use_cuda:
+                word_idx = word_idx.cuda()
```
alexholdenmiller: it'd be better if we could (1) never call `cuda()` and (2) never construct a Tensor in this function. we shouldn't need to--can't we just pass word_idx directly?

it looks like `word_idx[end_idxs[j]-1][j]` might be accessing a specific element not a sub-tensor, in which case it would copy that back off the GPU to the CPU. it'd be better if we could wait to copy things back from the GPU to the CPU until we finish completely--can we just append tensors to the list, concatenate them all at the end, and then copy that back? that should run faster as well.

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,480 @@
...
+            else:
+                value, word_idx = torch.max(word_weights, 0)
+            # mark end indices for items in batch
+            for k in range(word_idx.size(0)):
+                if not done[k]:
+                    if int(word_idx[k]) == self.END_IDX:
+                        done[k] = True
+                        total_done += 1
+            token_list.append(word_idx.view(bsz, 1))
+            i += 1
```
alexholdenmiller: isn't this essentially the same as above? can't we just start with `i = 0` and cut out the previous section?

parlai/agents/language_model/modules.py
```diff
@@ -0,0 +1,83 @@
...
+                raise ValueError('When using the tied flag, nhid must be equal to emsize')
+            self.decoder.weight = self.encoder.weight
+
+        self.init_weights()
+
+        self.rnn_type = rnn_type
+        self.nhid = nhid
+        self.nlayers = nlayers
+
+    def init_weights(self):
```
alexholdenmiller: could probably remove this function--each of the layers have their own default initialization schemes and we aren't surfacing the scale of the initialization as a commandline param

parlai/agents/language_model/modules.py
```diff
@@ -0,0 +1,83 @@
...
+        initrange = 0.1
+        self.encoder.weight.data.uniform_(-initrange, initrange)
+        self.decoder.bias.data.fill_(0)
+        self.decoder.weight.data.uniform_(-initrange, initrange)
+
+    def forward(self, input, hidden, no_pack=False):
+        emb = self.drop(self.encoder(input))
+        # if eval, pack padded sequence
+        if not self.training and not no_pack:
```
alexholdenmiller: should add comment `we don't pack during training because we have no padding in our input samples`

parlai/agents/language_model/language_model.py
emilydinan: Not exactly. i = 0 is a special case, because we get all of the outputs from the input sequence and have to select the correct tokens from the end indices. I could write this as one big loop and have special cases for i=0, but i thought this was slightly more clear.

emilydinan:
@alexholdenmiller Made all the changes we discussed. Can you review again before I merge?

parlai/agents/language_model/language_model.py
```diff
@@ -371,20 +356,22 @@ def batchify(self, observations, seq_len, is_training):
             for obs in observations:
                 if obs:
                     if 'text2vec' in obs:
-                        self.next_batch += obs['text2vec']
-            if len(self.next_batch) <= (seq_len + 1) * self.batchsize:
+                        for vec in obs['text2vec']:
```
alexholdenmiller: it's a list of vectors?

parlai/agents/language_model/language_model.py
```diff
@@ -0,0 +1,469 @@
...
+        loss = 0.0
+        bsz = data.size(0)
+
+        # feed in inputs without end token
+        output, hidden = self.model(data.transpose(0,1), hidden)
+        self.hidden = self.repackage_hidden(hidden)
+        # feed in end token
+        ends = Variable(torch.LongTensor([self.END_IDX for _ in range(bsz)]).view(1, bsz))
+        if self.use_cuda:
+            ends = ends.cuda()
```
alexholdenmiller: one last thing here--I think it would be better to just initialize this Tensor once (e.g. in __init__, including pushing it to the GPU if that's set) and then you can pack it in a variable every time here instead.
e.g. https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/seq2seq/modules.py#L64

jaseweston: i'm actually confused .. are you commenting on things rewritten by @emilydinan  -- or were these in the original code she started from?

parlai/agents/language_model/language_model.py
emilydinan: yes, not sure why i was iterating through it one by one...(i think i must have gotten rid of a check there). i'll fix to just add all at once.

parlai/agents/language_model/language_model.py
emilydinan: I thought of doing that, but the problem is that "bsz" isn't always the actual batchsize. Most of the time it is, but we get rid of examples that are not "valid." I guess it will be better to initialize it once, since it is used most of the time.

alexholdenmiller: my comment is on the current code above. re: actual bsz, you can use `expand` to size it to the current batch

PR #266, opened 2017-08-16
Add free automated flake8 testing on all pull requests

cclauss:
The owner of the this repo would need to go to https://travis-ci.com/profile and flip the repository switch on to enable free automated flake8 testing on each pull request.

flake8 testing of https://github.com/facebookresearch/ParlAI on Python 3.6.2

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./parlai/agents/hred/data_iterator.py:241:45: E999 SyntaxError: invalid syntax
        print 'Data Iterator Evaluate Mode: ', self.evaluate_mode
                                            ^

./parlai/agents/hred/utils.py:151:15: F821 undefined name 'xrange'
    for dx in xrange(sizeX):
              ^

./parlai/agents/hred/utils.py:190:15: F821 undefined name 'xrange'
    for dx in xrange(sizeX):
              ^

./parlai/mturk/core/data_model.py:174:27: F821 undefined name 'unicode'
        message_content = unicode(message_content)
                          ^

./parlai/mturk/core/handler_template.py:24:34: F821 undefined name 'rds_host'
data_model.setup_database_engine(rds_host, rds_db_name, rds_username, rds_password)
                                 ^

./parlai/mturk/core/handler_template.py:24:44: F821 undefined name 'rds_db_name'
data_model.setup_database_engine(rds_host, rds_db_name, rds_username, rds_password)
                                           ^

./parlai/mturk/core/handler_template.py:24:57: F821 undefined name 'rds_username'
data_model.setup_database_engine(rds_host, rds_db_name, rds_username, rds_password)
                                                        ^

./parlai/mturk/core/handler_template.py:24:71: F821 undefined name 'rds_password'
data_model.setup_database_engine(rds_host, rds_db_name, rds_username, rds_password)
                                                                      ^

./parlai/mturk/core/handler_template.py:76:52: F821 undefined name 'frame_height'
                template_context['frame_height'] = frame_height
                                                   ^

./parlai/tasks/booktest/agents.py:68:16: F821 undefined name 'obs'
        return obs
               ^

./parlai/tasks/wikiqa/build.py:24:28: F821 undefined name 'lq'
                s = '1 ' + lq + '\t' + ans.lstrip('|') + '\t\t' + cands.lstrip('|')
                           ^

./parlai/tasks/wikiqa/build.py:24:40: F821 undefined name 'ans'
[...]

PR #556, opened 2018-02-08
parser -> argparser

cclauss:
Found via #266


cclauss:
This change is not in __master__.

alexholdenmiller:
oh looks like it is--it's just also broken on a different line... (this fixed e.g. Line 191, but Line 246 is broken)

Issue #492, opened 2018-01-10
Missing Tasks


jaseweston:
This is a list of tasks not yet in ParlAI that would be great to have. Feel free to add more to the list also!  We will remove individual items when they are done.

    * Multi-hop Reading Comprehension : https://arxiv.org/pdf/1710.06481.pdf
    * SearchQA https://github.com/nyu-dl/SearchQA
    * Twitter data used in e.g. https://arxiv.org/pdf/1603.06155v2.pdf
    * Who Did What (cloze qa): https://arxiv.org/abs/1608.05457 
    * NewsQA https://datasets.maluuba.com/NewsQA

apsdehal:
@jaseweston How about AQuA dataset by deepmind for algebraic questions? https://github.com/deepmind/AQuA
I can add support for this if it is acceptable.

I can possibly work on others too.

jaseweston:
yes, sure -- it would be great if you can add it (and any others you want)!

dfcf93:
How about the Dureader dataset by Baidu for MRC? https://github.com/baidu/DuReader

jaseweston:
sure it would be great if you can add it!

BenjaminWinter:
I would add HotpotQA https://hotpotqa.github.io/, which my group are currently working on getting into ParlAI

jaseweston:
> I would add HotpotQA https://hotpotqa.github.io/, which my group are currently working on getting into ParlAI

awesome!

ricsinaruto:
I am commenting here to let interested people know that I have a working DailyDialog implementation running on this fork: https://github.com/Mrpatekful/ParlAI/tree/dialogwae.

I will submit a PR as soon as I get around to clean the code and run some tests. I have to note though that in my implementation I only use the chat dataset without any annotations, so I guess it's not a full implementation of DailyDialog, but I would be happy to collaborate if someone were up to it. I did implement two types of tasks though: a single-turn task, and a task where all dialog history is used, for agents that can handle multiple utterances.

abisee:
There's a large new grounded dialogue dataset that came out last month:
**Conversing by Reading: Contentful Neural Conversation with On-demand Machine Reading**
https://arxiv.org/pdf/1906.02738.pdf
Might be useful?

jaseweston:
Yes, it was already in our list..

PR #559, opened 2018-02-09
Add support for AQuA dataset

apsdehal:
- Partial #492 

So there is a rationale in this dataset which explains the answer. I have currently added to answers' array as second answer. What would be the best way to handle it?

### DIFF
--- a/parlai/tasks/aqua/__init__.py
+++ b/parlai/tasks/aqua/__init__.py
@@ -1,0 +1,5 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
--- a/parlai/tasks/aqua/agents.py
+++ b/parlai/tasks/aqua/agents.py
@@ -16,6 +16,8 @@ AQUA = 'AQuA'
 AQUA_QUESTION_KEY = 'question'
 AQUA_ANSWER_KEY = 'correct'
 AQUA_OPTIONS_KEY = 'options'
+AQUA_RATIONALE_KEY = 'rationale'
+RATIONALE_QUESTION_TEXT = 'Can you provide a rationale for your answer?'
 
 
 def _path(opt):
@@ -48,9 +50,12 @@ def setup_data(path):
             question_text = question[AQUA_QUESTION_KEY]
             answer = ord(question[AQUA_ANSWER_KEY]) - ord('A')
             labels = question[AQUA_OPTIONS_KEY]
-            answer = [labels[answer], question['rationale']]
+            answer = [labels[answer]]
             yield (question_text, answer, None, labels), True
 
+            # Ask for a rationale now
+            rationale = [question[AQUA_RATIONALE_KEY]]
+            yield (RATIONALE_QUESTION_TEXT, rationale, None, rationale), False
 
 class DefaultTeacher(DialogTeacher):
     def __init__(self, opt, shared=None):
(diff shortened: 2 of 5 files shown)


jaseweston:
You can add it as another field to the observation/action dict (the message sent to the agent). But this would be optional and many models might not use it. Field name 'rationale' perhaps? I'm not sure..

jaseweston:
is this good to check in as is, or do you need to add that stuff @apsdehal ?

apsdehal:
I checked up on the paper @jaseweston. Authors expect the models to generate answer as well as the rationale (so that they can explain how to reached the answers). So, rationale and actual answer are both answers in a way. Let me figure out a way to reflect that.

jaseweston:
could be in the same text field with a '\n' or.. ?  i dont know .. is the rationale text? can you put an example here?

apsdehal:
Yes, rationale is text.
```
Of the 200 students in a school , at least 45 % attended the prom night and at least 35 % took part in the debating session . What is the maximum number of students who could have neither attended the prom night nor the debating session ?

Options: A ) 27|B ) 81|C ) 90|D ) 99|E ) 110

Rationale: To maximize the number of students who did neither , we should minimize the number of students who debated or attended the prom .
Let ' s assume that all 35 % of students who debated also attended the prom .
Then 35 % did both , 10 % only attended prom , and 55 % did neither .
0.55 * 200 = 110
The answer is E .
```

apsdehal:
We can do that, put them together using `\n`.

jaseweston:
How does one evaluate the rationale? it is not multiple choice? you have to
generate a long text like that?

apsdehal:
Yes, kind of like machine translation and evaluated using with average sentence level perplexity and blue 4 score

apsdehal:
You have two answer two things:
- One option among 5 choices for answer
- Rationale explaining your option.

jaseweston:
hmm could do as two responses though because otherwise we can't eval the multiple choice easily... the second one couldnt be a second turn of dialog ('can you explain your answer?' -> rationale reply)

apsdehal:
Yes, that sounds good. I will update the PR accordingly.

apsdehal:
@jaseweston I have updated the PR.

alexholdenmiller:
```
~~
[AQuA]: Increasing the original price of an article by 5 percent and then increasing the new price by 5 percent is equivalent to increasing the original price by
[labels: A ) 10.25 %]
[cands: A ) 10.25 %|B ) 12.5 %|C ) 14.75 %|D ) 15.15 %|E ) 16.25 %]
   [RepeatLabelAgent]: A ) 10.25 %
~~
[AQuA]: Can you provide a rationale for your answer?
[labels: 1.05 * 1.05 * x = 1.1025 * x
The answer is A .]
[cands: 1.05 * 1.05 * x = 1.1025 * x
The answer is A .]
   [RepeatLabelAgent]: 1.05 * 1.05 * x = 1.1025 * x
The answer is A .
~~
[AQuA]: How much interest can a person get on Rs . 8200 at 17.5 % p . a . simple interest for a period of two years and six months ?
[labels: C ) 3587.5]
[cands: A ) 3000|B ) 4000.75|C ) 3587.5|D ) 3500|E ) 5020.75]
   [RepeatLabelAgent]: C ) 3587.5
~~
[AQuA]: Can you provide a rationale for your answer?
[labels: I = ( 8200 * 2.5 * 17.5 ) / 100 = ( 8200 * 5 * 35 ) / ( 100 * 2 * 2 ) = Rs . 3587.50
ANSWER C]
[cands: I = ( 8200 * 2.5 * 17.5 ) / 100 = ( 8200 * 5 * 35 ) / ( 100 * 2 * 2 ) = Rs . 3587.50
ANSWER C]
   [RepeatLabelAgent]: I = ( 8200 * 2.5 * 17.5 ) / 100 = ( 8200 * 5 * 35 ) / ( 100 * 2 * 2 ) = Rs . 3587.50
ANSWER C
```

I notice that for the rationale, we provide a single candidate--the correct answer. Should we remove the candidates instead in the rationale setting?

apsdehal:
I didn't add that. It is how it is in the dataset itself and it actually changes its format with different rationale, so not under our control really.

Issue #542, opened 2018-02-06
seq2seq bidirection with attention throws code error


hsgodhia:
Hey

When using the following command on pytorch 0.3 (py3.6)
```
harshals@tesla:~/ParlAI$ python3.6 examples/train_model.py -e 50 -m seq2seq -mf godz5 -t opensubtitles -dt train:stream -hist 100 -bs 1 -tr 48 --dict-maxexs=10000000 --gpu 2 --batch-sort false -hs 512 -esz 300 -nl 2 -emb glove -lm both -bi true -att dot

```

I get the following error stack

```
[loading fbdialog data:/home/harshals/ParlAI/data/OpenSubtitles/train.txt]
Traceback (most recent call last):
  File "examples/train_model.py", line 237, in <module>
    TrainLoop(setup_args()).train()
  File "examples/train_model.py", line 201, in train
    world.parley()
  File "/home/harshals/ParlAI/parlai/core/worlds.py", line 291, in parley
    acts[1] = agents[1].act()
  File "/home/harshals/ParlAI/parlai/agents/seq2seq/seq2seq.py", line 594, in act
    return self.batch_act([self.observation])[0]
  File "/home/harshals/ParlAI/parlai/agents/seq2seq/seq2seq.py", line 541, in batch_act
    _, _, loss = self.predict(xs, ys, lm=True)
  File "/home/harshals/ParlAI/parlai/agents/seq2seq/seq2seq.py", line 390, in predict
    predictions, scores, _ = self.model(xs, ys)
  File "/usr/local/lib/python3.6/site-packages/torch/nn/modules/module.py", line 325, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/harshals/ParlAI/parlai/agents/seq2seq/modules.py", line 80, in forward
    preds, score, _h = self.decoder(xi, hidden, enc_out, attn_mask)
  File "/usr/local/lib/python3.6/site-packages/torch/nn/modules/module.py", line 325, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/harshals/ParlAI/parlai/agents/seq2seq/modules.py", line 224, in forward
    xes = self.attention(xes, hidden, encoder_output, attn_mask)
  File "/usr/local/lib/python3.6/site-packages/torch/nn/modules/module.py", line 325, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/harshals/ParlAI/parlai/agents/seq2seq/modules.py", line 486, in forward
[...]

PR #560, opened 2018-02-09
fix mismatched tensors when using bidir encoder and attention

alexholdenmiller:


alexholdenmiller:
fix for #542

examples/train_model.py
```diff
@@ -145,6 +145,8 @@ def validate(self):
         valid_report, self.valid_world = run_eval(
             self.agent, opt, 'valid', opt['validation_max_exs'],
             valid_world=self.valid_world)
+        if opt.get('model_file'):
```
jaseweston: what's this for?

alexholdenmiller: in case you want to continue training from where you stopped instead of from the best valid so far

jaseweston: it won't have self.best_valid in either case will it? or will it?
i would make this an option as it doubles the space on disk

alexholdenmiller: we save the best valid result straight to the model_file arg with nothing appended to it. I'll add a flag for last valid though!

examples/train_model.py
```diff
@@ -62,6 +62,9 @@ def setup_args(model_args=None):
                        type=float, default=1.0,
                        help='value at which training will stop if exceeded by '
                             'training metric')
+    train.add_argument('--save-last-valid', type='bool', default=False
+                       help='Saves the model to model_file.last_valid after '
+                            'every validation (overwriting previous saves).')
```
jaseweston: needs to explain the best model is still saved in model_file

jaseweston: in fact i think something  like  'save-checkpoint'  might be clearer?  then the save-every-n-secs should save there too (and call it   .checkpt or something instead?)

parlai/agents/seq2seq/modules.py
```diff
@@ -117,7 +118,8 @@ def forward(self, xs, ys=None, cands=None, valid_cands=None):
 class Encoder(nn.Module):
     def __init__(self, num_features, padding_idx=0, rnn_class='lstm',
                  emb_size=128, hidden_size=128, num_layers=2, dropout=0.1,
-                 bidirectional=False, shared_lt=None, shared_rnn=None):
+                 bidirectional=False, shared_lt=None, shared_rnn=None,
+                 sparse=False):
```
emilydinan: Why do we have this sparse argument here (and in the decoder) when it is always initialized to False?

alexholdenmiller: I'd like for it to be an option, but there are some features still missing--this was the first step towards supporting it. maybe it's too soon to commit this section...

alexholdenmiller:
fixed jason's changes, I'd like to merge for now to get this up to date (it does have a bug fix)

alexholdenmiller:
happy to revisit any changes

PR #561, opened 2018-02-09
Wikipedia task

klshuster:
Adds task that provides wikipedia articles; can either choose 'summary' for only the first paragraphs, or 'full' for the whole article.

### DIFF
--- a/parlai/tasks/task_list.py
+++ b/parlai/tasks/task_list.py
@@ -367,6 +367,6 @@ task_list = [
         "task": 'wikipedia',
         "tags": [ "All" ],
         "description": "Dump of Wikipedia articles",
-        "notes": "Specify ':full' for the full articles to be returned, otherwise defaults to ':summary', which provides the first paragraphs. To have the article title show up in the labels rather than the text, specify ':label' at the end"
+        "notes": "Specify ':full' for the full articles to be returned, otherwise defaults to ':summary', which provides the first paragraphs. To put the article in the labels and the title in the text, specify ':key-value' at the end (for a title/content key-value association)"
     },
 ]
--- a/parlai/tasks/task_list.py
+++ b/parlai/tasks/task_list.py
@@ -367,6 +367,6 @@ task_list = [
         "task": 'wikipedia',
         "tags": [ "All" ],
         "description": "Dump of Wikipedia articles",
-        "notes": "You can just get the first paragraphs of wikipedia articles by specifying ':summary', i.e. 'wikipedia:summary'; this will save you a lot of time when building the data (note this is the default option)"
+        "notes": "Specify ':full' for the full articles to be returned, otherwise defaults to ':summary', which provides the first paragraphs"
     },
 ]
--- a/parlai/tasks/task_list.py
+++ b/parlai/tasks/task_list.py
@@ -367,6 +367,6 @@ task_list = [
         "task": 'wikipedia',
         "tags": [ "All" ],
         "description": "Dump of Wikipedia articles",
-        "notes": "Specify ':full' for the full articles to be returned, otherwise defaults to ':summary', which provides the first paragraphs"
+        "notes": "Specify ':full' for the full articles to be returned, otherwise defaults to ':summary', which provides the first paragraphs. To have the article title show up in the labels rather than the text, specify ':label' at the end"
     },
 ]
--- a/parlai/tasks/task_list.py
+++ b/parlai/tasks/task_list.py
@@ -360,5 +360,13 @@ task_list = [
         "task": "twitter",
         "tags": [ "All",  "ChitChat" ],
         "description": "Twitter data from: https://github.com/Marsan-Ma/chat_corpus/. No train/valid/test split was provided so 10k for valid and 10k for test was chosen at random."
-    }
+    },
+    {
+        "id": "Wikipedia",
+        "display_name": "Wikipedia",
+        "task": 'wikipedia',
+        "tags": [ "All" ],
+        "description": "Dump of Wikipedia articles",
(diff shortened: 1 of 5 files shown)


klshuster:
Rather than returning `None` for the labels, would it be better to return something like `[title]`?

jaseweston:
to make it into a supervised task of some kind? or so that we can make retrieval key,value pairs? what's the goal? 
one can certainly have different options of different tasks if we see reasons to make them..

klshuster:
My original thought was that it would be more helpful for retrieval (i.e. the key-value pairs)

jaseweston:
In that case you would want the key as the title, or?
Again, I think having different options would solve the issue.

klshuster:
Yeah I think so; I can add that as an option then

jaseweston:
ok!

parlai/tasks/task_list.py
```diff
@@ -360,5 +360,13 @@
...
         "tags": [ "All",  "ChitChat" ],
         "description": "Twitter data from: https://github.com/Marsan-Ma/chat_corpus/. No train/valid/test split was provided so 10k for valid and 10k for test was chosen at random."
-    }
+    },
+    {
+        "id": "Wikipedia",
+        "display_name": "Wikipedia",
+        "task": 'wikipedia',
+        "tags": [ "All" ],
+        "description": "Dump of Wikipedia articles",
```
jaseweston: This needs to say *which* dump (there are many dumps of wikipedia)

parlai/tasks/wikipedia/agents.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
```
jaseweston: i think some comment about which wikipedia dump here also would help.
also what the different variants do (i.e. copy from tast_list.py)

Hilal-Urun:
Hi, I want to ask that when we are working on wikipedia task, how can we get up to date wikipedia full data or is it possible ?

klshuster:
hi there, you'd need to manually download an up-to-date wiki dump from [this link](https://dumps.wikimedia.org/enwiki/), and then run the extraction command outlined in [this message](https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/wikipedia/agents.py#L67). More information can be found in the referenced repository in the instructions

Hilal-Urun:
> hi there, you'd need to manually download an up-to-date wiki dump from [this link](https://dumps.wikimedia.org/enwiki/), and then run the extraction command outlined in [this message](https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/wikipedia/agents.py#L67). More information can be found in the referenced repository in the instructions

Thanks for response.  @klshuster 
https://github.com/facebookresearch/ParlAI/blob/3cd5a8fe5f50c51ec71882a074c2a810d995024c/parlai/tasks/wikipedia/build.py#L12 
or here Can we just replace the new wikipedia link instead of the old one?

Hilal-Urun:
I downloaded the data and train model again with this;
python3 parlai/scripts/train_model.py -t blended_skill_talk,wikipedia --multitask-weights 1,3,3,3 -eps 10  --attention-dropout 0.0 --batchsize 64  --model transformer/generator --embedding-size 2560 --ffn-size 10240 --variant prelayernorm --n-heads 32 --n-positions 128 --n-encoder-layers 2 --n-decoder-layers 24 --history-add-global-end-token end --delimiter '  ' --dict-tokenizer bytelevelbpe  --dropout 0.1 --fp16 True --init-model zoo:blender/blender_3B/model --dict-file zoo:blender/blender_3B/model.dict --label-truncate 128 --log_every_n_secs 10 -lr 7e-06 --lr-scheduler reduceonplateau --lr-scheduler-patience 3 --optimizer adam --relu-dropout 0.0 --activation gelu --model-parallel true --save-after-valid True --text-truncate 128 --truncate 128 --warmup_updates 100 --fp16-impl mem_efficient --update-freq 2 --gradient-clip 0.1 --skip-generation True -vp 10 -vmt ppl -vmm min --model-file /tmp/test_train_27B

But when it starts the training it gives the error like ;
**Ran out of memory, skipping batch. if this happens frequently, decrease batchsize or truncate the inputs to the model.**
Its train on vastai with 2x3090 . I tried to decrease the bachsize and truncate but it gives same error again and again.

klshuster:
if you're only using 2 gpus to train the blender_3B model, you'll need to significantly reduce the batchsize; I just tried your above command on 2x32gb gpus with batchsize 16 and it seemed to work for me

Hilal-Urun:
> if you're only using 2 gpus to train the blender_3B model, you'll need to significantly reduce the batchsize; I just tried your above command on 2x32gb gpus with batchsize 16 and it seemed to work for me

Yes I tried to reduce batch size to 16 and than 8 but its still gave same error but okey I'm gonna try with 2x32 . Thanks a lot  :)

Hilal-Urun:
This time I get **RuntimeError: CUDA error: CUBLAS_STATUS_ALLOC_FAILED** when calling `cublasCreate(handle)` this error

Hilal-Urun:
Hi @klshuster  I have changed just https://github.com/facebookresearch/ParlAI/blob/3cd5a8fe5f50c51ec71882a074c2a810d995024c/parlai/tasks/wikipedia/build.py#L12 
link of this from my dataset from enwiki, but when downloading process starts, it gets the link from parlai modüle which is same as above. How can I handle this?

klshuster:
Hi @Hilal-Urun, would you mind opening up a separate issue in the repo (here: https://github.com/facebookresearch/ParlAI/issues), tagging me, and including a full stack trace of the issue?

PR #562, opened 2018-02-11
Opensubtitles 2018

urikz:
**Overview**
This PR add new data source - OpenSubtitles 2016. It contains way more data than the original, 2009 version of OpenSubtitles: 85M vs 2M examples.

The extraction process is a little more difficult for following reasons:
1. Unlike the OpenSubtitles2009, OpenSubtitles2016 contains several subtitles files per same movie. Thus, we need to deduplicate it properly (movie based), as well as make sure the split between train/valid/test is done correctly.
2. The tokenization is not not completely compatible with nltk. For example, the apostrophe is often separated from both left and right part of the word, which is inconsistent with nltk (and nltk alone cannot fix it).
3. The data in general is more dirty, thus more preprocessing is required.
4. The amount of data is enormous, so extraction needs to be parallelized to be efficient.

**Test plan**

```
$ python examples/display_data.py -t opensubtitles_2016 -n 100
[ optional arguments: ]
[  num_examples: 100 ]
[ Main ParlAI Arguments: ]
[  task: opensubtitles_2016 ]
[  download_path: /home/urikz/projects/ParlAI/downloads ]
[  datatype: train ]
[  image_mode: raw ]
[  numthreads: 1 ]
[  datapath: /home/urikz/projects/ParlAI/data ]
[ Batching Arguments: ]
[  batchsize: 1 ]
[creating task(s): opensubtitles_2016]
[building data: /home/urikz/projects/ParlAI/data/OpenSubtitles2016]
unpacking OpenSubtitles2016.tar.gz
[building fbformat]
[Found 106248 *.xml.gz movies within /home/urikz/projects/ParlAI/data/OpenSubtitles2016/OpenSubtitles2016/xml/en in 19 seconds]
[Data has been successfully extracted in 6115 seconds]103.8KiB
[loading fbdialog data:/home/urikz/projects/ParlAI/data/OpenSubtitles2016/train.txt]
[loaded 22526 episodes with a total of 170104864 examples]
[reward: 0]
[opensubtitles_2016]: and chocolate cake was 10 , rate these cheeses .
[labels: what 's worse than dog shit ?]
   [RepeatLabelAgent]: what 's worse than dog shit ?
~~
...
```

Do you think I should train some models to compare the performance?


urikz:
Run OpenSubtitles 2009 version 
`$ python examples/display_data.py -t opensubtitles:v2009 -n 100`

Note, that we don't need a FullTeacher in OpenSubtitles2018, because it's already done in the preprocessing. And we need to have it internally in order to deduplicate message/response pairs generated for the same movie, but from different subtitles files.

urikz:
I've added support to these versions:
The ones, which existed before
```
v2009_half
v2009
v2009_task100k
v2009_task10k
```
Their 2018 analogues
```
v2018_half
v2018
v2018_task100k
v2018_task10k
```

Two-turn dialogs from Open Subtitles 2018. Note, we deduplicate dialogs for the same movie from different subtitles files. Naturally, deduplication works better if we treat each message/response as a separate dialog. The amount of generated data will be smaller (and less skewed). 
```
v2018_no_history
v2018_no_history_task100k
v2018_no_history_task10k


```

jaseweston:
awesome... @alexholdenmiller ?

emilydinan:
Hi @urikz -- thanks again. Looks good to me, but one small change I would make is that opensubtitles:full and opensubtitles:half fail because "version" and "use_history" do not default to anything. I would set defaults for these if you are going to include these teachers in this way. (Perhaps set the default version to 2018 and use_history to True.)

emilydinan:
Hi @urikz -- thanks again. Looks good to me, but one small change I would make is that opensubtitles:full and opensubtitles:half fail because "version" and "use_history" do not default to anything. I would set defaults for these if you are going to include these teachers in this way. (Perhaps set the default version to 2018 and use_history to True.)

urikz:
@emilydinan, you mean, I should rename "v2018" and "v2009" to "v2018:full" and "v2009:full"? Sure thing!

emilydinan:
@urikz -- no, I actually meant that since there is a "FullTeacher" in there and a "HalfTeacher" that we should be able to specify "-t opensubtitles:full" or "-t opensubtitles:half" (both of which would default to some version/use_history of opensubtitiles without being specified). Does that make sense? Otherwise, maybe we could just add a comment clarifying that these are just helper teachers and can't actually be called directly.

urikz:
got it, so I'll add two more versions "half" and "full", which will correspond to "v2018_half" and "v2018".

emilydinan:
@urikz still not quite what I meant actually. If I were to type `python display_data.py -t opensubtitles:full` in the command line, it fails because `TypeError: __init__() missing 2 required positional arguments: 'version' and 'use_history'` All I'm saying is that you should specify defaults for 'version' and 'use_history' here: https://github.com/urikz/ParlAI/blob/bf82395be4676fb095f22148c6ffa97b974a0b98/parlai/tasks/opensubtitles/agents.py#L32 . The same is true for opensubtitles:half. Sorry for the confusion!

emilydinan:
Hi @urikz -- I'm noticing some behavior where the parsing seems to cut off the first word. For example, 

```
~~
[reward: 0]
[opensubtitles:v2018]: 'm fine .
[labels: you see to the cases .]
   [RepeatLabelAgent]: you see to the cases .
~~
[reward: 0]
[opensubtitles:v2018]: 'd break with all the child represents ?
[labels: yes .]
   [RepeatLabelAgent]: yes .
~~
[reward: 0]
[opensubtitles:v2018]: , it 's at home .
[labels: go and get it , then come to the consulate .]
   [RepeatLabelAgent]: go and get it , then come to the consulate .
~~
```

This is from running  `python examples/display_data.py -t opensubtitles:v2018 -dt train:stream`. Could you point me to where in your code you think this might be happening?

emilydinan:
@urikz after looking into it a bit more, I see that the problem is that V2018 teacher inherits from FbDialogTeacher, which deletes the first character because it expects the text to be in "fb_format" (i.e. the text/labels are prepended with some index). See here: https://github.com/facebookresearch/ParlAI/blob/ce1e33ad5f015c5948f64e2287d9ff6ccb8105fb/parlai/core/teachers.py#L922

urikz:
Thank you @emilydinan for approving the request! 

Btw, I see the problem that it takes a lot of time just to load the dataset into memory. However, what I did in my project, is I numberize corpus in advance and save it as one giant numpy array (plus two index arrays to remember start/end positions of every utterance). This way loading of this several GB of date takes 22 seconds (could possibly even faster with memory mapping). I can incorporate something like this into ParlAI. Would that be hard? Do you think it makes sense?

jaseweston:
you can train (or dispay) with "train:stream" as well..

urikz:
@jaseweston, I apologize for stupid question, but can train:stream create batches on the fly with the examples of the more less same length? (I saw something like this in ParlAI under "sort batches", but I'm not sure if that's exactly it).

klshuster:
Hi @urikz, you can utilize our [PytorchDataTeacher](https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/pytorch_data_teacher.py) to both stream the data and create batches on the fly with examples of the same length. The docstring has instructions on how to use it, but let me know if you have any questions about it.

To use the batch-sorting option, specify the option `--batch-sort-cache pop` or `--batch-sort-cache index`, and set `--batch-length-range` to any number you feel is good for differences in example length (i.e. `5` would indicate that examples in a batch will differ by at most length 5)

PR #563, opened 2018-02-12
txt2vec with utils, fix in metrics

ekmb:


parlai/core/utils.py
```diff
@@ -243,9 +247,9 @@ def parse(txt):
                 history['dialog'].extend(parse(reply))
         elif len(history['labels']) > 0:
             r = history['labels'][0]
-            history['dialog'].extend(parse(r))
+            history['dialog'].extend(parse(r, True))
```
alexholdenmiller: `True` => `splitSentences`?

ekmb: Oh, yes thanks for catching this!

PR #564, opened 2018-02-12
Change default max examples in Dictionary Agent

emilydinan:
Changed the default max examples for the DictionaryAgent to process to infinite.


jaseweston:
i think u might need to check some demo examples still work quickly?

emilydinan:
@jaseweston I checked the examples and all ran quickly with the exception of training DrQA on SQUAD (takes about 2 minutes to load dictionary as opposed to 1:45 with max examples set to 100,000). Not super noticeable since it's slow to begin with due to spacy parsing. Do you think I should set the flag there to max_exs=100,000?

PR #565, opened 2018-02-13
Added max_sorted_size flag to Dictionary Agent

emilydinan:
1) Added max sorted size flag to Dictionary Agent which trims dictionary to a set number of tokens after it has been sorted.
2) Changed line in examples README for consistency (all other examples are assumed to be run from examples directory)

### DIFF
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -68,6 +68,7 @@ class DictionaryAgent(Agent):
     default_lang = 'english'
     default_maxngram = -1
     default_minfreq = 0
+    default_maxsortedsize = -1
     default_null = '__NULL__'
     default_start = '__START__'
     default_end = '__END__'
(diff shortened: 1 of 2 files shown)


parlai/core/dict.py
```diff
@@ -99,6 +100,10 @@ def add_cmdline_args(argparser):
             '--dict-minfreq', default=DictionaryAgent.default_minfreq,
             type=int,
             help='minimum frequency of words to include them in sorted dict')
+        dictionary.add_argument(
+            '--dict-maxsortedsize', default=DictionaryAgent.default_maxsortedsize,
```
alexholdenmiller: maybe just `--dict-maxtokens`? @jaseweston

jaseweston: yes maxtokens is fine although im not sure how we do our spacing

btw, the switch between "_"  and "-" which i guess was alex's decision? gets me like several times a day..

parlai/core/dict.py
```diff
@@ -416,6 +431,7 @@ def sort(self):
             new_ind2tok[i] = tok
         self.tok2ind = new_tok2ind
         self.ind2tok = new_ind2tok
+        self.resize_to_max(self.maxtokens)
```
jaseweston: kind of weird to ever load them in the first place, but ok for now i guess?

PR #569, opened 2018-02-13
Metrics fix

emilydinan:
Fixes the metrics so that:
1) Each metric has an individually updated count
2) Hits@ only prints if text cands exist

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -249,7 +249,7 @@ class Metrics(object):
         if total > 0:
             if self.flags['print_prediction_metrics']:
                 m['accuracy'] = round_sigfigs(self.metrics['correct'] / self.metrics['correct_cnt'], 4)
-                m['f1'] = round_sigfigs(self.metrics['f1'] /self.metrics['f1_cnt'], 4)
+                m['f1'] = round_sigfigs(self.metrics['f1'] / self.metrics['f1_cnt'], 4)
                 if self.flags['has_text_cands']:
                     m['hits@k'] = {}
                     for k in self.eval_pr:
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -142,12 +142,10 @@ class Metrics(object):
         for k in self.eval_pr:
             self.metrics['hits@' + str(k)] = 0
         self.metrics['hits@_cnt'] = 0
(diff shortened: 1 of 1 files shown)


parlai/core/metrics.py
```diff
@@ -247,23 +250,22 @@ def report(self):
         m['total'] = total
         if total > 0:
             if self.print_prediction_metrics or self.metrics['f1'] > 0:
```
jaseweston: why is f1 > 0 in there?

alexholdenmiller: so the self.print_prediction_metrics is only there to make it so that if f1 == 0 because the model got everything completely wrong, then it will still print.

f1 > 0 is there because that flag is not shared during hogwild, and we definitely always want to enter this section when f1 > 0. We unfortunately aren't entering when f1 is zero but the model is trying, but that fix would be more complex and I didn't want @ ekmb to do that for now.

parlai/core/metrics.py
```diff
@@ -134,19 +134,19 @@ class Metrics(object):
...
             self.metrics[k + '_cnt'] = 0
         self.eval_pr = [1, 5, 10, 100]
         for k in self.eval_pr:
             self.metrics['hits@' + str(k)] = 0
+        self.metrics['hits@_cnt'] = 0
         if opt.get('numthreads', 1) > 1:
             self.metrics = SharedTable(self.metrics)
 
         self.datatype = opt.get('datatype', 'train')
+        self.has_text_cands = False
```
alexholdenmiller: note that these flags aren't shared between threads. we *could* store these in an internal field called "flags" which is a dict or shared table to share them. it looks messier, but if we are adding more then we probably should.

alexholdenmiller: @jaseweston thoughts?

parlai/core/metrics.py
jaseweston: i dont understand sorry. metrics has got too complicated lol.
under what circumstances do we not want to print?

jaseweston: i don't understand, but hopefully you know what you are doing :)
why does this has to depend on f1, and not some other metric?

parlai/core/metrics.py
emilydinan: @jaseweston @alexholdenmiller latest commit creates a SharedTable (in hogwild case) so that the flags are shared between threads if necessary

PR #572, opened 2018-02-14
fix map_predictions

### LINKED ISSUE #566: Bug report.
https://github.com/facebookresearch/ParlAI/blob/c6f03dc8d8f1df8a237542ce76e7d9be8d9e338d/parlai/core/utils.py#L435
This line, comparing Torch Variable to Int type, will issue this error: 
_RuntimeError: bool value of Variable objects containing non-empty torch.ByteTensor is ambiguous._

Modifying the former line as: `for c in predictions[i].data:` will solve this problem.

klshuster:
Fixes #566 by extracting data from predictions variable

### DIFF
--- a/parlai/core/utils.py
+++ b/parlai/core/utils.py
@@ -209,7 +209,7 @@ def make_batches(data, bsz):
 
 def maintain_dialog_history(history, observation, reply='',
                             historyLength=1, useReplies="labels",
-                            dict=None, useStartEndIndices=True, 
+                            dict=None, useStartEndIndices=True,
                             splitSentences=False):
     """Keeps track of dialog history, up to a truncation length.
     Either includes replies from the labels, model, or not all using param 'replies'."""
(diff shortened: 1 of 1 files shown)


parlai/core/utils.py
```diff
@@ -424,7 +424,7 @@ def map_predictions(cls, predictions, valid_inds, batch_reply, observations, dic
            using valid_inds.
            report_freq -- how often we report predictions
         """
-        predictions = predictions.cpu()
+        predictions = predictions.cpu().data
```
jaseweston: does there need to be some check whether you are cpu or gpu or this doesnt affect performance for cpu?

klshuster: I'm pretty sure this is a no-op if the variable is already on CPU; torch docs say that the method `Returns a CPU copy of this tensor if it's not already on the CPU`

jaseweston:
OK!

PR #574, opened 2018-02-14
Fix map_predictions to be compatible with LM as well as seq2seq

emilydinan:


alexholdenmiller:
cool for now

jaseweston:
i hope all the other models still work..

klshuster:
As far as i know, this function is currently only used by seq2seq and LM, so it should be fine for now

emilydinan:
@jaseweston yes, definitely only used by seq2seq and LM right now. going to refactor code later so that it just takes an iterable, just need a quick fix for now.

PR #581, opened 2018-02-15
Add non-python files in ParlAI package

### LINKED ISSUE #576: Can we keep a mturk task from outside parlai/mturk/tasks?
Hi @JackUrb, I have a few questions regarding the mturk evaluation:

1. This link (http://parl.ai/static/docs/mturk.html#running-a-task) says that 

> to run an MTurk task, first ensure that the task directory is in `parlai/mturk/tasks/`. 

 Is it by design? I tried to keep my task in some other directory (outside root parlai directory) and tried to import parlai as a package but that doesn't seem to work. Basically I am trying to use ParlAI as one of the dependency for one of my project for Human In the loop evaluation.

2. How easy/hard it is to provide support for keeping the mturk task anywhere?

deshraj:
Fixes #576.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,3 @@
+include parlai/mturk/core/server/html/*
+include parlai/mturk/core/server/server.js
+include parlai/mturk/core/server/package.json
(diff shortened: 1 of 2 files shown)


JackUrb:
Totally forgot to add these into the package, makes sense for why they wouldn't work outside the working directory. Thanks for the fix!

PR #588, opened 2018-02-22
Learning rate scheduler for language model

emilydinan:
1. Adds a function "receive_metrics" for Agent class which allows the agent to receive a dictionary
2. Calls this function after validation during the train_model loop
3. Implements this function in language_model so that the learning rate is decreased by a factor when the validation loss does not decrease. Defaults to 0.5
4. Also makes small change to "get_predictions" function in language_model so that it handles case when predictions are empty (this happens when there are no 'valid' examples in the batch)

### DIFF
--- a/parlai/agents/language_model/language_model.py
+++ b/parlai/agents/language_model/language_model.py
@@ -79,6 +79,7 @@ class LanguageModelAgent(Agent):
         self.use_cuda = not opt.get('no_cuda') and torch.cuda.is_available()
         self.batchsize = opt.get('batchsize', 1)
         self.use_person_tokens = opt.get('person_tokens', True)
+        self.valid_metrics = None
 
         if shared:
             # set up shared properties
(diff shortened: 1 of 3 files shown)


parlai/agents/language_model/language_model.py
```diff
@@ -487,6 +494,15 @@ def shutdown(self):
...
         super().shutdown()
 
+    def receive_metrics(self, metrics_dict):
+        if 'loss' in metrics_dict:
+            if self.best_val_loss is None:
+                self.best_val_loss = metrics_dict['loss']
+            else:
+                if metrics_dict['loss'] > self.best_val_loss:
+                    self.lr *= self.lr_factor
+                    print("Updating learning rate: lr =", self.lr)
```
alexholdenmiller: do you want to use loss or something like `opt.get('valid_metric', opt.get('accuracy'))`?

alexholdenmiller: nevermind

PR #589, opened 2018-02-23
Fix bug where best_val_loss is not saved in language model

emilydinan:
Fixes bug where best_val_loss is not saved in language model. Also saves to model file.

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -78,6 +78,9 @@ def setup_args(model_args=None):
     train.add_argument('-dbf', '--dict-build-first',
                        type='bool', default=True,
                        help='build dictionary first before training agent')
+    train.add_argument('-lfc', '--load-from-checkpoint',
+                       type=bool, default=False,
+                       help='load from checkpoint if available')
     return parser
 
 def run_eval(agent, opt, datatype, max_exs=-1, write_log=False, valid_world=None):
@@ -140,6 +143,8 @@ class TrainLoop():
             build_dict(opt)
         # Create model and assign it to the specified task
         self.agent = create_agent(opt)
+        if opt['load_from_checkpoint'] and opt.get('model_file') and os.path.is_file(opt['model_file'] + '.checkpoint'):
+            self.agent.load(opt['model_file'] + '.checkpoint')
         self.world = create_task(opt, self.agent)
         self.train_time = Timer()
         self.validate_time = Timer()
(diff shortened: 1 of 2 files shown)


examples/train_model.py
```diff
@@ -67,6 +68,9 @@ def setup_args(model_args=None):
     train.add_argument('-vmt', '--validation-metric', default='accuracy',
                        help='key into report table for selecting best '
                             'validation')
+    train.add_argument('-vopt', '--validation-metric-optim', default='max',
```
alexholdenmiller: `--validation-metric-mode` I think is better

examples/train_model.py
```diff
@@ -157,15 +172,21 @@ def validate(self):
         if opt.get('model_file') and opt.get('save_after_valid'):
             print("[ saving model checkpoint: " + opt['model_file'] + ".checkpoint ]")
             self.agent.save(opt['model_file'] + '.checkpoint')
+            self.world.save_agents()
+            if self.best_valid is not None:
+                print("[ saving best valid metric: " + opt['model_file'] + ".checkpoint.best_valid ]")
+                save_best_valid(opt['model_file'] + '.checkpoint', self.best_valid)
```
alexholdenmiller: I think maybe better to not save here. when loading a checkpoint, you should probably `cp/mv model.checkpoint model` and then run again, or else the next run will be saving to model.checkpoint.checkpoint and so on... not good. this also makes it look like maybe saved model at model.checkpoint got the model.checkpoint.best_valid score, which isn't true.

(so let's remove these four lines. also just realized that line 175 is unfortunately saving over the best valid model.)

examples/train_model.py
```diff
@@ -233,6 +254,9 @@ def train(self):
                 if self.save_time.time() > self.save_every_n_secs and opt.get('model_file'):
                     print("[ saving model checkpoint: " + opt['model_file'] + ".checkpoint ]")
                     self.agent.save(opt['model_file'] + '.checkpoint')
+                    if self.best_valid is not None:
+                        print("[ saving best valid metric: " + opt['model_file'] + ".checkpoint.best_valid ]")
+                        save_best_valid(opt['model_file'] + '.checkpoint', self.best_valid)
```
alexholdenmiller: can also remove this like above

examples/train_model.py
alexholdenmiller: (matches the arg name from optim)

PR #592, opened 2018-02-23
drqa bug with contiguous

alexholdenmiller:
ignore the word_dict change, it's already committed


alexholdenmiller:
merging, just some `contiguous`  calls

PR #593, opened 2018-02-23
add lr scheduling to s2s (and report loss during validation), opensubtitles bug fix

alexholdenmiller:
- small logging fix in train_model
- reduce learning rate on plateau for s2s
- report loss / ppl during valid as well as training now
- opensubtitles keyword params were misordered


alexholdenmiller:
I'll start with this for now, maybe later

PR #595, opened 2018-02-23
Load from checkpoint

emilydinan:
1. Adds option to load from checkpoint in the train_model example; models are still saved to original model file path
2. Gets rid of the "save_agents" function in worlds.py and replaces calls to this function with self.agent.save()

### DIFF
--- a/parlai/agents/drqa/config.py
+++ b/parlai/agents/drqa/config.py
@@ -95,7 +95,7 @@ def set_defaults(opt):
 
     # Make sure fix_embeddings and embedding_file are consistent
     if opt['fix_embeddings']:
-        if not opt.get('embedding_file') and not opt.get('pretrained_model'):
+        if not opt.get('embedding_file') and not opt.get('init_model'):
             print('Setting fix_embeddings to False as embeddings are random.')
             opt['fix_embeddings'] = False
 
--- a/parlai/agents/drqa/config.py
+++ b/parlai/agents/drqa/config.py
@@ -16,7 +16,7 @@ def add_cmdline_args(parser):
     # Basics
     agent.add_argument('--embedding_file', type=str, default=None,
                         help='File of space separated embeddings: w e1 ... ed')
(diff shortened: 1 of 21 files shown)


parlai/agents/drqa/drqa.py
```diff
@@ -146,10 +146,6 @@ def _init_from_saved(self, fname):
         saved_params = torch.load(fname,
             map_location=lambda storage, loc: storage)
 
-        if 'word_dict' in saved_params:
-            # for compatibility with old saves
-            self.word_dict.copy_dict(saved_params['word_dict'])
-
```
alexholdenmiller: this section needs to stay

parlai/agents/drqa/config.py
```diff
@@ -16,7 +16,7 @@ def add_cmdline_args(parser):
     # Basics
     agent.add_argument('--embedding_file', type=str, default=None,
                         help='File of space separated embeddings: w e1 ... ed')
-    agent.add_argument('--pretrained_model', type=str, default=None,
```
alexholdenmiller: why change this?

alexholdenmiller:
fixed the merge conflict with my diff

parlai/agents/drqa/drqa.py
emilydinan: was not deleted on purpose... thanks for catching!

parlai/agents/drqa/config.py
emilydinan: the new train loop sets opt['init_model'] so that we can load from checkpoint, and @jaseweston suggested to name the parameter 'init_model' instead of 'pretrained_model'

emilydinan:
Most recent changes:
1. Add 'init_model' flag to seq2seq and memnn
2. Fix bugs in memnn related to initialization

alexholdenmiller:
fixed merge conflict with my diff

PR #597, opened 2018-02-26
Add more learning rate scheduling features to language model

emilydinan:
Adds the following features for better learning rate scheduling for the language model
- learning rate patience 
- learning rate minimum


emilydinan:
@alexholdenmiller yeah, I agree... I'll add that before merging

Issue #596, opened 2018-02-24
Bug report


Jackberg:
1. There's a bug when using attention layer. In this line:
https://github.com/facebookresearch/ParlAI/blob/55fcf6127309f3c0e2f15c1fe6eae1fd71afcbcb/parlai/agents/seq2seq/modules.py#L80
new `hidden` states are returned, but never used for getting next prediction. This is the reason why attention model performs extremely bad.  Here's the result for just 30 mins training:
```
TEXT:  I get to read the articles of extradition acordind to the European Court of human rights .
PREDICTION:  i was just a little bit of a lot of people .
~
TEXT:  Yes , you are the very monster I created
PREDICTION:  i will be a good thing
~
TEXT:  Hello , detective Spooner .
PREDICTION:  i don' t know .
~
TEXT:  I' m a tiger .
PREDICTION:  i don' t know .
~
TEXT:  What' ve you got ?
PREDICTION:  i don' t know .
~
TEXT:  We are going to change the way we see the road .
PREDICTION:  i don' t know what you are .
```
What's more, attention model (using `local` for Twitter and `general` for Opensubtitles) can really make loss lower.

2. The default value of lookuptable
https://github.com/facebookresearch/ParlAI/blob/55fcf6127309f3c0e2f15c1fe6eae1fd71afcbcb/parlai/agents/seq2seq/seq2seq.py#L107
will cause much more memory usage, but I didn't find out the reason. Old value `all` works fine.

3. In this line of `vectorize()` function,
https://github.com/facebookresearch/ParlAI/blob/55fcf6127309f3c0e2f15c1fe6eae1fd71afcbcb/parlai/agents/seq2seq/seq2seq.py#L403
it only returns 6 values, but newer version needs 7.

alexholdenmiller:
1) wow, thanks. that was some copypasta from the line above it but was really hurting training with attention. thanks for the catch.

2) `unique` uses 3x more memory than `all`, intentionally. `all` shares the same tensor for the `weight` of the encoder Embedding layer, the decoder Embedding layer, and the final Linear layer producing an output token. `unique` keeps them separate, and `enc_dec` and `dec_out` share the mentioned pairs.

3) fixing, thanks.

Jackberg:
> unique uses 3x more memory than all, intentionally. 
I see... But 3x more is a tremendous amount. Any methods to reduce memory usage?

alexholdenmiller:
[...]

PR #598, opened 2018-02-26
fix attention

alexholdenmiller:
#596

### DIFF
--- a/parlai/agents/seq2seq/modules.py
+++ b/parlai/agents/seq2seq/modules.py
@@ -231,6 +231,9 @@ class Decoder(nn.Module):
         xes = F.dropout(self.lt(xs), p=self.dropout, training=self.training)
         if self.attn_time == 'pre':
             xes = self.attention(xes, hidden, encoder_output, attn_mask)
+        if xes.dim() == 2:
+            # if only one token inputted, sometimes needs unsquezing
+            xes.unsqueeze_(1)
         output, new_hidden = self.rnn(xes, hidden)
         if self.attn_time == 'post':
             output = self.attention(output, new_hidden, encoder_output, attn_mask)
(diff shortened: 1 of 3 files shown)


alexholdenmiller:
added post-attention as well, which appears to do better more frequently. (instead of modifying the embeddings of the inputs to the rnn, it modifies the outputs of the rnn before they are fed to the output layer, as described in )

with a few runs each (so there's variability in sampling...), babi task 19, with validation every 30 secs and an impatience of 10 (so not very patient).

no attention: 9% valid, 8% test || 17% valid, 14% test || 9% valid, 8% test

general pre-attention: 12% valid, 10% test || 13% valid, 12% test || 10% valid, 9% test 
general post-attention: 77% valid, 74% test || 80% valid, 81% test || 11% valid, 9% test || 7% valid, 8% test

dot pre-attention: 25% valid, 26% test || 12% valid, 11% test || 12% valid, 9% test
dot post-attention: 52% valid, 53% test || 12% valid, 10% test || 11% valid, 11% test || 10% valid, 11% test || 12% valid, 11% test || 41% valid, 41% test

concat pre-attention: 10% valid, 8% test || 12% valid, 10% test || 9% valid, 10% test || 9% valid, 8% test
concat post-attention: 14% valid, 13% test || 39% valid, 38% test || 12% valid, 9% test

in hindsight, there's more variability here than I expected, but post-attention appears to be the most reliable for getting above random chance. running on personachat now.

parlai/agents/seq2seq/modules.py
```diff
@@ -434,15 +442,23 @@ def forward(self, input):
 
 class AttentionLayer(nn.Module):
     def __init__(self, attn_type, hidden_size, emb_size, bidirectional=False,
-                 attn_length=-1):
+                 attn_length=-1, attn_time='pre'):
         super().__init__()
         self.attention = attn_type
 
         if self.attention != 'none':
             hsz = hidden_size
```
emilydinan: why do we rename this variable here?

alexholdenmiller: it made the lines below a lot shorter :)

alexholdenmiller:
quick note: after ~15 min, dot post-attention is producing pretty decent looking personachat sentences already, with no attetion, general post, and concat post all looking pretty poor.

alexholdenmiller:
general post looks better on the 4th validation, and oops meant 30 min not 15

PR #599, opened 2018-02-26
MLB Model for VQA datasets

klshuster:
An implementation of the 'Multimodal Low-Rank Bilinear Attention Network' as outlined in https://arxiv.org/abs/1610.04325. Pytorch modules have been adapted from [here](https://github.com/Cadene/vqa.pytorch) (an implementation maintained by @cadene). 

Validation accuracy over 100 epochs of training the model in ParlAI is in line with results obtained via training the model from [here](https://github.com/Cadene/vqa.pytorch) over the same time frame and with the same configuration. That is, with a batch size of 512 and using the attention strategy, we get 57.5% accuracy in ParlAI vs. 54% accuracy [here](https://github.com/Cadene/vqa.pytorch). 

Instructions on how to best run the model within ParlAI (including how to extract images, and how to use the `PytorchDataTeacher`) can be found in the docstring of `mlb.py`


jaseweston:
cool! maybe call it mlb_vqa so that its more clear what it is?

alexholdenmiller:
would be great to also add a small unit test (like the train_model test) even just running a single example (or single small batch) through the model works

klshuster:
I've obtained some benchmarks for results after implementing suggestions from @cadene; to summarize:

1. Changing the tokenization max length to 16 and removing the 'pad' parameter led to a huge speedup with no attention (nearly ~2x), and a slight speedup with attention.
2. Using the correct attention implementation resulted in worse validation score performance (54% vs 57%); thus, I will be keeping the old attention implementation as the default

PR #601, opened 2018-02-27
add ibm seq2seq model, some other small changes

alexholdenmiller:
- added ibm's sequence to sequence model
- changed the padutils to have keyword args for null and end idx (if you dont include the endidx, it won't append it to the targets)

some results on babi:task10k:1 with this seq2seq vs the one at parlai/agents/seq2seq, listed as `num_epochs e: time_taken s / test accuracy %`

```
ibm_s2s (otherwise default params)
(lr 0.005)
10e: 28.5s / 50%
10e: 29.5s / 51%
20e: 57.7s / 47%
20e: 56.8s / 51%
30e: 88.2s / 50%
30e: 86.3s / 50%
(lr 0.001)
30e: 87.5s / 81%
30e: 87.8s / 89%
40e: / 81%
(lr 0.0005)
30e: 89.2s / 83%
30e: 89.1s / 82%
30e: 88.4s / 84%
(lr 0.0001)
30e: 86.9s / 53%
(lr 0.00005)
30e: 87.1s / 52%

ahm_s2s (all default params)
10e: 26.3s / 84%
10e: 24.7s / 58%
20e: 49.6s / 84%
20e: 52.1s / 100%
30e: 79.4s / 96%
30e: 75.8s / 100%
```

in addition to that, training with validation every 30 seconds for an unlimited number of epochs showed the ibm_s2s reach nearly 100% training accuracy but with valid accuracy stagnating. When I set dropout to 50% (runs above were 10%) with lr=0.001, it was eventually able to break 95% valid acc, with resulting 94% test acc. I'll have 50% be the default for this model.

Here are other parameters in case the defaults change:
```
[ IBM Seq2Seq Arguments: ]
[  init_model: None ]
[  hiddensize: 128 ]
[  embeddingsize: 128 ]
[  numlayers: 2 ]
[  gradient_clip: -1 ]
[  bidirectional: False ]
[  attention: True ]
[  maxlength_in: 100 ]
[  maxlength_out: 2 ]
[  no_cuda: False ]
[  gpu: -1 ]
[  truncate: -1 ]
[  rnncell: gru ]
[  optimizer: adam ]
```


parlai/agents/language_model/language_model.py
```diff
@@ -444,7 +444,10 @@ def vectorize(self, observations, seq_len, is_training):
         else:
             # here we get valid examples and pad them with zeros
             xs, ys, labels, valid_inds, _, y_lens = PaddingUtils.pad_text(
-                observations, self.dict, self.END_IDX, self.NULL_IDX)
+                observations, self.dict, end_idx=self.END_IDX,
+                null_idx=self.NULL_IDX)
+            if xs is None:
+                return None, None, None, None, None
```
emilydinan: I think you are working on an old branch because I have a check for when xs is None in the current version. It's slightly different than just returning all None (returns [None]). So you should get rid of my checks when you merge this.

emilydinan: See here: https://github.com/facebookresearch/ParlAI/blob/cdf5adee9b07c675d9eb67456c3b0d9d7eeded0e/parlai/agents/language_model/language_model.py#L448

PR #602, opened 2018-02-27
[WIP] adding tensorflow seq2seq

alexholdenmiller:
still having some difficulties with evaluating the tensorflow model using ParlAI's metrics, but you can use the tensorflow metrics (e.g. bleu, etc) in the current version


alexholdenmiller:
closing for now. there are some good external examples of using tf, e.g. https://github.com/ryonakamura/parlai_agents has tf, chainer, and pytorch.

PR #604, opened 2018-02-28
Making websockets more robust, cleaning socket code

JackUrb:
Moves to using setIntervals rather than setTimeouts for heartbeats and worker threads

Wraps sends in functions that ensure they are completed without error, or handles the errors in appropriate ways rather than crashing entirely.

The socket now tries to reconnect for a minute if the server appears entirely dead. In the future this can be used to determine if the user's socket has died or if we crashed (via segfault or other).


JackUrb:
Will look into the stack trace once I find more context on where it's coming from

PR #607, opened 2018-03-01
Timeouts for waiting to pair in messenger

emilydinan:
1. Adds option to set a timeout for worlds while waiting for the needed number of agents. Once timed out, the agent is removed from the pool and returned to the overworld. The agent is marked as "removed" in the return_data field of AgentState. In order to access this data, I added a corresponding field and a function "set_state_data" to the MessengerAgent class. I could set other fields from AgentState to MessengerAgent here (like onboard_data etc.), but I wasn't sure of the use case--what do you think @JackUrb ?
2. Adds option to return to overworld while waiting

### DIFF
--- a/parlai/messenger/core/agents.py
+++ b/parlai/messenger/core/agents.py
@@ -24,7 +24,9 @@ class MessengerAgent(Agent):
         self.acted_packets = {}
         self.observed_packets = {}
         self.msg_queue = Queue()
-        self.return_data = None
+        self.stored_data = {}
+        # initialize stored data
+        self.set_stored_data()
 
     def observe(self, act):
         """Send an agent a message through the mturk manager"""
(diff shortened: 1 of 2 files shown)


parlai/messenger/core/messenger_manager.py
```diff
@@ -184,10 +206,23 @@ def _on_new_message(self, message):
...
+                for world_type in to_remove:
+                    self.remove_agent_from_pool(agent_state, world_type, mark_removed=False)
+                # put agent back in overworld
+                agent_state.set_active_agent(agent_state.get_overworld_agent())
+            else:
+                self.observe_message(
+                    agent_id,
+                    "We are trying to pair you with another person, please wait. "
+                    "If you wish to return to the Overworld, click *[EXIT]*",
+                    quick_replies=['[EXIT]']
```
JackUrb: 👍

parlai/messenger/core/agents.py
```diff
@@ -74,6 +75,12 @@ def put_data(self, message):
             }
             self.msg_queue.put(action)
 
+    def set_state_data(self):
```
JackUrb: Considering this is to some degree just _linking_ the agent's state data and the manager's state data on this agent, I'd suggest just calling it in the `__init__` function and initializing this data field to an empty dict on the AgentState initialization. This way it's more clearly an alias to the shared data pool between all instances of ParlAI agents from this particular user. As now it's somewhat unclear (I'd also suggest a more generic name, as anything can be put here, perhaps `stored_data` or `persistent_data`?)

emilydinan: Good call! I'll make the changes and call it `stored_data'.

PR #611, opened 2018-03-05
Personachat models

emilydinan:
Adds saizheng's generative models from the personachat paper with a few example scripts for
1. training models 
2. evaluating f1 and hits@1 metrics for some pretrained models
3. human evaluation (i.e. interacting with the pretrained models)


jaseweston:
dict: this shouldn't be in the diff, but in some model 'zoo' in aws. we should have a way of pulling in the model if you need it and it's not there, just like we do for tasks ideally?

PR #614, opened 2018-03-06
Pass opt to onboarding world in Messenger

emilydinan:
- Added ability to pass opt to onboarding world in messenger
- Changed overworld demo so that it initializes onboard world with opt instead of None (was not necessary to change QA data collection demo since there is no onboard world)
- Changed run function in OnboardWorld base class to take opt as an argument

### DIFF
--- a/parlai/messenger/core/messenger_manager.py
+++ b/parlai/messenger/core/messenger_manager.py
@@ -162,7 +162,7 @@ class MessengerManager():
                 )
                 return
 
-        def _overworld_function(opt, agent_id, task_id):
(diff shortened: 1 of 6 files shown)


emilydinan:
@jaseweston no-- no changes need to be made to your chatbot demo since you do not have an onboard world

jaseweston:
ok cool

emilydinan:
Added ability to pass opt to overworld through static function (not just onboard world), involved changes to QA data collection, overworld demo, and @jaseweston 's chatbot demo

Issue #609, opened 2018-03-02
Suspecious bug


Jackberg:
I've noticed an abnormal phenomenon for seq2seq model: the validation loss is always lower than training loss, both on Twitter and OpenSubtitles dataset.

And I found this part of codes quite suspicious:
https://github.com/facebookresearch/ParlAI/blob/09351eec1076319a9f5ea0f6a0c242d5f35fa43d/parlai/agents/seq2seq/seq2seq.py#L403-L411

So here we see when doing pure prediction, we don't use true labels `ys`, but when computing loss we use, which I think will bias the validation loss. Are there any reasons for computing loss like this?

alexholdenmiller:
Well, during training it has dropout enabled which should hurt its loss.

During validation, you're calculating the average loss on each token, so you do want to feed in the correct token not the one generated at the previous time step.

So both of these are expected!

Jackberg:
I think you're right @alexholdenmiller . Do you know whether the CrossEntropyLoss is averaged over sequence length or not? I'm not sure about this from the PyTorch documentation.

alexholdenmiller:
By default it is, but `size_average=False` disables that. A better implementation here would set size average to false and then sum the loss over all tokens in the entire validation set, then average, then calculate perplexity. I'll keep this issue open as a reminder to implement that instead of calculating perplexity per minibatch and averaging those.

Jackberg:
Well, I'm not sure about the `loss1` by averaging `total loss` over `total # words`, but I would say `loss2` (averaged by `total # words per batch`) is an unbiased estimation of the `loss1`. So I think it is not a bug of this implementation?

Jackberg:
And thanks @alexholdenmiller ! With `dropout=0.0`, the validation loss is roughly the same as training loss.

alexholdenmiller:
Great!

re the averaging: it doesn't matter for the loss, but for the perplexity the exponent gives you a different result if you average per minibatch or over the whole set

alexholdenmiller:
Updated perplexity scoring

PR #623, opened 2018-03-09
small logging changes to train, more accurate ppl

alexholdenmiller:
- perplexity is now calculated using average of the average loss (this is not fully accurate but better than current)
- train_model now tells you what the previous best valid is when a new best is found
- documentation for personachat options

### DIFF
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -101,8 +101,7 @@ def run_eval(agent, opt, datatype, max_exs=-1, write_log=False, valid_world=None
 
     if valid_world is None:
         valid_world = create_task(opt, agent)
-    else:
-        valid_world.reset()
+    valid_world.reset()
     cnt = 0
     while not valid_world.epoch_done():
         valid_world.parley()
@@ -115,6 +114,7 @@ def run_eval(agent, opt, datatype, max_exs=-1, write_log=False, valid_world=None
             # full depending on the structure of the data
             break
     valid_report = valid_world.report()
+    valid_world.reset()  # this makes sure agent doesn't remember valid data
 
     metrics = datatype + ':' + str(valid_report)
     print(metrics)
@@ -214,12 +214,8 @@ class TrainLoop():
             print(self.world.display() + '\n~~')
         logs = []
         # get report
-        if hasattr(self.agent, 'report'):
-            train_report = self.agent.report()
-            self.agent.reset_metrics()
-        else:
-            train_report = self.world.report(compute_time=True)
-            self.world.reset_metrics()
+        train_report = self.world.report(compute_time=True)
+        self.world.reset_metrics()
 
         # time elapsed
         logs.append('time:{}s'.format(math.floor(self.train_time.time())))
--- a/examples/train_model.py
+++ b/examples/train_model.py
@@ -180,11 +180,14 @@ class TrainLoop():
             self.agent.save(opt['model_file'] + '.checkpoint')
         if hasattr(self.agent, 'receive_metrics'):
             self.agent.receive_metrics(valid_report)
-        if self.best_valid is None or self.valid_optim * valid_report[opt['validation_metric']] > self.valid_optim * self.best_valid:
-            self.best_valid = valid_report[opt['validation_metric']]
+        new_valid = valid_report[opt['validation_metric']]
+        if self.best_valid is None or self.valid_optim * new_valid > self.valid_optim * self.best_valid:
+            print('[ new best {}: {}{} ]'.format(
+                opt['validation_metric'], new_valid,
+                ' (previous best was {})'.format(self.best_valid)
+                    if self.best_valid is not None else ''))
+            self.best_valid = new_valid
             self.impatience = 0
-            print('[ new best {}: {} ]'.format(
-                opt['validation_metric'], self.best_valid))
             if opt.get('model_file'):
                 print("[ saving best valid model: " + opt['model_file'] + " ]")
                 self.agent.save(opt['model_file'])
--- a/parlai/agents/language_model/language_model.py
+++ b/parlai/agents/language_model/language_model.py
@@ -399,9 +399,7 @@ class LanguageModelAgent(Agent):
 
 
     def predict(self, data, hidden, targets=None, is_training=True, y_lens=None):
-        """Produce a prediction from our model.
-        """
-        loss_dict = None
+        """Produce a prediction from our model."""
         output = None
         predictions = None
         if is_training:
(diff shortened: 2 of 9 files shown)


alexholdenmiller:
old ppl was sensitive to batch size, new one still has a slight sensitivity but much less.

ran personachat:self eval on a given train model, my_ppl is the new calculation:
```
bs 1   'ppl': 155.8, 'my_ppl': 66.09
bs 2   'ppl': 88.89, 'my_ppl': 63.43
bs 8   'ppl': 69.31, 'my_ppl': 64.01
bs 32  'ppl': 66.01, 'my_ppl': 64.33
bs 64  'ppl': 65.45, 'my_ppl': 64.33
bs 128 'ppl': 65.09, 'my_ppl': 64.33
bs 128 'ppl': 65.09, 'my_ppl': 64.33
bs 256 'ppl': 65.05, 'my_ppl': 64.39
bs 512 'ppl': 65.11, 'my_ppl': 64.59
bs 512 'ppl': 65.11, 'my_ppl': 64.59
```

parlai/core/metrics.py
```diff
@@ -258,6 +258,9 @@ def report(self):
             for k in self.metrics_list:
                 if self.metrics[k + '_cnt'] > 0 and k != 'correct' and k != 'f1':
                     m[k] = round_sigfigs(self.metrics[k] / self.metrics[k + '_cnt'], 4)
+        # if 'loss' in m and 'ppl' not in m:
+        if 'loss' in m:
+            m['ppl'] = round_sigfigs(math.exp(m['loss']), 4)
```
jaseweston: this doesn't make sense in that loss could mean anything (0/1, ranking, cross ent, mse).. this should only be for a specific kind of loss

alexholdenmiller: yeah I was a little worried about that too, hmm I think I found a better way to do this

alexholdenmiller:
quick note: will fix #609

alexholdenmiller:
New perplexity results are much more consistent:
```
1:    'loss': 4.218, 'ppl': 67.89
2:    'loss': 4.167, 'ppl': 64.55
512:  'loss': 4.167, 'ppl': 64.55
```

There's a small difference with bs = 1 which I currently believe is due to rounding errors in the decimals. Still trying to check.

parlai/agents/language_model/language_model.py
```diff
@@ -403,8 +425,8 @@ def predict(self, data, hidden, targets=None, is_training=True, y_lens=None):
             if bsz != self.batchsize:
                 self.hidden = self.model.init_hidden(bsz)
             loss = self.get_target_loss(data, self.hidden, targets, y_lens)
-            loss_dict['loss'] = loss
-            loss_dict['ppl'] = math.exp(loss)
+            self.metrics['loss'] += loss
+            self.metrics['num_tokens'] += sum(y_lens)
 
         return output, hidden, loss_dict, predictions
```
emilydinan: seems like you can just get rid of `loss_dict` and not return it since it will always return `None`. You should also get rid of all of the `loss_dict` code in the `batch_act` function

parlai/core/worlds.py
```diff
@@ -359,12 +367,21 @@ def episode_done(self):
...
+        for a in self.agents:
+            if hasattr(a, 'report'):
+                m = a.report()
+                for k, v in m.items():
+                    if k not in metrics:
+                        # first agent gets priority in settings values for keys
+                        # this way model can't e.g. override accuracy to 100%
+                        metrics[k] = v
+        if metrics:
+            if compute_time and 'total' in metrics:
```
emilydinan: are there cases when we do not count the total?

alexholdenmiller: eh not that I can think of but seems reasonable to check

parlai/agents/language_model/language_model.py
alexholdenmiller: good catch

alexholdenmiller:
confirmed that the bs = 1 is not rounding error in summing losses, will continue to investigate

examples/train_model.py
```diff
@@ -101,8 +101,7 @@ def run_eval(agent, opt, datatype, max_exs=-1, write_log=False, valid_world=None
 
     if valid_world is None:
         valid_world = create_task(opt, agent)
-    else:
-        valid_world.reset()
+    valid_world.reset()
```
jaseweston: is this the fix we were talking about before?

alexholdenmiller:
Update: got completely consistent results training s2s on babi:task10k:1.

Trained model with
```
py examples/train_model.py -t babi:task10k:1 --dict-file /tmp/dict_babi:task10k:1 -bs 32 -vtim 30 -vcut 0.95 -m seq2seq -mf /tmp/testbabi
```

Eval with two modes: regular 1000-ex eval, then do eval multitasking on the same data over and over.
```
py examples/eval_model.py -t babi:task10k:1{,babi:task10k:1...} --dict-file /tmp/dict_babi:task10k:1 -m seq2seq -mf /tmp/testbabi --gpu 1 -dt {mode} -bs {bsz}
```

Results:
```
valid1k
bs 1    'loss': 0.02498, 'ppl': 1.025
bs 1024 'loss': 0.02498, 'ppl': 1.025

test1k
bs 1    'loss': 0.02205, 'ppl': 1.022
bs 1024 'loss': 0.02205, 'ppl': 1.022

valid 10k
bs 1    'loss': 0.02498, 'ppl': 1.025
bs 1024 'loss': 0.02498, 'ppl': 1.025

test10k
bs    1 'loss': 0.02205, 'ppl': 1.022
bs 1024 'loss': 0.02205, 'ppl': 1.022

valid50k
bs 1    'loss': 0.02498, 'ppl': 1.025
bs 1024 'loss': 0.02498, 'ppl': 1.025
```

So this is looking good to me, let's land.

examples/train_model.py
alexholdenmiller: this makes it so that the agent forgets any partial observations of training examples going into validation

jaseweston: cool, yea that was the known bug that's in the TO DO, right?

alexholdenmiller: that's right! the custom metrics made it really obvious that there was carryover, because e.g. the lmloss in the language model (training loss) was being reported during the validation, and vice-versa. so these resets on both sides keep things clean, since the metrics are also reset.

alexholdenmiller: validation should finish episodes, though this does mean that training may have some examples truncated. that doesn't seem like a huge cost unless you're validating like every dozen examples though...

PR #628, opened 2018-03-09
Personachat paper: add mturk evaluations, language model trained on opensubtitles 2018, data collection task to mturk

emilydinan:
Sample files for how we used Mechanical Turk for human evaluation of our models in the personachat paper.


emilydinan:
@jaseweston this is just the Turk job to compare models. I can add one for the original data collection, but the instructions would be quite similar (just replace the model with another MTurk agent essentially).

jaseweston:
cool, yes adding that would be most good.

emilydinan:
added script to interact with language model trained on Opensubtitles2018. added a flag to language model that allows for sampling of distribution during interactive mode (and does not produce UNK token)

emilydinan:
Added personachat data collection task to parlai/mturk/tasks

emilydinan:
Changed the scripts for interacting with pretrained models so that we build using parlai `build_data` tools; this involved a small change to `examples/interactive.py`

emilydinan:
@jaseweston merging now!

PR #632, opened 2018-03-13
convai2 task

emilydinan:
Task for ConvAI2 challenge.

Data is taken from personachat dataset with some edits:
1. Contractions are added back in ~50% of the time (i.e. "I am" is sometimes changed to "I'm")
2. Data is doubled: to each file, we append the dataset again where the first utterance in the conversation is `<SILENCE>` and then the previous first utterance becomes the label, etc. Personas are switched in the appended data so that they make sense. We use the candidates from the original data, but switch out the true label in the list of candidates to the correct one.

### DIFF
--- a/parlai/tasks/convai2/agents.py
+++ b/parlai/tasks/convai2/agents.py
@@ -13,7 +13,11 @@ import os
 def _path(opt, persona):
     # Build the data if it doesn't exist.
     build(opt)
-    dt = opt['datatype'].split(':')[0] + '_' + persona
+    datatype =  opt['datatype'].split(':')[0]
+    if datatype == 'test':
+        print("Test set not included. Setting to valid.")
+        datatype = 'valid'
+    dt = datatype + '_' + persona
     return os.path.join(opt['datapath'], 'ConvAI2', dt + '.txt')
 
 class NoneTeacher(FbDialogTeacher):
--- a/parlai/tasks/convai2/agents.py
+++ b/parlai/tasks/convai2/agents.py
@@ -37,35 +37,5 @@ class SelfRevisedTeacher(FbDialogTeacher):
         opt['datafile'] = _path(opt, 'self_revised')
         super().__init__(opt, shared)
 
-class OtherOriginalTeacher(FbDialogTeacher):
-    def __init__(self, opt, shared=None):
-        opt = copy.deepcopy(opt)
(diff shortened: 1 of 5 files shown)


jaseweston:
we should probably make totally random candidates, even if they are not used usually in training.
for valid and test we don't need to double.

emilydinan:
@jaseweston Latest version on AWS is such that 
- candidates for "appended data" are pulled randomly from the training set, excluding the conversation. we check to make sure the label is not duplicated within the candidate set.
- only train sets are doubled (valid and test are left as is from personachat, but with contractions sometimes replaced as described above)

parlai/tasks/convai2/agents.py
```diff
@@ -0,0 +1,71 @@
...
+class SelfTeacher(SelfOriginalTeacher):
+    pass
+
+class SelfRevisedTeacher(FbDialogTeacher):
+    def __init__(self, opt, shared=None):
+        opt = copy.deepcopy(opt)
+        opt['datafile'] = _path(opt, 'self_revised')
+        super().__init__(opt, shared)
+
+class OtherOriginalTeacher(FbDialogTeacher):
```
jaseweston: oh i think we should get rid of the other and both actually for the competition

alexholdenmiller:
can we do __SILENCE__ instead to match e.g. our dictionary class's special tokens?

parlai/tasks/convai2/agents.py
alexholdenmiller: +1 on this

emilydinan:
- Changed `<SILENCE>` token to `__SILENCE__`
- Got rid of "other" and "both" teachers

emilydinan:
@jaseweston @alexholdenmiller 
- latest version on AWS combines test and validation sets to just be the validation set. `convai/agents.py` edited to issue a warning when you try to use a test set and sets the datatype to "valid."
- Also includes small spelling fix ("cant" was not included in list of original contractions that were fixed; changed to sometimes be "can't").

jaseweston:
super! any thing left to do? if not lets get it in.

emilydinan:
@jaseweston the only minor thing to decide is what we want the default teacher to be -- right now set to selfOriginal but we had discussed setting to selfRevised. this can obviously be changed later if we want to merge now.

jaseweston:
ok, yes let's check in and think about it

PR #639, opened 2018-03-14
KVmemNN interactive example

jaseweston:
eval and train to come next

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -275,9 +275,9 @@ def name_to_agent_class(name):
 
 def load_agent_module(opt):
     model_file = opt['model_file']
-    if model_file.startswith('model:'):
+    if model_file.startswith('models:'):
         # load model from the ParlAI model zoo directory
-        model_file = os.path.join(opt['datapath'], 'models', model_file[6:])
+        model_file = os.path.join(opt['datapath'], 'models', model_file[7:])
     optfile =  model_file + '.opt'
     if os.path.isfile(optfile):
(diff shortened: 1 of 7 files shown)


parlai/core/agents.py
```diff
@@ -274,12 +274,16 @@ def name_to_agent_class(name):
     return class_name
 
 def load_agent_module(opt):
-    optfile =  opt['model_file'] + '.opt'
+    model_file = opt['model_file']
+    if model_file.startswith('models:'):
```
emilydinan: cool -- i will have to change my scripts to do the same!

projects/personachat/scripts/kvmemnn_interactive.py
```diff
@@ -0,0 +1,27 @@
...
+        model_file='models:personachat/kvmemnn/kvmemnn/persona-self_rephraseTrn-True_rephraseTst-False_lr-0.1_esz-500_margin-0.1_tfidf-False_shareEmb-True_hops1_lins0_model',
+    )
+    opt = parser.parse_args()
+    # build all profile memory models
+    fnames = ['kvmemnn.tgz']
+    opt['model_type'] = 'kvmemnn' # for builder
+    build(opt, fnames)
+
+    # add additional model args
+    opt['interactive_mode'] = True
```
emilydinan: do you have interactive_mode as a flag for your model? maybe i missed. i added this as a flag for saizheng's model but you may not need it.

jaseweston:
ok i think this is good now

PR #647, opened 2018-03-16
Status partner disconnect early

klshuster:
This PR allows you to specify a minimum number of messages required to be sent by the turker before accepting their HIT if their partner disconnects from the chat

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -157,7 +157,7 @@ class ParlaiParser(argparse.ArgumentParser):
                  'unlimited.'
         )
         mturk.add_argument(
-            '--min-messages', dest='min-messages',
+            '--min-messages', dest='min_messages',
             default=0, type=int,
             help='number of messages required to be sent by MTurk agent when '
                  'considering whether to approve a HIT in the event of a '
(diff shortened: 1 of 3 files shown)


parlai/core/params.py
```diff
@@ -156,6 +156,14 @@ def add_mturk_args(self):
             help='number of HITs that can be launched at the same time, 0 is '
                  'unlimited.'
         )
+        mturk.add_argument(
+            '--min-messages', dest='min-messages',
```
JackUrb: dest should be `min_messages`

parlai/mturk/core/mturk_manager.py
```diff
@@ -290,7 +291,10 @@ def _handle_partner_disconnect(self, worker_id, assignment_id):
         elif not agent.state.is_final():
             # Update the assignment state
             agent.some_agent_disconnected = True
-            agent.state.status = AssignState.STATUS_PARTNER_DISCONNECT
+            if len(agent.state.messages) < self.minimum_messages:
```
JackUrb: Technicality here - messages will include both messages sent by the worker and their partner. You should parse messages and only count the messages that have `message_data.id == agent.id` before making this.

Admittedly this seems confusing, as `message_data.id` is actually storing the sender id of the message, but that's my bad for some poor variable naming. Packets have an id and a data field, and only the data field is stored in messages.

PR #649, opened 2018-03-16
update offensive language filtering

alexholdenmiller:


alexholdenmiller:
uses a token trie now so should be both more robust and faster

alexholdenmiller:
```
from parlai.core.utils import OffensiveLanguageDetector
o.contains_offensive_language('ah! classical music is so soothing.')
o = OffensiveLanguageDetector()
>>> False
o.contains_offensive_language('ah! classical music is so wet dream worthy')
>>> True
o.contains_offensive_language('ah! classical music makes me go apeshit sometimes')
>>> True
```

emilydinan:
This looks great!!

PR #652, opened 2018-03-23
Fbdialog updates

alexholdenmiller:
- the fbdialog parser now resets the episode if the next id is less than the current, rather than 1

e.g.
```
1 hello world
2 second entry
3 third entry
2 new episode
```

this maintains the old behavior but also handles small bugs in numbering in the files

+ a few unit test fixes


alexholdenmiller:
I'm going to merge because I'd like to use the new parser

PR #656, opened 2018-03-27
Messenger agents typing indicator functions

emilydinan:
1. Adds `observe_typing_on` for messenger agent to observe typing indicator
2. Rephrases message from onboarding world for pairing to be more generic / not reference overworld, added typing indicator
3. If an agent is in a pool for more than 30 seconds, we send a message that pairing is taking longer than expected and add a typing indicator.

### DIFF
--- a/parlai/messenger/core/messenger_manager.py
+++ b/parlai/messenger/core/messenger_manager.py
@@ -266,9 +266,10 @@ class MessengerManager():
                 self.observe_message(
                     agent_id,
                     "Please wait while we pair you with another person. "
-                    "If you wish to return to the Overworld, click *EXIT*",
+                    "If you wish to exit, click *EXIT*",
                     quick_replies=['EXIT']
                 )
+                self.message_socket.typing_on(agent_id)
         else:
             # If an agent is in a solo world, we can put a typing indicator
             # and mark the message as read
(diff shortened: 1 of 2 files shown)


parlai/messenger/core/messenger_manager.py
```diff
@@ -266,9 +266,10 @@ def _on_new_message(self, message):
                 self.observe_message(
                     agent_id,
                     "Please wait while we pair you with another person. "
-                    "If you wish to return to the Overworld, click *EXIT*",
+                    "If you wish to exit, click *EXIT*.",
```
JackUrb: Note - we should probably drop the quick reply and ask a user to manually type exit, as we've seen that people like just clicking quick replies without reading

parlai/messenger/core/messenger_manager.py
```diff
@@ -480,6 +481,22 @@ def _task_function(opt, agents, conversation_id, world_type):
...
+                            elif time_in_pool and time.time() - time_in_pool \
+                                    > 30:
+                                # tell agent that a match is taking longer than
+                                # expected
+                                if not agent_state.stored_data.get('seen_wait_message') \
+                                        or not agent_state.stored_data['seen_wait_message']:
+                                    self.observe_message(
+                                        agent_state.messenger_id,
+                                        "Pairing is taking longer than expected. "
+                                        "If you wish to exit, click *EXIT*.",
```
JackUrb: Re: the above

PR #660, opened 2018-03-27
fix scripts to collect from model zoo

emilydinan:
- Fix personachat scripts so that they collect from model zoo
- Add lines to language_model.py and persona_seq2seq.py to check for this.

### DIFF
--- a/projects/personachat/persona_seq2seq.py
+++ b/projects/personachat/persona_seq2seq.py
@@ -167,19 +167,6 @@ class Seq2seqAgent(Agent):
                 print('[ Using CUDA ]')
                 torch.cuda.set_device(opt['gpu'])
 
-            def check_in_model_zoo(filename):
-                if filename.startswith('models:'):
-                    filename = os.path.join(opt['datapath'], 'models', filename[7:])
-                return filename
-
-            if opt.get('model_file'):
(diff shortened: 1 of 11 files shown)


emilydinan:
1. Moved check for 'models:' prefix in model file into the `create_agent` function. Also added check for dict_file named similarly there. 
2. Small changes to language model sampling mode to make sure our sampling distribution norm does not blow up.
3. Added back a script for training Saizheng's profile memory model with parameters that produce better results.

projects/personachat/persona_seq2seq.py
```diff
@@ -1664,6 +1668,7 @@ def _encode(self, xs, is_training=False):
             if type(self.decoder) != nn.LSTM:
                 hidden = hidden[0]
         else:
+            xes_packed = pack_padded_sequence(xes.transpose(0, 1), x_lens)
```
alexholdenmiller: if you use pack_padded_sequence then I think you'll need to use pad_packed_sequence on the encoder_output

alexholdenmiller: does this actually run as is?

emilydinan:
@alexholdenmiller got rid of pack_padded for profile memory model

emilydinan:
@jaseweston yes i checked before, still works!

PR #662, opened 2018-03-27
KV Profile Mem training

jaseweston:


abhinavkashyap:
Is the code added in this commit correspond to **_"Section 4.5 Generative Profile Memory Network"_** of the paper??

jaseweston:
no, this is the retrieval model part

abhinavkashyap:
Oh thanks for that. Could you please point me to the part where the generative profile memory network as mentioned in section 4.5 is implemented?

jaseweston:
Explained here:

https://github.com/facebookresearch/ParlAI/tree/master/projects/personachat

specifically:

Evaluate pre-trained profile memory model trained to optimize f1 metric on
Persona-Chat using persona 'self':

python projects/personachat/scripts/profilememory_eval_f1.py

Train a profile memory model on Persona-Chat using persona 'self':

bash projects/personachat/scripts/profilememory_train.sh

abhinavkashyap:
Thank you!

abhinavkashyap:
If I run the script and check the perplexity on test dataset it is around 88.33. I ran it for around 30 epochs. The value in the paper for persona:self is around 38. Is there anything that I am missing here. Should I run it for more?

emilydinan:
Hi @abhinavkashyap -- there are currently some bugs with this model that have not been fixed yet: please see #1453 for more details and to follow progress. 

In general, the code in our projects folder is not as actively maintained as the rest of ParlAI. If you don't want to wait, for training a generative model on personachat or convai2, I would suggest following the train script here: ParlAI/projects/convai2/baselines/seq2seq/train.py or using the latest version of seq2seq in ParlAI with similar parameters.

uralik:
@abhinavkashyap , I would advice you to use general seq2seq agent with RNNs (as @emilydinan suggested) instead of the thing you mentioned earlier.
 
With params like this:
```
python train.py \
    -dt train \
    -t convai2:self \
    -m seq2seq \
    -bs 128 \
    -dp "${DATA_PATH}" \
    --model-file "${RUN_DIR}"/model \
    --dict-file "${DATA_PATH}"/convai2.dict \
    --numlayers 2 \
    --hiddensize 512 \
    --dropout 0.5 \
    --optimizer adam \
    -lr 0.001 \
    -esz 300 \
    --attention general \
    -lt enc_dec \
    -emb glove \
    -veps 5
```
you can quickly get to PPL below 30 (depending on early stopping etc) on validation data without any additional pretraining.

abhinavkashyap:
@emilydinan @uralik  Thank you for the reply. I am particularly interested in the sequence sequence Profile Memory Network where the attention over the memories are used. As I see it `profile_memory_train.sh` uses `general attention`  mechanism over the persona sentences to generate the replies.  

@emilydinan : #1453 fixes the interactive script for chatbot. However, I would like to understand the `Profile Memory Network` training.  Is there any other repository that has tried to reproduce the results of the paper "Personalizing Dialogue Agents?"

@uralik : The baselines for the convai2 challenge does not consider the profile memory network. It is the basic Seq2Seq and Seq2Seq with attention if I am right.  How can I run the `Profile Memory Network` on convai2 dataset.

It would be really helpful if you could point out any bugs in the `Profile Memory Network` generative model and other scripts that I can use.

Thank you.

emilydinan:
Hi @abhinavkashyap, we are actually going to deprecate this code shortly and remove it from the codebase as it contains many bugs and we will no longer be supporting this model. I'm not aware of any other repository that has tried to reproduce these results. 

If you still wish to use this model, you will have to keep the copy of this code locally and fix the bugs. Alternatively, you could edit the latest Seq2seq code to do something similar.

abhinavkashyap:
Hi @emilydinan thank you for this. Does the result from the paper that references this model still hold and what about the model architecture itself ?? This model is one of the major contributions of the work.

emilydinan:
Yes the results from the paper are still correct.

abhinavkashyap:
@emilydinan Thank you Emily for this conversation. 
I have added a tl;dr for anyone who is following this conversation. 

tldr; 
What I glean from this is that `Profile Memory Network` as discussed by the paper, is buggy in the code-base and has been deprecated as of now. And there are no other repositories (including this one) that has tried to reproduce the results as of yet (and may be has to be implemented in the latest `SeqSeq` code)

stephenroller:
Yeah that's roughly right. The reason the profile memory network code is buggy is because the platform has changed *substantially*, and we didn't make an effort to keep that model up to date as the platform evolved.

PR #665, opened 2018-03-29
change convai2 version, update readme

emilydinan:


alexholdenmiller:
yes, but note that the way the warning is currently printed, it prints `batchsize` times :)

PR #668, opened 2018-03-29
bug fix

alexholdenmiller:
since each model needs to be in it's own folder for the current build marking to verify whether to download something, need to have the double name


alexholdenmiller:
jason approved over messenger

PR #673, opened 2018-04-02
update torch.load calls and model_file interpretation

alexholdenmiller:


alexholdenmiller:
the map_location changes will enforce the desired behavior of torch.load (just load to cpu) without needing the try_except, fixing both the bug I was trying to fix and the one jason and kurt experienced

alexholdenmiller:
there's a side effect here that when the model is saved the model_file path will be the actual path not the virtual one models:path

examples/eval_model.py
```diff
@@ -30,6 +31,10 @@ def setup_args():
 def eval_model(parser, printargs=True):
     random.seed(42)
     opt = parser.parse_args(print_args=False)
+    # check to make sure the model file exists
```
jaseweston: if model file is none that could be ok, because we have some models that don't have a file at all (IR, repeatQuery )

parlai/core/params.py
```diff
@@ -333,6 +343,14 @@ def parse_args(self, args=None, namespace=None, print_args=True):
         if self.opt.get('datapath'):
             os.environ['PARLAI_DATAPATH'] = self.opt['datapath']
 
+        # nothing changes if model_file
```
jaseweston: can you make this comment more clear?

alexholdenmiller: woops I think I was interrupted halfway through writing that, thanks for catching

examples/eval_model.py
alexholdenmiller: let me know what you think about the new version. too messy?

alexholdenmiller: (check for `load` function, raise runtime error if model has it and no model file found)

examples/eval_model.py
```diff
@@ -30,8 +31,17 @@ def setup_args():
...
+    nomodel = Flase
```
jaseweston: False

examples/eval_model.py
```diff
@@ -30,8 +31,17 @@ def setup_args():
...
+    if opt.get('model_file') is None or not os.path.isfile(opt['model_file']):
```
jaseweston: i think if model file is set and file doesn't exist, you can bail immediately.
it's only the 'not set' case you have to check if it has a load function or so?

parlai/core/params.py
```diff
@@ -333,6 +348,15 @@ def parse_args(self, args=None, namespace=None, print_args=True):
         if self.opt.get('datapath'):
             os.environ['PARLAI_DATAPATH'] = self.opt['datapath']
 
+        # nothing changes if model_file does not start with 'models:'
```
jaseweston: i think a 1-line comment that says:

'map filenames that start with 'models:' to point to the model zoo'  

would be better?

alexholdenmiller:
./run_tests_short.sh all passed

PR #674, opened 2018-04-03
Improve compatibility with PyTorch master (0.4)

colesbury:
- Use isinstance instead of exact type checks
 - Use torch.is_tensor instead of string matching the type

The test_train_model.py script still fails. I think it's due to
var.squeeze() returning a 0-d value in a place  where it used to return
a 1-d 1-element value.


alexholdenmiller:
You're on current pytorch master? `test_train_model.py` is working for me for master from ~1-2 weeks ago, I'll update and try again!

Thanks for the PR though!

PR #678, opened 2018-04-06
Fix loading seq2seq model in cuda mode

pemazare:
This fixes a bug when loading an existing seq2seq model along with its optimizer: all tensors for the optimizer state are mapped to CPU when loaded, and the training then crashes because the optimizer uses CPU-stored tensors although the underlying model parameters are on GPU.

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -311,6 +311,12 @@ class Seq2seqAgent(Agent):
                           'changed.')
                 else:
                     self.optimizer.load_state_dict(states['optimizer'])
+                    if self.use_cuda:
+                        for state in self.optimizer.state.values():
+                            for k, v in state.items():
+                                if isinstance(v, torch.Tensor):
+                                    state[k] = v.cuda()
+
             self.scheduler = optim.lr_scheduler.ReduceLROnPlateau(
                 self.optimizer, 'min', factor=0.5, patience=3, verbose=True)
 


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -311,6 +311,12 @@ def __init__(self, opt, shared=None):
                           'changed.')
                 else:
                     self.optimizer.load_state_dict(states['optimizer'])
+                    if self.use_cuda:
+                        for state in self.optimizer.state.values():
+                            for k, v in state.items():
+                                if isinstance(v, torch.Tensor):
+                                    state[k] = v.cuda()
+
```
alexholdenmiller: maybe better to do states['optimizer'].pop('model') before `load_state_dict`? the optimizer should already have the model params loaded in from line 308

pemazare: Actually, looking at the generic code for [the Optimizer class](http://pytorch.org/docs/master/_modules/torch/optim/optimizer.html#Optimizer), the class actually needs the part of the state containing the old parameter list in order to correctly link together the tensors loaded from the saved optimizer to the tensors containing the model parameters (see the update_group method). So my impression is that it's preferable not to edit the state dict before we load it.

emilydinan: @pemazare looks good to me. @alexholdenmiller ?

Thanks for fixing this!

alexholdenmiller: hm not yet I'm still concerned about iterating over a private field of the optimizer one sec

PR #687, opened 2018-04-10
ConvAI2: adapt training, evaluation, and interactive code to the same setup

alexholdenmiller:


projects/personachat/persona_seq2seq.py
```diff
@@ -32,24 +32,35 @@
 import torch.nn as nn
 import torch.nn.functional as F
 import torch
-import torchtext.vocab as vocab
-Glove = vocab.GloVe(name='840B', dim=300)
+try:
```
emilydinan: nice

alexholdenmiller: thanks! yeah now you can run it from any directory too

PR #709, opened 2018-04-18
metrics: protect against div by zero

jaseweston:


parlai/core/metrics.py
```diff
@@ -250,15 +250,15 @@ def report(self):
         m['total'] = total
         if total > 0:
             if self.flags['print_prediction_metrics']:
-                m['accuracy'] = round_sigfigs(self.metrics['correct'] / self.metrics['correct_cnt'], 4)
-                m['f1'] = round_sigfigs(self.metrics['f1'] / self.metrics['f1_cnt'], 4)
+                m['accuracy'] = round_sigfigs(self.metrics['correct'] / max(1, self.metrics['correct_cnt'], 4))
```
emilydinan: I think your parentheses are in the wrong place (here and for the other changes down below)

PR #713, opened 2018-04-20
Fix: typo requireFileExists

LaceyChen17:
A typo in parlai/core/agents which might cause failure in creating agent.


LaceyChen17:
#714

PR #715, opened 2018-04-20
Starspace improvements: linear layer (helps a lot!) + using fixed candidate set in test works

jaseweston:


parlai/agents/starspace/starspace.py
```diff
@@ -95,6 +94,8 @@ def add_cmdline_args(argparser):
                            help='max norm of word embeddings')
         agent.add_argument('-shareEmb', '--share-embeddings', type='bool', default=True,
                            help='whether LHS and RHS share embeddings')
+        agent.add_argument('--lins', default=1, type=int,
+                           help='If set to 1, add a linear layer between lhs and rhs.')
```
alexholdenmiller: you can use `type='bool'`?

jaseweston: was leaving open the possibility of having multilayer nonlinearities..

PR #723, opened 2018-04-25
Decoupling message receives from message sends

JackUrb:
Splits out the message send logic from the message receive logic, as these follow two different pipelines. Having them separate means that messages can be sent without having to set up a socket, which is useful in circumstances when you're subclassing MessageManager because you have a different route for receiving messages and don't want the socket.

### DIFF
--- a/parlai/messenger/core/message_sender.py
+++ b/parlai/messenger/core/message_sender.py
@@ -1,0 +1,279 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+import json
+import logging
+import requests
+
+import parlai.mturk.core.shared_utils as shared_utils
+
+MAX_QUICK_REPLIES = 10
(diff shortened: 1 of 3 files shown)


parlai/messenger/core/message_sender.py
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+
+class MessageSender():
+    """MesageSender is a wrapper around the facebook messenger requests that
+    simplifies the process of sending content.
+    """
+
+    """MessageSocket is a wrapper around websocket to simplify message sends
```
emilydinan: you might have left this docstring here by accident

parlai/messenger/core/messenger_manager.py
```diff
@@ -476,6 +483,32 @@ def set_overworld_func(self, overworld_func):
     def set_agents_required(self, max_agents_for):
         self.max_agents_for = max_agents_for
 
+    def check_timeout_in_pool(self, world_type, agent_pool, max_time_in_pool):
```
emilydinan: nice--good call putting this in a separate function

parlai/messenger/core/message_sender.py
JackUrb: Good catch here!

PR #724, opened 2018-04-25
PytorchDataTeacher Documentation

klshuster:


klshuster:
So right now the title of the tutorial/docs page is "Batching and Hogwild" - should we change that to something like "Batching, Hogwild, and Dataloading"? I think we can put links at the top down to each section (note in my changes, it the 'Multiprocessed Pytorch Dataloader' shows up as its own subsection on the side)

jaseweston:
Maybe  'Data handling, batching and hogwild'  and even include some simple things about data loading at the top, like about :stream ?

klshuster:
I changed the title and added links/a short description of the data handling at the top

klshuster:
Is this ok to merge and revisit later? Or wait for @alexholdenmiller comments?

jaseweston:
let's merge and update later, it is adding value rn!

Issue #721, opened 2018-04-24
Seq2Seq on Reddit movie task throws RuntimeError midway through training


atalreja:
Hi, when running the following command with python 3.6.5 and pytorch 0.3.1 

```python examples/train_model.py -t "#moviedd-reddit" -dt train:stream -bs 1 -clen 3 -m seq2seq -mf /tmp/s2s -ltim 30 -vtim 30 -stim 30 -vcut 0.95 --dict-maxtokens 50000``` 

The model starts to train but after a few printouts it fails with the following error stack

```Traceback (most recent call last):
  File "examples/train_model.py", line 275, in <module>
    TrainLoop(setup_args()).train()
  File "examples/train_model.py", line 252, in train
    stop_training = self.validate()
  File "examples/train_model.py", line 176, in validate
    valid_world=self.valid_world)
  File "examples/train_model.py", line 106, in run_eval
    valid_world.parley()
  File "/home/atalreja/code/ParlAI/parlai/core/worlds.py", line 286, in parley
    acts[1] = agents[1].act()
  File "/home/atalreja/code/ParlAI/parlai/agents/seq2seq/seq2seq.py", line 570, in act
    return self.batch_act([self.observation])[0]
  File "/home/atalreja/code/ParlAI/parlai/agents/seq2seq/seq2seq.py", line 546, in batch_act
    predictions, text_cand_inds = self.predict(xs, ys, cands, valid_cands, is_training)
  File "/home/atalreja/code/ParlAI/parlai/agents/seq2seq/seq2seq.py", line 459, in predict
    _, scores, _ = self.model(xs, ys)
  File "/home/atalreja/anaconda3/lib/python3.6/site-packages/torch/nn/modules/module.py", line 357, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/atalreja/code/ParlAI/parlai/agents/seq2seq/modules.py", line 73, in forward
    y_in = ys.narrow(1, 0, ys.size(1) - 1)
RuntimeError: invalid argument 5: out of range at /pytorch/torch/lib/THC/generic/THCTensor.c:468
```

I think it might have to do with the context length setting because I was trying different values for that. Sorry for the lack of details, I'll continue trying different options

Thanks!

alexholdenmiller:
Thanks for reporting! I'm able to reproduce and will work on a fix

alexholdenmiller:
fixed, please pull and try again!

alexholdenmiller:
[...]

PR #726, opened 2018-04-25
moviedd reddit is splitting on pipes in labels

alexholdenmiller:
currently labels for moviedialog:task:4 are getting split on labels (e.g. "I am annoyed : |" creates two labels, "I am annoyed :" and "")

this builds a pipeless one where pipes are replaced with `__PIPE__`


alexholdenmiller:
fix for #721

PR #729, opened 2018-05-02
Pipe-less fix for Twitter task, minor fix for build_dict log part, language_model agent and pytorch 0.4 fixes

uralik:
The twitter task has similar problem as mentioned in [#721](https://github.com/facebookresearch/ParlAI/issues/721), so now "|" is replaced with "__PIPE__" in
 the text.
In general, when one uses build_dict script, then with whatever options are specified (dict-maxtokens or dict-minfreq) the log says that the dict file contains max possible num words, because the print is located before the sort function, where resizing happens. So print happens afterwards now.

### DIFF
--- a/parlai/agents/language_model/language_model.py
+++ b/parlai/agents/language_model/language_model.py
@@ -235,7 +235,7 @@ class LanguageModelAgent(Agent):
 
     def update_params(self):
         """Do one optimization step."""
-        torch.nn.utils.clip_grad_norm(self.model.parameters(), self.clip)
+        torch.nn.utils.clip_grad_norm_(self.model.parameters(), self.clip)
         self.optimizer.step()
 
     def reset(self):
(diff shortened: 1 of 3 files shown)


uralik:
language_model agent did not start properly with pytorch 0.4, changes from [
72cd483](https://github.com/facebookresearch/ParlAI/pull/729/commits/72cd48308cc714ffbe68a63d8ea2a48e784b8e33) fix it and agent works.

I tried to test those changes with pytorch 0.3.1-post2, it failed to run, but actually language_model agent fails to run even with current master branch **only when batch size > 1.** It gives the following error during the valid phase:

> File "/home/kulikov/code/ParlAI-orig/parlai/agents/language_model/language_model.py", line 356, in get_target_loss                                                             
>     loss += self.criterion(output_flat, targets.select(1,0).view(-1)).data
> RuntimeError: invalid argument 2: input is not contiguous at /opt/conda/conda-bld/pytorch_1518244507981/work/torch/lib/THC/generic/THCTensor.c:227

PR #733, opened 2018-05-07
Messenger agent bug fix

emilydinan:
Make messenger agent `observe_typing_on` function compatible with changes in https://github.com/facebookresearch/ParlAI/pull/723


JackUrb:
Ah good catch! Missed this one :(

PR #735, opened 2018-05-08
Move torchtext vectors to model zoo

emilydinan:
Moves torchtext vectors to ParlAI model zoo and changes name to `glove_vectors` or `fasttext_vectors`


emilydinan:
@jaseweston yes just checked it will create the models directory if it doesn't exist

PR #737, opened 2018-05-09
Language model options loading

emilydinan:
Save a `.opt` file which will be loaded when `load_agent_module` is called. If this file doesn't exist, the opt is loaded from the `opt` key in the dictionary in the model file (to make compatible with older models for which a `.opt` file is not saved)

### DIFF
--- a/parlai/agents/language_model/language_model.py
+++ b/parlai/agents/language_model/language_model.py
@@ -139,6 +139,13 @@ class LanguageModelAgent(Agent):
                 print('[ Setting opt from {} ]'.format(
                     init_model
                 ))
+                # since .opt file does not exist, save one for future use
+                with open(init_model + ".opt", 'wb') as handle:
+                    pickle.dump(
+                        new_opt,
+                        handle,
+                        protocol=pickle.HIGHEST_PROTOCOL
+                    )
                 opt = self.override_opt(new_opt)
 
             if (init_model is not None and
--- a/parlai/agents/language_model/language_model.py
+++ b/parlai/agents/language_model/language_model.py
@@ -130,10 +130,10 @@ class LanguageModelAgent(Agent):
             if opt.get('model_file') and os.path.isfile(opt['model_file']):
                 init_model = opt['model_file']
 
-            if init_model is not None and \
-                    not os.path.isfile(init_model + '.opt'):
-                # this will only be called for loading older models before
-                # .opt file is saved
+            # for backwards compatibility: will only be called for older models
+            # for which .opt file does not exist
+            if (init_model is not None and
+                    not os.path.isfile(init_model + '.opt')):
                 new_opt = self.load_opt(init_model)
                 # load model parameters if available
                 print('[ Setting opt from {} ]'.format(
@@ -149,9 +149,9 @@ class LanguageModelAgent(Agent):
                     )
                 opt = self.override_opt(new_opt)
 
-            if (init_model is not None and
-                    os.path.isfile(init_model + '.dict')) or \
-                    opt['dict_file'] is None:
+            if ((init_model is not None and
+                    os.path.isfile(init_model + '.dict')) or
+                    opt['dict_file'] is None):
                 opt['dict_file'] = init_model + '.dict'
 
             # load dictionary and basic tokens & vectors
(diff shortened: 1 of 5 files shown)


parlai/agents/language_model/language_model.py
```diff
@@ -132,8 +135,10 @@ def __init__(self, opt, shared=None):
                 # load model parameters if available
                 print('Loading existing model params from ' + init_model)
                 new_opt, self.states = self.load(init_model)
-                # override model-specific options with stored ones
-                opt = self.override_opt(new_opt)
+                # override model-specific options with stored ones if not
+                # already overriden with .opt file
+                if not os.path.isfile(init_model + '.opt'):
```
jaseweston: so this will only be for old models (saved before this PR)? maybe say that..

parlai/agents/language_model/language_model.py
```diff
@@ -132,8 +135,10 @@ def __init__(self, opt, shared=None):
                 # load model parameters if available
                 print('Loading existing model params from ' + init_model)
                 new_opt, self.states = self.load(init_model)
-                # override model-specific options with stored ones
-                opt = self.override_opt(new_opt)
+                # override model-specific options with stored ones if not
```
jaseweston: i think the code at 126 might cause a problem because parlai load agent will pass in the saved opts, which may include an init_model and a model_file option, and it will load from the init_model rather than model_file..  shouldnt it load from model_file if it exists (overrides 'init_model')?

emilydinan: good catch... i thought we wanted to load from init_model if it exists (and save into model_file). if that's the case, we should check for an init_model in `load_agent_module` in `parlai/core/agents.py`. but if it should be the other way around, why do we have init_model? what do you think?

jaseweston: i think when you begin training you might start from an init_model, but then after you have a model_file you would use that instead.  i'm not sure if this is model specific code, or if we want  to migrate that agents.py

i've tried to make a PR for seq2seq here, note some other changes with dict also: https://github.com/facebookresearch/ParlAI/pull/739/files

emilydinan:
Setting model states is moved to load function.

parlai/agents/language_model/language_model.py
```diff
@@ -574,7 +578,18 @@ def receive_metrics(self, metrics_dict):
...
+        return states['opt']
+
+    def load(self, path):
+        """Load model states."""
+        if os.path.isfile(path):
+            # load model parameters if available
+            print('[ Loading existing model params from {} ]'.format(path))
+            self.states = torch.load(path, map_location=lambda cpu, _: cpu)
+
+        if self.states:
```
jaseweston: can be in same if above?

parlai/agents/language_model/language_model.py
```diff
@@ -119,29 +122,29 @@ def __init__(self, opt, shared=None):
...
+            # next check for 'model_file', this would override init_model
+            if opt.get('model_file') and os.path.isfile(opt['model_file']):
                 init_model = opt['model_file']
-            else:
-                init_model = None
 
-            if init_model is not None:
+            if init_model is not None and \
+                    not os.path.isfile(init_model + '.opt'):
+                # this will only be called for loading older models before
```
jaseweston: maybe say the words 'backwards compatible' or something

parlai/agents/language_model/language_model.py
```diff
@@ -119,29 +122,29 @@ def __init__(self, opt, shared=None):
...
-            # next check for 'model_file'
-            elif opt.get('model_file') and os.path.isfile(opt['model_file']):
+            # next check for 'model_file', this would override init_model
+            if opt.get('model_file') and os.path.isfile(opt['model_file']):
                 init_model = opt['model_file']
-            else:
-                init_model = None
 
-            if init_model is not None:
+            if init_model is not None and \
```
jaseweston: isn't it better to use parentheses than "\" ?

parlai/agents/language_model/language_model.py
emilydinan: not sure what you mean by this

jaseweston: oh ok i see the load might fail so states wouldnt be set.. should that error or ?

emilydinan: oh i see what you're saying. i don't think that should ever happen. will get rid of it.

PR #739, opened 2018-05-09
Seq2Seq saves the opt, and loads from it (making eval much simpler), plus other small fixes

jaseweston:


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -630,7 +625,8 @@ def shutdown(self):
     def load(self, path):
         """Return opt and model states."""
         states = torch.load(path, map_location=lambda cpu, _: cpu)
-        return states['opt'], states
+        self.opt = self.override_opt(states['opt'])
```
emilydinan: should we only override if a `.opt` file is not present?

Issue #720, opened 2018-04-24
Trivial issue: Unnecessary loops?


Jackberg:
I noticed some unnecessary loops, e.g. these snippets:
https://github.com/facebookresearch/ParlAI/blob/25a19b8559e513105c956d8d8f09ff71e7e370d4/parlai/core/worlds.py#L660-L662
https://github.com/facebookresearch/ParlAI/blob/25a19b8559e513105c956d8d8f09ff71e7e370d4/parlai/core/worlds.py#L668-L670

When you're using, say seq2seq model, with 2 agents, these `if` statements will be hit `3*batch_size` times in each parley, but the body will never be executed. For a task with huge dataset, I think these snippets will consume considerable time.

I wonder if there are situations that these `if` statements will be true for just a subset of the batch worlds? If not, maybe add an `else` statement to end the loop earlier will slightly reduce the training time?
```
else:
    break
```

alexholdenmiller:
Thanks for the suggestion! I tried using different batch sizes just to display the data and couldn't find any difference in the runtime unfortunately, but the worlds are the same type as the initial world so it's safe to make the change anyways. (#740)

PR #740, opened 2018-05-09
small change for issue #720

alexholdenmiller:
inner world has same functions as batch instances so can do the check once instead of bsz times

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -657,16 +657,16 @@ class BatchWorld(World):
         num_agents = len(self.world.get_agents())
         batch_observations = self.batch_observations
 
-        for w in self.worlds:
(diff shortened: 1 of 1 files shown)


parlai/core/worlds.py
```diff
@@ -657,16 +657,16 @@ def parley(self):
         num_agents = len(self.world.get_agents())
         batch_observations = self.batch_observations
 
-        for w in self.worlds:
-            if hasattr(w, 'parley_init'):
```
jaseweston: is it not possible some worlds could have the init, but others not?

jaseweston: they are all identical for sure?

jaseweston:
ping

jaseweston:
ping

PR #741, opened 2018-05-09
add mixture of experts to seq2seq, fix some casting/item() calls for pytorch 0.4

alexholdenmiller:


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -31,7 +31,7 @@ class Seq2seqAgent(Agent):
     highest probability token at each time step.
 
     For more information, see Sequence to Sequence Learning with Neural
-    Networks `(Sutskever et al. 2014) <https://arxiv.org/abs/1409.3215>`_.
+    Networks `(Sutskever et al. 2014) <arxiv.org/abs/1409.3215>`_.
```
jaseweston: probably needs a link to attention also: https://arxiv.org/abs/1409.0473

parlai/tasks/twitter/build.py
```diff
@@ -5,9 +5,13 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 # Download and build the data if it does not exist.
 
+try:
```
jaseweston: wait did u guys (with ilia) change the twitter data itself recently? that's not good practice in general.. we need versions/naming for that

alexholdenmiller: #729

PR #742, opened 2018-05-10
remove observation from next_word_prob

alexholdenmiller:
also removed the caching from seq2seq since I was having a bug with it after these changes, could add back in later

### DIFF
--- a/projects/convai2/baselines/seq2seq/eval_ppl.py
+++ b/projects/convai2/baselines/seq2seq/eval_ppl.py
@@ -48,11 +48,15 @@ class Seq2seqEntry(Seq2seqAgent):
         obs = self.observation
         obs['eval_labels'] = [' '.join(partial_out)]
         batch = self.vectorize([obs])
+        if self.prev_enc is not None and batch[0].shape[1] != self.prev_enc[0].shape[1]:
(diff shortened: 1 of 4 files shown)


jaseweston:
oh is this gonna slow it down a lot? that wouldn't be good

projects/convai2/baselines/seq2seq/eval_ppl.py
```diff
@@ -43,23 +45,14 @@ def next_word_probability(self, observation, partial_out):
         e.g.
         {'text': 'Run test program.'}, ['hello'] => {'world': 1.0}
         """
-        if not hasattr(self, 'prev_enc'):
-            self.prev_enc = None
-            self.last_text = None
-        if observation['text'] != self.last_text:
```
jaseweston: is it a big speed difference? you could definitely keep this by simply storing the last text as before, i dont see the difference?

alexholdenmiller: fixed!

alexholdenmiller:
there may have been one exchange where the person said the same thing twice in a row and that broke it, but regardless I found a safe way to calculate whether it should reset the cache

PR #744, opened 2018-05-10
Seq2seq opt compatibility

emilydinan:
- Save an opt file if one doesn't exist (this is so eventually we don't need to load opts if an opt file doesn't exist)
- Added opt file to AWS and altered scripts to download it so that scripts are compatible with this change

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -23,11 +23,13 @@ import pickle
 
 class Seq2seqAgent(Agent):
     """Agent which takes an input sequence and produces an output sequence.
+
     This model supports encoding the input and decoding the output via one of
     several flavors of RNN. It then uses a linear layer (whose weights can
     be shared with the embedding layer) to convert RNN output states into
     output tokens. This model currently uses greedy decoding, selecting the
     highest probability token at each time step.
+
     For more information, see the following papers:
     - Neural Machine Translation by Jointly Learning to Align and Translate
       `(Bahdanau et al. 2014) <arxiv.org/abs/1409.0473>`_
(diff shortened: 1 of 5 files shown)


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -645,6 +645,9 @@ def load(self, path):
         if not os.path.isfile(path + '.opt'):
             # backwards compatible to old models
             self.opt = self.override_opt(states['opt'])
+            # save .opt file to make compatible
```
jaseweston: hmm im not sure this shutdown is being called all the time.. i think just put it in the load function (in the init)?

projects/convai2/baselines/seq2seq/eval_f1.py
```diff
@@ -22,6 +22,8 @@
     )
     opt = parser.parse_args(print_args=False)
     opt['model_type'] = 'seq2seq'
-    fnames = ['convai2_self_seq2seq_model.tgz', 'dict_convai2_self']
-    download_models(opt, fnames, 'convai2')
+    opt['override'] = ['dict_file', 'dict_lower', 'batchsize']
```
jaseweston: oh is this the same prob we talked about before? we should just fix it by replacing  "set_defaults" function as we said?

parlai/agents/seq2seq/seq2seq.py
emilydinan: it is called in the load function which is called in init

emilydinan: (it just appears that it is being called in shutdown on github because they are hiding a bunch of lines)

PR #752, opened 2018-05-14
Seq2Seq model sharing change

emilydinan:
- Update seq2seq so that it shares the model (this is necessary for doing human evaluation of the model on Mturk).
- Small change to evaluation task to include model name


alexholdenmiller:
can you run ParlAI/projects/convai2/baselines/seq2seq/eval_ppl.py and eval_f1.py and just make sure they start logging examples without breaking? (that covers both batching and hogwild then, no need to run to completion)

emilydinan:
@alexholdenmiller yep-- still works!

PR #755, opened 2018-05-15
fix bug in dict (ansfreq)

backpropper:
Fixes small bug in ansfreq and remove redundancy in adding answers to dict.
@klshuster can you also remove the unnecessary condition on cuda [here](https://github.com/facebookresearch/ParlAI/blob/ec3c6f0153f16b031e7e70625f809dc4df47ceb7/parlai/agents/mlb_vqa/mlb_vqa.py#L550) and the irrelevant optim states as we discussed [here](https://github.com/facebookresearch/ParlAI/blob/ec3c6f0153f16b031e7e70625f809dc4df47ceb7/parlai/agents/mlb_vqa/mlb_modules.py#L356)?


jaseweston:
can we merge these if they are done?

PR #760, opened 2018-05-15
maintain dialog history fixes

emilydinan:
Fixes to `maintain_dialog_history`. The default option is 'label_else_model' which defaults to 'label' if 'labels' is in the observation and 'model' otherwise. Changed agents in ParlAI to have this as the default option.

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -146,6 +146,11 @@ class Seq2seqAgent(Agent):
                                 'softmax (see arxiv.org/abs/1711.03953).')
         agent.add_argument('-rf', '--report-freq', type=float, default=0.001,
                            help='Report frequency of prediction during eval.')
+        agent.add_argument('-histr', '--history-replies',
+                           default='label_else_model', type=str,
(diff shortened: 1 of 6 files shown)


parlai/core/utils.py
```diff
@@ -245,7 +245,8 @@ def parse(txt, splitSentences=False):
         history['episode_done'] = False
 
     if useReplies != 'none':
-        if useReplies == 'model':
+        if useReplies == 'model' or (useReplies == 'label_else_model' and
+                                     'labels' not in observation):
```
jaseweston: also needs to check eval_labels

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -459,7 +464,7 @@ def observe(self, observation):
                 self.history, obs,
                 reply=self.answers[batch_idx],
                 historyLength=self.truncate,
-                useReplies=self.opt['include_labels'],
+                useReplies=self.opt.get('history_replies', 'label_else_model'),
```
jaseweston: you already set that as the default, but ok

PR #762, opened 2018-05-15
New MTurk timeout error message; fix for giving qualifications

klshuster:
Updated the timeout message so that in tasks with no inter-turker communication, the turker is not confused by the timeout message.

### DIFF
--- a/parlai/mturk/core/mturk_manager.py
+++ b/parlai/mturk/core/mturk_manager.py
@@ -1006,8 +1006,8 @@ class MTurkManager():
         within the expected window.
         """
         # Expire the hit for the disconnected user
-        text = ('You haven\'t entered a message in too long, leaving the other'
-                ' participant unable to complete the HIT. Thus this hit has '
+        text = ('You haven\'t entered a message in too long. As these HITs '
+                ' often require real-time interaction, this hit has '
                 'been expired and you have been considered disconnected. '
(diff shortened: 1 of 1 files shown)


parlai/mturk/core/mturk_manager.py
```diff
@@ -1310,8 +1310,7 @@ def soft_block_worker(self, worker_id):
         """Soft block a worker by giving the worker the block qualification"""
         qual_name = self.opt['block_qualification']
         assert qual_name != '', ('No block qualification has been specified')
-        self.give_worker_qualification(worker_id, qual_name,
-                                       is_sandbox=self.is_sandbox)
+        self.give_worker_qualification(worker_id, qual_name)
```
JackUrb: Why did this change?

JackUrb:
This is looking good to me! Other than that one comment

parlai/mturk/core/mturk_manager.py
klshuster: ``is_sandbox`` is not a keyword arg for ``self.give_worker_qualification`` (though it is an argument for the ``mturk_client``; that's handled in ``self.give_worker_qualification``).

JackUrb: Oh oops I don't think I've ever directly used manager.give_worker_qualification, hence the bug. Thanks for the fix!

PR #766, opened 2018-05-17
Fixes to ConvAI2 PPL eval  + small KvmemNN change

jaseweston:
- make sure build dict doesnt use command line args
- make sure label candidates arent available
- in kvmemnn, allow access externally to attention weights


projects/convai2/eval_ppl.py
```diff
@@ -151,6 +151,7 @@ def parley(self):
 
         # hide labels from model
         labels = action.pop('eval_labels', action.pop('labels', None))
+        action.pop('label_candidates')
```
alexholdenmiller: If you add “, None” after label candidates then there wont be an error if label candidates arent there

PR #778, opened 2018-05-23
Fixes squeeze in kvmemnn that prevented training from working

JackUrb:
Running train with kvmemnn without setting an explicit dimension for the squeeze resulted in some 0 dimension tensor comparisons, which pytorch wasn't happy with, causing the whole thing to error out. 

This fixes the issue (and repairs a ".data[0] will be an error in pytorch 0.5" error).


JackUrb:
@alexholdenmiller metrics are being reported successfully for my current run, so I think all is well

PR #781, opened 2018-05-24
memnn fix for pytorch0.4 and bug in utils split sentence

alexholdenmiller:
- remove Variable
- use item() instead of .data[0]
- use dim=X for softmax
- splitSentences flag was missing in call in maintain_dialog_history, gave strange bug during validation

### DIFF
--- a/parlai/core/utils.py
+++ b/parlai/core/utils.py
@@ -248,7 +248,7 @@ def maintain_dialog_history(history, observation, reply='',
         if useReplies == 'model' or (useReplies == 'label_else_model' and
                                      len(history['labels']) == 0):
             if reply != '':
-                history['dialog'].extend(parse(reply))
+                history['dialog'].extend(parse(reply, splitSentences))
         elif len(history['labels']) > 0:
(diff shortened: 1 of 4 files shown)


parlai/core/utils.py
```diff
@@ -248,7 +248,7 @@ def parse(txt, splitSentences=False):
         if useReplies == 'model' or (useReplies == 'label_else_model' and
                                      len(history['labels']) == 0):
             if reply != '':
-                history['dialog'].extend(parse(reply))
+                history['dialog'].extend(parse(reply, splitSentences))
```
jaseweston: hmm, would this be affecting our kvmemnn too then or what?

alexholdenmiller: I'm not 100% certain. this actually caused an error here or I wouldn't have found it because it was returning:
`[ int, list, list, list, list, list ]`

this section should only be entered when the model reply is being used in the history. which, actually shouldn't be happening for memnn here either: it should use the labels. let me double check that.

PR #782, opened 2018-05-24
Opt fix

emilydinan:
Add keys to opt loaded from model file that aren't there.

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -276,7 +276,7 @@ def name_to_agent_class(name):
 
 def load_agent_module(opt):
     model_file = opt['model_file']
-    optfile =  model_file + '.opt'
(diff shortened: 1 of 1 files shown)


parlai/core/agents.py
```diff
@@ -287,6 +287,9 @@ def load_agent_module(opt):
                 print("[ warning: overriding opt['" + str(k) + "'] to " + str(v) +
```
alexholdenmiller: actually can you change the loop to `for k, v in opt['override'].items()`? a little cleaner that way

alexholdenmiller:
good catch!

alexholdenmiller:
helps protect people who didn't use `opt.get` (good reminder--if you add params to your model code, you should try to use `opt.get` instead of `opt[]` with a sensible default value)

PR #785, opened 2018-05-25
Update DictionaryAgent with methods to standardize data preparation

vgonzalez88:
Added methods `vectorize`, `permute` and `unpermute` in an attempt to standardize the data preparation process across different agents.

To test correctness as well as provide an example implementation, I added a new `batch_act` method to `example_seq2seq.py` and confirmed that it trained.

### DIFF
--- a/parlai/agents/example_seq2seq/example_seq2seq.py
+++ b/parlai/agents/example_seq2seq/example_seq2seq.py
@@ -287,7 +287,7 @@ class ExampleSeq2seqAgent(Agent):
 
         return xs, ys, labels, valid_inds, is_training
 
-    def batch_act(self, observations):
+    def batch_act_deprecated(self, observations):
         batchsize = len(observations)
         # initialize a table of replies with this agent's id
         batch_reply = [{'id': self.getID()} for _ in range(batchsize)]
@@ -312,6 +312,41 @@ class ExampleSeq2seqAgent(Agent):
             self.dict, self.END_IDX, labels=labels,
             answers=labels, ys=ys.data if ys is not None else None,
             report_freq=self.opt.get('report_freq', 0))
+
+        return batch_reply
+
+    def batch_act(self, observations):
+        batch_size = len(observations)
+        # initialize a table of replies with this agent's id
+        batch_reply = [{'id': self.getID()} for _ in range(batch_size)]
+
+        is_training = any(['labels' in obs for obs in observations])
+
+        mode = 'train' if is_training else 'eval'
+
+        vec_obs = [self.dict.vectorize(obs, mode, self.use_cuda)
+                   for obs in observations]
+
+        xs, ys, labels, valid_inds, _, _ = self.dict.permute(vec_obs, use_cuda=self.use_cuda)
+
+        if xs is None:
+            return batch_reply
+
+        predictions = self.predict(xs, ys, is_training)
+
+        unpermute_pred = self.dict.unpermute(predictions.cpu().data,
+                                             valid_inds, batch_size)
+        # Format the predictions into reply format
+        for rep, pred in zip(batch_reply, unpermute_pred):
+            if pred is not None:
+                output_tokens = []
+                # Remove the final END_TOKEN that is appended to predictions
+                for i, token in enumerate(pred):
+                    if token == self.END_IDX and i != 0:
+                        break
+                    else:
+                        output_tokens.append(token)
+                rep['text'] = self.dict.vec2txt(output_tokens)
 
         return batch_reply
 
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -11,7 +11,11 @@ import argparse
 import copy
 import numpy as np
 import os
-
+import random
+try:
+    import torch
+except Exception as e:
+    raise ModuleNotFoundError('Need to install Pytorch: go to pytorch.org')
 
 def escape(s):
     """Replace potential special characters with escaped version.
@@ -492,6 +496,115 @@ class DictionaryAgent(Agent):
                     if text:
                         self.add_to_dict(self.tokenize(text))
         return {'id': 'Dictionary'}
+
+    def vectorize(self, obs, mode='train', use_cuda=True):
+        """
+        Converts 'text' and 'label'/'eval_label' field to vectors
+        kwargs:
+        -mode: which label should be passed through, is either train, eval or test
+        -use_cuda: whether to use cuda
+        """
+        if 'text' not in obs:
+            return obs
+        # convert 'text' field to vector using self.txt2vec and then a tensor
+        obs['text'] = torch.LongTensor(self.txt2vec(obs['text']))
+        if use_cuda:
+            obs['text'] = obs['text'].cuda()
+
+        if mode == 'train':
+            assert 'labels' in obs
+            new_labels = []
+            for label in obs['labels']:
+                vec_label = self.txt2vec(label)
+                vec_label.append(self[self.end_token])
+                new_label = torch.LongTensor(vec_label)
+                if use_cuda:
+                    new_label = new_label.cuda()
+                new_labels.append(new_label)
+            obs['labels'] = new_labels
+            # remove the evaluation label if included
+            obs.pop('eval_labels', None)
+        elif mode == 'eval' or mode == 'test':
+            if 'eval_labels' in obs:
+                new_labels = []
+                for label in obs['eval_labels']:
+                    vec_label = self.txt2vec(label)
+                    vec_label.append(self[self.end_token])
+                    new_label = torch.LongTensor(vec_label)
+                    if use_cuda:
+                        new_label = new_label.cuda()
+                    new_labels.append(new_label)
+                obs['eval_labels'] = new_labels
+            obs.pop('labels', None)
+        else:
+            raise RuntimeError('Mode {} not supported. Mode must be \'train\', \'eval\', or \'test\'.'.format(mode))
+
+        return obs
+
+    def permute(self, obs_batch, sort=True, is_valid=lambda obs: 'text' in obs, use_cuda=True):
+        """Creates a batch of valid observations from an unchecked batch.
+        Assumes each observation has been vectorized by vectorize function.
+        """
+        try:
+            # valid examples and their indices
+            valid_inds, exs = zip(*[(i, ex) for i, ex in
+                                    enumerate(obs_batch) if is_valid(ex)])
+        except ValueError:
+            # zero examples to process in this batch, so zip failed to unpack
+            return None, None, None, None, None, None
+        x_text = [ex['text'] for ex in exs]
+        x_lens = [ex.shape[0] for ex in x_text]
+
+        if sorted:
+            ind_sorted = sorted(range(len(x_lens)), key=lambda k: -x_lens[k])
+
+            exs = [exs[k] for k in ind_sorted]
+            valid_inds = [valid_inds[k] for k in ind_sorted]
+            x_text = [x_text[k] for k in ind_sorted]
+            end_idxs = [x_lens[k] for k in ind_sorted]
+
+        padded_xs = torch.LongTensor(len(exs),
+                                     max(x_lens)).fill_(self[self.null_token])
+        if use_cuda:
+            padded_xs = padded_xs.cuda()
+
+        for i, ex in enumerate(x_text):
+            padded_xs[i, :ex.shape[0]] = ex
+
+        xs = padded_xs
+
+        eval_labels_avail = any(['eval_labels' in ex for ex in exs])
+        labels_avail = any(['labels' in ex for ex in exs])
+        some_labels_avail = eval_labels_avail or labels_avail
+
+        # set up the target tensors
+        ys = None
+        labels = None
+        y_lens = None
+        if some_labels_avail:
+            # randomly select one of the labels to update on (if multiple)
+            if labels_avail:
+                labels = [random.choice(ex.get('labels', [''])) for ex in exs]
+            else:
+                labels = [random.choice(ex.get('eval_labels', [''])) for ex in exs]
+            y_lens = [y.shape[0] for y in labels]
+            padded_ys = torch.LongTensor(len(exs),
+                                         max(y_lens)).fill_(self[self.null_token])
+            if use_cuda:
+                padded_ys = padded_ys.cuda()
+            for i, y in enumerate(labels):
+                padded_ys[i, :y.shape[0]] = y
+            ys = padded_ys
+        return xs, ys, labels, valid_inds, end_idxs, y_lens
+
+    def unpermute(self, predictions, valid_inds, batch_size):
+        """Re-order permuted predictions to the initial ordering.
+        """
+        unpermuted = [None]*batch_size
+        for pred, idx in zip(predictions, valid_inds):
+            unpermuted[idx] = pred
+        return unpermuted
+
 
     def share(self):
         shared = {}


parlai/core/dict.py
```diff
@@ -493,6 +497,115 @@ def act(self):
...
+            for label in obs['labels']:
+                vec_label = self.txt2vec(label)
+                vec_label.append(self[self.end_token])
+                new_label = torch.LongTensor(vec_label)
+                if use_cuda:
+                    new_label = new_label.cuda()
+                new_labels.append(new_label)
+            obs['labels'] = new_labels
+            # remove the evaluation label if included
+            obs.pop('eval_labels', None)
```
alexholdenmiller: don't need this

parlai/core/dict.py
```diff
@@ -493,6 +497,115 @@ def act(self):
...
+                new_labels = []
+                for label in obs['eval_labels']:
+                    vec_label = self.txt2vec(label)
+                    vec_label.append(self[self.end_token])
+                    new_label = torch.LongTensor(vec_label)
+                    if use_cuda:
+                        new_label = new_label.cuda()
+                    new_labels.append(new_label)
+                obs['eval_labels'] = new_labels
+            obs.pop('labels', None)
```
alexholdenmiller: don't need this

parlai/core/dict.py
```diff
@@ -493,6 +497,115 @@ def act(self):
...
+        """
+        if 'text' not in obs:
+            return obs
+        # convert 'text' field to vector using self.txt2vec and then a tensor
+        obs['text'] = torch.LongTensor(self.txt2vec(obs['text']))
+        if use_cuda:
+            obs['text'] = obs['text'].cuda()
+
+        if mode == 'train':
+            assert 'labels' in obs
```
alexholdenmiller: this assertion only applies to some models, e.g. unsupervised training may not use text-label pairs but could still want to train on the text.

I think actually you can cut `mode` from the arguments completely and just parse the labels or the eval labels if they are there, otherwise not.

parlai/core/dict.py
```diff
@@ -11,7 +11,11 @@
 import copy
 import numpy as np
 import os
-
+import random
+try:
+    import torch
+except Exception as e:
+    raise ModuleNotFoundError('Need to install Pytorch: go to pytorch.org')
```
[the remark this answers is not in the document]
alexholdenmiller: update: talked with kurt about moving all of these functions from DictAgent to a new TorchAgent, parlai/agents/torch_agent.py (maybe jason will prefer it somewhere else but that's a good start).

parlai/core/dict.py
```diff
@@ -493,6 +497,115 @@ def act(self):
...
+                padded_ys = padded_ys.cuda()
+            for i, y in enumerate(labels):
+                padded_ys[i, :y.shape[0]] = y
+            ys = padded_ys
+        return xs, ys, labels, valid_inds, end_idxs, y_lens
+
+    def unpermute(self, predictions, valid_inds, batch_size):
+        """Re-order permuted predictions to the initial ordering.
+        """
+        unpermuted = [None]*batch_size
```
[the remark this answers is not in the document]
vgonzalez88: Isn't it possible that we have a sequence of [{}, {'text':'this is a question'}, {}]? I thought I was supposed to get back this sequence with unpermute, but I wouldn't know how many trailing empty observations there were without batch size.

parlai/agents/torch_agent/torch_agent.py
```diff
@@ -0,0 +1,154 @@
...
+        self.END_IDX = self.dict[self.dict.end_token]
+        self.START_IDX = self.dict[self.dict.start_token]
+
+    def share(self):
+        shared = super().share()
+        shared['opt'] = self.opt
+        shared['dict'] = self.dict
+        return shared
+
+    def vectorize(self, obs, use_cuda=True):
```
jaseweston: default cuda on true might not be good, can't it just take this from the parent  self.use_cuda or something? it seems to be there?

parlai/agents/torch_agent/torch_agent.py
```diff
@@ -0,0 +1,154 @@
...
+            new_labels = []
+            for label in obs['labels']:
+                vec_label = self.dict.txt2vec(label)
+                vec_label.append(self.END_IDX)
+                new_label = torch.LongTensor(vec_label)
+                if use_cuda:
+                    new_label = new_label.cuda()
+                new_labels.append(new_label)
+            obs['labels'] = new_labels
+        elif 'eval_labels' in obs:
```
jaseweston: repeated code

parlai/agents/torch_agent/torch_agent.py
```diff
@@ -0,0 +1,154 @@
...
+        # convert 'text' field to vector using self.txt2vec and then a tensor
+        obs['text'] = torch.LongTensor(self.dict.txt2vec(obs['text']))
+        if use_cuda:
+            obs['text'] = obs['text'].cuda()
+
+        if 'labels' in obs:
+            new_labels = []
+            for label in obs['labels']:
+                vec_label = self.dict.txt2vec(label)
+                vec_label.append(self.END_IDX)
```
jaseweston: makes an assumption that you use this, some models don't

parlai/agents/torch_agent/torch_agent.py
```diff
@@ -0,0 +1,154 @@
...
+                new_label = torch.LongTensor(vec_label)
+                if use_cuda:
+                    new_label = new_label.cuda()
+                new_labels.append(new_label)
+            obs['eval_labels'] = new_labels
+
+        return obs
+
+    def permute(self, obs_batch, sort=True, is_valid=lambda obs: 'text' in obs, use_cuda=True):
```
jaseweston: permute is an unclear name

parlai/agents/torch_agent/torch_agent.py
jaseweston: also in general i think it's bad we are encoding the eval_labels by default, we shouldn't be using these most of the time.. need to think about this but can go with this for now

parlai/agents/torch_agent/torch_agent.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        Converts 'text' and 'label'/'eval_label' field to vectors
+        kwargs:
+        -mode: which label should be passed through, is either train, eval or test
+        -use_cuda: whether to use cuda
+        """
+        if 'text' not in obs:
+            return obs
+        # convert 'text' field to vector using self.txt2vec and then a tensor
+        obs['text'] = torch.LongTensor(self.dict.txt2vec(obs['text']))
```
alexholdenmiller: maybe better to move these to vector fields instead of overwriting the original, in case the subclass wants the plain text as well. (e.g. we use text2vec, maybe text_vec & label_vec and eval_label_vec etc is even better)

parlai/agents/torch_agent/torch_agent.py
alexholdenmiller: any choice like this should be parameterized in the function keywords

parlai/agents/torch_agent/torch_agent.py
alexholdenmiller: any suggestions? samuel suggested permute (i.e. by using it)

some other options: map_valid & unmap_valid?

parlai/agents/torch_agent/torch_agent.py
```diff
@@ -0,0 +1,218 @@
...
+                labels = [random.choice(ex.get('eval_labels', [''])) for ex in exs]
+            y_lens = [y.shape[0] for y in labels]
+            padded_ys = torch.LongTensor(len(exs),
+                                         max(y_lens)).fill_(self.NULL_IDX)
+            if self.use_cuda:
+                padded_ys = padded_ys.cuda()
+            for i, y in enumerate(labels):
+                padded_ys[i, :y.shape[0]] = y
+            ys = padded_ys
+        return xs, ys, labels, valid_inds, end_idxs, y_lens
```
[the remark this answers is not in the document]
vgonzalez88: So I looked through the repo and it seems none of our models uses that output from `pad_text`, though LanguageModel does use `y_lens`. I'll remove it from this function since it isn't used.

vgonzalez88:
I implemented all of the changes discussed here and offline. The added unit tests are being passed and should cover the utility functions, except for `maintain_dialog_history`.

parlai/agents/torch/torch.py
```diff
@@ -0,0 +1,257 @@
+# Copyright (c) 2017-present, Facebook, Inc.
```
jaseweston: should it be in core?

parlai/agents/torch/torch.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+
+    @staticmethod
+    def dictionary_class():
+        return DictionaryAgent
+
+    @staticmethod
+    def add_cmdline_args(argparser):
+        agent = argparser.add_argument_group('TorchAgent Arguments')
+        agent.add_argument('-hist', '--history-length', default=10, type=int,
```
jaseweston: i think often you want to parameterize on number of utterances in history also.. i guess could be another param

jaseweston: default 10 is weird

parlai/agents/torch/torch.py
```diff
@@ -0,0 +1,257 @@
...
+                                     max(x_lens)).fill_(self.NULL_IDX)
+        if self.use_cuda:
+            padded_xs = padded_xs.cuda()
+
+        for i, ex in enumerate(x_text):
+            padded_xs[i, :ex.shape[0]] = ex
+
+        xs = padded_xs
+
+        eval_labels_avail = any(['eval_labels_vec' in ex for ex in exs])
```
jaseweston: having eval_labels everywhere is not good. the whole point of eval_labels was to protect models from having access to the labels.. i think we should switch to the observe(own action) thing now?

vgonzalez88: During testing/validation don't those labels get sent through via eval_labels? What is this observe thing?

parlai/agents/torch/torch.py
vgonzalez88: I wasn't sure as it's specific, but I could see it living in core because it's task-independent. I'm leaning towards moving it to core.

parlai/agents/example_seq2seq/example_seq2seq.py
```diff
@@ -262,56 +242,36 @@ def predict(self, xs, ys=None, is_training=False):
...
-        xs = Variable(xs)
-
-        if ys is not None:
-            ys = torch.LongTensor(ys)
-            if self.use_cuda:
-                ys = ys.cuda()
-            ys = Variable(ys)
-
-        return xs, ys, labels, valid_inds, is_training
-
```
alexholdenmiller: all this red makes me happy

parlai/agents/example_seq2seq/example_seq2seq.py
```diff
@@ -262,56 +242,36 @@ def predict(self, xs, ys=None, is_training=False):
...
         predictions = self.predict(xs, ys, is_training)
 
-        # maps returns predictions back to the right `valid_inds`
-        # in the example above, a prediction `world` should reply to `hello`
-        PaddingUtils.map_predictions(
-            predictions.cpu().data, valid_inds, batch_reply, observations,
-            self.dict, self.END_IDX, labels=labels,
-            answers=labels, ys=ys.data if ys is not None else None,
-            report_freq=self.opt.get('report_freq', 0))
+        unpermute_pred = self.unmap_valid(predictions.cpu().data,
```
alexholdenmiller: "permute"

`predictions.cpu().data` => `predictions`, and then can probably move this call onto one line

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+
+        :param obs: single observation from observe function
+        :param addEndIdx: default True, adds the end token to each label
+        """
+        if 'text' not in obs:
+            return obs
+        # convert 'text' field to vector using self.txt2vec and then a tensor
+        obs['text_vec'] = torch.LongTensor(self.dict.txt2vec(obs['text']))
+        if self.use_cuda:
+            obs['text'] = obs['text'].cuda()
```
alexholdenmiller: `text_vec`

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+
+        label_type = None
+        if 'labels' in obs:
+            label_type = 'labels_vec'
```
alexholdenmiller: nit: I'd prefer this be `labels` without vec

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+        elif 'eval_labels' in obs:
+            label_type = 'eval_labels_vec'
+
+        if label_type is not None:
+            new_labels = []
+            for label in obs[label_type[:-4]]: # leave off the _vec portion
```
alexholdenmiller: then this slicing isn't needed

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+                vec_label = self.dict.txt2vec(label)
+                if addEndIdx:
+                    vec_label.append(self.END_IDX)
+                new_label = torch.LongTensor(vec_label)
+                if self.use_cuda:
+                    new_label = new_label.cuda()
+                new_labels.append(new_label)
+            obs[label_type] = new_labels
```
alexholdenmiller: and this can be label_type + '_vec'

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+try:
+    import torch
+except Exception as e:
+    raise ModuleNotFoundError('Need to install Pytorch: go to pytorch.org')
+
+from collections import deque, namedtuple
+import random
+
+Batch = namedtuple("Batch", ["text_vecs", "label_vecs", "labels",
+                             "valid_inds"])
```
alexholdenmiller:
```
Batch = namedtuple("Batch", [
    "text_vec",  # bsz x seqlen tensor containing the parsed text data
    "label_vec", # bsz x seqlen tensor containing the parsed label (one per batch row)
    "labels", # list of length bsz containing the selected label for each batch row (some datasets have multiple labels per input example)
    "valid_indices",  # list of length bsz containing the original indices of each example in the batch. we use these to map predictions back to their proper row, since e.g. we may sort examples by their length or some examples may be invalid.
])
```

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+            obs[label_type] = new_labels
+        return obs
+
+    def map_valid(self, obs_batch, sort=True, is_valid=lambda obs: 'text_vec' in obs):
+        """Creates a batch of valid observations from an unchecked batch.
+        Assumes each observation has been vectorized by vectorize function.
+
+        Returns a namedtuple Batch that contains, in order, batched text
+        vectors, batched label vectors, batched labels, indices of valid
+        observations, and lengths of label vectors.
```
alexholdenmiller: out of date

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+
+        :param obs_batch: list of vectorized observations
+        :param sort: default True, orders the observations by length of vector
+        :param is_valid: default lambda, determines if an observation is valid
```
alexholdenmiller: lambda isn't a default :)

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+            else:
+                num_choices = [len(ex.get('eval_labels_vec', [])) for ex in exs]
+                choices = [random.choice(range(num)) if num != 0 else -1
+                           for num in num_choices]
+                label_vecs = [ex['eval_labels_vec'][choices[i]]
+                              if choices[i] != -1 else torch.LongTensor([])
+                              for i, ex in enumerate(exs)]
+                labels = [ex['eval_labels'][choices[i]]
+                          if choices[i] != -1 else ''
+                          for i, ex in enumerate(exs)]
```
alexholdenmiller: should combine this repeated code unless we move eval_labels completely (@jaseweston and I had discussed a way of doing this but it's not in the works yet)

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+        unpermuted = [None]*batch_size
+        for pred, idx in zip(predictions, valid_inds):
+            unpermuted[idx] = pred
+        return unpermuted
+
+    def maintain_dialog_history(self, observation, reply='',
+                                useStartEndIndices=True,
+                                splitSentences=False):
+        """Keeps track of dialog history, up to a truncation length.
+        Either includes replies from the labels, model, or not all using param 'replies'."""
```
alexholdenmiller: actually, can you update this function to use the current maintain_dialog_history? made a couple changes in another diff to the useStartEndIndices functionality (which should default to False)

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+    def dictionary_class():
+        return DictionaryAgent
+
+    @staticmethod
+    def add_cmdline_args(argparser):
+        agent = argparser.add_argument_group('TorchAgent Arguments')
+        agent.add_argument('-histk', '--history-tokens', default=10000, type=int,
+                           help='Number of past tokens to remember.')
+        agent.add_argument('-histd', '--history-dialog', default=10, type=int,
+                           help='Number of past dialog examples to remember.')
```
alexholdenmiller: use -1 (all) as default

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @staticmethod
+    def dictionary_class():
+        return DictionaryAgent
+
+    @staticmethod
+    def add_cmdline_args(argparser):
+        agent = argparser.add_argument_group('TorchAgent Arguments')
+        agent.add_argument('-histk', '--history-tokens', default=10000, type=int,
+                           help='Number of past tokens to remember.')
```
alexholdenmiller: same here

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+        unpermuted = [None]*batch_size
+        for pred, idx in zip(predictions, valid_inds):
+            unpermuted[idx] = pred
+        return unpermuted
+
+    def maintain_dialog_history(self, observation, reply='',
+                                useStartEndIndices=False,
+                                splitSentences=False):
+        """Keeps track of dialog history, up to a truncation length.
```
klshuster: might want to also list the :params here as well (just to keep in line with the previous functions in this class)

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,258 @@
...
+
+Batch = namedtuple("Batch", [
+    "text_vec",  # bsz x seqlen tensor containing the parsed text data
+    "label_vec", # bsz x seqlen tensor containing the parsed label (one per batch row)
+    "labels", # list of length bsz containing the selected label for each batch row (some datasets have multiple labels per input example)
+    "valid_indices",  # list of length bsz containing the original indices of each example in the batch. we use these to map predictions back to their proper row, since e.g. we may sort examples by their length or some examples may be invalid.
+])
+
+class TorchAgent(Agent):
+    """Base Agent for all models which use Torch.
```
jaseweston: this might be a little too dramatic..  maybe add something about how you don't actually HAVE to use it..

parlai/core/torch_agent.py
```diff
@@ -0,0 +1,258 @@
...
+                    vec_label.append(self.END_IDX)
+                new_label = torch.LongTensor(vec_label)
+                if self.use_cuda:
+                    new_label = new_label.cuda()
+                new_labels.append(new_label)
+            obs[label_type + "_vec"] = new_labels
+        return obs
+
+    def map_valid(self, obs_batch, sort=True, is_valid=lambda obs: 'text_vec' in obs):
+        """Creates a batch of valid observations from an unchecked batch.
```
jaseweston: explain what 'valid' means?

parlai/agents/example_seq2seq/example_seq2seq.py
```diff
@@ -262,56 +242,35 @@ def predict(self, xs, ys=None, is_training=False):
...
+        vec_obs = [self.vectorize(obs)
+                   for obs in observations]
 
-        # convert the observations into batches of inputs and targets
-        # `labels` stores the true labels returned in the `ys` vector
-        # `valid_inds` tells us the indices of all valid examples
-        # e.g. for input [{}, {'text': 'hello'}, {}, {}], valid_inds is [1]
-        # since the other three elements had no 'text' field
-        xs, ys, labels, valid_inds, is_training = self.vectorize(observations)
+        xs, ys, labels, valid_inds = self.map_valid(vec_obs)
```
alexholdenmiller: can you confirm that this still works with the namedtuple?

alexholdenmiller: if not this may change to `batch = self.map_valid(vec_obs)` and below can do e.g. `self.predict(batch.text_vecs, ...`

alexholdenmiller: actually, yeah can you run e.g.
```
py examples/train_model.py -t babi:task10k:1 --dict-file /tmp/dict_babi:task10k:1 -bs 32 -vtim 30 -vcut 0.95 -m example_seq2seq
```
(and again with --no-cuda set) on your dev machine and make sure both work?

vgonzalez88: Confirmed that it trains both with and without GPU!

Issue #749, opened 2018-05-14
Torch 0.4 migration


jsedoc:
Is there an ETA for the migration to torch 0.4?

jsedoc:
I actually use ParlAI for a lot, so seq2seq, language_model, MemNN, Fairseq, and starspace

alexholdenmiller:
seq2seq and lm should be good, fairseq will be a little longer as we need to update to current fairseq not just update for pytorch 0.4 (significant changes to the current agent). memnn and starspace might be behind but are probably easy to fix if you want to submit a PR :)

zhaoguangxiang:
What about persona-chat and convAI2?

jsedoc:
upvote for both.

Also, for the LM and seq2seq, is there a 0.4 branch to pull from?

alexholdenmiller:
@jsedoc for lm and seq2seq, should be good on master

for the kvmemnn and persona_seq2seq models, same status as memnn and starspace--will do when we can!

alexholdenmiller:
fix for memnn: #781

jaseweston:
fix for starspace https://github.com/facebookresearch/ParlAI/pull/786

alexholdenmiller:
all models mentioned in this thread except for fairseq are updated. going to close for a fairseq specific thread.

PR #786, opened 2018-05-25
fix starspace for new pytorch

klshuster:
Fix for starspace to work in torch (>=) 0.4; #749

### DIFF
--- a/parlai/agents/starspace/starspace.py
+++ b/parlai/agents/starspace/starspace.py
@@ -240,7 +240,7 @@ class StarspaceAgent(Agent):
     def same(self, y1, y2):
         if len(y1.squeeze(0)) != len(y2.squeeze(0)):
             return False
-        if abs((y1.squeeze()-y2.squeeze()).sum().data.sum()) > 0.00001:
+        if abs((y1.squeeze(0)-y2.squeeze(0)).sum().data.sum()) > 0.00001:
             return False
         return True
 
--- a/parlai/agents/starspace/starspace.py
+++ b/parlai/agents/starspace/starspace.py
@@ -15,7 +15,6 @@ from parlai.core.thread_utils import SharedTable
 from .modules import Starspace
 
 import torch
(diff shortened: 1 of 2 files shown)


parlai/agents/starspace/starspace.py
```diff
@@ -243,7 +238,7 @@ def observe(self, observation):
         return obs
 
     def same(self, y1, y2):
-        if len(y1.squeeze()) != len(y2.squeeze()):
+        if len(y1.squeeze(0)) != len(y2.squeeze(0)):
             return False
         if abs((y1.squeeze()-y2.squeeze()).sum().data.sum()) > 0.00001:
```
alexholdenmiller: the squeeze here maybe breaks

parlai/agents/starspace/starspace.py
```diff
@@ -386,9 +381,8 @@ def valid(obs):
 
         max_x_len = max([len(x) for x in parsed_x])
         for x in parsed_x:
-            x += [[self.NULL_IDX]] * (max_x_len - len(x))
+            x += [self.NULL_IDX] * (max_x_len - len(x))
```
alexholdenmiller: are you sure about this change? :)

jaseweston:
this seems to be idle..

parlai/agents/starspace/starspace.py
klshuster: I'll change it to `squeeze(0)` to just be safe

parlai/agents/starspace/starspace.py
klshuster: I'm pretty sure this is correct; `x` is a `deque` of `int`s here, so it would not make sense to add a list of a lists to a deque of ints (rather, we should add a list of ints to a deque of ints)

klshuster:
made the fix to `squeeze` and ensured that this works with hogwild

jaseweston:
cool let this be got in!!

PR #792, opened 2018-05-30
add ability to use start & end tokens as p1 / p2 in context

alexholdenmiller:
- plus some general cleanup
- removing of Variables from s2s
- don't download convai2 models when a different model file is specified (e.g. don't download remote when using local)

### DIFF
--- a/parlai/core/utils.py
+++ b/parlai/core/utils.py
@@ -249,12 +249,7 @@ def maintain_dialog_history(history, observation, reply='',
     obs = observation
     if 'text' in obs:
         if useStartEndIndices:
-            if '\n' in obs['text']:
-                split = obs['text'].split('\n')
-                split[-1] = dict.end_token + ' ' + split[-1]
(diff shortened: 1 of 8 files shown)


parlai/core/utils.py
```diff
@@ -252,14 +246,26 @@ def parse(txt, splitSentences):
...
-        history['labels'] = observation['eval_labels']
+    obs = observation
+    if 'text' in obs:
+        if useStartEndIndices:
+            if '\n' in obs['text']:
+                split = obs['text'].split('\n')
+                split[-1] = dict.end_token + ' ' + split[-1]
+                obs['text'] = '\n'.join(split)
+            else:
+                obs['text'] = dict.end_token + ' ' + obs['text']
```
jaseweston: hmm..did you get my comment on this? seems lost
i dont get it.. how is splitting the 'text' in obs related to the people talking ..
and i think it should be based on id field

PR #796, opened 2018-05-30
move eval_ppl to a parent script

alexholdenmiller:
- move `projects/convai2/eval_ppl.py` generic functionality to `parlai/scripts/eval_ppl.py`
- convai2-specific functionality and documentation stays in `projects/convai2/eval_ppl.py`

question: should we move the example model to scripts? I kept it in convai2 right now since it is dependent on the `build_dict()` function being available, which I made a dependency on the eval_ppl in scripts


alexholdenmiller:
this is a prereq for adding twitter project folder with similar eval/build_dict/train functions

PR #797, opened 2018-05-31
Retrieve and Read

klshuster:
Retrieve and read agent (also adds open squad teacher)

### DIFF
--- a/parlai/agents/drqa/layers.py
+++ b/parlai/agents/drqa/layers.py
@@ -204,12 +204,14 @@ class BilinearSeqAttn(nn.Module):
         Wy = self.linear(y) if self.linear is not None else y
         xWy = x.bmm(Wy.unsqueeze(2)).squeeze(2)
         xWy.data.masked_fill_(x_mask.data, -float('inf'))
+        self.xWy = xWy
         if self.training:
             # In training we output log-softmax for NLL
             alpha = F.log_softmax(xWy)
         else:
             # ...Otherwise 0-1 probabilities
             alpha = F.softmax(xWy)
+            # alpha = xWy.exp()
         return alpha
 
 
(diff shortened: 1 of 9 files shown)


parlai/agents/drqa/layers.py
```diff
@@ -204,12 +204,14 @@ def forward(self, x, y, x_mask):
...
         xWy.data.masked_fill_(x_mask.data, -float('inf'))
+        self.xWy = xWy
         if self.training:
             # In training we output log-softmax for NLL
             alpha = F.log_softmax(xWy)
         else:
             # ...Otherwise 0-1 probabilities
-            alpha = F.softmax(xWy)
+            # alpha = F.softmax(xWy)
+            alpha = xWy.exp()
```
alexholdenmiller: just curious, why this change?

parlai/core/agents.py
```diff
@@ -327,6 +327,13 @@ def create_agent(opt, requireModelExists=False):
     the options file if it exists (the file opt['model_file'] + '.opt' must exist and
     contain a pickled dict containing the model's options).
     """
+    if opt.get('datapath', None) is None:
+        # add datapath, it is missing
+        from parlai.core.params import ParlaiParser
+        parser = ParlaiParser(add_model_args=True)
```
alexholdenmiller: I think you can change this to just `parser = ParlaiParser()` (don't need the model args)

alexholdenmiller: or even `parser = ParlaiParser(add_parlai_args=False); parser.add_parlai_datapath()` since that's all we need

parlai/tasks/squad/agents.py
```diff
@@ -115,6 +115,40 @@ def setup_data(self, path):
                     yield (context + '\n' + question, answers), True
 
 
+class OpenSquadTeacher(DialogTeacher):
+    """This version of SQuAD inherits from the core Dialog Teacher, which just
+    requires it to define an iterator over its data `setup_data` in order to
+    inherit basic metrics, a default `act` function.
+    For SQuAD, this does not efficiently store the paragraphs in memory.
```
alexholdenmiller: You can remove this comment about the paragraphs

PR #798, opened 2018-05-31
Tensorboard log with tensorboardx package in train_model.py script

uralik:
It is quite comfortable to compare curves rather than numbers. TensorboardX provides single api to log any kind of numbers in TB and can be used separate from tensorflow.

However, in order to run actual tensorboard you need to install any basic tensorflow from pip.

Right now there is an option to turn logging on (off by default), and one can also configure the metrics to track, options to keep in the event file name (useful to distinguish different setups).

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -14,13 +14,18 @@ SPHINXPROJ    = ParlAI
 SOURCEDIR     = source
 BUILDDIR      = build
 
+
 # Put it first so that "make" without argument is like "make help".
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
 .PHONY: help Makefile
 
+clean:
+	rm -rfv "${BUILDDIR}"
+
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
+	cd "${SOURCEDIR}"; python generate_task_list.py
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
--- a/docs/source/tutorial_worlds.rst
+++ b/docs/source/tutorial_worlds.rst
@@ -263,18 +263,18 @@ Multiprocessed Pytorch Dataloader
 For large datasets, where it is best to stream from disk during training
 rather than load initially into memory, we provide a teacher that utilizes pytorch data loading.
 
-(Note: the module `here <https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/pytorch_data_teacher.py>`_
+(Note: the module `here <https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/pytorch_data_teacher.py>`__
 contains all of the code discussed in this tutorial)
 
 Pytorch Dataloading Intro
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 A Pytorch ``DataLoader`` is a dataloading mechanism that provides multiprocessed
-loading of data from disk (as described `here <http://pytorch.org/tutorials/beginner/data_loading_tutorial.html>`_).
+loading of data from disk (as described `here <http://pytorch.org/tutorials/beginner/data_loading_tutorial.html>`__).
 A ``DataLoader`` can be initialized with a variety of different options; the only
 ones that concern us are ``dataset`` and ``collate_fn``.
 
 The ``dataset`` is a
-Pytorch ``Dataset`` (as described `here <http://pytorch.org/tutorials/beginner/data_loading_tutorial.html>`_),
+Pytorch ``Dataset`` (as described `here <http://pytorch.org/tutorials/beginner/data_loading_tutorial.html>`__),
 which is a class that implements two functions: ``__getitem__(self, idx)`` and ``__len__(self)``.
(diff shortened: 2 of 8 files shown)


uralik:
related to this [223](https://github.com/facebookresearch/ParlAI/pull/223)

parlai/tasks/twitter/build.py
```diff
@@ -8,8 +8,8 @@
 try:
     from emoji.unicode_codes import UNICODE_EMOJI
     import unidecode
-except ModuleNotFound:
```
uralik: this is not related to tensorboard, but also a bug

alexholdenmiller: nice

parlai/core/logs.py
```diff
@@ -0,0 +1,76 @@
...
+        self.__dict__ = self._shared_state
+
+
+class TensorboardLogger(Shared):
+    def __init__(self, opt):
+        Shared.__init__(self)
+
+        if isinstance(opt, ParlaiParser):
+            print('[ Deprecated Warning: TrainLoop should be passed opt not Parser ]')
+            opt = opt.parse_args()
```
alexholdenmiller: shouldn't need this

parlai/core/logs.py
```diff
@@ -0,0 +1,76 @@
...
+        assert opt['tensorboard_log'] is True, 'TensorboardLogger class requires tensorboard_log to be True'
```
alexholdenmiller: I think you don't need this

parlai/core/logs.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+"""
+This file provides interface to log any metrics in tensorboard, could be extended to any other tool like visdom
+"""
```
alexholdenmiller: Maybe good to add a comment on how to access the logs here?

parlai/scripts/train_model.py
```diff
@@ -81,6 +82,12 @@ def setup_args(parser=None):
     train.add_argument('-lfc', '--load-from-checkpoint',
                        type='bool', default=False,
                        help='load model from checkpoint if available')
+    train.add_argument('-tblog', '--tensorboard-log', type=bool, default=False,
```
jaseweston: we can add these from the tensorboardlogger

jaseweston: like how we add from models or other things (you extend the args)

uralik: got it

parlai/scripts/train_model.py
```diff
@@ -81,6 +82,12 @@ def setup_args(parser=None):
...
+                       help="Tensorboard logging of metrics")
+    train.add_argument('-tbtag', '--tensorboard-tag', type=str, default=None,
+                       help='Specify all opt keys which you want to be presented in in TB name')
+    train.add_argument('-tbmetric', '--tensorboard-metric', type=str, default=None,
+                       help="Specify metrics which you want to track, it will be extracrted from report dict.")
```
alexholdenmiller: Can you move these to `@staticmethod TensorboardLogger.add_cmdline_args(parser)`? Then you can call that function here.

parlai/core/logs.py
```diff
@@ -0,0 +1,76 @@
...
+            os.makedirs(tbpath)
+        self.writer = SummaryWriter(log_dir='{}/{}'.format(tbpath, tensorboard_tag))
+        if opt['tensorboard_metrics'] is None:
+            self.tbmetrics = ['ppl', 'loss']
+        else:
+            self.tbmetrics = opt['tensorboard_metrics'].split(',')
+
+    def add_metrics(self, setting, step, report):
+        """
+        setting - ['train', 'valid']
```
alexholdenmiller: `, 'test']` I guess

parlai/scripts/train_model.py
alexholdenmiller: I hadn't seen jason's comment but mine was the same thing

parlai/scripts/train_model.py
uralik: yes, this is what @jaseweston wrote above, agreed

docs/Makefile
```diff
@@ -14,13 +14,18 @@ SPHINXPROJ    = ParlAI
...
 
 .PHONY: help Makefile
 
+clean:
+	rm -rfv "${BUILDDIR}"
+
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
+	cd "${SOURCEDIR}"; python generate_task_list.py
```
uralik: this is just to avoid running python by hand

docs/source/tutorial_worlds.rst
```diff
@@ -263,18 +263,18 @@ Multiprocessed Pytorch Dataloader
 For large datasets, where it is best to stream from disk during training
 rather than load initially into memory, we provide a teacher that utilizes pytorch data loading.
 
-(Note: the module `here <https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/pytorch_data_teacher.py>`_
+(Note: the module `here <https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/pytorch_data_teacher.py>`__
```
uralik: double underscore for anonymous references (gave warnings before)

PR #799, opened 2018-05-31
ParlAI Model Zoo !! 

jaseweston:
@

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -54,6 +54,15 @@ def modelzoo_path(datapath, path):
     if not path.startswith('models:'):
         return path
     else:
+        # Check if we need to download the model
+        animal = path[7:path.rfind('/')].replace('/', '.')
+        module_name = "parlai.zoo.%s" % (animal)  
+        my_module = importlib.import_module(module_name)
+        try:
+            download = getattr(my_module, 'download')
+            download(datapath)
+        except AttributeError:
+            pass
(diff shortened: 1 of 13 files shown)


parlai/core/params.py
```diff
@@ -54,6 +54,17 @@ def modelzoo_path(datapath, path):
     if not path.startswith('models:'):
         return path
     else:
+        # Check if we need to download the model
+        animal = path[7:path.rfind('/')].replace('/', '.')
+        module_name = "parlai.zoo.%s" % (animal)
```
alexholdenmiller: also this is cool, I saw this in pierre-emmanuel's diff, you can do this instead:
```
>>> w = 'world'
>>> f'hello {w}'
'hello world'
```

(e.g. `f'parlai.zoo.{animal}'` which I think is super clean and nice)

parlai/core/params.py
```diff
@@ -54,6 +54,17 @@ def modelzoo_path(datapath, path):
...
+        # Check if we need to download the model
+        animal = path[7:path.rfind('/')].replace('/', '.')
+        module_name = "parlai.zoo.%s" % (animal)  
+        try:
+            my_module = importlib.import_module(module_name)
+            download = getattr(my_module, 'download')
+            download(datapath)
+        except ModuleNotFoundError:
+            pass
+        except AttributeError:
```
alexholdenmiller: can do `except (ModuleNotFoundError, AttributeError)` if you want

PR #805, opened 2018-06-01
ParlAI text data format 

jaseweston:


parlai/core/build_data.py
```diff
@@ -48,7 +48,7 @@ def download(url, path, fname, redownload=False):
     will not download tar file again if it is present (default ``True``)."""
     outfile = os.path.join(path, fname)
     download = not os.path.isfile(outfile) or redownload
-
+    print("[ downloading: " + download + " to " + outfile + " ]")
```
klshuster: did you mean `fname` instead of `download` here?

jaseweston:
fixed print statement, some other small changes

PR #811, opened 2018-06-04
Opt fix

emilydinan:
Opt loading fix: fix for case when you don't specify a model (and instead load the model name from an opt file) but new arguments were added to model since the opt file was saved. In this case, we need to be able to add the command line args for the model at parsing time.

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -45,6 +45,7 @@ import pickle
 import random
 import os
 
+
 class Agent(object):
     """Base class for all other agents."""
 
@@ -275,6 +276,7 @@ def name_to_agent_class(name):
     class_name += 'Agent'
     return class_name
 
+
 def load_agent_module(opt):
     model_file = opt['model_file']
     optfile = model_file + '.opt'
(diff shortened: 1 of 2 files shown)


parlai/core/params.py
```diff
@@ -332,6 +333,16 @@ def add_extra_args(self, args=None):
...
+        else:
+            # try to get model name from model opt file
+            model_file = parsed.get('model_file', None)
+            if model_file is not None:
+                optfile = model_file + '.opt'
+                if os.path.isfile(optfile):
+                    with open(optfile, 'rb') as handle:
+                        new_opt = pickle.load(handle)
+                        if 'model' in new_opt:
+                            self.add_model_subargs(new_opt['model'])
```
alexholdenmiller: can you double-check this doesn't kill the convai2 eval_ppl? it might get angry since we try to load the model params into a subclass of the model

emilydinan: ah good call -- checked and it doesn't break! in this case, 'model' should already be set so it will not load it from the opt file

parlai/core/agents.py
```diff
@@ -343,8 +343,21 @@ def create_agent(opt, requireModelExists=False):
         from parlai.core.params import ParlaiParser
         parser = ParlaiParser(add_parlai_args=False)
         parser.add_parlai_data_path()
+        # add model args if they are missing
```
jaseweston: thnx! maybe we can move some code out into another function somehow? this is getting awful long!

PR #819, opened 2018-06-05
fix unit tests

alexholdenmiller:
- add __init__ file to zoo directory
- add back removed argument to tfidf_retriever


jaseweston:
yes that was my bad sorry, fixed

alexholdenmiller:
merging as this also includes the added __init__.py

PR #824, opened 2018-06-06
Zoo model list  + add classic drqa retriever to zoo

jaseweston:


parlai/zoo/model_list.py
```diff
@@ -0,0 +1,34 @@
...
+ python examples/interactive.lua --model-file "models:wikipedia_2016-12-21/tfidf_retriever/drqa_docs"
+"""
+
+model_list = [
+    {
+        "id": "wikipedia_2016-12-21_retriever",
+        "path": "models:wikipedia_2016-12-21/tfidf_retriever/drqa_docs",
+        "agent":  "tfidf_retriever",
+        "task": "wikipedia_2016-12-21",
+        "description": "retrieval over Wikipedia dump, used for DrQA on the open squad dataset"
```
klshuster: Maybe mention here that it's the dump from the original DrQA paper? Just to differentiate it from the one below

jaseweston: ok!

PR #827, opened 2018-06-06
slightly prettier display in interactive and beyond

jaseweston:


parlai/core/utils.py
```diff
@@ -563,10 +563,11 @@ def contains_offensive_language(self, text):
         return None
 
 
-def display_messages(msgs, prettify=False):
+def display_messages(msgs, prettify=False,ignore_fields=''):
     """Returns a string describing the set of messages provided"""
```
alexholdenmiller: can you update the docstring with what these args do?

parlai/core/utils.py
```diff
@@ -563,10 +598,14 @@ def contains_offensive_language(self, text):
         return None
 
 
-def display_messages(msgs, prettify=False):
-    """Returns a string describing the set of messages provided"""
+def display_messages(msgs, prettify=False,ignore_fields=''):
+    """Returns a string describing the set of messages provided
+    If prettify is true, candidates are displayed using prettytable.
```
klshuster: Note that this is currently only true for `text_candidates`

jaseweston: right

PR #832, opened 2018-06-08
Make persona_seq2seq work without gpu

emilydinan:
Load models to CPU and small bug fix


projects/personachat/persona_seq2seq.py
```diff
@@ -2119,7 +2120,7 @@ def valid(obs):
                 self.xs_persona.copy_(xs_persona, )
                 xs_persona = Variable(self.xs_persona)
             else:
-                xs_persona = parsed_persona
+                xs_persona = Variable(xs_persona)
```
alexholdenmiller: ouch

PR #835, opened 2018-06-08
Offensive language detector script

jaseweston:


parlai/scripts/detect_offensive_language.py
```diff
@@ -0,0 +1,83 @@
...
+`python parlai/scripts/detect_offensive_language.py -t "convai_chitchat" --display-examples True -dt train`
+"""
+from parlai.core.params import ParlaiParser
+from parlai.core.agents import create_agent
+from parlai.core.worlds import create_task
+from parlai.agents.repeat_label.repeat_label import RepeatLabelAgent
+from parlai.core.utils import Timer, OffensiveLanguageDetector
+
+import random
+import os
```
emilydinan: don't need to import this

parlai/scripts/detect_offensive_language.py
```diff
@@ -0,0 +1,83 @@
...
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+"""Basic example which iterates through the tasks specified and
+checks them for offensive language.
+
+For example:
+`python parlai/scripts/detect_offensive_language.py -t "convai_chitchat" --display-examples True -dt train`
+"""
+from parlai.core.params import ParlaiParser
+from parlai.core.agents import create_agent
```
emilydinan: don't need to import this

parlai/scripts/detect_offensive_language.py
```diff
@@ -0,0 +1,83 @@
...
+import random
+import os
+
+def setup_args(parser=None):
+    if parser is None:
+        parser = ParlaiParser(True, True)
+    # Get command line arguments
+    parser.add_argument('-ltim', '--log-every-n-secs', type=float, default=2)
+    parser.add_argument('-d', '--display-examples', type='bool', default=False)
+    parser.set_defaults(datatype='valid')
```
emilydinan: should we set default to train?

jaseweston: yes

parlai/scripts/detect_offensive_language.py
jaseweston: i updated so you can actually put a model through it as well

jaseweston:
updated

PR #837, opened 2018-06-08
Update TorchAgent with standard functions and fix small bug

vgonzalez88:
Fixes small bug where GPU wasn't specified and moves in the `save`, `load`, and `shutdown` functions with some very basic functionality.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -13,6 +13,7 @@ except Exception as e:
     raise ModuleNotFoundError('Need to install Pytorch: go to pytorch.org')
 
 from collections import deque, namedtuple
+import pickle
 import random
 
 Batch = namedtuple("Batch", [
@@ -77,8 +78,6 @@ class TorchAgent(Agent):
         self.history_tokens = opt['history_tokens']
         self.history_dialog = opt['history_dialog']
         self.history_replies = opt['history_replies']
-        self.model = None
-        self.optimizer = None
 
     def share(self):
         shared = super().share()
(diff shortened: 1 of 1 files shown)


parlai/core/torch_agent.py
```diff
@@ -271,3 +275,34 @@ def parse(txt, splitSentences):
                 self.history['labels'] = labels
 
         return self.history['dialog']
+
+        def save(self, path):
```
jaseweston: save should save the opt file as well, see e.g. seq2seq

parlai/core/torch_agent.py
```diff
@@ -75,6 +77,8 @@ def __init__(self, opt, shared=None):
         self.history_tokens = opt['history_tokens']
         self.history_dialog = opt['history_dialog']
         self.history_replies = opt['history_replies']
+        self.model = None
+        self.optimizer = None
```
alexholdenmiller: I don't think we should set these to None

parlai/core/torch_agent.py
```diff
@@ -271,3 +275,34 @@ def parse(txt, splitSentences):
...
+            """Save model parameters if model_file is set.
+
+            Override this method for more specific saving.
+            """
+            path = self.opt.get('model_file', None) if path is None else path
+
+            if path and hasattr(self, 'model'):
+                model = {}
+                model['model'] = self.model.state_dict()
+                model['optimizer'] = self.optimizer.state_dict()
```
alexholdenmiller: this would need a `hasattr(self, 'optimizer')`

parlai/core/torch_agent.py
alexholdenmiller: specifically, the opt should be saved in a separate file at modelfile.opt
let's also remove line 290 where we save it within the `model` dict

vgonzalez88: oh good point. do you think it's fine that `load` doesn't load in the opt file? I saw in seq2seq it does so, but didn't know if I should put that in the parent class

parlai/core/torch_agent.py
vgonzalez88: Should I leave them out and check for both 'model' and 'optimizer' in the save function? Or is it safe to assume that the child class will have a 'model' field?

vgonzalez88: Oh I already was checking for 'model'. I'll just remove these.

alexholdenmiller:
made a few changes myself :) looks good now!

vgonzalez88:
Thanks Alex!

Issue #839, opened 2018-06-10
the result of  baseline profilememorynet of task convai2 is so bad


zhaoguangxiang:
prediction: i love to do , but i am a nurse __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__ __START__
test:{'hits@10': 0.525, 'total': 7801, 'loss': 4.539, 'hits@1': 0.0495, 'f1': 0.0324, 'accuracy': 0, 'hits@5': 0.267, 'hits@100': 1.0, 'ppl': 93.59}

when i run the train.py under the profilememory folder

zhaoguangxiang:
None of the other parameters have changed, except that the batch is set to 64

emilydinan:
It's possible that you need to train for longer. It looks like in that [training script](https://github.com/facebookresearch/ParlAI/blob/master/projects/convai2/baselines/profilememory/train.py) it is set to validate every 3 seconds which is something it definitely should not do. I will put out a PR to fix this.

If you look at this script [here](https://github.com/facebookresearch/ParlAI/blob/master/projects/convai2/baselines/profilememory/interactive.py) you can interact with the profile memory model trained on convai2 that seems (at least qualitatively) better and was trained for much longer.

alexholdenmiller:
closing for now, feel free to reopen if you have more questions

PR #841, opened 2018-06-11
Fix training script for persona_seq2seq model

emilydinan:
Was set to validate every 3 seconds


emilydinan:
Related to issue #839

alexholdenmiller:
oops

jaseweston:
we never put this one in the leaderboard, so.. i guess we could? but it's
pretty similar numbers right? but good that the code works decently if it's
in that dir..

On Mon, Jun 11, 2018 at 10:33 AM, Alexander Miller <notifications@github.com

Issue #844, opened 2018-06-11
Seq2seq model loads glove vectors to parlai_home


fgvbrt:
Hello, I would like to suggest improvement for seq2seq model. 
In this model you download glove/fasttext vectors in parlai home directory:
https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/seq2seq/seq2seq.py#L252

But for example glove vectors after downloading and extracting occupy about 10 GB which is quite a lot. I thinks that it would be better to add option for downloading it custom directory (for example specified  with datapath option, I also saw that you use this option in function modelzoo_path, so this option could be a good choice).

alexholdenmiller:
Thanks you're right this should have been using the datapath, updating!

alexholdenmiller:
Fixed!

Issue #843, opened 2018-06-11
memory issues for convai2 seq2seq baseline


fgvbrt:
Could you please help me what are memory requirements for seq2seq baseline for convai2 competition:
https://github.com/facebookresearch/ParlAI/blob/master/projects/convai2/baselines/seq2seq/train.py
I got the following situation:
1) Batch size 64 and sgd optimizer crashes after first  evaluation
2) Batch size 32 and sgd optimizer crashes after few hours of work
3) Batch size 64 and adam optimizer crashes before first evaluation
4) Batch size 32 and adam in first few minutes.

I am using ParlAI commit aadd9886816b579e34bbeb3f2a85ffe4fb2eca5b
Pytorch 0.4.0,
Titan X 12gb
cuda-9.0, cudnn with cudnn v7.1

Is it normal behavior? It is very frustrating, that after few hours of training training everything could just crash. 

Example of crash below:
[crash.txt](https://github.com/facebookresearch/ParlAI/files/2091781/crash.txt)

Thanks in advance!

alexholdenmiller:
The batch sizes there work for a 16gb GPU, but I'm not sure exactly what's pushing it over the edge after a few hours of training, as every example would have have been seen by then. It's possible it's getting ahead of pytorch's garbage collector. I pushed the GPU memory to the limit in order to not truncate the input sizes at all.

I'll dig around and see if I can find any places where memory usage might spike, but otherwise I've managed the memory usage just by setting the truncate parameter (e.g. `-tr 40`) or lowering the batchsize (however note that this can cut the persona from the history on this dataset).

alexholdenmiller:
Memory during training should be reduced, and I do an dummy forward which should identify too-large batch/truncation settings before training starts and give a warning to reduce batch size / truncation size.

PR #845, opened 2018-06-11
utility improvements to seq2seq

alexholdenmiller:
- still testing the oom code to see if it helps reduce memory spikes, it's based on fairseq's code and was suggested by myle
- changed the vector caches to reference the parlai data path instead of parlai_home

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -246,16 +246,13 @@ class Seq2seqAgent(Agent):
                     raise ex
                 if opt['embedding_type'].startswith('glove'):
                     init = 'glove'
-                    embs = vocab.GloVe(
-                        name='840B',
-                        dim=300,
+                    embs = vocab.GloVe(name='840B', dim=300,
                         cache=modelzoo_path(self.opt.get('datapath'),
                                             'models:glove_vectors')
                     )
                 elif opt['embedding_type'].startswith('fasttext'):
                     init = 'fasttext'
-                    embs = vocab.FastText(
-                        language='en',
+                    embs = vocab.FastText(language='en',
                         cache=modelzoo_path(self.opt.get('datapath'),
                                             'models:fasttext_vectors')
                     )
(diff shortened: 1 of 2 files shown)


alexholdenmiller:
looks like the oom code is working! this clears out pytorch's GPU memory cache whenever it gets oom during the forward and backward pass during training (*not* when the network weights are being updated and *not* during validation) and just moves on to the next batch (logging the oom to the metrics and printing a warning)

alexholdenmiller:
data change: #844
oom: #843 and others

alexholdenmiller:
trained with batchsize 350 for a bit and was able to catch some spikes during training and continue

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -248,23 +249,15 @@ def __init__(self, opt, shared=None):
                     embs = vocab.GloVe(
                         name='840B',
                         dim=300,
-                        cache=os.path.join(
```
jaseweston: just checking: is this the same as 
https://github.com/facebookresearch/ParlAI/blob/master/parlai/zoo/glove_vectors/build.py ?


    opt  = { 'datapath': datapath }
    fnames = ['glove.840B.300d.zip']
    download_models(opt, fnames, 'glove_vectors', use_model_type=False,
                    path = "http://nlp.stanford.edu/data")

not clear it is..

alexholdenmiller: we should probably just remove parlai/zoo/glove_vectors right? torchtext has its own code for downloading its vectors.

jaseweston: it is being used by drqa, if you can make drqa work with the other then yes! fine..! be good to get drqa to work with fasttext as well, anyway..

alexholdenmiller: but yes it is the same, download_models uses `os.path.join(opt['datapath'], 'models', model_folder)` where model_folder here is the `glove_vectors` string in that call

jaseweston: i thought one might be a binary file and one a text file or something? i guess just check drqa still works please

jaseweston:
see comment

PR #853, opened 2018-06-13
[MTurk] Fixing thread safety for time lock file

JackUrb:
Before the lock file was acquired separately from the step that checked for existence. Using a context manager and some C-style syntax, this is now an atomic action, meaning it should be *actually* thread-safe rather than just weakly thread-safe.

### DIFF
--- a/parlai/mturk/core/mturk_manager.py
+++ b/parlai/mturk/core/mturk_manager.py
@@ -11,6 +11,7 @@ import pickle
 import threading
 import time
 import uuid
+import errno
 
 from botocore.exceptions import ClientError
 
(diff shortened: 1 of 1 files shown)


parlai/mturk/core/mturk_manager.py
```diff
@@ -53,6 +54,28 @@
...
+        self.fd = None
+
+    def __enter__(self):
+        while self.fd is None:
+            try:
+                self.fd = os.open(self.filename, self.flags)
+            except OSError as e:
+                if e.errno == errno.EEXIST:  # Failed as the file exists.
+                    pass
+            time.sleep(shared_utils.THREAD_SHORT_SLEEP)
```
klshuster: what does the sleep do here? just curious

JackUrb: Just here to prevent a tight loop if a thread is waiting on the lock.

PR #856, opened 2018-06-14
Logging includes %done and ETA via a util function

jaseweston:


parlai/core/utils.py
```diff
@@ -101,6 +101,8 @@ def load_cands(path, lines_have_ids = False, cands_are_replies = False):
     return cands
 
 
+
+
```
klshuster: nit - added a few blank lines here :)

PR #858, opened 2018-06-14
total -> exs

jaseweston:
With this change we get things like this:
python train_model.py -t squad -m drqa -bs 32 --dict-file /tmp/dictydict

[ time:10s total_exs:32 ] {'exs': 32, 'train_loss': 330.1}
[ time:21s total_exs:64 ] {'exs': 32, 'train_loss': 327.9}
[ time:32s total_exs:96 ] {'exs': 32, 'train_loss': 326.3}
[ time:36s total_exs:128 ] {'exs': 32, 'train_loss': 319.7}

which is much clearer

### DIFF
--- a/projects/personachat/kvmemnn/kvmemnn.py
+++ b/projects/personachat/kvmemnn/kvmemnn.py
@@ -387,7 +387,7 @@ class KvmemnnAgent(Agent):
             return round_sigfigs(f)
 
         metrics = self.metrics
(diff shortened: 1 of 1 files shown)


jaseweston:
ok, this may have some cases where someone uses 'total' somewhere .. prepare for having to fix some things somewhere ok?!!! but otherwise..let's go..

PR #859, opened 2018-06-14
Messenger docs

JackUrb:
Now featuring a chat with a personachat model! (the below photo specifically)
personachat_example


JackUrb:
screen shot 2018-06-14 at 2 50 26 pm
Looks like this currently. Perhaps I should add a caption?

jaseweston:
<3 <3 <3

PR #872, opened 2018-06-15
Squad-fulldoc teacher +  generic ParlAIformat

jaseweston:


parlai/core/teachers.py
```diff
@@ -1084,3 +1087,90 @@ def setup_data(self, path):
                     reward = 0
             if x:
                 yield [x, None, reward], start
+
+
+class ParlAIDialogTeacher(FixedDialogTeacher):
+    """This module provides access to data in the ParlAI Text Dialog format.
+
+
+    Subclasses ``DialogTeacher`` for functionality and provides an
```
klshuster: It actually subclasses `FixedDialogTeacher`

parlai/core/teachers.py
```diff
@@ -1084,3 +1087,90 @@ def setup_data(self, path):
...
+        return self.examples[episode_idx]
+
+    def _setup_data(self, path):
+        print("[loading parlAI text data:" + path + "]")
+        self.examples = []
+        with open(path) as read:
+            for line in read:
+                msg = str_to_msg(line.rstrip('\n'))
+                if msg:
+                    self.examples.append(msg)
```
klshuster: need to count `episode_done`

parlai/core/teachers.py
```diff
@@ -1084,3 +1087,90 @@ def setup_data(self, path):
...
+        self.id = opt.get('parlaidialogteacher_datafile', 'teacher')
+        self.reset()
+        
+    def num_examples(self):
+        return len(self.examples)
+
+    def num_episodes(self):
+        return self.num_examples()
+
+    def get(self, episode_idx, entry_idx=None):
```
klshuster: do streaming stuff

parlai/core/teachers.py
jaseweston: fixed

parlai/core/teachers.py
```diff
@@ -1084,3 +1087,90 @@ def setup_data(self, path):
...
+
+    Lines 3 and 6 contain a query, a label, a reward for getting the question
+    correct, and three label candidates.
+
+    Since both of these examples are part of the same episode, the information
+    provided in the first example is relevant to the query in the second
+    example and therefore the agent must remember the first example in order to
+    do well.
+
+    In general dialog in this format can be any speech, not just QA pairs:
```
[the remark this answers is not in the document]
jaseweston: fixed

parlai/core/teachers.py
jaseweston: will do soon..maybe in another PR?

parlai/core/teachers.py
```diff
@@ -1084,3 +1087,84 @@ def setup_data(self, path):
...
+        self.id = opt.get('parlaidialogteacher_datafile', 'teacher')
+        self.reset()
+        
+    def num_examples(self):
+        return len(self.examples)
+
+    def num_episodes(self):
+        return self.episodes
+
+    def get(self, episode_idx, entry_idx=None):
```
klshuster: In the case where there are multiple examples per episode (e.g. in your example above), it might be better to keep track of a list of episodes, rather than a list of examples. That is, have self.episodes be a list of episodes, where you can index into the episode with the entry_idx. In this current setup, during random training you may get examples in the middle of an episode, and then jump to a new episode.

parlai/tasks/fromfile/agents.py
```diff
@@ -53,62 +53,7 @@ def __init__(self, opt, shared=None):
         
 class ParlaiformatTeacher(FixedDialogTeacher):
```
klshuster: would this class now subclass `ParlAIDialogTeacher` rather than `FixedDialogTeacher`?

parlai/core/teachers.py
jaseweston: ah ok, i didnt understand the interface thanks

jaseweston:
ok done

PR #876, opened 2018-06-18
Update fairseq compatibility to v0.5.0.

### LINKED ISSUE #826: Fairseq model out of date

stephenroller:
Fixes #826. Still could use some unit tests.


stephenroller:
Some more TODOS:
- [ ] add PPL tracking
- [ ] clean up some comments

jaseweston:
is the plan to check in and iterate or wait till everything is complete? @alexholdenmiller  @stephenroller

stephenroller:
It's ready for review from my perspective. I'm using this branch for all my experiments.

stephenroller:
I've filed tasks for the handful of missing features (multi-GPU) and some other TorchAgent desires. I should write unit tests, but whether we want this to block this merge or be another issue is up to y'all.

jaseweston:
I think we should start moving ahead..

parlai/agents/fairseq/fairseq.py
```diff
@@ -4,346 +4,412 @@
...
-from .fairseq_py.fairseq import criterions
-from .fairseq_py.fairseq import dictionary
-from .fairseq_py.fairseq.sequence_generator import SequenceGenerator
-from .fairseq_py.fairseq import options
-
-from torch.autograd import Variable
-
-from collections import deque
+    raise RuntimeError(
+        "Please run \"pip install 'git+https://github.com/pytorch/"
```
klshuster: This command doesn't do anything for me; it says `Requirement already satisfied: fairseq from git+https://github.com/pytorch/fairseq.git@v0.5.0#egg=fairseq in ./ParlAI/parlai/agents/fairseq/fairseq_py` - are there additional steps i need to take to get this running?

stephenroller: You may have to `pip uninstall fairseq`, especially if you had done the `cd fairseq_py; python setup.py develop` previously.

klshuster: great that worked, thanks

klshuster:
I get the following error when i run `python examples/train_model.py -m fairseq -t babi:task1k:1 -a fconv_self_att -mf /tmp/test_fairseq`: 
image

parlai/agents/fairseq/fairseq.py
stephenroller: I should change it to `pip install --upgrade`

stephenroller:
@klshuster good catch! I thought I had worked around that but I must've messed up.

stephenroller:
Fixed this bug. Still doesn't work with `-a fconv_self_att` because of an upstream bug, but works with `-a fconv` and `-a transformer_wmt_en_de`.

stephenroller:
Also took care of Kurt's issue with upgrading fairseq in case old version was symlinked into the conda env.

klshuster:
Cool, this works for me now

parlai/agents/fairseq/fairseq.py
```diff
@@ -4,346 +4,434 @@
...
+
+            # Next generate freely to create our response
+            if self.args.skip_generation:
+                # skip the generation step
+                for i in valid_inds:
+                    batch_reply[i]["text"] = ""
+            else:
+                # actually do the generation
+                for i, response in zip(valid_inds, self._generate(samples)):
+                    batch_reply[i]["text"] = response
```
alexholdenmiller: this setup looks like a good opportunity for torch agent batch_act to be generic and call e.g. `self.train_step` or `self.valid_step` and hoave those be overridable by the children, I think--with the rest being boilerplate.

stephenroller: Agreed, this is one way we could improve the usability of torch_agent

stephenroller:
wow

thanks for the testing! Fixed a bug around interactive mode, will merge when unit tests are go.

stephenroller:
For posterity, here's an example command:

`rm -rf /tmp/test_fairseq*; python examples/train_model.py -m fairseq -t babi:task1k:1 -a transformer -mf /tmp/test_fairseq --dict-file /tmp/test_fairseq.dict --lr 0.1 --optimizer sgd --momentum 0.9`

`-a fconv` seems to have some upstream problems at the moment.

stephenroller:
The following also works:

`python examples/train_model.py -m fairseq -t iwslt14 -a fconv_iwslt_de_en --dict-lower True --dict-file /tmp/tmpdict -ltim 10 --lr 0.25  --dropout 0.2 --optimizer sgd  --momentum 0.9 --clip-norm 1 --momentum 0.99 -bs 16 -vtim 30  --validation-metric ppl`

PR #878, opened 2018-06-18
[Messenger] Fixing message sender splits

JackUrb:
Turns out message splitting was suffering from an off-by-one error that I had carried over into the `[*SPLIT*]` feature, so now that error has been repaired in both locations.


JackUrb:
No worries! I totally missed it too

PR #883, opened 2018-06-18
update arg for pytorch buildteacher

klshuster:
Can now do `-pyt` instead of having to write out `--pytorch-buildteacher` (docs are also updated)


klshuster:
can now specify task with `-pyt` or dataset with `-pytd`

example:

`py display_data.py -pyt babi:task1k:1`
`py display_data.py -pytd vqa_v1`

PR #893, opened 2018-06-20
Ability to pull parlai config details from a parlai_internal file

JackUrb:
Basic flow here is that if you have a `parlai_internal.mturk.config` file, you can specify some defaults to use in your opts and in your qualifications. This doesn't break any existing flow.

Additional fixes:
- Cleans up default parameters for qualification names
- Cleans up some underscores and makes soft_block functions safer.
- Makes it so that locale qualifications are only added in the mturk utils if they haven't already been provided beforehand

### DIFF
--- a/parlai/mturk/core/mturk_utils.py
+++ b/parlai/mturk/core/mturk_utils.py
@@ -276,7 +276,7 @@ def remove_worker_qualification(worker_id, qualification_id,
 def create_hit_type(hit_title, hit_description, hit_keywords, hit_reward,
                     assignment_duration_in_seconds, is_sandbox,
                     qualifications=None,
-                    auto_approve_delay=4*7*24*3600,  # default 4 weeks
+                    auto_approve_delay=7*24*3600,  # default 1 week
                     ):
     """Create a HIT type to be used to generate HITs of the requested params"""
     client = get_mturk_client(is_sandbox)
--- a/parlai/mturk/core/mturk_utils.py
+++ b/parlai/mturk/core/mturk_utils.py
@@ -283,16 +283,16 @@ def create_hit_type(hit_title, hit_description, hit_keywords, hit_reward,
 
     # If the user hasn't specified a location qualification, we assume to
     # restrict the HIT to some english-speaking countries.
-    local_requirements = []
+    locale_requirements = []
     has_locale_qual = False
     if qualifications is not None:
         for q in qualifications:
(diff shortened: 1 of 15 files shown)


parlai/mturk/core/mturk_utils.py
```diff
@@ -276,26 +276,34 @@ def remove_worker_qualification(worker_id, qualification_id,
 def create_hit_type(hit_title, hit_description, hit_keywords, hit_reward,
                     assignment_duration_in_seconds, is_sandbox,
                     qualifications=None,
-                    auto_approve_delay=4*7*24*3600, # default 4 weeks
+                    auto_approve_delay=4*7*24*3600,  # default 4 weeks
```
emilydinan: should we consider making this shorter since people often forget to set it?

JackUrb: Yeah probably, maybe I'll default 1 week instead? Will also update docs

klshuster: +1

parlai/mturk/core/mturk_utils.py
```diff
@@ -276,26 +276,34 @@ def remove_worker_qualification(worker_id, qualification_id,
...
+    local_requirements = []
+    has_locale_qual = False
     if qualifications is not None:
-        localRequirements += qualifications
+        for q in qualifications:
+            if q['QualificationTypeId'] == '00000000000000000071':
+                has_locale_qual
+        local_requirements += qualifications
+
+    if not has_locale_qual:
```
klshuster: ❤️

parlai/mturk/core/mturk_utils.py
```diff
@@ -276,26 +276,34 @@ def remove_worker_qualification(worker_id, qualification_id,
...
-            {'Country': 'CA'},
-            {'Country': 'GB'},
-            {'Country': 'AU'},
-            {'Country': 'NZ'}
-        ],
-        'RequiredToPreview': True
-    }]
+    # If the user hasn't specified a location qualification, we assume to
+    # restrict the HIT to some english-speaking countries.
+    local_requirements = []
```
klshuster: super nit - do we wanna call this `locale_requirements` instead of `local_requirements`?

JackUrb:
Right now I don't have an example `config` file would look like - it can implement either of the depicted methods here though.

PR #894, opened 2018-06-20
Updates to TorchAgent

stephenroller:
- Move some logic from example_seq2seq to TorchAgent so it can be more readily used (by fairseq) for example
- fix a minor bug that came from copying the original maintain_dialog_history code
- Rename the TorchAgent `--history-tokens` option to `--truncate` for consistency with seq2seq and others.

Note that the `--truncate` bit does *not* truncate the output space, so there's an inconsistency with the option, so we should debate this.

Requesting both Victor and Alex review.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -234,14 +234,11 @@ class TorchAgent(Agent):
         """
 
         def parse(txt, splitSentences):
-            if dict is not None:
-                if splitSentences:
-                    vec = [self.dict.txt2vec(t) for t in txt.split('\n')]
-                else:
-                    vec = self.dict.txt2vec(txt)
-                return vec
+            if splitSentences:
+                vec = [self.dict.txt2vec(t) for t in txt.split('\n')]
             else:
-                return [txt]
+                vec = self.dict.txt2vec(txt)
+            return vec
 
         allow_reply = True
 
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -18,12 +18,23 @@ import random
 import copy
 
 Batch = namedtuple("Batch", [
-    "text_vec",  # bsz x seqlen tensor containing the parsed text data
-    "text_lengths", # bsz x 1 tensor containing the lengths of the text in same order as text_vec; necessary for pack_padded_sequence
-    "label_vec", # bsz x seqlen tensor containing the parsed label (one per batch row)
-    "labels", # list of length bsz containing the selected label for each batch row (some datasets have multiple labels per input example)
-    "valid_indices",  # list of length bsz containing the original indices of each example in the batch. we use these to map predictions back to their proper row, since e.g. we may sort examples by their length or some examples may be invalid.
+    # bsz x seqlen tensor containing the parsed text data
+    "text_vec",
+    # bsz x 1 tensor containing the lengths of the text in same order as
+    # text_vec; necessary for pack_padded_sequence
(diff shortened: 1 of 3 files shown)


parlai/core/torch_agent.py
```diff
@@ -239,7 +249,7 @@ def parse(txt, splitSentences):
         allow_reply = True
 
         if 'dialog' not in self.history:
-            self.history['dialog'] = deque(maxlen=self.history_tokens)
+            self.history['dialog'] = deque(maxlen=self.truncate)
```
alexholdenmiller: `maxlen=self.truncate if self.truncate >= 0 else None`

parlai/core/torch_agent.py
```diff
@@ -311,3 +323,26 @@ def shutdown(self):
         if path is not None:
             self.save(path + '.shutdown_state')
         super().shutdown()
+
+    def reset(self):
+        """Reset observation and episode_done."""
+        self.observation = None
+        self.episode_done = True
```
alexholdenmiller: Actually I think we want this to look like the seq2seq reset:

https://github.com/facebookresearch/ParlAI/blob/2951872be31f842cea861bcd52c34ed17e247f23/parlai/agents/seq2seq/seq2seq.py#L400-L406

parlai/core/torch_agent.py
```diff
@@ -311,3 +323,26 @@ def shutdown(self):
...
+    def observe(self, observation):
+        observation = copy.deepcopy(observation)
+        if not self.episode_done:
+            # if the last example wasn't the end of an episode, then we need to
+            # recall what was said in that example
+            prev_dialogue = self.observation['text']
+            observation['text'] = prev_dialogue + '\n' + observation['text']
+        self.observation = observation
+        self.episode_done = observation['episode_done']
+        return observation
```
alexholdenmiller: we should copy the more complex version of observe in parlai/agents/seq2seq.py which uses maintain_dialog_history instead of this simpler `'\n'`-concat.
(and by extension, remove self.episode_done as a field, e.g. in reset as well)
probably needs a little cleanup though
https://github.com/facebookresearch/ParlAI/blob/2951872be31f842cea861bcd52c34ed17e247f23/parlai/agents/seq2seq/seq2seq.py#L455-L475

parlai/core/torch_agent.py
```diff
@@ -227,7 +237,7 @@ def maintain_dialog_history(self, observation, reply='',
         """
 
         def parse(txt, splitSentences):
-            if dict is not None:
```
alexholdenmiller: actually I think we can remove this check, currently we expect agents to call super().__init__ and in the init we set up self.dict

parlai/core/torch_agent.py
stephenroller: Alright, but it should it actually use the `maintain_dialogue_history` in this TorchAgent class, right?

stephenroller: Going to put off this for another pull request, since this is just literally copying code from one class to another.

parlai/core/torch_agent.py
stephenroller: See comment w.r.t. observe below.

stephenroller:
Denied requests moved into #901.

alexholdenmiller:
Thanks for the issue, please double-check running `python tests/test_torch_agent.py` before merging

stephenroller:
Rebased to check new unit tests post #895 and #899.

PR #895, opened 2018-06-20
Fix some bugs in image captioning framework

vgonzalez88:
Small PR to fix issues that came up during development of generative image captioning model.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -18,6 +18,7 @@ import random
 
 Batch = namedtuple("Batch", [
     "text_vec",  # bsz x seqlen tensor containing the parsed text data
+    "text_lengths", # bsz x 1 tensor containing the lengths of the text in same order as text_vec; necessary for pack_padded_sequence
     "label_vec", # bsz x seqlen tensor containing the parsed label (one per batch row)
     "labels", # list of length bsz containing the selected label for each batch row (some datasets have multiple labels per input example)
     "valid_indices",  # list of length bsz containing the original indices of each example in the batch. we use these to map predictions back to their proper row, since e.g. we may sort examples by their length or some examples may be invalid.
@@ -85,7 +86,7 @@ class TorchAgent(Agent):
         shared['dict'] = self.dict
         return shared
 
-    def vectorize(self, obs, addEndIdx=True):
+    def vectorize(self, obs, addStartIdx=True, addEndIdx=True):
         """
         Converts 'text' and 'label'/'eval_label' field to vectors.
 
@@ -109,6 +110,8 @@ class TorchAgent(Agent):
             new_labels = []
             for label in obs[label_type]:
                 vec_label = self.dict.txt2vec(label)
+                if addStartIdx:
+                    vec_label.insert(0, self.START_IDX)
                 if addEndIdx:
                     vec_label.append(self.END_IDX)
                 new_label = torch.LongTensor(vec_label)
@@ -132,12 +135,12 @@ class TorchAgent(Agent):
                           observation, determines if an observation is valid
         """
         if len(obs_batch) == 0:
-            return Batch()
+            return Batch(None, None, None, None, None)
 
         valid_obs = [(i, ex) for i, ex in enumerate(obs_batch) if is_valid(ex)]
 
         if len(valid_obs) == 0:
-            return Batch()
+            return Batch(None, None, None, None, None)
 
         valid_inds, exs = zip(*valid_obs)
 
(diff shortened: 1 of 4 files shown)


vgonzalez88:
I just repeated the changes from the COCO Caption Task in the Flickr Task as all the errors would be the same.

parlai/core/torch_agent.py
```diff
@@ -132,12 +135,12 @@ def map_valid(self, obs_batch, sort=True, is_valid=lambda obs: 'text_vec' in obs
                           observation, determines if an observation is valid
         """
         if len(obs_batch) == 0:
-            return Batch()
+            return Batch(None, None, None, None, None)
```
alexholdenmiller: is this necessary? I thought namedtuples fill in Nones automatically

parlai/core/torch_agent.py
```diff
@@ -168,6 +172,7 @@ def map_valid(self, obs_batch, sort=True, is_valid=lambda obs: 'text_vec' in obs
         # set up the target tensors
         ys = None
         labels = None
+        y_lens = None
```
alexholdenmiller: this isn't used, right?

parlai/tasks/coco_caption/agents.py
```diff
@@ -71,7 +71,7 @@ def _path(opt, version):
 
 class DefaultDataset(Dataset):
     """A Pytorch Dataset utilizing streaming"""
-    def __init__(self, opt, version='2014'):
+    def __init__(self, opt, shared=None, version='2014'):
```
alexholdenmiller: do you need this?

parlai/tasks/coco_caption/agents.py
```diff
@@ -258,13 +260,13 @@ def next_example(self):
...
-            if self.image_mode != 'none':
+            if self.image_mode != 'none' and 'image_id' in self.example:
                 # move the image we loaded in the background into the example
                 image = self.data_queue.get()
                 self.example['image'] = image
-            ready = (self.example, self.epochDone)
+            ready = (self.example, self.imageEpochDone)
         # get the next base example: super().next_example() calls self.get()
-        self.example, self.epochDone = super().next_example()
+        self.example, self.imageEpochDone = super().next_example()
```
alexholdenmiller: do you use this anywhere?

parlai/tasks/flickr30k/agents.py
```diff
@@ -35,7 +35,7 @@ def _path(opt):
 
 class FlickrDataset(Dataset):
     """A Pytorch Dataset utilizing streaming"""
-    def __init__(self, opt):
+    def __init__(self, opt, shared=None):
```
alexholdenmiller: do datasets need the shared param?

klshuster: I must've missed these - Datasets do not need the `shared` param

alexholdenmiller:
@klshuster @vgonzalez88

parlai/tasks/coco_caption/agents.py
vgonzalez88: Yes. We use `imageEpochDone` instead of `epochDone` as FixedDialogTeacher takes the return of this `next_example` function and sets `self.epochDone`. So before this change, the loader livelocked as `self.epochDone` was being set to True in this line, but the prior result returned by `ready` contained `self.epochDone = False`. Therefore, FixedDialogTeacher would set `self.epochDone` to `False` immediately after this function exited. The program would see that the epoch wasn't completed and call this function again, and `ready` would once again return `False`, causing livelock even though `super().next_example()` kept returning True. Simply changing this result to `_` would mean that the trigger for `epochDone` would never be sent through to the system.

parlai/core/torch_agent.py
alexholdenmiller: (I rescind this comment, I was wrong)

parlai/core/torch_agent.py
alexholdenmiller: confirm: I think this line is unnecessary

parlai/core/torch_agent.py
```diff
@@ -194,7 +199,7 @@ def map_valid(self, obs_batch, sort=True, is_valid=lambda obs: 'text_vec' in obs
                     padded_ys[i, :y.shape[0]] = y
             ys = padded_ys
 
-        return Batch(xs, ys, labels, valid_inds)
+        return Batch(xs, torch.LongTensor(x_lens), ys, labels, valid_inds)
```
alexholdenmiller: I'd prefer this to be put into a longtensor earlier in the function

parlai/tasks/coco_caption/agents.py
alexholdenmiller: ah you're right! rescinded

parlai/core/torch_agent.py
vgonzalez88: Yes, it isn't used except within the if statement, will remove.

parlai/core/torch_agent.py
vgonzalez88: Fixed

PR #898, opened 2018-06-20
testing installing torch on travis

alexholdenmiller:


alexholdenmiller:
actually torch installs pretty quickly, so going to keep it in. the failure is due to a bug being fixed in #899

PR #900, opened 2018-06-20
[MTurk] Fixing time resets

JackUrb:
Explicitly ensures freeing workers by freeing all the workers being removed from a refreshed worker_time file. Ensures repeated delete_qualification calls don't crash the system.

### DIFF
--- a/parlai/mturk/core/mturk_manager.py
+++ b/parlai/mturk/core/mturk_manager.py
@@ -153,12 +153,11 @@ class MTurkManager():
             if os.path.exists(file_path):
                 with open(file_path, 'rb+') as time_log_file:
                     existing_times = pickle.load(time_log_file)
-                    if time.time() - existing_times['last_reset'] > \
-                            24 * 60 * 60:  # Reset if more than a day old
-                        pass
-                    elif time.time() - existing_times['last_reset'] < \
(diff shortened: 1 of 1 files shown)


parlai/mturk/core/mturk_manager.py
```diff
@@ -151,16 +151,18 @@ def _reset_time_logs(self, force=False):
...
-                        if time.time() - existing_times['last_reset'] > \
-                                24 * 60 * 60:  # Reset if more than a day old
-                            pass
-                        elif time.time() - existing_times['last_reset'] < \
-                                RESET_TIME_LOG_TIMEOUT:
-                            # no need to reset, another thread did recently
-                            return
+                with open(file_path, 'rb+') as time_log_file:
+                    existing_times = pickle.load(time_log_file)
+                    if time.time() - existing_times['last_reset'] > \
```
klshuster: sorry, maybe i am confused - so this checks if the last reset was greater than a day, and then does nothing if so?

JackUrb: If the time has been more than a day since the last reset as listed by the file, then it passes out of this conditional and resets the file. 
If the file has been reset really recently, it returns and doesn't reset the logs.

This logic is indeed opaque, I'll update

PR #905, opened 2018-06-22
TorchAgent needs to actually load the model state it saves.

stephenroller:


parlai/core/torch_agent.py
```diff
@@ -322,6 +322,10 @@ def load(self, path):
         Override this method for more specific loading.
         """
         states = torch.load(path, map_location=lambda cpu, _: cpu)
+        if 'model' in states:
+            self.model.load_state_dict(states['model'])
+        if 'optimizer' in states:
+            self.optimizer.load_state_dict(states['optimizer'])
```
vgonzalez88: I think you need to also force some of the components to be `cuda` if we've enabled cuda already as seen [here](https://github.com/facebookresearch/ParlAI/blob/f64a694d1e10797a1b18c7a56633a6d90c4bf983/parlai/agents/seq2seq/seq2seq.py#L339).

alexholdenmiller:
this was intentionally set up this way due to a desire to load the option file first (in the states returned by `load`), set up the nn.Modules using parameters from those loaded options, and then fill them with with the loaded state dict.

however, now we load the options file automatically earlier in the pipeline, so we could move to this version as the standard in torch agent

parlai/core/torch_agent.py
alexholdenmiller: I think this may have been a hack for torch 0.3 having some missing functionality, let's try without it and see if it works :)

vgonzalez88: Okay! Wasn't sure so just wanted to point it out.

PR #907, opened 2018-06-22
SQuAD sentence teachers

emilydinan:
Index and default teachers with the labels as the sentences that contain the true answers.


parlai/tasks/squad/agents.py
```diff
@@ -260,7 +260,7 @@ def setup_data(self, path):
                     labels = []
                     for sentence in sentences:
                         for answer in answers:
-                            if answer in sentence:
+                            if answer in sentence and sentence not in labels:
```
stephenroller: This is an O(n) check, but maybe it's necessary. Can labels be a set instead? Ditto below

PR #911, opened 2018-06-25
Create Documentation_License

aszlam:


projects/personachat/Documentation_License
```diff
@@ -0,0 +1,384 @@
+Attribution 4.0 International
```
jaseweston: i think you want this in tasks/personachat for the dataset?
the projects directory is more the set of code to train the task

PR #918, opened 2018-06-26
[MTurk] ensuring more strict ordering of messages, improving connectivity

JackUrb:
- Decreases acknowledgement wait time in order to raise expectation of ordering. Too long of a wait time slows the system
- Skips the step of putting something back into the queue to send the message later. Retries in a timeout loop instead, thus preventing leaving the "blocked" state, ensuring nothing advances while the line is blocked by an outgoing message.
- Prevents receiving of messages that have actions while an outgoing message is still yet to be acknowledged (thus ensuring a more strict ordering). Before you could receive a message alongside a new "send_message" command before your message got acknowledged. When the message was eventually ack'd, the input field would just drop.

Tested using a total packet drop rate of *30%* while retaining almost no disruption of service. This is a much higher rate than we'd expect from an active user, as drop rates are closer to the poles of ~1% or ~100% (totally disconnected).

### DIFF
--- a/parlai/mturk/core/socket_manager.py
+++ b/parlai/mturk/core/socket_manager.py
@@ -157,7 +157,7 @@ class SocketManager():
 
     # Time to acknowledge different message types
     ACK_TIME = {Packet.TYPE_ALIVE: 2,
-                Packet.TYPE_MESSAGE: 2}
+                Packet.TYPE_MESSAGE: 0.5}
 
     # Default time before socket deemed dead
     DEF_SOCKET_TIMEOUT = 30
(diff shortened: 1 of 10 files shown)


parlai/mturk/core/socket_manager.py
```diff
@@ -157,7 +157,7 @@ class SocketManager():
 
     # Time to acknowledge different message types
     ACK_TIME = {Packet.TYPE_ALIVE: 2,
-                Packet.TYPE_MESSAGE: 2}
+                Packet.TYPE_MESSAGE: 0.5}
```
klshuster: I'm assuming changing this to 0.5 vs. the 0.3 in the file above is intentional?

JackUrb: Yup - the commands sent by the server have to have a slightly longer delay to be safe, but the javascript doesn't have any command/control from it's end so it can be more aggressive.

PR #921, opened 2018-06-27
refactor messenger_manager to clean up redundancies

scoutsaachi:
Adding hooks to MessengerManager to allow it to be extended without duplicating the entire file

### DIFF
--- a/parlai/messenger/core/messenger_manager.py
+++ b/parlai/messenger/core/messenger_manager.py
@@ -264,9 +264,9 @@ class MessengerManager():
         self.agent_id_to_overworld_thread[agent_id] = overworld_thread
         pass
 
-    # override if extra bookkeeping must be done when removing agent
     def after_agent_removed(self, agent_id):
-        """ Perform any changes to metadata on agent removal"""
+        """ Perform any changes to metadata on agent removal
+        override if extra bookkeeping must be done when removing agent"""
         pass
 
     def _on_new_message(self, message):
--- a/parlai/messenger/core/messenger_manager.py
+++ b/parlai/messenger/core/messenger_manager.py
@@ -80,6 +80,7 @@ class MessengerManager():
         self.running = True
         self.conversation_index = 0
         self.shutting_down = True
+        self.need_server_setup = True  # is server/setup socket needed
 
         # Messaging interaction functions that determine what to do when
         # messages are confirmed as delivered, marked as read by a user, and
@@ -109,7 +110,8 @@ class MessengerManager():
 
     def remove_agent_from_pool(self, agent, world_type='default',
                                mark_removed=True):
-        """Remove agent from the pool"""
+        """Remove agent from the pool. Returns whether the agent needed to be
+           removed"""
         with self.agent_pool_change_condition:
             shared_utils.print_and_log(
                 logging.DEBUG,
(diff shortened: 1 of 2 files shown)


parlai/messenger/core/messenger_manager.py
```diff
@@ -422,23 +429,27 @@ def setup_server(self):
...
-        """Set up socket to start communicating to workers"""
-        shared_utils.print_and_log(logging.INFO,
-                                   'Local: Setting up WebSocket...',
-                                   should_print=True)
-        self.app_token = None
-        if self.opt.get('force_page_token'):
-            pass
-        else:
+    # override if permission needed externally
+    def get_app_token(self):
```
JackUrb: Nit: function description styled with `"""comment"""` as in other functions

parlai/messenger/core/messenger_manager.py
```diff
@@ -258,14 +264,19 @@ def _overworld_function(opt, agent_id, task_id):
         self.agent_id_to_overworld_thread[agent_id] = overworld_thread
         pass
 
+    # override if extra bookkeeping must be done when removing agent
+    def after_agent_removed(self, agent_id):
+        ''' Perform any changes to metadata on agent removal'''
```
emilydinan: use double quotation marks here instead of single quotation marks

parlai/messenger/core/messenger_manager.py
```diff
@@ -618,10 +638,9 @@ def observe_message(self, receiver_id, text, quick_replies=None):
         return self.message_sender.send_fb_message(receiver_id, text, True,
                                                    quick_replies=quick_replies)
 
-    def observe_payload(self, receiver_id, data, quick_replies=None):
+    def observe_payload(self, receiver_id, data):
```
emilydinan: why did we remove quick replies here?

parlai/messenger/core/messenger_manager.py
```diff
@@ -422,23 +429,27 @@ def setup_server(self):
...
+        """Set up socket to start communicating to workers"""
+        if self.need_server_setup:
+            shared_utils.print_and_log(logging.INFO,
+                                       'Local: Setting up WebSocket...',
+                                       should_print=True)
+
+        self.app_token = self.get_app_token()
         # cache the app token
         if self.app_token is None:
             self.app_token = input(
```
JackUrb: This logic can also go into the `self.get_app_token()` call, as it is still part of the flow of getting an app token.

parlai/messenger/core/messenger_manager.py
```diff
@@ -618,10 +638,9 @@ def observe_message(self, receiver_id, text, quick_replies=None):
         return self.message_sender.send_fb_message(receiver_id, text, True,
                                                    quick_replies=quick_replies)
 
-    def observe_payload(self, receiver_id, data, quick_replies=None):
```
JackUrb: This and the below are occurring because the internal parlai hasn't been synced since that change. Keep what it was before.

parlai/messenger/core/messenger_manager.py
```diff
@@ -80,6 +80,7 @@ def __init__(self, opt):
         self.running = True
         self.conversation_index = 0
         self.shutting_down = True
+        self.need_server_setup = True  # is server/setup socket needed
```
emilydinan: perhaps we should set this variable with opt (and have it default to True)

scoutsaachi:
Added an argument for bypassing setting up server and fixed nits.

parlai/messenger/core/messenger_manager.py
JackUrb: Can also remove the comment above, or move it into the `"""` comment

JackUrb:
This is looking good from my end!

PR #925, opened 2018-06-27
move twitter dict and model to model zoo

alexholdenmiller:


parlai/scripts/eval_ppl.py
```diff
@@ -180,7 +181,19 @@ def eval_ppl(opt, build_dict):
...
+    if not build_dict and not dict_file:
+        raise RuntimeError('eval_ppl script either needs a dictionary build '
+                           'function or a dictionary file.')
+
+    if build_dict:
+        dict_agent = build_dict()
+    else:
+        dict_opt = copy.deepcopy(opt)
+        dict_opt['model'] = dict_opt.get('dictionary_class', 'parlai.core.dict:DictionaryAgent')
+        dict_opt['model_file'] = dict_file
```
jaseweston: if dict_file is set?  otherwise shouldnt it just load from opt?

alexholdenmiller: as it is now, either build_dict or dict_file must be set (assuming we don't trust the model's dict_file)

jaseweston: ok

parlai/scripts/eval_ppl.py
```diff
@@ -180,7 +181,19 @@ def eval_ppl(opt, build_dict):
         but rather should have hardcoded settings for its dictionary.
```
jaseweston: i dont understand the comment for "build_dict"

and...where is the comment  for dict_file ?

PR #929, opened 2018-06-28
fairseq logging rounding

alexholdenmiller:
added `round_sigfigs` to fairseq logs


stephenroller:
Not that it matters much, but I wonder if it's better to do the rounding on display side; these rounded values could, for example, affect the impatience during early stopping.

alexholdenmiller:
yeah that's fair, then we wouldn't have to have this code in every function--I'm not *super* concerned about early stopping being affected by four sigfigs, but wouldn't mind not having to include this in each model

PR #934, opened 2018-06-28
Moving to new metric for determining disconnects

JackUrb:
Migrates from a time-dependent timeout model (which isn't very strong in a 3 point system where the middle one acts as a router) to one that relies more strongly on heartbeats and ping/pong semantics. Changes are as follows:
- The ParlAI Python thread now keeps track of it's own heartbeat time rather than just responding to heartbeats with heartbeats.
- The router both forwards heartbeats and responds to them with a pong message (which ensures connectivity with the router).
- Both the javascript and python clients now keep track of the number of pongs they've received from the router since they've received a heartbeat from one another. This is a proxy way of telling when the router is up but the client on the other side of the router is dead.

Why do this at all?
- Increases the number of handled failure states
- Makes it easier to detect which node in the system is having issues
- Opens the door for better online indicators on the Turker's side and better feedback as to what happened in a crash.
- Opens the door for proper handling of a disconnected task once ParlAI-MTurk switches to a base-pay + bonus system.
- Potentially allows *restarting a HIT* when there's a segfault on the python client once HIT state is stored in a DB rather than in python mem.

### DIFF
--- a/parlai/mturk/core/socket_manager.py
+++ b/parlai/mturk/core/socket_manager.py
@@ -32,6 +32,7 @@ class Packet():
     TYPE_ALIVE = 'alive'
     TYPE_MESSAGE = 'message'
     TYPE_HEARTBEAT = 'heartbeat'
+    TYPE_PONG = 'pong'
 
     def __init__(self, id, type, sender_id, receiver_id, assignment_id, data,
                  conversation_id=None, requires_ack=True, blocking=True,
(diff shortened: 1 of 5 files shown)


parlai/mturk/core/socket_manager.py
```diff
@@ -197,7 +199,9 @@ def __init__(self, server_url, port, alive_callback, message_callback,
         self.queues = {}
         self.threads = {}
         self.run = {}
-        self.last_heartbeat = {}
+        self.last_sent_heartbeat = {}  # time of last heartbeat sent
+        self.last_gotten_heartbeat = {}  # actual last received heartbeat
```
klshuster: nit: maybe `last_received_heartbeat`? something about `last_gotten` sounds awk to me (and in the comment you write received)

PR #936, opened 2018-06-29
packed assigned to False in except branch

uralik:
in line 197 it could happen that packed does not exist otherwise


alexholdenmiller:
merging

PR #938, opened 2018-06-29
Move display script, add max length arg

emilydinan:
Moved most logic from `examples/display_data.py` over to `scripts/display_data.py` and added argument for the maximum display length for the text field (before, you couldn't change it from 1000, which isn't sufficient to view the data for tasks like SQuAD where the text field is quite long)

### DIFF
--- a/parlai/core/utils.py
+++ b/parlai/core/utils.py
@@ -626,7 +626,7 @@ class OffensiveLanguageDetector(object):
         return None
 
 
-def display_messages(msgs, prettify=False, ignore_fields=''):
+def display_messages(msgs, prettify=False, ignore_fields='', max_len=1000):
     """Returns a string describing the set of messages provided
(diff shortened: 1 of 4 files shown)


parlai/core/utils.py
```diff
@@ -650,8 +650,8 @@ def display_messages(msgs, prettify=False, ignore_fields=''):
             lines.append(msg['image'])
         if msg.get('text', ''):
             text = msg['text']
-            if len(text) > 1000:
-                text = text[:1000] + '...'
+            if len(text) > max_len:
+                text = text[:max_len] + '...'
```
jaseweston: in my experience it would actually be nicer to see a bit of the beginning and a bit of the end..

emilydinan: good point i'll put out a diff to add that functionality

PR #939, opened 2018-07-01
logs naming fix and train_model script arg help fix

uralik:
tensorboard folder now is put in the same folder as the model file.

help string in the train_model script had wrong default value.


parlai/scripts/train_model.py
```diff
@@ -57,7 +57,7 @@ def setup_args(parser=None):
     train.add_argument('-sval', '--save-after-valid', type='bool',
                        default=False,
                        help='Saves the model to model_file.checkpoint after '
-                            'every validation (default True).')
+                            'every validation (default False).')
```
stephenroller: You can also write `%(default)s` instead and argparse will insert the default.

uralik: I' ll change that, thanks!

PR #943, opened 2018-07-03
BPE support.

stephenroller:
This could be more memory efficient by passing a specialized format (pseudofile of word <tab> count) into the learn_bpe. The API calls would be more complex. For now, submitting functional version that works fine.

If we need to run this on something with an extremely large training data set, we could run into problems.

We could make it *much* more memory efficient and cleaner API by reimplementing or forking the subword-nmt repo, but I think we probably want to stick to the official version.

### DIFF
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -15,6 +15,13 @@ import numpy as np
 import os
 import re
 
+try:
+    from subword_nmt import learn_bpe, apply_bpe
+    # Don't explicitly throw the runtime error unless the user needs it
+    BPE_INSTALLED = True
+except ImportError:
+    BPE_INSTALLED = False
+
 RETOK = re.compile(r'\w+|[^\w\s]|\n', re.UNICODE)
 
 def escape(s):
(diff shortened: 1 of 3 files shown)


parlai/core/dict.py
```diff
@@ -129,6 +136,12 @@ def add_cmdline_args(argparser):
             dictionary.add_argument(
                 '--dict-lower', default=DictionaryAgent.default_lower, type='bool',
                 help='Whether or not to lowercase all text seen.')
+            dictionary.add_argument(
```
jaseweston: how do you turn off bpe?

stephenroller: BPE is only enabled if you pass `--dict-tokenizer bpe` as an argument.

parlai/core/dict.py
```diff
@@ -365,6 +387,28 @@ def tokenize(self, text, building=False):
...
+        Finalize and freeze the dictionary.
+
+        If using BPE tokenization, performs the codec learning.
+        """
+
+        if self.tokenizer != 'bpe':
+            # only BPE needs the second pass
+            return
+
+        # Now that we have a second pass,
```
stephenroller: should update this comment.

parlai/core/dict.py
```diff
@@ -365,6 +387,31 @@ def tokenize(self, text, building=False):
...
+        Finalize and freeze the dictionary.
+
+        If using BPE tokenization, performs the codec learning.
+        """
+
+        if self.tokenizer != 'bpe':
+            # only BPE needs the second pass
+            return
+
+        # Let the BPE model learn its codecs, then use the encodings to to
```
klshuster: nit: extra "to"

stephenroller:
Fixed the comment bug. Will merge after CI fix.

PR #935, opened 2018-06-29
speed up s2s ~2x

alexholdenmiller:
seq2seq was returning its predictions as texts to the teacher to get accuracy and f1 metrics, this was hurting it a surprising amount.

this diff removes that, losing access to accuracy and f1 during training logs, but seq2seq now returns token-accuracy as a useful proxy metric instead.

---
babi:task10k:1 experiments

seq2seq command:
`py examples/train_model.py -t babi:task10k:1 --dict-file /tmp/dict_babi:task10k:1 -bs 128 -eps 20 -dt train:ordered -m seq2seq -hs 128 -esz 128 -nl 2`

fairseq command:
`py examples/train_model.py -t babi:task10k:1 --dict-file /tmp/dict_babi:task10k:1 -bs 128 -eps 20 -dt train:ordered -m fairseq --arch lstm --encoder_embed_dim 128 --encoder-hidden-size 128 --encoder-layers 2 --decoder-embed-dim 128 --decoder-hidden-size 128 --decoder-layers 2 --decoder-out-embed-dim 128 --decoder-attention 0
`

seq2seq before: 77s, 78s
seq2seq after: 40s, 40s
fairseq2seq: 62s, 66s

---
convai2 experiments

seq2seq command:
`py examples/train_model.py -t convai2 --dict-file /tmp/dict_convai2 -bs 128 -m seq2seq -eps 1 -dt train:ordered`

fairseq command:
`py examples/train_model.py -t convai2 --dict-file /tmp/dict_convai2 -bs 128 -eps 1 -dt train:ordered -m fairseq --arch lstm --encoder_embed_dim 128 --encoder-hidden-size 128 --encoder-layers 2 --decoder-embed-dim 128 --decoder-hidden-size 128 --decoder-layers 2 --decoder-out-embed-dim 128 --decoder-attention 0`

seq2seq before: 182s, 174s
seq2seq after: 84s, 82s
fairseq: 125s, 125s

PR #944, opened 2018-07-03
Display model should use be eval mode.

stephenroller:
Important since #935, display_model essentially works in training mode and won't generate responses.

### DIFF
--- a/examples/display_model.py
+++ b/examples/display_model.py
@@ -24,6 +24,8 @@ def main():
     # Get command line arguments
     parser = ParlaiParser(True, True)
     parser.add_argument('-n', '--num-examples', default=10)
+    # by default we want to display info about the validation set
+    parser.set_defaults(datatype='valid')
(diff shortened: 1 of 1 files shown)


examples/display_model.py
```diff
@@ -25,9 +25,10 @@ def main():
     parser = ParlaiParser(True, True)
     parser.add_argument('-n', '--num-examples', default=10)
     opt = parser.parse_args()
+    opt['datatype'] = 'valid'
 
     # Create model and assign it to the specified task
-    agent = create_agent(opt)
+    agent = create_agent(opt, requireModelExists=True)
```
jaseweston: some models don't use a model file, e.g. ir_baseline -- will it still work?

stephenroller: Oh that's a great point, we can drop this part of the patch.

stephenroller:
Fixed, thanks!

PR #951, opened 2018-07-05
Beam search

uralik:
There is Beam class now which keeps all the information about beam and can also dump nice graphs for you if you want to analyze different paths. It also supports extraction of nbest list of hypothesis and is doing length penalty as it was suggested by Google NMT paper.

This beam is intended to use with mini-batches, so its quite fast. The class structure inspired by OpenNMT project, though it is not that powerful as there.

It is model/agent agnostic , already works good for my internal agent. This is dump graph example:

In the following PR I will put changes to make it work in seq2seq agent.

### DIFF
--- a/parlai/core/beam.py
+++ b/parlai/core/beam.py
@@ -1,22 +1,36 @@
 import torch
+import math
 
 class Beam(object):
-    """
-    Beam class serves as a structure to keep beam of hypothesis with corresponding scores, backtracking info etc.
-    """
-    def __init__(self, beam_size, min_length, padding_token, bos_token, eos_token, cuda):
+    def __init__(self, beam_size, min_length=3, padding_token=0, bos_token=1, eos_token=2, min_n_best=3, cuda=False):
+        """
+        Generic beam class. It keeps information about beam_size hypothesis.
+        :param beam_size: number of hypothesis in the beam
+        :param min_length: minimum length of the predicted sequence
+        :param padding_token: Set to 0 as usual in ParlAI
+        :param bos_token: Set to 1 as usual in ParlAI
+        :param eos_token: Set to 2 as usual in ParlAI
+        :param min_n_best: Beam will not be done unless this amount of finished hypothesis (with EOS) is done
+        :param cuda: What device to use for computations
+        """
         self.beam_size = beam_size
         self.min_length = min_length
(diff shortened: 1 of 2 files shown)


stephenroller:
Super cool dialog tree

parlai/core/beam.py
```diff
@@ -0,0 +1,184 @@
...
+        self.min_n_best = min_n_best
+
+    def get_output_from_current_step(self):
+        return self.outputs[-1]
+
+    def get_backtrack_from_current_step(self):
+        return self.bookkeep[-1]
+
+    def advance(self, softmax_probs):
+        voc_size = softmax_probs.size(-1)
```
alexholdenmiller: can you add a comment on the dimensions of softmax probs?

parlai/core/beam.py
```diff
@@ -0,0 +1,184 @@
...
+            if self.eos_top_ts is None:
+                self.eos_top_ts = len(self.outputs) - 1
+
+    def done(self):
+        return self.eos_top and self.n_best_counter >= self.min_n_best
+
+    def get_top_hyp(self):
+        """
+        Helper function to get single best hypothesis
+        :return: hypothesis sequence represented as t{}-w{}-sc{:.{prec}f}
```
alexholdenmiller: what does `t{}-w{}-sc{:.{prec}f}` mean?

parlai/core/beam.py
```diff
@@ -0,0 +1,184 @@
...
+        tok_id = best_idxs % voc_size
+
+        self.outputs.append(tok_id)
+        self.bookkeep.append(hyp_id)
+
+        #  check new hypos for eos label, if we have some, add to finished
+        for i in range(self.beam_size):
+            if self.outputs[-1][i] == self.eos:
+                #  this is finished hypo, adding to finished
+                self.finished.append((self.scores[i], len(self.outputs)-1, i))
```
alexholdenmiller: could you actually switch these to a `namedtuple` (`Hypothesis`?)

parlai/core/beam.py
uralik: done

parlai/core/beam.py
uralik: nvm, it does not exist anymore

PR #972, opened 2018-07-11
Fairseq enable sampling. DictionaryAgent slightly DRYer.

stephenroller:
Thoughts from everyone here would be welcome.

### DIFF
--- a/parlai/agents/fairseq/fairseq.py
+++ b/parlai/agents/fairseq/fairseq.py
@@ -131,14 +131,12 @@ class _FairseqDictionary(DictionaryAgent):
         # this to be the best way.
 
         # add the token to the dictionary
-        self.add_token(self._LUA)
-        self.freq[self._LUA] = self.freq[self.null_token] + 1
-        # Sort the list. Note that sort() calls remove_tail, so we need to
-        # ensure we don't throw anything away
-        tmp = self.minfreq
-        self.minfreq = -1
-        self.sort()
-        self.minfreq = tmp
+        self.add_token(_FairseqDictionary._LUA)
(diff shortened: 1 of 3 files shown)


parlai/core/dict.py
```diff
@@ -261,6 +253,12 @@ def __init__(self, opt, shared=None):
             if opt.get('dict_file'):
                 self.save_path = opt['dict_file']
 
+    def add_token(self, word):
+        if word not in self.tok2ind:
+            index = len(self.tok2ind)
+            self.tok2ind[word] = index
+            self.ind2tok[index] = word
+
```
alexholdenmiller: great!

parlai/agents/fairseq/fairseq.py
```diff
@@ -107,7 +107,38 @@ def _fairseq_opt_wrapper(opt, skip_pretrained_embedding_loading=False):
...
+
+        # add the token to the dictionary
+        self.add_token(self._LUA)
+        self.freq[self._LUA] = self.freq[self.null_token] + 1
+        # Sort the list. Note that sort() calls remove_tail, so we need to
+        # ensure we don't throw anything away
+        tmp = self.minfreq
+        self.minfreq = -1
+        self.sort()
+        self.minfreq = tmp
```
alexholdenmiller: maybe better: let's add a kwarg to sort which says `trim=True`, which decides whether to call self.remove_tail as well as self.resize_to_max (which trims to the max number of total tokens). Then we can just do `self.sort(trim=False)` here.

stephenroller:
Updated

PR #974, opened 2018-07-11
Refactor TorchAgent with more functionality and boilerplate

alexholdenmiller:
in this diff:
- switch map_valid to batchify (it returns a Batch)
- switch unmap_valid to match_batch (it takes in the list of message dicts now)
- add default batch_act to torch agent (less boilerplate in child)
- move train and eval functionality to `train_step` and `eval_step`, called by batch_act, which each take a batch as input
- update vectorize function a bit
- update child classes (fairseq and example_seq2seq) to use the new setup

still to come:
- use maintain_dialog_history in observe
- update maintain_dialog_history
- add candidates to vectorize and batchify
- move seq2seq to subclass of torchagent
- move languagemodel to subclass of torchagent
- move memnn to subclass of torchagent
- move starspace to subclass of torchagent


parlai/agents/example_seq2seq/example_seq2seq.py
```diff
@@ -150,107 +145,82 @@ def share(self):
...
-
-        Update the model using the targets if available.
-        """
+    def v2t(self, vector):
+        """Converts vector to text."""
+        if vector.dim() == 1:
+            output_tokens = []
+            # Remove the final END_TOKEN that is appended to predictions
+            for i, token in enumerate(vector):
+                if token == self.END_IDX:
```
stephenroller: Do we also need to check for START_IDX?

parlai/agents/example_seq2seq/example_seq2seq.py
```diff
@@ -150,107 +145,82 @@ def share(self):
...
+                    break
+                else:
+                    output_tokens.append(token)
+            return [self.dict.vec2txt(output_tokens)]
```
stephenroller: It feels weird that (1d input) -> list[str] and (2d input) -> list[str]. I feel like you can always see it as a reduction of the last axis.

Illustration of the usage of [0] is where it particularly feels strange.

parlai/agents/example_seq2seq/example_seq2seq.py
```diff
@@ -150,107 +145,82 @@ def share(self):
 
         return shared
 
-    def predict(self, xs, ys=None, is_training=False):
-        """Produce a prediction from our model.
-
-        Update the model using the targets if available.
-        """
+    def v2t(self, vector):
```
stephenroller: We should document params, since this method has a complicated type signature. Or hide it with `_v2t`?

parlai/agents/example_seq2seq/example_seq2seq.py
```diff
@@ -150,107 +145,82 @@ def share(self):
...
+        """Converts vector to text."""
+        if vector.dim() == 1:
+            output_tokens = []
+            # Remove the final END_TOKEN that is appended to predictions
+            for i, token in enumerate(vector):
```
stephenroller: `_` instead of `i`?

parlai/agents/example_seq2seq/example_seq2seq.py
```diff
@@ -150,107 +145,82 @@ def share(self):
...
+
+        scores = decoder_output.view(-1, decoder_output.size(-1))
+        loss = self.criterion(scores, ys.view(-1))
+        loss.backward()
+        self.update_params()
+
+        _max_score, predictions = decoder_output.max(2)
+        return self.v2t(predictions)
+
+    def eval_step(self, xs, *args, **kwargs):
```
stephenroller: The eval also needs `ys` in order to evaluate PPL? Maybe this method should be renamed `generate` and then `eval_step` can call `generate`.

parlai/core/torch_agent.py
```diff
@@ -214,20 +209,35 @@ def map_valid(self, obs_batch, sort=True, is_valid=lambda obs: 'text_vec' in obs
                     padded_ys[i, :y.shape[0]] = y
             ys = padded_ys
 
-        return Batch(xs, x_lens, ys, labels, valid_inds)
+        cands = None
+        # TODO: return candidates
+        return Batch(xs, ys, labels, valid_inds, cands)
+
+    def match_batch(self, batch_reply, valid_inds, predictions=None,
```
stephenroller: nice.

parlai/core/torch_agent.py
```diff
@@ -121,21 +130,21 @@ def vectorize(self, obs, addStartIdx=True, addEndIdx=True):
...
-                if addStartIdx:
-                    vec_label.insert(0, self.START_IDX)
-                if addEndIdx:
-                    vec_label.append(self.END_IDX)
-                new_label = torch.LongTensor(vec_label)
-                if self.use_cuda:
-                    new_label = new_label.cuda()
-                new_labels.append(new_label)
-            obs[label_type + "_vec"] = new_labels
+            label = random.choice(obs[label_type])
```
stephenroller: we might want to keep an RNG around so seeds can ensure this is deterministic.

parlai/agents/fairseq/fairseq.py
```diff
@@ -401,45 +401,29 @@ def reset(self):
...
-        # torchagent boilerplate
-        self.is_training = any(["labels" in obs for obs in observations])
-        vec_obs = [self.vectorize(obs) for obs in observations]
-        xs, _, ys, _, valid_inds = self.map_valid(vec_obs)
-        if xs is None:
-            return batch_reply
-
-        # here begins fairseq specific stuff
-        samples = self._make_sample(xs, ys)
+    def batchify(self, *args, **kwargs):
```
stephenroller: So I don't know of a reason that `fairseq` needs `sort=True`. That was the default behavior from @vgonzalez88 because something else expected the batch sorted, but I can't remember the details.

parlai/core/torch_agent.py
```diff
@@ -356,11 +366,48 @@ def observe(self, observation):
...
+            predictions = output
+            candidate_preds = None
+
+        self.match_batch(batch_reply, batch.valid_indices,
+                         predictions=predictions,
+                         candidate_preds=candidate_preds)
+
+        return batch_reply
+
+    def train_step(self, xs, ys=None, cands=None, *args, **kwargs):
```
stephenroller: We should document this function well, as the user is going to have to implement it.

parlai/core/torch_agent.py
```diff
@@ -356,11 +366,48 @@ def observe(self, observation):
...
+        raise NotImplementedError('Abstract class: user must implement train_step')
+
+    def eval_step(self, xs, ys=None, cands=None, *args, **kwargs):
```
stephenroller: We should document this function well, as the user is going to have to implement it.

parlai/core/torch_agent.py
```diff
@@ -356,11 +366,48 @@ def observe(self, observation):
...
         return self.batch_act([self.observation])[0]
 
-    def batch_act(self):
-        raise NotImplementedError("Abstract class: user must implement batch_act")
+    def batch_act(self, observations):
+        batch_size = len(observations)
+        # initialize a table of replies with this agent's id
+        batch_reply = [{'id': self.getID()} for _ in range(batch_size)]
+
+        is_training = any(['labels' in obs for obs in observations])
```
stephenroller: We should do high level comments in this method, as I expect users to be falling back to it to help understand what is happening around their implementation.

parlai/core/torch_agent.py
```diff
@@ -356,11 +366,48 @@ def observe(self, observation):
...
+
+        if is_training:
+            output = self.train_step(batch.text_vec, batch.label_vec,
+                                     batch.candidates)
+        else:
+            output = self.eval_step(batch.text_vec, batch.label_vec,
+                                    batch.candidates)
+
+        if isinstance(output, tuple):
+            predictions = output[0]
```
stephenroller: Consider tuple unpacking? `predictions, candidate_preds = output`? Also, it's a bit worrying to have the user implement a method with potentially multiple return types.

parlai/agents/example_seq2seq/example_seq2seq.py
alexholdenmiller: sure that makes sense, adjusting

parlai/agents/example_seq2seq/example_seq2seq.py
alexholdenmiller: eh this is simpler and it shouldn't actually output START_IDX if it's learning properly

alexholdenmiller:
oops forgot to update the unit test for torch agent to the new api

parlai/core/dict.py
```diff
@@ -383,8 +383,6 @@ def tokenize(self, text, building=False):
 
     def bpe_tokenize(self, text):
         """Returns a sequence of BPE-tokens from the text."""
-        if self.lower:
```
jaseweston: why remove?

parlai/core/torch_agent.py
```diff
@@ -19,66 +19,100 @@
 import math
 from operator import attrgetter
 
-Batch = namedtuple("Batch", [
+Batch = namedtuple('Batch', [
```
jaseweston: comment above this explaining what it is

parlai/core/torch_agent.py
```diff
@@ -345,6 +415,10 @@ def reset(self):
         self.episode_done = True
 
     def observe(self, observation):
+        """Process incoming message in preparation for producing a response.
+
+        This includes remembering the past history of the conversation.
+        """
         observation = copy.deepcopy(observation)
```
jaseweston: possibly slow?

parlai/core/torch_agent.py
```diff
@@ -356,23 +430,67 @@ def observe(self, observation):
...
+        is_training = any(['labels' in obs for obs in observations])
+
+        # convert the observations into vectors
+        vec_obs = [self.vectorize(obs) for obs in observations]
+
+        # create a batch from the vectors
+        batch = self.batchify(vec_obs)
+
+        if is_training:
+            output = self.train_step(batch)
```
jaseweston: maybe  "batch_train_step"  and   "batch_eval_step"  ?

parlai/core/dict.py
alexholdenmiller: this wasn't really supposed to be there, this check is supposed to be in the tokenize function which will then call bpe_tokenize after lower'ing the text

parlai/core/torch_agent.py
alexholdenmiller: agreed, I want to replace this function with calling maintain_dialog_history too but happy to remove this in the meantime

parlai/core/torch_agent.py
```diff
@@ -15,41 +15,54 @@
...
-])
+"""
+Batch is a namedtuple containing data being sent to an agent.
+
+This is the input type of the train_step and eval_step functions.
+Agents can override the batchify function to return an extended namedtuple
+with additional fields if they would like, though we recommend calling the
+parent function to set up these fields as a base.
+
+:field text_vec:      bsz x seqlen tensor containing the parsed text data.
```
stephenroller: I really appreciate this, thank you.

If we want to be really thorough, we could annotate with types too. (`:field Tensor[int64] text_vec`). This isn't really necessary.

Shame it isn't 2020 so that we have [Data Classes](https://docs.python.org/3/library/dataclasses.html#module-dataclasses).

parlai/agents/example_seq2seq/example_seq2seq.py
stephenroller: (y)

parlai/agents/fairseq/fairseq.py
stephenroller: Ahh, yes, sadly the fairseq lstm is using this (but not Transformer). Thanks for checking into this.

parlai/agents/fairseq/fairseq.py
```diff
@@ -402,28 +402,49 @@ def reset(self):
...
+        If --skip-generation is not set, return a prediction for each input.
+
+        :param batch: parlai.core.torch_agent.Batch, contains tensorized
+                      version of observations.
+        """
+        if batch.text_vec is None:
+            return
         self.is_training = False
-        samples = self._make_sample(xs, ys)
+        samples = self._make_sample(batch.text_vec, batch.label_vec)
```
stephenroller: I really like the decision to use this Batch object.

parlai/agents/fairseq/fairseq.py
```diff
@@ -477,31 +499,32 @@ def report(self):
...
 
     def receive_metrics(self, metrics_dict):
-        """Used to update lr scheduler."""
+        """Update lr scheduler with validation loss."""
         self.trainer.lr_step(-1, metrics_dict["valid_loss"])
 
     # Helper functions
     def _seq_length(self, xs):
-        """Computes length of the sequence (non-padded size)"""
+        """Compute length of the sequence (non-padded size)."""
```
stephenroller: I think we can delete this whole method now that the seq length is coming in on the Batch. I'll wait to do this after this patch lands though.

parlai/core/torch_agent.py
```diff
@@ -144,26 +195,39 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None):
             obs[label_type + '_choice'] = label
         return obs
 
-    def batchify(self, obs_batch, sort=False, is_valid=lambda obs: 'text_vec' in obs):
-        """Creates a batch of valid observations from an unchecked batch, where
-        a valid observation is one that passes the lambda provided to the function.
-        Assumes each observation has been vectorized by vectorize function.
+    def batchify(self, obs_batch, sort=False,
+                 is_valid=lambda obs: 'text_vec' in obs):
```
stephenroller: Do we have any idea of when the default `is_valid` implementation may not be useful? I'm inclined to think having a `lambda` as a default parameter is a terrible idea, and this patch might be a good excuse to remove the behavior.

parlai/core/torch_agent.py
```diff
@@ -144,26 +195,39 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None):
...
+        this function and return your own "Batch" namedtuple: copy the Batch
+        namedtuple at the top of this class, and then add whatever additional
+        fields that you want to be able to access. You can then call
+        super().batchify(...) to set up the original fields and then set up the
+        additional fields in your subclass and return that batch instead.
+
         :param obs_batch: list of vectorized observations
-        :param sort:      default False, orders the observations by length of vector
+        :param sort:      default False, orders the observations by length of
+                          vector. set to true when using
```
stephenroller: nit s/set/Set/

parlai/core/torch_agent.py
```diff
@@ -211,31 +274,41 @@ def batchify(self, obs_batch, sort=False, is_valid=lambda obs: 'text_vec' in obs
...
-        return Batch(xs, ys, labels, valid_inds, cands)
+        return Batch(xs, x_lens, ys, y_lens, labels, valid_inds, cands)
 
-    def match_batch(self, batch_reply, valid_inds, predictions=None,
-                    candidate_preds=None):
+    def match_batch(self, batch_reply, valid_inds, output=None):
         """Match sub-batch of predictions to the original batch indices.
 
-        Batches may be only partially filled (e.g. when completing the remainder
+        Batches may be only partially filled (i.e when completing the remainder
```
stephenroller: nit s/i.e/i.e./

parlai/core/torch_agent.py
```diff
@@ -211,31 +274,41 @@ def batchify(self, obs_batch, sort=False, is_valid=lambda obs: 'text_vec' in obs
...
         :param batch_reply: full-batchsize list of message dictionaries to put
             responses into
-        :param predictions: sub-batchsize list of text outputs from model. may
-            be None (default) if model chooses to not answer.
         :param valid_inds: original indices of the predictions
-        :param candidate_preds: sub-batchsize list of lists of text outputs
-            ranked by the model. may be None (default) if model isn't ranking.
+        :param output: namedtuple which contains sub-batchsize list of text
+            outputs from model. may be None (default) if model chooses not to
+            answer. this method will check for ``predictions`` and
```
stephenroller: Nit, uppercase beginning of sentences in all the `:param:`'s

parlai/core/torch_agent.py
stephenroller: Awesome, thank you!

parlai/core/torch_agent.py
```diff
@@ -211,31 +274,41 @@ def batchify(self, obs_batch, sort=False, is_valid=lambda obs: 'text_vec' in obs
...
-        e.g. the length of the input sequences if using pack_padded_sequence.
+        e.g the length of the input sequences if using pack_padded_sequence.
+
         This matches rows back with their original row in the batch for
         calculating metrics like accuracy.
 
+        If output is None (model choosing not to provide any predictions), we
+        will just return the batch of replies.
+
+        Otherwise, output should be a namedtuple, which can provide predictions
```
stephenroller: Let's say that output is a `parlai.core.torch_agent.Output`, not a namedtuple.

parlai/core/torch_agent.py
```diff
@@ -366,60 +442,67 @@ def observe(self, observation):
...
+    def eval_step(self, batch):
+        """Process one batch but do not train on it."""
         raise NotImplementedError('Abstract class: user must implement eval_step')
 
 
 class Beam(object):
-    def __init__(self, beam_size, min_length=3, padding_token=0, bos_token=1, eos_token=2, min_n_best=3, cuda='cpu'):
-        """
-        Generic beam class. It keeps information about beam_size hypothesis.
+    """Generic beam class. It keeps information about beam_size hypothesis."""
```
stephenroller: IMHO these formatting changes/doc/comment changes for `Beam` should be in a different PR.

parlai/core/torch_agent.py
alexholdenmiller: intentional to keep it below 79 characters :)

parlai/core/torch_agent.py
alexholdenmiller: yeah I mean already this won't work for image captioning tasks, since there won't be text. I'm not sure if there's a better default.

parlai/core/torch_agent.py
alexholdenmiller: wow! I'm already starting to be ready to add types soon but we aren't quite yet forcing python 3.6 over some earlier 3.Xs

PR #975, opened 2018-07-11
Drop BPE codecs file and always base it on the dict_file.

stephenroller:


alexholdenmiller:
wait for the check before merging, it had an error but I think it was probably transient

stephenroller:
Things look green now.

PR #980, opened 2018-07-11
eval_model uses bsz when counting num_examples, batchsort loaded from saved opt

alexholdenmiller:


alexholdenmiller:
@emilydinan a change to override opt!

parlai/core/agents.py
```diff
@@ -300,6 +300,8 @@ def load_agent_module(opt):
                 new_opt[k] = v
         new_opt['model_file'] = model_file
         model_class = get_agent_module(new_opt['model'])
+        if 'batch_sort' in new_opt:
+            opt['batch_sort'] = new_opt['batch_sort']
```
emilydinan: is this backwards? the model is created with `new_opt` so why do we need to change `opt`?

uralik: exactly, but the batch sorting is happening not in the model, but we want it to use setting which was use during the model training

jaseweston: this is  weird and ugly to some special case flags here like that..  we need to find a nicer way of doing what you want i think.. e.g. why only batch_sort
and as emily said, i dont think it's expected this function is changing its input

jaseweston: can u please unmerge this, and work out a nicer way

Issue #969, opened 2018-07-11
Squad 2.0


theSage21:
Would it be ok if I submit a PR to update the SQuAD task to the 2.0 dataset?

alexholdenmiller:
We would be happy to have a PR for that, although don't erase the existing one--you can create a new version of the task which points to the 2.0 data instead.

theSage21:
You mean copy the existing squad task and call the new one squad2 after making changes? I saw a `version = None` inside the classes in the squad task and thought I needed to make it work with that? So that somewhere we can specify the version (maybe cli) and it's taken care of from there.

theSage21:
Does https://github.com/facebookresearch/ParlAI/pull/983 satisfy what's needed?

alexholdenmiller:
Thanks for opening the PR, I'll take a look when I get the chance! A quick skim looked good overall though.

PR #983, opened 2018-07-12
Add separate squad2.0 task for #969

theSage21:
Differences from squad

- If answering the question is impossible, answer is an empty string as per the dataset.
- `plausible_answers` key is populated in the event of an impossible question.

Everything else remains the same as the squad 1.1 task

### DIFF
--- a/parlai/tasks/squad2/agents.py
+++ b/parlai/tasks/squad2/agents.py
@@ -1,0 +1,391 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+from parlai.core.teachers import FixedDialogTeacher, DialogTeacher
+from .build import build
+
+import json
+import os
+
+
+class IndexTeacher(FixedDialogTeacher):
+    """Hand-written SQuAD teacher, which loads the json squad data and
+    implements its own `act()` method for interacting with student agent,
+    rather than inheriting from the core Dialog Teacher. This code is here as
+    an example of rolling your own without inheritance.
+
+    This teacher also provides access to the "answer_start" indices that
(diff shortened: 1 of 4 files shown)


parlai/tasks/squad2/agents.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def get(self, episode_idx, entry_idx=None):
+        article_idx, paragraph_idx, qa_idx = self.examples[episode_idx]
+        article = self.squad[article_idx]
+        paragraph = article['paragraphs'][paragraph_idx]
+        qa = paragraph['qas'][qa_idx]
+        context = paragraph['context']
+        question = qa['question']
+
+        answers = ['']
```
alexholdenmiller: here we'll want to start with an empty list, as `'' in 'any string'` returns true, which would mark every sentence as containing the label in this case

parlai/tasks/squad2/agents.py
```diff
@@ -0,0 +1,391 @@
...
+            context = context.replace(edited_answers[i], answers[i])
+
+        labels = []
+        label_starts = []
+        for sentence in sentences:
+            for answer in answers:
+                if answer in sentence and sentence not in labels:
+                    labels.append(sentence)
+                    label_starts.append(context.index(sentence))
+                    break
```
alexholdenmiller: then down here you can do:
```
if len(labels) == 0:
    labels.append('')
```

alexholdenmiller:
took a quick look with the following and it looked good:
```
import parlai.tasks.squad2.agents as s2
agents = [c for s, c in s2.__dict__.items() if s.endswith('Teacher') and 'squad2' in str(c)]
p = ParlaiParser()
p.set_params(task='squad2')
opt = p.parse_args()
print('\n\n'.join([str(a(opt).act()) for a in agents]))
```

muni2773:
@theSage21 @alexholdenmiller I understand the hit but I think the performance suffers because the underlying drqa NN is still built for Squad 1.1. Thoughts ?

PR #991, opened 2018-07-16
topk sampling from the seq2seq model, make softmax_layer_bias argument to configure bias in the last decoder layer

uralik:
So now one can do topk sampling for search (so greedy, beam and topk is supported). K is specified with --topk cmd arg. Default value is 1, which goes down to greedy search.

Another change is a bias in the last decoder layer. Now one can turn it on by using --softmax-layer-bias True argument. It is False by default.


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -161,6 +161,8 @@ def add_cmdline_args(argparser):
         agent.add_argument('--beam-size', type=int, default=1, help='Beam size, if 1 then greedy search')
         agent.add_argument('--beam-log-freq', type=float, default=0.0,
                            help='The portion of beams to dump from minibatch into model_name.beam_dump folder')
+        agent.add_argument('--topk', type=int, default=1, help='Top k sampling from renormalized softmax, default 1 means simple greedy max output')
```
jaseweston: at test time

PR #994, opened 2018-07-17
update bpe tokenizer, lower-case special tokens

alexholdenmiller:
- switch bpe helper to use final dictionary counts instead of remembering all training data
- replace --bpe-num-symbols with --dict-maxtokens
- use --dict-minfreq for bpe's min frequency
- switch special tokens to lower case, as upper case causes frequent issues with lookups (if using --dict-lower(), `dict['__START__']` returns the index of the start token but `dict.txt2vec('__START__')` returns the unknown index because it doesn't recognize `'__start__'`)
- convert loaded upper-cased special tokens to lower case

to handle the bug described we could put a `lower()` in the indexing function, but the place it is now makes reasonable sense (at the beginning of tokenizing)


parlai/core/dict.py
```diff
@@ -383,46 +377,45 @@ def tokenize(self, text, building=False):
...
-                # other dicts can be modified as we go
-                idx = self.tok2ind.pop(token)
-                del self.ind2tok[idx]
+
         for token in to_remove:
             del self.freq[token]
+            idx = self.tok2ind.pop(token)
+            del self.ind2tok[idx]
+
+    def remove_non_bpe(self):
```
stephenroller: should this be `_remove_non_bpe`? It seems like a good idea to signal to users that this isn't something they should call themselves.

parlai/core/dict.py
```diff
@@ -579,82 +586,69 @@ class _BPEHelper(object):
...
-    def _apply(self, tokens):
-        return self.bpe.segment_tokens(tokens)
+        if hasattr(self, 'bpe'):
+            return self.bpe.segment_tokens(tokens)
+        else:
+            return tokens
 
-    def finalize(self):
-        """Build the codecs"""
-        if self.built:
```
stephenroller: I'm worried about this, b/c without testing this code, it looks to me like loading a model from disk will still cause the BPE encoder to retrain. This is a very expensive step, and so if the dict and codecs files are already built, we should use them.

parlai/core/dict.py
```diff
@@ -116,7 +116,7 @@ def add_cmdline_args(argparser):
             dictionary.add_argument(
                 '--dict-maxtokens', default=DictionaryAgent.default_maxtokens,
                 type=int,
-                help='max number of tokens to include in sorted dict')
+                help='max number of tokens to include in dictionary or bpe codecs.')
```
stephenroller: should we make a similar note for minfreq?

parlai/core/dict.py
```diff
@@ -460,8 +459,15 @@ def save(self, filename=None, append=False, sort=True):
         """
         filename = self.opt['dict_file'] if filename is None else filename
         print('Dictionary: saving dictionary to {}'.format(filename))
-        if sort:
-            self.sort()
+
+        if self.tokenizer == 'bpe':
+            self.bpehelper.finalize((f'{k} {v}' for k, v in self.freq.items()),
```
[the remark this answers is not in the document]
alexholdenmiller: good call!

parlai/core/dict.py
alexholdenmiller: agreed, good call

parlai/core/dict.py
alexholdenmiller: done

PR #998, opened 2018-07-18
Simple dataset creation docs

jaseweston:
it's still pretty confusing that fbdialog teacher is in there, when we also now have parlaidialogteacher,  but at least the top  of the doc is simple now?


docs/source/tutorial_task.rst
```diff
@@ -5,23 +5,75 @@
...
+If you have a dialogue, QA or other text-only dataset that you can put
+in a text file in the format we will now describe, you can just load it directly from
+there, with no extra code!
+
+Here's an example dataset with a single episode with 2 examples:
+
+::
+
+	text:hello how are you today?	label:i'm great thanks! what are you doing?
+	text:i've just been biking.	label:oh nice, i haven't got on a bike in years!	episode_done:True
```
alexholdenmiller: labels

PR #1006, opened 2018-07-23
update qangeroo to share data

alexholdenmiller:


alexholdenmiller:
merging for @taineleau

PR #1007, opened 2018-07-23
Convai2 dataset update

emilydinan:
Small updates to convai2 dataset:
- self_original remains the same in train/valid set
- other datasets are edited to have the exact same editing as self_original (i.e. "i'm" vs "i am" switches)


emilydinan:
@jaseweston and it shouldn't change any of the files they are working with for the competition unless they are using the revised personas

PR #1008, opened 2018-07-24
[fairseq] Allow setting default arguments in an overridable way.

stephenroller:


stephenroller:
The error looks transient.

stephenroller:
Just needed to rebase to get rid of the error.

The normal way of overriding `add_cmdline_args` to `argparser.set_defaults` fails to work because `FairseqAgent.add_cmdline_args` uses `parse_known_args` in order to add sub-options from fairseq. The result is that overriding the normal way causes these sub-options to be picked using the old defaults.

PR #1010, opened 2018-07-24
better display_data

jaseweston:
- doesn't show repeatLabelAgent (we already have the 'labels' field anyway)
- shows other hidden fields if they exist (unless you ignore them). clips the text of those too if needed

### DIFF
--- a/parlai/scripts/display_data.py
+++ b/parlai/scripts/display_data.py
@@ -22,8 +22,9 @@ def setup_args(parser=None):
     if parser is None:
         parser = ParlaiParser(True, True)
     # Get command line arguments
-    parser.add_argument('-ne', '--num-examples', type=int, default=10)
-    parser.add_argument('-mdl', '--max-display-len', type=int, default=1000)
+    parser.add_argument('-ne', '--num_examples', type=int, default=10)
+    parser.add_argument('-mdl', '--max_display_len', type=int, default=1000)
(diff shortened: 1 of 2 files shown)


parlai/scripts/display_data.py
```diff
@@ -22,8 +22,9 @@ def setup_args(parser=None):
     if parser is None:
         parser = ParlaiParser(True, True)
     # Get command line arguments
-    parser.add_argument('-ne', '--num-examples', type=int, default=10)
-    parser.add_argument('-mdl', '--max-display-len', type=int, default=1000)
+    parser.add_argument('-ne', '--num_examples', type=int, default=10)
```
emilydinan: changing this flag just here could be confusing because we use the same name elsewhere but with '-' instead of '_' (see for instance scripts/display_model.py, scripts/eval_model.py, scripts/convert_data_to_parlai_format.py, etc). maybe we should change it everywhere or nowhere?

jaseweston: yes i would like to have the "_" everywhere, less confusing in the code, but i don't want to do it in this PR -- so you want me to change it back for now, or not?

dykang:
In case I use fromfile and fromfile_datapath, the task name looks like the filename, which seems not good. 

Also, as Jason said, I would like to use an additional tag (e.g., different id or name for each episode). I look forward to seeing this PR!

PR #1012, opened 2018-07-25
Add tblogger parser, now --tensorboard-log False works when you start…

uralik:
… the script

We need this part in order to be able to rewrite tensorboard-log to False during eval (that is what we usually want)


stephenroller:
AFAIK, the usual semantics we'd have is that we'd use `opt.get('tensorboard_log', None)`, so that these command line args aren't needed in the script.

stephenroller:
Talked with @uralik offline and realized I misunderstood his intention with this. All looks well.

PR #1015, opened 2018-07-26
Random Candidate Agent

emilydinan:
Agent that chooses a random label candidate as the reply if label candidates are available. If they are not, it repeats the label.

It also randomly shuffles the label candidates to return as text candidates.

### DIFF
--- a/parlai/agents/random_candidate/random_candidate.py
+++ b/parlai/agents/random_candidate/random_candidate.py
@@ -1,0 +1,40 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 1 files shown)


parlai/agents/random_candidate/random_candidate.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+"""Simple agent which chooses a random label from the label candidates if
+they are available. If candidates are not available, it repeats the label.
```
jaseweston: not sure it should repeat the label? should more get it wrong, no?

PR #1032, opened 2018-07-27
add candidates to torch agent vectorization, cleaner vectorize

alexholdenmiller:
- move vectorization of each text element to a separate private function
- switch from deque to list, simpler and faster in this current implementation (it was faster in seq2seq when there was more appending of elements)
- add candidates to vectorization
- return candidates in batchify


parlai/core/torch_agent.py
```diff
@@ -151,6 +156,44 @@ def share(self):
         shared['dict'] = self.dict
         return shared
 
+    def _vectorize_text(self, text, use_cuda=False, add_start=False,
```
vgonzalez88: Wooo! I like the helper for readability.

parlai/core/torch_agent.py
```diff
@@ -162,37 +205,43 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None):
...
-            vec_label += self.dict.txt2vec(label)
-            if add_end:
-                vec_label.append(self.END_IDX)
-            new_label = torch.LongTensor(vec_label)
-            if self.use_cuda:
-                new_label = new_label.cuda()
-            obs[label_type + '_vec'] = new_label
+            vec_label = self._vectorize_text(label, self.use_cuda, add_start,
+                                             add_end, truncate, False)
+            # print(label, vec_label, truncate)
```
vgonzalez88: nit: remove comment

alexholdenmiller: thanks good catch

PR #1036, opened 2018-07-31
ParlAI-Mturk Refactor with new testing

JackUrb:
This is a prerequisite to changing to the sqlite db and then integrating bonus-per-interaction as a first class feature.

Viewing the individual commits most of the code changes are moving things around into places that make more sense, with a slight improvement to stability and reduction of redundancy. This also standardizes naming of variables to have consistent logic throughout the codebase, where a **worker** refers to an individual who is using the MTurk service and tracks their state across assignments, and an **agent** is and individual instance of that worker that tracks the state in an individual assignment for a world.

The end goal is being able to have testable and more segmented code for unit tests (Rather than only being able to rely on integration testing). It is also required to segment agent state out from `MTurkManager` as agent state is going to be moved to being sqlite backed, so `MTurkManager` has to use indirect accesses.

Agent management stuff has been moved out of `MTurkManager` into `MTurkAgent`, and worker state management has been moved out of `MTurkManager` into a new `WorkerManager` class. Individual changes can be seen in the individual commits. 

Testing:
- [x] Worker Manager unit tests
- [x] MTurk Agent unit tests
- [x] Socket Manager unit tests 
- [x] Socket Manager integration tests 
- [x] MTurk Manager unit tests
- [x] MTurk Manager Integration tests 

Additional
- [x] fix regression of #1058 caused by the initial refactor.


parlai/mturk/core/worker_manager.py
```diff
@@ -0,0 +1,344 @@
...
+        self.is_sandbox = mturk_manager.is_sandbox
+
+    def _create_agent(self, hit_id, assignment_id, worker_id):
+        """Initialize an agent and return it"""
+        return MTurkAgent(
+            self.opt, self.mturk_manager, hit_id, assignment_id, worker_id)
+
+    def _get_worker(self, worker_id):
+        """A safe way to get a worker by worker_id"""
+        if worker_id in self.mturk_workers:
```
klshuster: why not e.g. `self.mturk_workers.get(worker_id, None)`

parlai/mturk/core/worker_manager.py
```diff
@@ -0,0 +1,344 @@
...
+            # Clear the send message command, as a message was recieved
+            agent.set_last_command(None)
+            agent.put_data(pkt.id, pkt.data)
+
+    def map_over_agents(self, map_function, filter_func=None):
+        """Take an action over all the agents we have access to, filters if
+        a filter_func is given"""
+        for worker in self.mturk_workers.values():
+            for agent in worker.agents.values():
+                if filter is None or filter_func(agent):
```
klshuster: i think `filter` should be `filter_func` here

parlai/mturk/core/worker_manager.py
```diff
@@ -0,0 +1,344 @@
...
+            conversation_id = agent.conversation_id
+            if conversation_id in self.conv_to_agent:
+                conv_participants = self.conv_to_agent[conversation_id]
+                if agent in conv_participants:
+                    for other_agent in conv_participants:
+                        if agent.assignment_id != other_agent.assignment_id:
+                            partner_callback(other_agent)
+                if len(conv_participants) > 1:
+                    # The user disconnected from inside a conversation with
+                    # another turker, record this as bad behavoir
```
klshuster: nit: `behavior`

parlai/mturk/core/worker_manager.py
```diff
@@ -0,0 +1,344 @@
...
+
+    def get_complete_hits(self):
+        """Returns the list of all currently completed HITs"""
+        hit_ids = []
+        for hit_id, agent in self.hit_id_to_agent.items():
+            if agent.hit_is_complete:
+                hit_ids.append(hit_id)
+        return hit_ids
+
+    def get_agent_work_status(self, assignment_id):
```
klshuster: should this function return something by default? or is `None` handled? looks like if there's an error that's not the one you specified, there's no return branch

parlai/mturk/core/agents.py
```diff
@@ -22,6 +21,133 @@
...
+    STATUS_IN_TASK = 'in task'
+    STATUS_DONE = 'done'
+    STATUS_DISCONNECT = 'disconnect'
+    STATUS_PARTNER_DISCONNECT = 'partner disconnect'
+    STATUS_PARTNER_DISCONNECT_EARLY = 'partner disconnect early'
+    STATUS_EXPIRED = 'expired'
+    STATUS_RETURNED = 'returned'
+
+    def __init__(self, status=None):
+        """Create an AssignState with the given assignment_id.
```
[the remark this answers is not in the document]
JackUrb: Ah yeah I should update this - the `assignment_id` was duplicated between this class and the `MTurkAgent` class and the only class that accesses the `AssignState` class was the `MTurkAgent` so I realized it was unneeded complication.

parlai/mturk/core/agents.py
```diff
@@ -356,9 +529,11 @@ def wait_for_hit_completion(self, timeout=None):
...
         sync_attempts = 0
-        while not self.hit_is_complete and \
-                self.manager.get_agent_work_status(self.assignment_id) != \
-                self.ASSIGNMENT_DONE:
+        while (
+            not self.hit_is_complete and
+            self.mturk_manager.get_agent_work_status(self.assignment_id) !=
+            self.ASSIGNMENT_DONE
+        ):
             if sync_attempts < 8:
```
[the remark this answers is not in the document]
JackUrb: Ah this actually scales, I should document this better and correct the implementation (as it's incorrect given the new THREAD_X_SLEEP durations). Every iter is a THREAD_SHORT_SLEEP duration, which is now to 0.1 seconds. The time right now is 0.1 + 0.2 + 0.4 + 0.8 + 1.8 + 3.2 + 6.4 + 12.8 ≠ 2 hours hahaha.

parlai/mturk/core/mturk_manager.py
```diff
@@ -190,233 +193,152 @@ def _log_working_time(self, mturk_worker):
...
+
+        self.worker_manager.map_over_agents(expire_func, is_onboard)
 
-    def _expire_worker_pool(self):
+    def _expire_agent_pool(self):
         """Expire all workers in the worker pool"""
-        for agent in self.worker_pool:
+        for agent in self.agent_pool.copy():
             self.force_expire_hit(agent.worker_id, agent.assignment_id)
+            self._remove_from_agent_pool(agent)
```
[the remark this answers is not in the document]
JackUrb: Technically no, as this function is only called after the agent pool is no longer being queried, but for completeness yes. Thanks for the catch!

parlai/mturk/core/mturk_manager.py
```diff
@@ -1257,7 +1183,7 @@ def get_qualification_list(self, qualifications=None):
             qualifications.append({
                 'QualificationTypeId': block_qual_id,
                 'Comparator': 'DoesNotExist',
-                'RequiredToPreview': True
+                'ActionsGuarded': 'DiscoverPreviewAndAccept'
```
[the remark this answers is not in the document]
JackUrb: Gotta keep up with the MTurk API sometime right?

parlai/mturk/core/mturk_manager.py
```diff
@@ -1536,7 +1477,7 @@ def create_qualification(self, qualification_name, description,
         if not can_exist:
             qual_id = mturk_utils.find_qualification(qualification_name,
                                                      self.is_sandbox)
-            if qual_id is not None:
+            if qual_id is not None and not can_exist:
```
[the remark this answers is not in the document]
JackUrb: Oops I somehow missed the outer condition - thanks for this catch!

parlai/mturk/core/test/test_full_system.py
```diff
@@ -0,0 +1,914 @@
...
+        self.disconnected = False
+        self.closed = False
+        self.ws = None
+        self.should_heartbeat = True
+        self.fake_workers = []
+        self.launch_socket()
+        self.handlers = {}
+
+    def send(self, packet):
+        self.ws.send_message_to_all(packet)
```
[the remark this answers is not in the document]
JackUrb: Almost true - `__init__` sets `self.ws`, but it happens in a separate thread. Will add a check in the init calls

parlai/mturk/core/worker_manager.py
JackUrb: copypasta - yours is cleaner

parlai/mturk/core/worker_manager.py
JackUrb: Good catch! Turns out I started failing this test. Oops...

parlai/mturk/core/worker_manager.py
JackUrb: None of the callsites for this function do anything other than an `==` compare to `ASSIGNMENT_DONE` or `ASSIGNMENT_NOT_DONE`, so it hasn't come up, but yeah I should probably have this be an explicit return.

PR #1043, opened 2018-08-03
Update SQuAD Task to Allow Non-Random Loading

vgonzalez88:
Necessary for an MTurk Task. 

Allows us to control if SQuAD can load examples sequentially during training.


vgonzalez88:
I was not aware of `-dt train:ordered` this accomplishes the same thing.

PR #1052, opened 2018-08-07
Implement fairseq candidate ranking using generator.

### LINKED ISSUE #924: FairSeq grade candidates
Fairseq can currently generate and evaluate a single label, but it ignores `label_candidates` for when there's multiple labels.

The change belongs here:
https://github.com/facebookresearch/ParlAI/blob/5994d437089892ec626fcd510b6fefd2869d8a80/parlai/agents/fairseq/fairseq.py#L348-L349

stephenroller:
Uses avg probability per token of the sequence (should be same thing `seq2seq` uses). This is not configurable; it is just a dumb call to fairseq's grader.

Supports variable number of candidates. The way this is done is with multiple calls to scorer, so the batch size isn't as exploited as it could be. We could increase throughput by streamlining, but it would come at the cost of complexity. Seems fine for now.

Holding until #1049 is merged, as there will likely need to be unit test updates and resolving conflicts.

Fixes #924.

### DIFF
--- a/parlai/agents/fairseq/fairseq.py
+++ b/parlai/agents/fairseq/fairseq.py
@@ -468,10 +468,13 @@ class FairseqAgent(TorchAgent):
             # Interactive mode won't have a gold label
             self.trainer.valid_step(samples)
 
+        # Output placeholders
+        reranked_cands = None
+        generated_output = None
+
         # Grade each of the candidate sequences
         if batch.candidate_vecs is not None:
             bsz = len(batch.text_vec)
-            best_cands = []
             reranked_cands = []
             # score the candidates for each item in the batch separately, so that
             # we can support variable number of candidates
@@ -479,7 +482,6 @@ class FairseqAgent(TorchAgent):
                 cands = batch.candidate_vecs[i]
                 if not cands:
                     reranked_cands.append(None)
-                    best_cands.append(None)
                     continue
                 ncand = len(cands)
                 # repeat the input many times
@@ -495,12 +497,24 @@ class FairseqAgent(TorchAgent):
                 # intentional hanging comma here; argsort returns a list
                 ranked, = self._argsort(scores, batch.candidates[i], descending=True)
                 reranked_cands.append(ranked)
-                # best candidate is the highest scoring of the items
-                best_cands.append(ranked[0])
-            return Output(best_cands, reranked_cands)
-        elif not self.args.skip_generation:
-            # Next generate freely to create our response
-            return Output(self._generate(samples), None)
+
+        # Next generate freely to create our response
+        if not self.args.skip_generation:
+            generated_output = self._generate(samples)
+        elif reranked_candidates:
+            # we're skiping generation, but we're also grading candidates
+            # so output the highest ranked candidate
+            # In the case of zero candidates, we don't have something to rank,
+            # so we may need to pass on that None
+            generated_output = [
+                ranked and ranked[0] or None for ranked in reranked_cands
+            ]
+        else:
+            # no output at all
+            pass
+
+        return Output(generated_output, reranked_cands)
+
 
     def _generate(self, samples):
         src_tokens = samples["net_input"]["src_tokens"]
--- a/parlai/agents/fairseq/fairseq.py
+++ b/parlai/agents/fairseq/fairseq.py
@@ -5,6 +5,7 @@
 # of patent rights can be found in the PATENTS file in the same directory.
 
 from parlai.core.dict import DictionaryAgent
+from parlai.core.utils import argsort, padded_tensor
 
 try:
     from fairseq import models, optim, criterions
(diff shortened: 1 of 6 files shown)


parlai/agents/fairseq/fairseq.py
```diff
@@ -448,10 +451,31 @@ def eval_step(self, batch):
         if batch.label_vec is not None:
             # Interactive mode won't have a gold label
             self.trainer.valid_step(samples)
-        # Grade each of the candidate sequences
-        # TODO: grade everything in observations[i]['label_candidates']
 
-        if not self.args.skip_generation:
+        # Grade each of the candidate sequences
+        if batch.candidate_vecs is not None:
```
stephenroller: Review question: would we rather this be self.rank_candidates?

parlai/agents/fairseq/fairseq.py
```diff
@@ -448,10 +451,31 @@ def eval_step(self, batch):
...
+                ys, _ = self._padded_tensor(cands)
+                s = self._make_sample(xs, ys)
+                # perform the actual grading
+                scored = list(self.scorer.score_batched_itr([s], cuda=self.use_cuda))
+                scores = [s[3][0]['score'].item() for s in scored]
+                ranked, = self._argsort(scores, batch.candidates[i], descending=True)
+                reranked_cands.append(ranked)
+                # best candidate is the highest scoring of the items
+                best_cands.append(ranked[0])
+            return Output(best_cands, reranked_cands)
```
stephenroller: Should we support generation *and* ranking at the same time?

parlai/core/torch_agent.py
```diff
@@ -250,6 +250,45 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None):
 
         return obs
 
+    def _padded_tensor(self, items):
```
stephenroller: TODO: unit test

parlai/core/torch_agent.py
```diff
@@ -250,6 +250,45 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None):
...
+        lens = [len(item) for item in items]
+        t = max(lens)
+        output = torch.LongTensor(n, t).fill_(self.NULL_IDX)
+        for i in range(len(items)):
+            output[i, :lens[i]] = items[i]
+        if self.use_cuda:
+            output = output.cuda()
+        return output, lens
+
+    def _argsort(self, keys, *lists, descending=False):
```
stephenroller: TODO: unit test.

Both these utility methods could be useful as general utilities rather than being stuffed into this module here.

parlai/core/torch_agent.py
```diff
@@ -250,6 +250,45 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None):
...
+        """Reorder each list in lists by the (descending) sorted order of keys.
+
+        :param iter keys: sort indices
+        :param list[list] lists: lists to reordered by keys's order
+        :param bool descending: Use descending order if true
+        :return list[list]]: The reordered items
+        """
+        ind_sorted = sorted(range(len(keys)), key=lambda k: keys[k])
```
stephenroller: In general I'd usually call `np.argsort` here but numpy arrays aren't common in the code.

parlai/core/torch_agent.py
```diff
@@ -250,6 +250,45 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None):
...
+        if descending:
+            ind_sorted = list(reversed(ind_sorted))
+        output = []
+        for l in lists:
+            if isinstance(l, set):
```
stephenroller: This should be dropped. It was a hack for `babi`, but we should fix this somewhere else.

parlai/core/torch_agent.py
```diff
@@ -250,6 +250,45 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None):
...
+        t = max(lens)
+        output = torch.LongTensor(n, t).fill_(self.NULL_IDX)
+        for i in range(len(items)):
+            output[i, :lens[i]] = items[i]
+        if self.use_cuda:
+            output = output.cuda()
+        return output, lens
+
+    def _argsort(self, keys, *lists, descending=False):
+        """Reorder each list in lists by the (descending) sorted order of keys.
```
stephenroller: Document that it's happy with Tensors too.

parlai/agents/fairseq/fairseq.py
alexholdenmiller: I think this is a fine way to do it: rank the candidates if they're there.

parlai/agents/fairseq/fairseq.py
alexholdenmiller: I prefer that but it's probably not *that* useful

parlai/agents/fairseq/fairseq.py
```diff
@@ -538,6 +562,7 @@ def _make_sample(self, xs, ys):
         # TODO: should the right/left padding thing be in torch agent?
         repadded = convert_padding_direction(xs, self.dict.pad(), right_to_left=True)
         sample = {}
+        sample["id"] = torch.range(0, len(xs) - 1)
```
alexholdenmiller: arange so torch doesn't get mad

stephenroller:
Ready for review.

parlai/core/torch_agent.py
alexholdenmiller: let's move argsort to utils too but I'll leave handling the optional torch import up to you

alexholdenmiller:
core changes look great!

stephenroller:
sounds great, I'll get those changes in.

stephenroller:
Updated and rebased

stephenroller:
Rebase.

PR #1054, opened 2018-08-09
Fixes to FP16 with fairseq master.

stephenroller:


stephenroller:
needs rubber stamp plz

PR #1063, opened 2018-08-13
Fix a bug when using model replies and ending episodes.

stephenroller:


stephenroller:
@alexholdenmiller should also look at this when he's back from PTO.

alexholdenmiller:
looks good, thanks @stephenroller!

PR #1064, opened 2018-08-13
Cache txt2vec items.

stephenroller:
Surpringly impactful. For example, with a weak model, training time for
2 epochs of convai2 goes from 335s to 211s (~40% speedup).


parlai/core/dict.py
```diff
@@ -101,6 +102,7 @@ class DictionaryAgent(Agent):
     default_tok = 're'
     default_lower = False
     default_textfields = 'text,labels'
+    default_cachesize = 1048576
```
jaseweston: why this?

stephenroller: Cache is most powerful if it's a power of 2. I'll add a comment.

alexholdenmiller:
great idea!

Issue #1056, opened 2018-08-10
ConvAI2 wild evaluation dataset contribution


litinsky:
Hi, folks!

I represent ConvAI2 organizing team. We obtained conversations dataset during the challenge. How can I contribute it to ParlAI?

jaseweston:
Hi! There is documentation how to add a task here:

http://www.parl.ai/static/docs/tutorial_task.html

and you can look at all the  other tasks already there as examples:

https://github.com/facebookresearch/ParlAI/tree/master/parlai/tasks

jaseweston:
i think this is done right?

PR #1071, opened 2018-08-15
Convai wild evaluation dataset task

litinsky:
Dataset collected during the wild evaluation of ConvaAI2 participants bots (http://lnsigo.mipt.ru/export/datasets/convai/convai2_wild_evaluation_0.2.tgz). 1921 convaersations. 60% train, 20% valid and 20% test is chosen at random from the whole dataset.

### DIFF
--- a/parlai/tasks/convai2_wild_evaluation/build.py
+++ b/parlai/tasks/convai2_wild_evaluation/build.py
@@ -21,6 +21,7 @@ def make_parlai_format(data: list, dpath: str):
     assert valid_p > 0
     assert test_p > 0
 
+    random.seed()
     random.shuffle(data)
     data_len = len(data)
 
--- a/parlai/tasks/convai2_wild_evaluation/build.py
+++ b/parlai/tasks/convai2_wild_evaluation/build.py
@@ -1,0 +1,80 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+# Download and build the data if it does not exist.
+
+import os
+import json
+import random
+
+import parlai.core.build_data as build_data
+
+
+def make_parlai_format(data: list, dpath: str):
+    train_p = 0.6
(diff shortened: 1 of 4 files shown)


parlai/tasks/task_list.py
```diff
@@ -411,4 +411,11 @@
         "tags": ["All", "Visual"],
         "description": "COCO annotations derived from the 2015 COCO Caption Competition. Link to dataset: http://cocodataset.org/#download",
     },
+    {
+        "id": "ConvAI2_wild_evaluation",
+        "display_name": "ConvAI2_wild_evaluation",
+        "task": "convai2_wild_evaluation",
+        "tags": ["All",  "ChitChat" ],
+        "description": "Dataset collected during the wild evaluation of ConvaAI2 participants bots (http://lnsigo.mipt.ru/export/datasets/convai/convai2_wild_evaluation_0.2.tgz). 60% train, 20% valid and 20% test is chosen at random from the whole dataset."
```
jaseweston: best not to include the tgz here as it will likely be updated, but could include http://convai.io perhaps?

litinsky: jaseweston, shure, fixed it.

parlai/tasks/convai2_wild_evaluation/build.py
```diff
@@ -0,0 +1,81 @@
...
+def make_parlai_format(data: list, dpath: str):
+    train_p = 0.6
+    valid_p = 0.2
+    test_p = 1 - (train_p + valid_p)
+
+    assert train_p > 0
+    assert valid_p > 0
+    assert test_p > 0
+
+    random.seed()
```
jaseweston: we don't want a random seed changing the dataset for everyone who downloads it here

jaseweston: probably better to just take every other n lines or so?

stephenroller:
Does this fix #1056?

parlai/tasks/convai2_wild_evaluation/build.py
litinsky: jaseweston, sorry for being absent - was on my vacation. Removed random seed from the dataset extraction.

Issue #875, opened 2018-06-16
fix pytorch UserWarnings for softmax, item(), etc


christqoh:
Not an actual issue, just user warnings you might want to know about so it doesn't turn into one:

`ParlAI/projects/personachat/persona_seq2seq.py:1738: UserWarning: Implicit dimension choice for softmax has been deprecated. Change the call to include dim=X as an argument.
  attn_weights = F.softmax(attn_w_premask)`

`ParlAI/projects/personachat/persona_seq2seq.py:1486: UserWarning: Implicit dimension choice for log_softmax has been deprecated. Change the call to include dim=X as an argument.
  scores = F.log_softmax(scores)`

as found through running `python projects/convai2/baselines/profilememory/interactive.py`

PR #1073, opened 2018-08-15
Get rid of persona_seq2seq Convai2 baseline code

emilydinan:
As it is not an official baseline for the ConvAI2 competition we are going to remove it from the convai2/baselines folder. 

NOTE: We still have scripts for evaluating and interacting with this model (pre-trained on the personachat dataset) located in the projects/personachat folder.


stephenroller:
Makes #875 no longer necessary.

emilydinan:
@stephenroller for #875 we will still have this model and the scripts for the model trained on personachat (and these warnings will still show)

PR #1074, opened 2018-08-15
Throw away empty vector items from the batch.

stephenroller:
This causes problems for all of the fairseq models, and will probably cause problems for future models once they are built on top of TorchAgent.

The issue is that on the final batch (but only when using `-bsrt True`), you can get a batch with items that are only pad tokens. This has a tendency to either break attention mechanisms (Transformer) or break the calls to `pad_packed_sequence` (lstm, including seq2seq). By filtering these examples, we don't have to build model-specific work arounds.


stephenroller:
Added the docstring back in.

PR #1088, opened 2018-08-22
Add ability to specify number of cands in Flickr30k

klshuster:
Similar to functionality in coco_caption


parlai/tasks/flickr30k/agents.py
```diff
@@ -128,6 +130,10 @@ def add_cmdline_args(argparser):
                            help='Include an intro question with each image \
                                 for readability (e.g. for coco_caption, \
                                 Describe the above picture in a sentence.)')
+        agent.add_argument('--num_cands', type=int,
```
alexholdenmiller: the underscore makes its debut!!

PR #1092, opened 2018-08-22
Touch it, fix it. Add strict linting to changed files.

stephenroller:
As discussed.

Marking as [wip] because I want to test it with travis.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8]
+ignore =
+    C
+
+max-line-length=88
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+ignore =
+    C
+
+max-line-length=88
```
alexholdenmiller: why 88?

stephenroller: https://github.com/ambv/black/blob/master/README.md#command-line-options matches internal defaults

stephenroller:
All ready

PR #1095, opened 2018-08-23
Clean up squad sentence teachers

emilydinan:
Delete some unnecessary squad sentence teachers. Now there are two "sentence teachers: `squad:sentence` and `squad:fulldocsentence`. 

Use the flag `--include-context` to have the data formatted as `'text': <context>\n<question>` (if `True`) or `'context': <context>, 'text': <question>` (if `False`, this is the default behavior)

### DIFF
--- a/parlai/tasks/squad/agents.py
+++ b/parlai/tasks/squad/agents.py
@@ -222,71 +222,35 @@ class FulldocTeacher(ParlAIDialogTeacher):
 class SentenceTeacher(IndexTeacher):
     """Teacher where the label(s) are the sentences that contain the true
     answer.
-    """
-    def __init__(self, opt, shared=None):
-        super().__init__(opt, shared)
-        self.sent_tok = get_sentence_tokenizer()
-
-    def get(self, episode_idx, entry_idx=None):
-        article_idx, paragraph_idx, qa_idx = self.examples[episode_idx]
-        article = self.squad[article_idx]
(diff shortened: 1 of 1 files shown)


parlai/tasks/squad/agents.py
```diff
@@ -383,15 +304,40 @@ class FulldocsentenceTeacher(FulldocTeacher):
...
+                           'separate field')
 
     def get(self, episode_idx, entry_idx=None):
+        action = {}
         episode = self.episodes[episode_idx][entry_idx]
-        entry = {'episode_done': episode['episode_done']}
-        context = ' '.join(episode['text'].split('\n')[:-1]).replace('\xa0', ' ')
+        context = ' '.join(
+            episode['text'].split('\n')[:-1]
+        ).replace('\xa0', ' ')
```
stephenroller: what's up with the weird character replacement?

emilydinan: great question...this is a non breaking space character and they were left in the text. in some cases, the correct answer span in the text contains this character and so if we don't remove it, the model has no possibility of getting the answer correct (if it returns an answer span)

stephenroller: Makes sense. It'd be nice if we had a comment here briefly explaining that.

emilydinan: added with commit ff68e3

PR #1096, opened 2018-08-23
Add flag for multiple GPUs. User must manually support this.

stephenroller:


stephenroller:
One thing we could do is add a stub method called `setup_multigpu()` or similar which just throws a NotImplementedError. What's problematic about this is we can't call it in `TorchAgent.__init__ because` the subclass probably hasn't set up its model yet (it's running `super().__init__() now, but yet to run its own `__init__()`).

PR #1097, opened 2018-08-23
[hold] migrate seq2seq agent to TorchAgent parent class

alexholdenmiller:
- params.py had bug: wouldn't model-zoo-ify model_file and dict_file args in 'override' dict
- update padding_tensor in util to also support incoming list of tensors
- remove code from seq2seq which is now in torchagent
- update seq2seq to define torchagent sub-functions (train_step and eval_step)
- fix now-broken seq2seq code (new batch format, perplexity evaluator needs different vectorization)
- add some new features to torchagent (optimizer dict, bug in history, call reset-metrics in reset)


alexholdenmiller:
I'll double-check this further tomorrow, pseudo-wip at the moment. @uralik have a few things we want to do as well which maybe should be in another diff, but also this should NOT land until existing code has been backed up into a legacy folder (I'll do in another diff), and code here must add version check.

parlai/core/torch_agent.py
```diff
@@ -110,6 +120,10 @@ def dictionary_class():
     def add_cmdline_args(argparser):
         """Add the default commandline args we expect most agents to want."""
         agent = argparser.add_argument_group('TorchAgent Arguments')
+        agent.add_argument(
+            '-opt', '--optimizer', default='sgd',
+            help='Choose between pytorch optimizers. Any member of torch.optim'
```
stephenroller: nit: space needed after torch.optim

parlai/core/torch_agent.py
```diff
@@ -253,8 +276,9 @@ def vectorize(self, obs, add_start=True, add_end=True, truncate=None,
         :param split_lines: If set, returns list of vectors instead of a single
                             vector for input text, one for each substring after
                             splitting on newlines.
+        :param nocache:     default False, recalculates vecs even if present
         """
-        if 'text_vec' in obs:
+        if 'text_vec' in obs and not nocache:
```
stephenroller: Interesting. Why?

alexholdenmiller: I manually modified the eval_labels field in the PerplexityEvaluator and wanted to re-vectorize. However I realized I could just call the _vectorize_text function... I'll revert this.

stephenroller:
On another note, I've noticed this span appearing as copy-pasta in a few places:

https://github.com/facebookresearch/ParlAI/blob/519f7695aecfdc2a06498a0e32a83aff937380a9/parlai/agents/seq2seq/seq2seq.py#L181-L225

This might might be a good opportunity to refactor this out into some sort of TorchAgent utils method.

parlai/core/utils.py
```diff
@@ -914,14 +914,24 @@ def padded_tensor(items, pad_idx=0, use_cuda=False, left_padded=False):
...
+        output = torch.LongTensor(n, t)
+    output.fill_(pad_idx)
+
+    for i, item in enumerate(items):
+        if not isinstance(item, torch.Tensor):
+            item = torch.LongTensor(item)
         if left_padded:
             # place at end
-            output[i, t - lens[i]:] = torch.LongTensor(items[i])
+            output[i, t - lens[i]:] = item
```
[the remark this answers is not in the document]
alexholdenmiller:
```
if not isinstance(item, torch.Tensor):
    item = torch.LongTensor(item)
```
should do it right?

alexholdenmiller:
yeah good call--I had just moved the whole init-model code out in my copy but moving the pretrained-embedding code up to torch agent is great

alexholdenmiller:
TODO: add version check to print warning to use legacy seq2seq

alexholdenmiller:
ready for migration, although open to suggestions for any more good places to split up functionality.

I'd like to also actually make same changes to clean up seq2seq which break backwards compatibility (handled by the new version flag), but this diff is already big... I may do as a diff onto this branch.

alexholdenmiller:
marking [hold] as I shouldn't land this until we incorporate all other breaking changes into this branch.

alexholdenmiller:
moving all non-s2s changes to separate diffs

alexholdenmiller:
this diff also weirdly picked up other file changes from one of the merge commits, so I may remove this completely in the end

alexholdenmiller:
abandoning in favor of #1110

PR #1100, opened 2018-08-24
Fixes small bug in acting after a disconnect (disconnect comes first)

JackUrb:
If an agent disconnects while a world is still pulling info, the disconnect may cause the message queue to be removed before act will return that the agent has disconnected, causing a crash. This fixes that bug.


JackUrb:
Seems these tests are flakier on travis than locally. Will update further.

PR #1104, opened 2018-08-24
add check for model version when loading opt

alexholdenmiller:
- loading an opt file will check for the model's version if the model has a `model_version` function
- `legacy:seq2seq:0` (`legacy:model:version`) is now a new shortcut type, which is written out for you if you try to load an unsupported model

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -319,6 +319,16 @@ def load_agent_module(opt):
                 new_opt[k] = v
         new_opt['model_file'] = model_file
         model_class = get_agent_module(new_opt['model'])
+        if hasattr(model_class, 'model_version'):
+            curr_version = new_opt.get('model_version', 0)
+            if curr_version != model_class.model_version():
+                raise RuntimeError(
+                    'It looks like you are trying to load an older version of '
+                    'the selected model. Change your model argument to use '
+                    'the old version from parlai/legacy_agents: for example, '
+                    '`--model legacy:model_name:{}` or '
+                    '`-m parlai.legacy_agents.model.model_v0:ModelAgent`'
+                    ''.format(curr_version))
         return model_class(new_opt)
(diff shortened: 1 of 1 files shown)


parlai/core/agents.py
```diff
@@ -319,6 +319,16 @@ def load_agent_module(opt):
...
         model_class = get_agent_module(new_opt['model'])
+        if hasattr(model_class, 'model_version'):
+            curr_version = new_opt.get('model_version', 0)
+            if curr_version != model_class.model_version():
+                raise RuntimeError(
+                    'It looks like you are trying to load an older version of '
+                    'the selected model. Change your model argument to use '
+                    'the old version from parlai/legacy_agents: for example, '
+                    '`--model legacy:model_name:{}` or '
+                    '`-m parlai.legacy_agents.model.model_v0:ModelAgent`'
```
klshuster: did you wanna do `'parlai.legacy_agents.model.model_v{}:ModelAgent'.format(curr_version)` here as well?

parlai/core/agents.py
```diff
@@ -319,6 +319,16 @@ def load_agent_module(opt):
...
         model_class = get_agent_module(new_opt['model'])
+        if hasattr(model_class, 'model_version'):
+            curr_version = new_opt.get('model_version', 0)
+            if curr_version != model_class.model_version():
+                raise RuntimeError(
+                    'It looks like you are trying to load an older version of '
+                    'the selected model. Change your model argument to use '
+                    'the old version from parlai/legacy_agents: for example, '
+                    '`--model legacy:model_name:{v}` or '
+                    '`-m parlai.legacy_agents.model.model_v{v}:ModelAgent`'
```
jaseweston: can you please put this in agents/legacy_agents rather than polluting the top level directory structure?

PR #1110, opened 2018-08-27
switch seq2seq parent class to TorchAgent

alexholdenmiller:
all seq2seq.py changes from #1097

- makes TorchAgent the parent class of Seq2seq
- updates all functions to use TorchAgent API / util functions
- does not yet break backwards compatibility

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -407,7 +407,7 @@ class Seq2seqAgent(TorchAgent):
             # save opt file
             with open(path + ".opt", 'wb') as handle:
                 # save version string
-                self.opt['seq2seq_version'] = Seq2seqAgent.VERSION
+                self.opt['model_version'] = Seq2seqAgent.VERSION
                 pickle.dump(self.opt, handle, protocol=pickle.HIGHEST_PROTOCOL)
 
     def load(self, path):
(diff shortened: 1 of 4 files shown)


alexholdenmiller:
I'll make diff on top of this which will break its backwards compatibility, but this will currently successfully load old convai2 models with the same numbers if you disable the version check

alexholdenmiller:
going to ignore lint error for now, as should be fixed when #1119 is merged into this branch

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -462,185 +293,102 @@ def share(self):
...
-        if predictions is not None:
-            PaddingUtils.map_predictions(
-                predictions, valid_inds, batch_reply, observations,
-                self.dict, self.END_IDX, report_freq=report_freq, labels=labels,
-                answers=self.answers, ys=ys.data if ys is not None else None)
-
+        except RuntimeError as e:
+            # catch out of memory exceptions during fwd/bck (skip batch)
+            if 'out of memory' in str(e):
+                print('| WARNING: ran out of memory, skipping batch. '
```
stephenroller: I wonder if there's a cleaner way. Maybe doing it in TorchAgent around the train_step call? It's such a frustratingly large try statement.

It's fine though.

PR #1113, opened 2018-08-28
TorchAgent and Fairseq fixes/enhancements

stephenroller:
- Support FastText common crawl. These are decidedly better than normal fasttext
- Default momentum to 0; using -1 causes a bug in fairseq, and 0 works just fine for TorchAgent
- Random projection should be scaled according to the [size of the output dimension](http://scikit-learn.org/stable/modules/random_projection.html#gaussian-random-projection). Otherwise it breaks the norms! This change dramatically lowers PPL in early iterations
- Use explicit padding index when vectorizing text. Otherwise breaks fairseq. Also use enhancements of torch_padded.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -143,7 +143,7 @@ class TorchAgent(Agent):
             help='Choose between pytorch optimizers. Any member of torch.optim'
                  ' should be valid.')
         agent.add_argument(
-            '-mom', '--momentum', default=-1, type=float,
+            '-mom', '--momentum', default=0, type=float,
             help='if applicable, momentum value for optimizer.')
         # preprocessing arguments
         agent.add_argument(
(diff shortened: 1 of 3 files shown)


parlai/core/torch_agent.py
```diff
@@ -533,14 +543,12 @@ def batchify(self, obs_batch, sort=False,
         xs, x_lens = None, None
         if any('text_vec' in ex for ex in exs):
             _xs = [ex.get('text_vec', self.EMPTY) for ex in exs]
-            xs, x_lens = padded_tensor(_xs)
+            xs, x_lens = padded_tensor(_xs, self.NULL_IDX, self.use_cuda)
```
stephenroller: I'd really like to add a `left_padded=True` here, so that the `xs` are left padded instead of right padded. This is more "standard" in seq2seq architectures, but I think y'all report it not mattering much?

alexholdenmiller: great!

stephenroller:
Added the left padding

stephenroller:
Manually checked that the build passed; TravisCI is having partially degraded service ATM.

PR #1114, opened 2018-08-28
Change how make_samples is done.

stephenroller:
Change how samples are created. Mostly a nitty-gritty fairseq change, but makes an internal extension easier.


alexholdenmiller:
oh it was marked hold before

stephenroller:
It doesn't pass fairseq test right now so I didn't want to merge it yet.

stephenroller:
#1179 fixes the bug that was holding me back on this, so I'll merge after travis checks.

PR #1118, opened 2018-08-28
Database State for ParlAI-MTurk

JackUrb:
Adds a bunch of SQL and a SQLite database that will back the state for ParlAI-MTurk. Database contents can be found in the logging file. Includes getters/setters, and testing as well!

### DIFF
--- a/parlai/mturk/core/mturk_data_handler.py
+++ b/parlai/mturk/core/mturk_data_handler.py
@@ -114,6 +114,7 @@ class MTurkDataHandler():
         return self.conn
 
     def _force_task_group_id(self, task_group_id):
+        '''Throw an error if a task group id is neither provided nor stored'''
         if task_group_id is None:
             task_group_id = self.task_group_id
         assert task_group_id is not None, 'Default task_group_id not set'
--- a/parlai/mturk/core/mturk_data_handler.py
+++ b/parlai/mturk/core/mturk_data_handler.py
@@ -1,0 +1,294 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+import logging
+import os
+import sqlite3
+import time
+
+import parlai.mturk.core.shared_utils as shared_utils
+from parlai.mturk.core.agents import AssignState
+
+parent_dir = os.path.dirname(os.path.abspath(__file__))
+
+# Run data table:
+CREATE_RUN_DATA_SQL_TABLE = (
+    """CREATE TABLE IF NOT EXISTS runs (
+        id string PRIMARY KEY,
+        created integer NOT NULL,
+        maximum integer NOT NULL,
+        completed integer NOT NULL,
+        failed integer NOT NULL,
+    );
+    """)
+
+# Worker data table:
+CREATE_WORKER_DATA_SQL_TABLE = (
+    """CREATE TABLE IF NOT EXISTS workers (
+        id string PRIMARY KEY,
+        accepted integer NOT NULL,
+        disconnected integer NOT NULL,
+        completed integer NOT NULL,
+        approved integer NOT NULL,
+        rejected integer NOT NULL
+    );
+    """)
+
+# HIT data table:
+CREATE_HIT_DATA_SQL_TABLE = (
+    """CREATE TABLE IF NOT EXISTS hits (
+        id string PRIMARY KEY,
+        expiration integer NOT NULL,
+        hit_status string,
+        assignments_pending int,
+        assignments_available int,
+        assignments_complete int,
+        FOREIGN KEY (run_id) REFERENCES runs (id)
+    );
+    """)
+
+# Assignment data table: (as one HIT can technically have multiple assignments)
+CREATE_ASSIGN_DATA_SQL_TABLE = (
+    """CREATE TABLE IF NOT EXISTS assignments (
+        id string PRIMARY KEY,
+        status string,
+        approve_time int,
+        FOREIGN KEY (worker_id) REFERENCES workers (id),
+        FOREIGN KEY (hit_id) REFERENCES hits (id),
+    );
+    """)
+
+# pairing data table: (reflects one worker<->assignment pairing)
+CREATE_PAIRING_DATA_SQL_TABLE = (
+    """CREATE TABLE IF NOT EXISTS pairings (
+        status string,
+        onboarding_start int,
+        onboarding_end int,
+        task_start int,
+        task_end int,
+        conversation_id string,
+        bonus_amount int,
+        bonus_text string,
+        bonus_paid boolean,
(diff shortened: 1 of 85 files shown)


parlai/mturk/core/mturk_data_handler.py
```diff
@@ -0,0 +1,474 @@
...
+            except sqlite3.Error as e:
+                shared_utils.print_and_log(
+                    logging.ERROR,
+                    "Could not get db connection, failing: {}".format(repr(e)),
+                    should_print=True)
+                raise e
+        return self.conn
+
+    def _format_pairing(self, pairing_result):
+        return {
```
klshuster: if you wanted to be fancy, you could do `dict(zip(['status', ..., 'run_id'], pairing_result))`

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -0,0 +1,474 @@
...
+            task_group_id = self.task_group_id
+        assert task_group_id is not None, 'Default task_group_id not set'
+        conn = self._get_connection()
+        c = conn.cursor()
+        c.execute('INSERT INTO runs VALUES (?,?,?,?,?);',
+                  (task_group_id, 0, target_hits, 0, 0))
+        conn.commit()
+
+    def log_hit_status(self, mturk_hit_creation_response, task_group_id=None):
+        '''Create or update an entry in the hit sattus table'''
```
klshuster: nit: 'status'

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -0,0 +1,474 @@
...
+                      (task_group_id, ))
+
+        c.execute('REPLACE INTO hits VALUES (?,?,?,?,?,?,?);',
+                  (id, expiration, status, assignments_pending,
+                   assignments_available, assignments_complete, task_group_id))
+        conn.commit()
+
+    def log_worker_accept_assignment(self, worker_id, assignment_id, hit_id,
+                                     task_group_id=None):
+        if task_group_id is None:
```
klshuster: since this is repeated a bunch, might just wanna make it a method e.g. `task_group_id = self.get_task_group_id(task_group_id)` and do the assert in there

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -0,0 +1,474 @@
...
+            c.execute('INSERT INTO workers VALUES (?,?,?,?,?,?);',
+                      (worker_id, 1, 0, 0, 0, 0))
+        else:
+            # Increment number of assignments the worker has accepted
+            c.execute(
+                'UPDATE workers SET accepted = accepted + 1 WHERE id = ?;',
+                (worker_id, )
+            )
+
+        # Ensure the assignment exsits, mark the current worker
```
klshuster: nit: `exists`

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -0,0 +1,474 @@
...
+                  (complete_type, time.time(), worker_id, assignment_id))
+
+        # Update run data to have another completed
+        c.execute('UPDATE runs SET completed = completed + 1 WHERE id = ?;',
+                  (task_group_id, ))
+        conn.commit()
+
+    def log_abandon_assignment(self, worker_id, assignment_id, approve_time,
+                               disconnect_type, task_group_id=None):
+        '''Note that an assignment was completed'''
```
klshuster: nit: `abandoned` instead of `completed`

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -0,0 +1,474 @@
...
+        conn.commit()
+
+    def get_worker_data(self, worker_id):
+        conn = self._get_connection()
+        c = conn.cursor()
+        c.execute("SELECT * FROM workers WHERE id = ?;", (worker_id, ))
+        results = c.fetchone()
+        if results is None:
+            return None
+        return {
```
klshuster: looking at https://docs.python.org/3/library/sqlite3.html#sqlite3.Row, it looks like you can set the `row_factory` to be `sqlite3.Row`, from which you can access the keys when you call `fetchone()`. might be helpful instead of these long dicts :)

klshuster: also, i noticed you returned the dict here rather than making this a `_format..` method

parlai/mturk/core/test/test_db_interactions.py
```diff
@@ -0,0 +1,547 @@
...
+            worker_id_1, assignment_id_1)
+        pair_2 = db_logger.get_worker_assignment_pairing(
+            worker_id_2, assignment_id_2)
+        pair_3 = db_logger.get_worker_assignment_pairing(
+            worker_id_2, assignment_id_3)
+        for f in ['onboarding_start', 'onboarding_end', 'task_start',
+                  'task_end', 'conversation_id']:
+            for pair in [pair_1, pair_2, pair_3]:
+                self.assertIsNone(pair[f])
+        self.assertEqual(pair_1['status'], AssignState.STATUS_NONE)
```
klshuster: why are these `STATUS_NONE` here and not `Accepted`?

parlai/mturk/core/test/test_db_interactions.py
```diff
@@ -0,0 +1,547 @@
...
+        assignment_1_data = db_logger.get_assignment_data(assignment_id_1)
+        assignment_2_data = db_logger.get_assignment_data(assignment_id_2)
+        assignment_3_data = db_logger.get_assignment_data(assignment_id_3)
+        self.assertEqual(assignment_1_data['assignment_id'], assignment_id_1)
+        self.assertEqual(assignment_1_data['status'], 'Reviewable')
+        self.assertIsNotNone(assignment_1_data['approve_time'])
+        self.assertEqual(assignment_1_data['worker_id'], worker_id_1)
+        self.assertEqual(assignment_1_data['hit_id'], HIT1['HIT']['HITId'])
+        self.assertEqual(assignment_2_data['assignment_id'], assignment_id_2)
+        self.assertEqual(assignment_2_data['status'], 'Reviewable')
```
klshuster: so are abandoned HITs now reviewable?

parlai/mturk/core/mturk_data_handler.py
JackUrb: Yeah I think I can make the changes to have the row_factory solution work, thanks for finding this!

parlai/mturk/core/test/test_db_interactions.py
JackUrb: Pairs track the assign state from the perspective of MTurkManager, where `STATUS_NONE` is essentially the valid state of `"Accepted"`. These should soon transition to `STATUS_ONBOARDING` or `STATUS_WAITING`, which is the more relevant info to track in a pairing.

parlai/mturk/core/test/test_db_interactions.py
JackUrb: In the long run yes, but in the short term not yet. I suppose I should do that transition when the time comes (basic idea is the only HIT that will not be Reviewable is one that was explicitly abandoned, but disconnects can still submit HITs because we'll have bonus per interaction).

PR #1119, opened 2018-08-28
new version of seq2seq

alexholdenmiller:
- primary change is massive updates to further modularize and clean out modules.py

- successfully trains & evals & ranks on babi:task10k:1
- successfully trains & evals & ranks on convai2

### DIFF
--- a/parlai/agents/seq2seq/modules.py
+++ b/parlai/agents/seq2seq/modules.py
@@ -125,7 +125,7 @@ class Seq2seq(nn.Module):
             # TODO: do we need to do this? actually shouldn't need to since we
             # don't do input feeding
             for i in range(seqlen):
-                xi = xs.select(1, i)
+                xi = xs.select(1, i).unsqueeze(1)
                 output, hidden = self.decoder(xi, hidden, attn_params)
                 score = self.output(output)
                 scores.append(score)
@@ -404,14 +404,14 @@ class RNNDecoder(nn.Module):
 
         if self.attn_time == 'pre':
             # modify input vectors with attention
-            xes = self.attention(xes, hidden, attn_params)
+            xes, _attw = self.attention(xes, hidden, attn_params)
 
         # feed tokens into rnn
         output, new_hidden = self.rnn(xes, hidden)
 
         if self.attn_time == 'post':
             # modify output vectors with attention
-            output = self.attention(output, new_hidden, attn_params)
+            output, _attw = self.attention(output, new_hidden, attn_params)
 
         return output, new_hidden
 
@@ -617,7 +617,7 @@ class AttentionLayer(nn.Module):
             # calculate activation scores, apply mask if needed
             if attn_mask is not None:
                 # remove activation from NULL symbols
-                import pdb; pdb.set_trace()  # is this the best operation?
+                # TODO: is this the best operation?
                 attn_w_premask -= (1 - attn_mask) * 1e20
             attn_weights = F.softmax(attn_w_premask, dim=1)
 
@@ -627,8 +627,5 @@ class AttentionLayer(nn.Module):
         merged = torch.cat((xes.squeeze(1), attn_applied.squeeze(1)), 1)
         # combine them with a linear layer and tanh activation
         output = F.tanh(self.attn_combine(merged).unsqueeze(1))
-        print('make sure last_hidden == xes')
-        import pdb; pdb.set_trace()
-        # TODO: remove tanh?
 
         return output, attn_weights
--- a/parlai/agents/seq2seq/modules.py
+++ b/parlai/agents/seq2seq/modules.py
@@ -13,6 +13,17 @@ from torch.nn.utils.rnn import pad_packed_sequence, pack_padded_sequence
 import torch.nn.functional as F
 
 
+def opt_to_kwargs(opt):
+    """Get kwargs for seq2seq from opt."""
+    kwargs = {}
+    for k in ['numlayers', 'dropout', 'bidirectional', 'rnn_class',
+              'lookuptable', 'decoder', 'numsoftmax',
+              'attention', 'attention_length', 'attention_time']:
+        if k in opt:
+            kwargs[k] = opt[k]
+    return kwargs
+
+
 def pad(tensor, length, dim=0):
     """Pad tensor to a specific length.
 
@@ -30,17 +41,6 @@ def pad(tensor, length, dim=0):
             dim=dim)
     else:
         return tensor
-
-
-def opt_to_kwargs(opt):
-    """Get kwargs for seq2seq from opt."""
-    kwargs = {}
-    for k in ['numlayers', 'dropout', 'bidirectional', 'rnn_class',
-              'lookuptable', 'decoder', 'numsoftmax',
-              'attention', 'attention_length', 'attention_time']:
-        if k in opt:
-            kwargs[k] = opt[k]
-    return kwargs
 
 
 class Seq2seq(nn.Module):
@@ -73,8 +73,7 @@ class Seq2seq(nn.Module):
             numlayers=numlayers, dropout=dropout,
             attn_type=attention, attn_length=attention_length,
             attn_time=attention_time,
-            bidir_input=bidirectional,
-            numsoftmax=numsoftmax)
+            bidir_input=bidirectional)
 
         shared_lt = (self.decoder.lt
                      if lookuptable in ('enc_dec', 'all') else None)
@@ -82,15 +81,15 @@ class Seq2seq(nn.Module):
         self.encoder = RNNEncoder(
             num_features, embeddingsize, hiddensize,
             padding_idx=self.NULL_IDX, rnn_class=rnn_class,
-            num_layers=numlayers, dropout=dropout,
+            numlayers=numlayers, dropout=dropout,
             bidirectional=bidirectional,
             shared_lt=shared_lt, shared_rnn=shared_rnn)
 
         shared_weight = (self.decoder.lt
                          if lookuptable in ('dec_out', 'all') else None)
         self.output = OutputLayer(
-            num_features, embeddingsize, hiddensize, numsoftmax=numsoftmax,
-            shared_weight=shared_weight)
+            num_features, embeddingsize, hiddensize, dropout=dropout,
+            numsoftmax=numsoftmax, shared_weight=shared_weight)
 
     def _encode(self, xs, prev_enc=None):
         if prev_enc is not None:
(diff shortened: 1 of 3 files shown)


alexholdenmiller:
trains / evals successfully on convai2, including ranking

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -38,17 +38,17 @@ class Seq2seqAgent(TorchAgent):
...
 
-    @staticmethod
-    def add_cmdline_args(argparser):
+    @classmethod
+    def add_cmdline_args(cls, argparser):
         """Add command-line arguments specifically for this agent."""
         agent = argparser.add_argument_group('Seq2Seq Arguments')
         agent.add_argument('--init-model', type=str, default=None,
-                           help='load dict/features/weights/opts from this file')
+                           help='load dict/model/opts from this path')
```
uralik: I really like the way when we have automatic default in the help with `%(default)s`, prolly we can fix that in seq2seq now

alexholdenmiller: yes let's do separate diff for those, I also want to reformat the whitespace on all the args

parlai/agents/seq2seq/modules.py
```diff
@@ -5,17 +5,34 @@
...
+
+    :param tensor: vector to pad
+    :param length: new length
+    :param dim: (default 0) dimension to pad
+
+    :returns: padded tensor if the tensor is shorter than length
+    """
     if tensor.size(dim) < length:
         return torch.cat(
             [tensor, tensor.new(*tensor.size()[:dim],
```
stephenroller: Just out of pure curiosity, did you benchmark this with other approaches? (namely init the full zero matrix, then assign the tensor into it?). Not that there's need to.

stephenroller: Just out of pure curiosity, did you benchmark this with other approaches? (namely init the full zero matrix, then assign the tensor into it?). Not that there's need to.

stephenroller: Also, should this actually fill_(null_idx)?

parlai/agents/seq2seq/modules.py
```diff
@@ -27,277 +44,291 @@ def pad(tensor, length, dim=0):
...
-        xs -- input to the encoder
-        ys -- expected output from the decoder
-        cands -- set of candidates to rank, if applicable
-        cand_indices -- indices to match candidates with their appropriate xs
-        prev_enc -- if you know you'll pass in the same xs multiple times and
-            the model is in eval mode, you can pass in the encoder output from
-            the last forward pass to skip recalcuating the same encoder output
-        rank_during_training -- (default False) if set, ranks any available
-            cands during training as well
+        :param xs:          (bsz x seqlen) LongTensor input to the encoder
```
stephenroller: <3

parlai/agents/seq2seq/modules.py
```diff
@@ -27,277 +44,291 @@ def pad(tensor, length, dim=0):
...
+            return prev_enc
+        else:
+            return self.encoder(xs)
+
+    def _starts(self, bsz):
+        """Return bsz start tokens."""
+        return self.START.detach().expand(bsz, 1)
+
+    def _decode_forced(self, ys, encoder_states):
+        """Decode with teacher forcing."""
```
stephenroller: I don't understand what this means

parlai/agents/seq2seq/modules.py
```diff
@@ -27,277 +44,291 @@ def pad(tensor, length, dim=0):
...
-        if isinstance(scores, list):
-            scores = torch.cat(scores, 1)
-
-        return predictions, scores, cand_preds, cand_scores, encoder_states, nbest_beam_preds, nbest_beam_scores
-
-
-class Encoder(nn.Module):
-    def __init__(self, num_features, padding_idx=0, rnn_class='lstm',
-                 emb_size=128, hidden_size=128, num_layers=2, dropout=0.1,
+            scores = self._decode(encoder_states, maxlen or self.longest_label)
```
stephenroller: All much cleaner now, great!

parlai/agents/seq2seq/modules.py
```diff
@@ -27,277 +44,291 @@ def pad(tensor, length, dim=0):
...
+
+    def __init__(self, num_features, embeddingsize, hiddensize,
+                 padding_idx=0, rnn_class='lstm', numlayers=2, dropout=0.1,
                  bidirectional=False, shared_lt=None, shared_rnn=None,
                  sparse=False):
+        """Initialize recurrent encoder."""
         super().__init__()
 
         self.dropout = nn.Dropout(p=dropout)
-        self.layers = num_layers
```
stephenroller: Why change all the names?

parlai/agents/seq2seq/modules.py
```diff
@@ -309,278 +340,199 @@ def forward(self, xs):
...
-                hidden = (hidden[0].view(-1, self.dirs, bsz, self.hsz).max(1)[0],
-                          hidden[1].view(-1, self.dirs, bsz, self.hsz).max(1)[0])
+                hidden = (hidden[0].view(-1, self.dirs, bsz, self.hsz).sum(1),
+                          hidden[1].view(-1, self.dirs, bsz, self.hsz).sum(1))
             else:
-                hidden = hidden.view(-1, self.dirs, bsz, self.hsz).max(1)[0]
+                hidden = hidden.view(-1, self.dirs, bsz, self.hsz).sum(1)
 
-        return encoder_output, hidden
+        return hidden, encoder_output, attn_mask
```
stephenroller: intentional swap?

parlai/agents/seq2seq/modules.py
```diff
@@ -599,49 +551,81 @@ def __init__(self, attn_type, hidden_size, emb_size, bidirectional=False,
...
-                    torch.bmm(hid, enc_out.transpose(1, 2)).squeeze(1))
-            # calculate activation scores
+                enc_t = enc_out.transpose(1, 2)
+                attn_w_premask = torch.bmm(hid, enc_t).squeeze(1)
+
+            # calculate activation scores, apply mask if needed
             if attn_mask is not None:
                 # remove activation from NULL symbols
+                # TODO: is this the best operation?
                 attn_w_premask -= (1 - attn_mask) * 1e20
```
stephenroller: Could do `th.masked_fill_(attn_w_premask, -1e20)`. That's what I've seen in fairseq.

Maybe make that `1e20` as a private constant, like `_ALMOST_INFINITY`? Don't think it matters.

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -312,12 +312,12 @@ def train_step(self, batch):
...
         self.model.train()
         self.zero_grad()
-        preds = None
         try:
             out = self.model(batch.text_vec, batch.label_vec)
 
             # generated response
-            preds, scores = out[0], out[1]
+            scores = out[0]
+            preds = scores.max(2)[1]
```
stephenroller: nit: i prefer `_, preds = scores.max(2)`

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -382,12 +381,13 @@ def eval_step(self, batch):
...
-        if cand_preds is not None:
+        if cand_scores is not None:
+            cand_preds = cand_scores.sort(1, True)[1]
             # now select the text of the cands based on their scores
-            cand_choices = self._pick_cands(cand_preds, cand_inds,
+            cand_choices = self._pick_cands(cand_preds, cand_params[1],
                                             batch.candidates)
 
-        text = [self._v2t(p) for p in preds]
+        text = [self._v2t(p) for p in scores.max(2)[1].cpu()]
```
stephenroller: IMHO hiding this max() call in here is a bit nefarious.

```python
_, preds = scores.max(2)
text = [self._v2t(p) for p in preds.cpu()]
```

parlai/core/torch_agent.py
```diff
@@ -775,7 +775,7 @@ def _init_cuda_buffer(self, model, criterion, batchsize, maxlen):
             try:
                 print('preinitializing pytorch cuda buffer')
                 dummy = torch.ones(batchsize, maxlen).long().cuda()
-                sc = model(dummy, dummy)[1]
+                sc = model(dummy, dummy)[0]
```
stephenroller: this needs an explanation. It's assuming model has a particular interface and that interface isn't clear to me right now.

parlai/agents/seq2seq/modules.py
stephenroller: nit TBH I think encoder_output, hidden, attn_mask is better.

parlai/core/torch_agent.py
stephenroller: (Assuming this is related to the hidden, encoder_out, attn_mask from earlier)

parlai/agents/seq2seq/modules.py
uralik: _decode_forced is putting tokens from the target as the next input to estimate the target score during training or during eval PPL computation

alexholdenmiller:
Note: I squashed old pretrained seq2seq module weights into the new modules and got exactly the same f1, ppl, hits on convai2.

parlai/agents/seq2seq/modules.py
alexholdenmiller: didn't benchmark, although 1) this is copied from pytorch source, with `dim` param added by me and 2) only used for candidate score padding so I'm not super concerned about the perf :)  oh I'll update with null_idx

parlai/agents/seq2seq/modules.py
alexholdenmiller: just for consistency, there was a mix of underscores and non and things didn't match with the opt args

parlai/agents/seq2seq/modules.py
alexholdenmiller: swap was intentional because later
```
hidden = encoder_states[0]
attn_params = (encoder_states[1], encoder_states[2])
```
but I could change that to [1] and ([0], [2])

it does match the rnn output more closely

parlai/agents/seq2seq/modules.py
alexholdenmiller: `self._YUGE`?

parlai/agents/seq2seq/seq2seq.py
alexholdenmiller: actually introduced a bug too because I used the teacher-forced scores on accident from the if block, I'll update

parlai/core/torch_agent.py
alexholdenmiller: agreed, I need to find a better way to extract the scores here or move it back to s2s

parlai/agents/seq2seq/modules.py
alexholdenmiller: "teacher-forcing" always inputs the true token at each time step, vs using the model's predicted

Issue #923, opened 2018-06-27
FairSeq unit tests


stephenroller:
Fairseq currently doesn't have any unit tests -- it would be a good idea to add them.

PR #1122, opened 2018-08-29
Integration test tasks

stephenroller:
Two simple tasks (repeat the input) which can test both ranking and generative models. The idea is small models should be able to learn these tasks in under a minute on CPU.

Actual tests incoming.


stephenroller:
Ready for review. Fixes #923 for all intents and purposes.

tests/test_torch_agent.py
```diff
@@ -95,6 +101,7 @@ def test_share(self):
         shared = agent.share()
         self.assertTrue('dict' in shared)
 
+    @unittest.skipIf(SKIP_TESTS, "Torch not installed.")
```
alexholdenmiller: ooh cool

PR #1126, opened 2018-08-29
fully isolate seq2seq legacy code from rest of parlai

alexholdenmiller:
- move all dependencies from the rest of parlai in legacy seq2seq to a copy in the legacy folder


alexholdenmiller:
just freezing all the code for this legacy model for reproducibility

alexholdenmiller:
training from scrach on babi:task10k:1 and convai2 scripts work with same old results

alexholdenmiller:
I'm not going to fix the lint errors :)

stephenroller:
party!

PR #1129, opened 2018-08-30
seq2seq bug fix

alexholdenmiller:


alexholdenmiller:
Added code to output very negative scores for the padding index. This is useful for weight tying especially because it prevents any gradient from getting through and changing the embedding of the padding token if weight tying with the output layer is enabled.

Time measurements
```
21.7s / 21.4s / 22.0s  (with dec_out weight sharing, and blocked scores)
21.8s / 23.2s / 21.5s  (block scores/gradient for NULL, no shared weight)
25.3s / 23.1s / 23.0s  (regular forward)
```

Looks like non-stat-sig difference for enabling/disabling the gradient / wiping out the scores for NULL index as far as timing goes, if anything slightly faster.

alexholdenmiller:
ready for review

alexholdenmiller:
added NEAR_INF to utils.py, signed up for around 150+ / 50- line changes for the linter in utils.py :)

alexholdenmiller:
will merge after checks complete if they pass

PR #1135, opened 2018-09-05
Move MemNN to TorchAgent

alexholdenmiller:
- move memnn parent to torch agent, remove tons of unnecessary or overly complex code
- currently removed the generator, I'll add it back in later
- now it ranks on the batch of labels instead of label_candidates field during training
- now reports mean_rank and loss, no longer calculates hits@1 during training

speed tests with bsz 32, esz 128, time features on, 3 hops on babi:task10k:1
~24x faster using 1 thread on the GPU
~8x faster using 1 thread on the CPU
~1.7x faster using 30 threads on the CPU

(not going to land until I add model versioning code and back current code up to legacy)

### DIFF
--- a/parlai/agents/memnn/memnn.py
+++ b/parlai/agents/memnn/memnn.py
@@ -9,7 +9,7 @@ from parlai.core.thread_utils import SharedTable
 from parlai.core.utils import round_sigfigs, padded_tensor
 
 import torch
-from torch.nn import CrossEntropyLoss
+from torch import nn
 
 import os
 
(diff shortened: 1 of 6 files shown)


alexholdenmiller:
interestingly: the model gets >75% accuracy on babi task 3 without positional encoding, whereas in the original paper https://arxiv.org/abs/1503.08895 the best reported is 29%.

alexholdenmiller:
TODO: add model_version() and make torchagent.save() save the model version

alexholdenmiller:
TODO done

alexholdenmiller:
ready for review

PR #1136, opened 2018-09-05
Db logging when using ParlAI MTurk

JackUrb:
Currently gated behind a flag b/c still experimental without much gain (until we have the tooling), but this adds the database logging functionality to MTurk, which tracks HIT status as well as some worker stats. 

Updates some tests as I changed a bit of the data_handler api. Also fixes threading for the DB and adds locks.

DB works so far though:
screen shot 2018-09-05 at 3 29 31 pm

screen shot 2018-09-05 at 3 29 47 pm

As a bonus, this now properly exits when all hits are confirmed as expired or completed, and doesn't need to wait to "expire all hits" for a forever.


emilydinan:
this looks amazing! quick question -- why is the first table called 'pairings'?

JackUrb:
Those specifically track a unique pairing between a worker and an assignment. Some assignment ids can be re-distributed to new workers. This is essentially what matches the "agent" semantics in the PMT codebase vs a "worker" or an "assignment"

parlai/mturk/core/mturk_manager.py
```diff
@@ -1354,7 +1393,13 @@ def create_hits(self, qualifications=None):
     def get_hit(self, hit_id):
         """Get hit from mturk by hit_id"""
         client = mturk_utils.get_mturk_client(self.is_sandbox)
-        return client.get_hit(HITId=hit_id)
+        hit = client.get_hit(HITId=hit_id)
+        if self.db_logger is not None:
+            try:
+                self.db_logger.log_hit_status(hit)
+            except Exception:
```
klshuster: when does this fail?

klshuster: got it, thanks

PR #1139, opened 2018-09-07
Support for preallocated buffers in padded_tensor.

stephenroller:
Small speedup for anything that uses TorchAgent by preallocating memory for inputs/labels, and reusing that memory for each batch_act.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -198,6 +198,12 @@ class TorchAgent(Agent):
         if opt.get('numthreads', 1) > 1:
             torch.set_num_threads(1)
 
+        # initialize buffers for calls to vectorize()
+        # note that the second dimension is just a guess, and the tensor will
+        # be resized if necessary
+        self.xs_buffer = torch.LongTensor(opt['batchsize'], opt.get('truncate') or 64)
+        self.ys_buffer = torch.LongTensor(opt['batchsize'], opt.get('truncate') or 64)
+
         # check for cuda
         self.use_cuda = not opt['no_cuda'] and torch.cuda.is_available()
         if self.use_cuda:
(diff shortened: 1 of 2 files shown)


alexholdenmiller:
were you able to measure a speedup? I use to use more buffers like this in the past and myle suggested it might not make a huge difference. it ended up getting removed in a refactor because those buffers were kind of messy, but I'm happy with the cleanliness of these.

stephenroller:
small, in my custom internal-only setup.  I'll do some more formal benchmarking

stephenroller:
Interestingly, it seemed to provide a negative impact:

invocation:

before:
```
[ num_epochs completed:1.0 time elapsed:755.9537072181702s ]
...
test:{'exs': 7801, 'accuracy': 0, 'f1': 0.135, 'bleu': 0.001643, 'loss': 4.049, 'ppl': 57.32}
python projects/convai2/baselines/seq2seq/train.py -eps 1 -mf /tmp/convtest    872.39s user 98.40s system 88% cpu 18:17.10 total
```

after:
```
[ num_epochs completed:1.0 time elapsed:764.8831079006195s ]
...
test:{'exs': 7801, 'accuracy': 0, 'f1': 0.1277, 'bleu': 0.0009912, 'loss': 4.07, 'ppl': 58.56}
python projects/convai2/baselines/seq2seq/train.py -eps 1 -mf /tmp/convtest    806.42s user 90.46s system 88% cpu 16:48.37 total
```

Official pytorch position is something along the lines of "it can help in some situations, but it's probably not worth it because the code is less clean anyway." The reason that it probably *hurts* in this case, is due to the fact that padded_tensor writes to each chunk line by line. If it, instead, created a fresh CPU tensor and then did the copy into the buffer all at once, it'd probably help/see no change.

That actually makes me wonder about whether the current call to `items[0].new(n, t)`

stephenroller:
Confirmed the `.new` creates a CPU tensor which is more efficient.

PR #1146, opened 2018-09-10
input dropout (UNK masking) in torch agent

uralik:
This function should be called appropriately where developer wants it to be called.
We can call it always in train_step of batch_act if-else branch if you want, but I would leave it for developer, i.e. to be called from {train, eval}_step functions


stephenroller:
I'm cautious about adding more and more utility methods into `TorchAgent` that isn't directly required or needed, making it unapproachably large. Is it possible to make this into a `nn.Module` layer which users could just insert in front of their `Embedding`?

alexholdenmiller:
I like the idea of this as an `nn.Module`, it'll probably run faster too.

alexholdenmiller:
I implemented the change and put it just in seq2seq for now since it's an nn.Module, we might want to make some kind of module repository? or can just import from other agents

alexholdenmiller:
ready for review

parlai/agents/seq2seq/modules.py
```diff
@@ -278,13 +281,39 @@ def forward(self, xs, ys=None, cand_params=None, prev_enc=None,
...
+        """
+        super().__init__()
+        self.unknown_idx = unknown_idx
+        self.prob = probability
+
+    def forward(self, input):
+        """If training and dropout rate > 0, masks input with unknown token."""
+        if self.training and self.prob > 0:
+            mask = input.new(input.size()).float().uniform_(0, 1) < self.prob
+            input.masked_fill_(mask, self.unknown_idx)
```
stephenroller: Did you happen to check whether the destructive fill could have problems here? I just wonder, if `text_vec` is cached (as in `-bsrt True`) that the `text_vec` could be thrashed and filled with UNKs.

Reason this shouldn't happen: We're applying this to the output of `padded_tensor`, which produces a new tensor every iteration, and just copies over from the (possibly cached) `text_vec`.

parlai/agents/seq2seq/modules.py
```diff
@@ -278,13 +281,39 @@ def forward(self, xs, ys=None, cand_params=None, prev_enc=None,
...
+
+
 class RNNEncoder(nn.Module):
     """RNN Encoder."""
 
     def __init__(self, num_features, embeddingsize, hiddensize,
                  padding_idx=0, rnn_class='lstm', numlayers=2, dropout=0.1,
                  bidirectional=False, shared_lt=None, shared_rnn=None,
-                 sparse=False):
+                 input_dropout=0, unknown_idx=4, sparse=False):
```
stephenroller: This magic `4` default seems really dangerous. Given that it depends on the actual dict, I feel like it would be safer to have a `None` default, and then throw a `RuntimeException` if `input_dropout != 0`. People who hope to just reuse this module without the `input_dropout` don't have to worry about it, but then people who DO want that feature are explicitly forced to correctly identify their `UNK` token.

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -211,6 +214,7 @@ def _init_model(self, states=None):
         self.model = Seq2seq(
             len(self.dict), opt['embeddingsize'], opt['hiddensize'],
             padding_idx=self.NULL_IDX, start_idx=self.START_IDX,
+            unknown_idx=self.dict[self.dict.unk_token],
```
stephenroller: good

parlai/agents/seq2seq/modules.py
stephenroller: Or at least leave a comment explaining why it's 4.

alexholdenmiller: yeah good call

PR #1147, opened 2018-09-10
Optional logging, fetch notices

JackUrb:
With the last wave of MTurk botting, it would've been good to put in some notices that can cause an interruption for people using the ParlAI-MTurk platform. This adds a call to our website to request potential notices for live or sandbox runs. 

As I'm expanding the platform to have more features, it'll be useful to get some new metrics as well about disconnect rates and such. This PR also introduces optionally logging worker statistics on runs. Users can opt in or snooze the warning for 2 weeks.

### DIFF
--- a/parlai/mturk/core/mturk_manager.py
+++ b/parlai/mturk/core/mturk_manager.py
@@ -13,7 +13,6 @@ import time
 import uuid
 import errno
 import requests
-import json
 
 from parlai.mturk.core.agents import AssignState
 from parlai.mturk.core.socket_manager import Packet, SocketManager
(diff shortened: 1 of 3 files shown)


parlai/mturk/core/mturk_manager.py
```diff
@@ -153,6 +161,51 @@ def _init_logging_config(self):
...
+            'allowed': selected in ['y', 'Y'],
+            'asked_time': time.time()
+        }
+
+        with open(PARLAI_MTURK_LOG_PERMISSION_FILE, 'wb+') as perm_file:
+            pickle.dump(permissions, perm_file)
+            return permissions['allowed']
+
+    def _upload_worker_data(self):
+        '''Uploads worker data acceptance and completion rates to the parlai
```
emilydinan: nit: double quotation marks here instead of single

PR #1154, opened 2018-09-12
Baselines (random and IR baseline) can use a given set of label candidates

jaseweston:


parlai/agents/ir_baseline/ir_baseline.py
```diff
@@ -139,20 +147,24 @@ def observe(self, obs):
...
         reply = {}
         reply['id'] = self.getID()
         
         # Rank candidates
+        cands = None
         if 'label_candidates' in obs and len(obs['label_candidates']) > 0:
-            # text = obs['text']
+            cands = obs['label_candidates']
+        if hasattr(self, 'label_candidates'):
+            cands = self.label_candidates
```
alexholdenmiller: can you add `# override label candidates with candidate file if set` to make it obvious the ordering is intentional?

PR #1157, opened 2018-09-13
--display-ignore-fields in display_model.

stephenroller:


examples/display_model.py
```diff
@@ -25,6 +25,7 @@ def main():
     # Get command line arguments
     parser = ParlaiParser(True, True)
     parser.add_argument('-n', '--num-examples', default=10)
+    parser.add_argument('--display-ignore-fields', type=str, default='agent_reply')
```
jaseweston: group arguments?

alexholdenmiller:
wait won't that default remove the model's predictions from the display?

stephenroller:
@alexholdenmiller no

but i'll remove the default

https://github.com/facebookresearch/ParlAI/blob/728b763d70a1eff5aa190f492d30bff83eee68c0/parlai/core/utils.py#L791-L800

examples/display_model.py
stephenroller: ?

alexholdenmiller:
https://github.com/facebookresearch/ParlAI/blob/728b763d70a1eff5aa190f492d30bff83eee68c0/parlai/core/utils.py#L778-L781

doesn't that skip the agent?

stephenroller:
I've put the argument in, and removed the `agent_reply` bug; then linted. It's ready for a second review.

stephenroller:
Rebased, waiting on tests to merge.

PR #1159, opened 2018-09-14
Adding python3 shebang to all of the python files

JackUrb:
Adding this shebang to all of the files means that the internal sync script will have 1/3 the number of linter errors, which is a step in the direction of getting ParlAI to not be broken internally. Externally it labels all the files as being written for python3, (though we were already pretty clear on that).

### DIFF
--- a/parlai/tasks/convai2/__init__.py
+++ b/parlai/tasks/convai2/__init__.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 # Copyright (c) 2017-present, Facebook, Inc.
 # All rights reserved.
 # This source code is licensed under the BSD-style license found in the
--- a/examples/base_train.py
+++ b/examples/base_train.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 # Copyright (c) 2017-present, Facebook, Inc.
 # All rights reserved.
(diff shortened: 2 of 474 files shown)


stephenroller:
Trying to match me for codemods eh?

parlai/tasks/convai2/__init__.py
```diff
@@ -1,5 +1,7 @@
-# Copyright (c) 2017-present, Facebook, Inc.
-# All rights reserved.
-# This source code is licensed under the BSD-style license found in the
```
stephenroller: what happened with this one?

stephenroller:
Fix the copyright errors, double check that one that seems like it has stuff deleted, then it will be good

parlai/tasks/convai2/__init__.py
JackUrb: Especially when the script literally just read the contents of a file and then rewrote them. Perhaps this file was encoded weirdly?

JackUrb:
Obviously I'm just competing to see who can alter more files in one commit for a passable reason 🤣

Though I'm going to skip lint on this one, the next person to touch these can deal with those

stephenroller:
Looks good

PR #1162, opened 2018-09-18
Adding ability to launch directly to heroku hobby servers

JackUrb:
This argument (`--hobby`) lets runs launch directly to the hobby tier rather than needing to transfer manually. Tested by launching with and without flag, and without access to hobby tiers.


JackUrb:
Failed lint on params, which I'm choosing not to be responsible for

emilydinan:
this is awesome

PR #1167, opened 2018-09-20
Offensive language detector update

emilydinan:
- spicy butter is no longer considered offensive 😢 
- better display for `detect_offensive_language` script


alexholdenmiller:
best ParlAI diff of 2018?

PR #1176, opened 2018-09-26
Remove a few double python shebangs.

stephenroller:


stephenroller:
Ignoring lint, it's fixed in another PR.

PR #1182, opened 2018-09-28
Organize tests

stephenroller:
A little bit of organization around tests.

Now tests are run with `python setup.py test`, which is a bit more standard in the python community. This only runs the travis-friendly tests by default. 

Travis tests include running mturk tests if any mturk files have changed. To implement this, and other tests, I used a python module for git. Users will need to run `pip install gitpython`.

Full, lengthy integration tests can be run with `python setup.py test -s tests.suite.full`

A few tests are refactored:
- TFIDF uses the skipIf pattern we're already using for TorchAgent
- Copyright and __init__.py tests are moved to a unified test class.


tests/test_code.py
```diff
@@ -0,0 +1,72 @@
...
+                    '{} missing copyright "{}"'.format(fn, msg)
+                )
+
+
+class TestInit(unittest.TestCase):
+    """Make sure all python packages have init files."""
+    def test_init_everywhere(self):
+        for folder in utils.git_ls_dirs('parlai'):
+            if os.path.basename(folder) == 'html':
+                continue
```
alexholdenmiller: this removed the actual test, right?

stephenroller: good catch

stephenroller:
Fix @alexholdenmiller 's comment and rebase (fixing merge conflict)

tests/test_code.py
alexholdenmiller: `if not root.endswith('__pycache__'):` is still missing right?  could put that into the util function since I don't think we'd ever want those directories included in a unit test...

stephenroller: No, because this only looks at files that are covered by git, and that's in .gitignore.

alexholdenmiller: oh super nice

PR #1183, opened 2018-09-28
Autogenerate docs for each of the CLI scripts we distribute.

stephenroller:
This isn't much more than just calling `--help` with each one. It's not the prettiest now, so maybe we should discuss it.

Also fix a bug with the help string of one script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ instance/
 docs/_build/
 docs/source/task_list.inc
 docs/source/zoo_list.inc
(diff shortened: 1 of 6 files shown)


stephenroller:
So the newer commit does include the info from the main docstring into the code. I wonder if we could adjust it to use an autogenerate... I'll play with it before merging.

stephenroller:
Yes, using autodoc + CLI usage works!

stephenroller:
Rebase for conflict.

PR #1185, opened 2018-10-01
Don't use a fixed CLI description.

stephenroller:


stephenroller:
Lint issues fixed by #1186, we can merge that first.

PR #1188, opened 2018-10-01
Ngram block for beam search, blocks all repeating ngrams up to n

uralik:
Simply counts ngrams of current hypothesis and blocks whatever was repeated. Solves local repetition issue quite well. Be default we do not block anything. 

PEP styles should be OK

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -31,7 +31,7 @@ except ImportError as e:
 
 
 from torch import optim
-from collections import deque, namedtuple
+from collections import deque, namedtuple, Counter
 import pickle
 import random
 import math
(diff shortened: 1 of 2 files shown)


parlai/core/torch_agent.py
```diff
@@ -897,6 +897,11 @@ def __init__(self, beam_size, min_length=3, padding_token=0, bos_token=1,
         self.eos_top_ts = None
         self.n_best_counter = 0
         self.min_n_best = min_n_best
+        self.block_ngram = block_ngram
+
+    @staticmethod
+    def find_ngrams(input_list, n):
```
stephenroller: wouldn't mind a proper docstring here, but (y)

parlai/core/torch_agent.py
```diff
@@ -924,6 +929,20 @@ def advance(self, softmax_probs):
...
+                if self.block_ngram > 0:
+                    current_ngrams = []
+                    for ng in range(self.block_ngram):
+                        ngrams = Beam.find_ngrams(current_hypo, ng)
+                        if len(ngrams) > 0:
+                            current_ngrams.extend(ngrams)
+                    counted_ngrams = Counter(current_ngrams)
+                    if any(v > 1 for k, v in counted_ngrams.items()):
+                        # block this hypothesis hard
+                        beam_scores[i] = -1e20
```
stephenroller: use `parlai.core.utils.NEAR_INF`

parlai/core/torch_agent.py
uralik: added

parlai/core/torch_agent.py
uralik: fixed

PR #1190, opened 2018-10-01
Autogenerated docs for model zoo (incomplete)

stephenroller:
As discussed, generates a list based on `parlai.zoo.model_list`.

However, model_list could use some improvements at this time.


jaseweston:
nice!

PR #1196, opened 2018-10-02
[wip] Test parlai.zoo.model_list

stephenroller:
Similar to parlai.tasks.task_list. One of the doc day changes.

Marked as WIP because the test currently doesn't pass: we need to add documentation.


emilydinan:
nice! why are the word embedding files not included in the model zoo list?

PR #1199, opened 2018-10-02
Reorganize how TorchAgent does vectorize/batchify.

stephenroller:
This also serves as an example of how to extend a TorchAgent for some custom fields now, and how specialize vectorization for just a particular class.

Thoughts?


stephenroller:
Adjusted and ready for review.

emilydinan:
nice! why are the word embedding files in the zoo not in the list?

PR #1201, opened 2018-10-03
[WIP] Add READMEs

emilydinan:
READMEs for core directory and several agents:
- language model


emilydinan:
latest change also makes a small change to the `--no-cuda` arg in torch agent. make this arg `store_true` to be consistent with flag elsewhere and also with the tutorial. cc @alexholdenmiller

stephenroller:
I'd feel more comfortable if you pulled the --no-cuda thing into a separate PR but I certainly won't hold this PR over that.

PR #1207, opened 2018-10-05
Unigram Agent

stephenroller:
Simple Unigram Agent, which just outputs the $n$ most common unigrams every time.

This code is *just* simple enough to make an excellent reference model, so I also made it a first class citizen in the autodocs.

### DIFF
--- a/parlai/agents/unigram/README.md
+++ b/parlai/agents/unigram/README.md
@@ -1,0 +1,13 @@
+# Unigram Agent
+
+Baseline model which predicts the $n$ most common unigrams regardless of input.
+
+## Examples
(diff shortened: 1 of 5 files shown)


stephenroller:
Added a README for Jason. :)

parlai/agents/unigram/README.md
```diff
@@ -0,0 +1,13 @@
+# Unigram Agent
+
+Baseline model which predicts the $n$ most common unigrams regardless of input.
+
+## Examples
```
emilydinan: can you change to the format of the other READMEs

PR #1210, opened 2018-10-05
Deprecate FDT Batch Sort in favor of PDT Batch Sort (on the fly)

klshuster:
**(EDIT: Ready to review)**

This PR will contain elements of deprecating the current batch sort functionality in `FixedDataTeacher` and will simplify the use of batch sort while using PytorchDataTeacher.

Steps necessary before review:

- [x] Simplify parameters for batch sorting in `PytorchDataTeacher`
- [x] Simplify data storage scheme when using the `PytorchDataTeacher`
- [x] Remove references to batch sort in `FixedDataTeacher`
- [x] Figure out appropriate serialization of examples so that the PytorchDataTeacher is 1. Fast and 2. Works
- [x] In line with second point, make sure that using `--pytorch-preprocess` to preprocess observations before sending to agent (e.g. calling `agent.observe(act)` works well with the new serialization method
- [x] Remove references to batch sort in documentation
- [ ] More Integration Tests (make sure we get same performance on tasks)

**CHANGES IN THIS PR**
1. Deprecate `--pytorch-datafile` in favor of `--pytorch-datapath` (which is ONLY specified if the data does not reside in the normal ParlAI Datapath)
2. (_New_) Now specify batch sorting via command line arg `-pybsrt true`
3. (_New_) Specify which field to use when sorting batches via command line arg `--batch-sort-field`
4. (_Fix_) When specifying `--pytorch-preprocess true`, Tensors are now correctly handled. 
5. The location of the saved data is now more straightforward - will be in `ParlAI/data/<task>_pyt_data/data/...`. If an agent is specified and the data is preprocessed, the path is `ParlAI/data/<task>_pyt_data_<agent>_preprocess/...`

**SPEED COMPARISON**
screenshot 2018-10-23 17 26 05


parlai/core/utils.py
```diff
@@ -311,7 +311,13 @@ def round_sigfigs(x, sigfigs=4):
...
 
 def flatten(teacher, context_length=-1, include_labels=True):
-    """Return a flattened version of a teacher's data.
+    """
+    ******************************************************************
+    DEPRECATED: If you would like to make use of batch sorting, please
+    use the PytorchDataTeacher
+    ******************************************************************
+
+    Return a flattened version of a teacher's data.
```
stephenroller: Humble recommendation: add

warnings.warn("flatten is deprecated. Please use PytorchDataTeacher.", warnings.DeprecationWarning)

stephenroller: Also make sure this renders reasonably in the autodocs.

klshuster: Thanks; will make sure the autodocs look good.

klshuster:
note: #1237 changes are incorporated here

stephenroller:
love that quip shot

parlai/core/params.py
```diff
@@ -362,17 +376,36 @@ def add_parlai_args(self, args=None):
...
         pytorch.add_argument(
-            '--batch-sort-cache', type=str,
-            choices=['pop', 'index', 'none'], default='none',
-            help='Whether to have batches of similarly sized episodes, and how'
-            'to build up the cache')
+            '-pybsrt', '--pytorch-teacher-batch-sort',
+            type='bool', default=False,
+            help='Whether to construct batches of similarly sized episodes'
+            'when using the PytorchDataTeacher (either via specifying `-pyt`'
+            'or `-pytd`')
```
alexholdenmiller: maybe we should actually turn it on by default for pytorch teacher?

before we disabled it just because of how it "corrupted" the data with newline-concat, but this version doesn't have that problem

parlai/core/params.py
```diff
@@ -362,17 +376,36 @@ def add_parlai_args(self, args=None):
...
         pytorch.add_argument(
             '--batch-length-range', type=int, default=5,
             help='degree of variation of size allowed in batch')
         pytorch.add_argument(
             '--shuffle', type='bool', default=False,
             help='Whether to shuffle the data')
-        self.add_parlai_data_path(parlai)
+        pytorch.add_argument(
+            '--batch-sort-field', type=str, default='text',
+            help='What field to use when determining the length of an episode')
```
alexholdenmiller: maybe set `choices=['text']`?

parlai/core/params.py
```diff
@@ -362,17 +376,36 @@ def add_parlai_args(self, args=None):
...
+                             help='Number of past utterances to remember when '
+                                  'building flattened batches of data in multi-'
+                                  'example episodes.'
+                                  '(For use with PytorchDataTeacher)')
+        pytorch.add_argument('-pyincl', '--pytorch-include-labels',
+                             default=True, type='bool',
+                             help='Specifies whether or not to include labels '
+                                  'as past utterances when building flattened '
+                                  'batches of data in multi-example episodes.'
+                                  '(For use with PytorchDataTeacher)')
```
alexholdenmiller: should we really rename these or just take over the existing args in the pytorch teacher?

parlai/core/params.py
alexholdenmiller: fwiw: the way batch sorting is implemented in the paper require more complex operations than you have below--other options wouldn't just be e.g. "labels" but could be something like "text then labels" (currently what the regular batch sorting is doing, first sort by text then break ties by average length of the label)

parlai/core/params.py
alexholdenmiller: I guess that makes it hard to phase out the other one rather than just replacing it

PR #1212, opened 2018-10-05
tqdm progress bar, smart num examples, random tmp files for output

uralik:
Some changes mentioned in the title.

is using tqdm OK here?

### DIFF
--- a/parlai/scripts/convert_data_to_parlai_format.py
+++ b/parlai/scripts/convert_data_to_parlai_format.py
@@ -18,10 +18,9 @@ Examples
 from parlai.core.params import ParlaiParser
 from parlai.agents.repeat_label.repeat_label import RepeatLabelAgent
 from parlai.core.worlds import create_task
-from parlai.core.utils import msg_to_str
+from parlai.core.utils import msg_to_str, TimeLogger
 import random
 import tempfile
-from tqdm import tqdm
 
 
 def dump_data(opt):
@@ -40,11 +39,12 @@ def dump_data(opt):
         num_examples = world.num_examples()
     else:
         num_examples = opt['num_examples']
+    log_timer = TimeLogger()
 
     print('[ starting to convert.. ]')
     print('[ saving output to {} ]'.format(outfile))
     fw = open(outfile, 'w')
-    for _ in tqdm(range(num_examples)):
+    for _ in range(num_examples):
         world.parley()
         world.acts[0]['labels'] = world.acts[0].get(
             'labels', world.acts[0].pop('eval_labels', None))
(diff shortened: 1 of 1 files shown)


parlai/scripts/convert_data_to_parlai_format.py
```diff
@@ -19,19 +19,32 @@
...
 
 def dump_data(opt):
     # create repeat label agent and assign it to the specified task
     agent = RepeatLabelAgent(opt)
     world = create_task(opt, agent)
     ignorefields = opt.get('ignore_fields', '')
+    if opt['outfile'] is None:
+        outfile = tempfile.mkstemp(
+            prefix='convert_data_to_parlai_output',
+            suffix='.txt')[1]
```
jaseweston: what was wrong with /tmp/dump?  what does this do?

parlai/scripts/convert_data_to_parlai_format.py
```diff
@@ -19,19 +19,32 @@
 from parlai.agents.repeat_label.repeat_label import RepeatLabelAgent
 from parlai.core.worlds import create_task
 from parlai.core.utils import msg_to_str
-
 import random
+import tempfile
+from tqdm import tqdm
```
jaseweston: whatdoes tqdm do? we actually have a logger also here:
https://github.com/facebookresearch/ParlAI/blob/master/parlai/scripts/build_dict.py#L113

parlai/scripts/convert_data_to_parlai_format.py
uralik: By default it was always using the same file, so overwriting the previous thing. Now file with prefix convert_data_to_parlai_output and some random suffix is created in /tmp folder by default if opt['outfile'] is not specified by the user. The full file name will be printed during the execution

parlai/scripts/convert_data_to_parlai_format.py
uralik: oh, I will fix that with TimeLogger()

parlai/scripts/convert_data_to_parlai_format.py
jaseweston: random is kinda weird but ok..  it could have the name of the actual task from "-t" and the type from "dt" in the name if u like?

uralik: agreed, will make /tmp/task_datatype_random.txt, the random is only suffix part such that if you run couple of times with some changed code inside of another agent or so and you forgot to specify - your old results will not be overwritten :)

jaseweston:
sounds good -- approving in advance

PR #1213, opened 2018-10-05
Making socket manager dual-threaded rather than multi

JackUrb:
Somewhat simplifies the `SocketManager` code (from a threading perspective) by making one singular sending thread rather than having a sending thread for each and every agent that is connected to the system at once. This reduces the number of threads that are handling heartbeats from N to 1, and heartbeats are sent in order ensuring nobody gets left out. 

This should further improve the disconnect issue, however I don't fully believe this will get everything yet (online indicators next). 

Tested by launching 20 hits and opening them all at once.

### DIFF
--- a/parlai/mturk/core/socket_manager.py
+++ b/parlai/mturk/core/socket_manager.py
@@ -204,6 +204,7 @@ class SocketManager():
 
         # initialize the state
         self.listen_thread = None
+        self.send_thread = None
         self.queues = {}
         self.threads = {}
         self.run = {}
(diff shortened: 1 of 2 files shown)


JackUrb:
Build is failing because it can't install `boto3`, unsure what @stephenroller change did that made it no longer download that dependency. Tests run fine locally

JackUrb:
Merging as the travis issue is with dependancies (as the old test script would download mturk deps when needed). Perhaps it's time to re-address how we keep ParlAI-MTurk dependancies

PR #1221, opened 2018-10-15
[wip] TorchRankerAgent abstract class

bhancock8:
[Hoping to start a discussion on the overall idea of adding this base parent class; an extra round of polish will be added before it's ready for a full code review.]

We currently have a number of models supporting ranking (e.g., Memnn, Kvmemnn, Transformer), but they each handle candidates separately--where they come from, how they're loaded/shared, what the defaults are, how they calculate metrics, etc. It makes sense to combine this shared functionality into an abstract class TorchRankerAgent to reduce confusion and code duplication, ensure fair quality/performance comparisons, and simplify the addition of new ranker models. 

Class Hierarchy:
- TorchAgent: handles core flags, setup, vectorization, history tracking, etc. that all torch models will have in common (as it currently does).
-TorchRankerAgent: implements ranker-related functionality: building candidates, tracking ranking-related metrics, assists with computing ranking-related metrics
-MemnnAgent/TransformerAgent/etc.: implements the specific handling of memories and calculation of ranks between candidates. This could be as simple as implementing two methods:

```
def score_candidates(self, batch, cand_vecs):
"""Given a batch and candidate set, return scores (for ranking)"""

def build_model(self):
"""Build a new model (implemented by children classes)"""
```

Much of the TorchRankerAgent code was pulled up from the existing implementation of MemnnAgent.

** Candidates:**
With the TorchRankerAgent, candidates can come from one of four places (and can be specified separately for train/eval using --train-candidates and --eval-candidates):

-batch [default]: use the other labels in the batch as negatives
-inline: use the candidate set provided with each example
-fixed: use the fixed candidate set from a user-provided file (i.e., for interactive mode)
-vocab: use every token in the vocab as a candidate (it was provided in MemNN, so I carried it through)


parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+from parlai.core.thread_utils import SharedTable
+from parlai.core.utils import round_sigfigs, padded_tensor
+
+class TorchRankerAgent(TorchAgent):
+    @staticmethod
+    def add_cmdline_args(argparser):
+        TorchAgent.add_cmdline_args(argparser)
+        agent = argparser.add_argument_group('TorchRankerAgent')
+        agent.add_argument(
+            '-tc', '--train_candidates', type=str, default='batch',
```
stephenroller: let's use `--train-candidates` instead, similar for others.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+            choices=['batch', 'inline', 'fixed', 'dict'],
+            help='the source of candidates during training')
+        agent.add_argument(
+            '-ec', '--eval_candidates', type=str, default='batch',
+            choices=['batch', 'inline', 'fixed', 'dict'],
```
stephenroller: default to fixed?

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+
+class TorchRankerAgent(TorchAgent):
+    @staticmethod
+    def add_cmdline_args(argparser):
+        TorchAgent.add_cmdline_args(argparser)
+        agent = argparser.add_argument_group('TorchRankerAgent')
+        agent.add_argument(
+            '-tc', '--train_candidates', type=str, default='batch',
+            choices=['batch', 'inline', 'fixed', 'dict'],
+            help='the source of candidates during training')
```
stephenroller: capitalization on help strings.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+            self.model = shared['model']
+            self.metrics = shared['metrics']
+        else:
+            # Must call _get_model_file first so that paths are updated if
+            # necessary (e.g., .dict file)
+            model_file = self._get_model_file(opt)
+            self.metrics = {'loss': 0.0, 'examples': 0, 'rank': 0}
+            opt['rank_candidates'] = True
+
+            super().__init__(opt, shared)
```
stephenroller: should the call to `super().__init__` be outside the if? Seems weird it's called separately.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+        if self.use_cuda:
+            self.model.cuda()
+            self.rank_loss.cuda()
+
+        optim_params = [p for p in self.model.parameters() if p.requires_grad]
+        self._init_optim(optim_params)
+
+    def score_candidates(self, batch, cand_vecs):
+        """Given a batch and candidate set, return scores (for ranking)"""
+        raise NotImplementedError(
```
stephenroller: Fit this on one line. We use 88 in parlai, so it should be easy.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+            'Abstract class: user must implement score()')
+
+    def build_model(self):
+        """Build a new model (implemented by children classes)"""
+        raise NotImplementedError(
+            'Abstract class: user must implement build_model()')
```
stephenroller: ditto

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+
+        self.rank_loss = nn.CrossEntropyLoss(reduction='sum')
+        self._set_fixed_candidates(shared)
+
+        if self.use_cuda:
+            self.model.cuda()
+            self.rank_loss.cuda()
+
+        optim_params = [p for p in self.model.parameters() if p.requires_grad]
+        self._init_optim(optim_params)
```
stephenroller: @alexholdenmiller I propose we change this to init_optim, etc.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+            self.metrics['examples'] += batchsize
+            for b in range(batchsize):
+                rank = (ranks[b] == label_inds[b]).nonzero().item()
+                self.metrics['rank'] += 1 + rank
+
+        cand_preds = [[cands[i] for i in row] for row in ranks]
+        preds = [row[0] for row in cand_preds]
+        return Output(preds, cand_preds)
+
+    def _build_candidates(self, batch, source, mode):
```
stephenroller: very nice. We might need to check that this docstring is RST friendly, but i like it a lot

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+        label_inds = None
+
+        if source == 'batch':
+            self._warn_once(
+                f'{mode}_batch_candidates',
+                f'[ Executing {mode} mode with batch labels as set of '
+                    'candidates. ]')
+            if batchsize == 1:
+                self._warn_once(
+                    f'{mode}_batchsize_1',
```
stephenroller: Need to get read of the fstrings. We're currently maintaining 3.5 compatibility. @alexholdenmiller might take the excuse to drop it now though.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+                    "candidate set.")
+
+            cands = batch.candidates[0]
+            cand_vecs, _ = padded_tensor(batch.candidate_vecs[0],
+                use_cuda=self.use_cuda)
+            if label_vecs is not None:
+                label_inds = label_vecs.new_empty(batchsize)
+                # Warning: This computation is O(batchsize x num_candidates x seqlen)
+                for i, label_vec in enumerate(label_vecs):
+                    # Option 1: matrix multiplies
```
stephenroller: Comment seems weird? How is it matrix multiplies? Anyway, let's drop the option 1/option 2 comments and just leave the version you actually chose.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+                m['loss'] = self.metrics['loss']
+                m['mean_loss'] = self.metrics['loss'] / examples
+            if self.metrics['rank'] > 0:
+                m['mean_rank'] = self.metrics['rank'] / examples
+        for k, v in m.items():
+            # clean up: rounds to sigfigs and converts tensors to floats
+            m[k] = round_sigfigs(v, 4)
+        return m
+
+    def vectorize(self, *args, **kwargs):
```
stephenroller: Vectorize's override should be in memnn, NOT in torch ranker agent.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+
+        if model_file is not None:
+            # if we are loading a model, should load its dict too
+            if (os.path.isfile(model_file + '.dict') or
+                    opt['dict_file'] is None):
+                opt['dict_file'] = model_file + '.dict'
+
+        return model_file
+
+    def _set_fixed_candidates(self, shared):
```
stephenroller: pass over this again with line length of 88, drop the fstrings.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+                self.fixed_candidates_vec, _ = padded_tensor(candidate_vecs,
+                    use_cuda=self.use_cuda)
+            else:
+                self.fixed_candidates = None
+                self.fixed_candidates_vec = None
+
+    def _warn_once(self, flag, msg):
+        if not hasattr(self, flag):
+            setattr(self, flag, True)
+            print(msg)
```
stephenroller: I vote `import warning; warning.warn` instead of print

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+                    self.fixed_candidates]
+                self.fixed_candidates_vec, _ = padded_tensor(candidate_vecs,
+                    use_cuda=self.use_cuda)
+            else:
+                self.fixed_candidates = None
+                self.fixed_candidates_vec = None
+
+    def _warn_once(self, flag, msg):
+        if not hasattr(self, flag):
+            setattr(self, flag, True)
```
stephenroller: Ehhh this setattr is scary to me. Can we change it to `setattr(self, '__warned_' + flag, True)` or maybe even have a warnings dictionary that  marks these or something? Many of those flag names could easily be variable names in a subclass (while it was previously more limited in just memnn)

parlai/core/torch_ranker_agent.py
bhancock8: So, I was torn on this. Defaulting to fixed seems a little scary to me because 'fixed' here means a single identical candidate set for all examples (e.g., when in interactive mode). So we'll want something different during training anyway. And not all tasks have a fixed candidate set, so our default would be something that's not always applicable, (possibly resulting in a person using whatever we backed off to when they think they're using fixed), whereas you'll always have a batch. Thoughts?

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+        batchsize = batch.text_vec.size(0)
+        self.model.train()
+        self.optimizer.zero_grad()
+
+        cands, cand_vecs, label_inds = self._build_candidates(batch,
+            source=self.opt['train_candidates'], mode='train')
+        scores = self.score_candidates(batch, cand_vecs)
+        loss = self.rank_loss(scores, label_inds)
+
+        self.metrics['loss'] += loss.item()
```
emilydinan: Note: ParlAI automatically calculates hits@k metrics IFF label_candidates are provided. If we are simply ranking with the negatives as other elements in the batch (or with a separate set of candidates), we must calculate accuracy, etc. ourselves.

parlai/core/torch_ranker_agent.py
stephenroller: perhaps `inline` is the behavior we want? If the dataset provides a set of `label_candidates`, the default should be to use them.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def _set_fixed_candidates(self, shared):
+        if shared:
+            self.fixed_candidates = shared['fixed_candidates']
+            self.fixed_candidates_vec = shared['fixed_candidates_vec']
+        else:
+            if self.opt['fixed_candidates_path']:
+                print(f"Vectorizing fixed candidates set from "
+                    f"{self.opt['fixed_candidates_path']}")
+                with open(self.opt['fixed_candidates_path'], 'r') as f:
```
emilydinan: A couple things here: the biggest use case for this is interactive mode, in which case the set of fixed candidates is often *huge*, so
- we might want to only encode the fixed candidates with the model *once* at init time
- we might want to save these encoded candidates to a file for loading, so we don't have to repeat this
- you might run into some memory issues, consider batching here and/or forcing this to happen on CPU
(you can see how this is handled in internal)

parlai/core/torch_ranker_agent.py
bhancock8: Yeah, it's honestly a little weird to me too that super().__init__() gets sandwiched between other code. I think it should probably always be called first and then have things overwritten by the children as necessary. I ddi in this way to keep it compatible with past saved MemNN models, but if we're willing to break that (and I think it's worth it here), I can move some things up to TorchAgent so TorchRankerAgent isn't doing anything special for loading existing models (and I don't think it needs to).

parlai/core/torch_ranker_agent.py
bhancock8: But fstrings are python's best feature. 😢 Will do, even though it pains me.

parlai/core/torch_ranker_agent.py
bhancock8: Yes please, @alexholdenmiller! f-strings are so much more readable and take up less space. 🙏

parlai/core/torch_ranker_agent.py
bhancock8: The way it's currently written, it does encode only once at init time. But it doesn't cache the encodings or handle batching. I'll pull up the heavy-lifting code from internal, since I imagine that class will eventually inherit from this one anyway (ideally).

parlai/core/torch_ranker_agent.py
stephenroller: No, those models need to remain working; these pretrained models are among our most popular downloads.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,337 @@
...
+                flag=(mode + '_batch_candidates'),
+                msg=('[ Executing {} mode with provided inline set of candidates '
+                    '(assumed to be identical for all candidates in a batch). ]'
+                    ''.format(mode)))
+            if batch.candidate_vecs is None:
+                raise ValueError("If using candidate source 'inline', then "
+                    "bath.candidate_vecs can not be None.")
+            if batchsize > 1 and (
+                torch.any(batch.candidate_vecs[0] != batch.candidate_vecs[1])):
+                raise ValueError("All examples in a batch must have the same "
```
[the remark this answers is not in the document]
bhancock8: I agree that if you pass in batch_size > 1 and choose `inline` candidates, then you'll get passed in a different choice of candidates for each example. But at least with the MemNN, I was getting errors before in that case, as I believe there were places that it was making that assumption. So I threw in the exception to make sure we didn't silently fail. But I absolutely agree that we should support varied candidates in a batch. It just means that our candidates can now be 2D or 3D, and we'll need to be extra clear in our documentation. I'll take a look tomorrow and see what changes would need to be made to support that for MemNN, and I assume you'll take care of it for the new-age transformer.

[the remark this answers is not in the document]
stephenroller: edit: noticed the TODO

parlai/core/torch_ranker_agent.py
bhancock8: Ok, so I guess I'll leave this for now to not break things. But I think long-term it's worth being explicit somewhere about how we expect inheritance to work here. Is the TorchAgent just an abstract class meant to ensure certain methods exist in children, or is it going to be a base class that handles most non-model specific details? Because if the latter is the case, I think it would make sense to move model loading and initialization calls up to TorchAgent. (e.g., check if 'model_file' and/or 'dict_file' are in opt, and if they are, load them. if they're not, call build_model() and instantiate a new dictionary). But I get that modifying TorchAgent would touch a lot of things, so fine to not mess with it in this PR.

parlai/core/torch_ranker_agent.py
bhancock8: Thank you for the note! I don't want to duplicate the metrics code, so I'll just make sure that we pass back the ranked candidates in a way that ParlAI will recognize.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,337 @@
...
+            self.fixed_candidates_vec = shared['fixed_candidates_vec']
+        else:
+            if self.opt['fixed_candidates_path']:
+                print("Vectorizing fixed candidates set from {}".format(
+                    self.opt['fixed_candidates_path']))
+                with open(self.opt['fixed_candidates_path'], 'r') as f:
+                    self.fixed_candidates = [line.strip() for line in
+                        f.readlines()]
+                vectorize_func = self._vectorize_text
+                candidate_vecs = [vectorize_func(cand) for cand in
```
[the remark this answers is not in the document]
bhancock8: Actually, unfortunately the truncate flag defaults to -1 and then self.truncate is set to None if truncate < 0, so instead of self.opt.get('truncate'), we need to use self.truncate here. My vote would be to just stick with None throughout; no -1. I assume it's there so we can set the type as `int` for the flag, but we can easily define a one-liner func `int_or_none` that validates the input is int or None and use that instead. -1's as sentinels are scary since they so often work where they shouldn't in python code. What do you think @alexholdenmiller?

parlai/core/torch_ranker_agent.py
```diff
@@ -31,9 +30,15 @@ def add_cmdline_args(argparser):
             choices=['batch', 'inline', 'fixed', 'vocab'],
             help='The source of candidates during training')
         agent.add_argument(
-            '-cands', '--fixed-candidates-path', type=str,
+            '-cands', '--fixed-candidates-path', type=str, default='',
```
stephenroller: Why set an empty default (and not leave it to None?)

parlai/core/torch_ranker_agent.py
```diff
@@ -31,9 +30,15 @@ def add_cmdline_args(argparser):
...
             help='A text file of fixed candidates to use for all examples, one '
-                'candidate per line')
+                 'candidate per line')
+        agent.add_argument(
+            '-candvecs', '--fixed-candidate-vecs-path', type=str, default='',
```
stephenroller: I'm tempted to suggest we just derive this path (use fixed_candidates_path or model_file with a special suffix), but that could be short sited. model_file + shorthash(fixed_candidate_paths)?

parlai/core/torch_ranker_agent.py
stephenroller: just marking resolved

parlai/core/torch_ranker_agent.py
```diff
@@ -195,52 +213,49 @@ def _build_candidates(self, batch, source, mode):
...
-            if batchsize > 1 and (
-                torch.any(batch.candidate_vecs[0] != batch.candidate_vecs[1])):
-                raise ValueError("All examples in a batch must have the same "
-                    "candidate set.")
-
-            cands = batch.candidates[0]
-            cand_vecs, _ = padded_tensor(batch.candidate_vecs[0],
-                use_cuda=self.use_cuda)
+                raise ValueError(
+                    "If using candidate source 'inline', then bath.candidate_vecs "
```

parlai/core/torch_ranker_agent.py
```diff
@@ -308,30 +319,83 @@ def _get_model_file(self, opt):
...
             else:
                 self.fixed_candidates = None
-                self.fixed_candidates_vec = None
+                self.fixed_candidate_vecs = None
+
+    def vectorize_fixed_candidates(self, cands_batch):
+        return [self._vectorize_text(cand, truncate=self.truncate, truncate_left=False)
+                for cand in cands_batch]
+
+    def _cat_and_pad(self, tensor_list, max_len=None, use_cuda=False):
```
stephenroller: use `parlai.core.utils.padded_tensor` instead

parlai/core/torch_ranker_agent.py
```diff
@@ -308,30 +319,83 @@ def _get_model_file(self, opt):
...
+                                   opt['fixed_candidate_vecs_path'])
+
+                if self.use_cuda:
+                    self.fixed_candidate_vecs = self.fixed_candidate_vecs.cuda()
             else:
                 self.fixed_candidates = None
-                self.fixed_candidates_vec = None
+                self.fixed_candidate_vecs = None
+
+    def vectorize_fixed_candidates(self, cands_batch):
```
stephenroller: make this private (`_vectorize_fixed_candidates`)

parlai/core/torch_ranker_agent.py
```diff
@@ -308,30 +319,83 @@ def _get_model_file(self, opt):
 
         return model_file
 
-    def _set_fixed_candidates(self, shared):
+    def set_fixed_candidates(self, shared):
```
stephenroller: Seems like private is better, no?

parlai/core/torch_ranker_agent.py
bhancock8: My thought in making it public was that if a user wants to pre-encode (not just pre-vectorize), then they'll override this function instead of the whole set_fixed_candidates method that does the batching, etc.

stephenroller: oic okay.

parlai/core/torch_ranker_agent.py
bhancock8: Yeah, it does feel silly to have them type two probably very similar paths. But allowing a human-passed filepath lets them pass in candidates that they've encoded themselves offline if they'd like. Maybe we just assume the vec path is the same as the cand path with the .vec extension?

parlai/core/torch_ranker_agent.py
bhancock8: the types and dimensions don't line up: it expects a list of lists, and we have a list of list of tensors.

parlai/core/torch_ranker_agent.py
alexholdenmiller: I like the cache-and-read-from fixed-candidates-path.vec

parlai/core/torch_ranker_agent.py
alexholdenmiller: I believe you copied this directly from memnn--keep the code from there. calculating the mean rank is a nice useful metric which is a little different than hits@ and is calculable MUCH faster than hits@k

parlai/core/torch_ranker_agent.py
alexholdenmiller: that's what parlai.core.utils.padded_3d is for

parlai/core/torch_ranker_agent.py
alexholdenmiller: MemNN had, basically:
```
model_file = self._get_model_file(opt)
super().__init__(opt, shared)
```
at the top of init, since we want to override the dict_file. it's fine if we do it on the shared instances too, even if a little wasted energy, to get it out of the if block.

We could move that to TorchAgent, seq2seq is doing the same thing.

parlai/core/torch_ranker_agent.py
stephenroller: My concern for that was that fixed candidates is likely dataset specific (e.g. all the training examples) but the encodings will be model specific (every run in a sweep will have different caches!)

parlai/core/torch_ranker_agent.py
bhancock8: Ok, I can revert back to the original memnn code. It seems unfortunate though that the ranking-related metrics are housed in the Metrics object that we don't have very direct access to. Code duplication is especially scary when it comes to metrics; maybe at sme point we can break the calculation of ranking-related metrics into a single helper method that both Metrics and TorchRankerAgent call.

parlai/core/torch_ranker_agent.py
bhancock8: How about this: -candvecs/--fixed-candidate-vectors is a string that's either 'reuse' [default], 'replace', or a path. 'reuse' will load from fixed-candidates-path.vec (or back off to 'replace' if that fails), 'replace' will generate a new vec file at fixed-candidates-path.vec, or if they pass a full path, we use whateever's there. So most of the time, if you're not getting fancy, you generated it once and then it always just reloads it. But if you really want to, you can keep separate vec files and switch between them by passing in their particular paths. Thoughts?

bhancock8: I went ahead and implemented what I proposed above. You can take a look.

bhancock8:
Alright, I believe I've made all suggested edits and it's passing the tests. Merge at will, I say! Unless there are any additional suggestions.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,445 @@
...
+from parlai.core.utils import round_sigfigs, padded_3d
+
+
+class TorchRankerAgent(TorchAgent):
+    @staticmethod
+    def add_cmdline_args(argparser):
+        TorchAgent.add_cmdline_args(argparser)
+        agent = argparser.add_argument_group('TorchRankerAgent')
+        agent.add_argument(
+            '-tc', '--train-candidates', type=str, default='batch',
```
stephenroller: I'll make a humble suggestion here to improve usability, but call it optional:

We should have this be `-cands/--candidates`, and it should default to `inline`. Then `-ec` should have a default of `None`. If `--eval-candidates` is None, then it is set to whatever `-cands` is set to. So if you want `batch` for both, you just need to set one flag. If you want differing behavior at train/test, then you have to specify the extra flag

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,445 @@
...
+            choices=['batch', 'inline', 'fixed', 'vocab'],
+            help='The source of candidates during training')
+        agent.add_argument(
+            '-ec', '--eval-candidates', type=str, default='batch',
+            choices=['batch', 'inline', 'fixed', 'vocab'],
+            help='The source of candidates during training')
+        agent.add_argument(
+            '-cands', '--fixed-candidates-path', type=str,
```
stephenroller: I suggest the short name for this be `-fcp`, or better yet, there is no short name for this.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,445 @@
...
+    @staticmethod
+    def add_cmdline_args(argparser):
+        TorchAgent.add_cmdline_args(argparser)
+        agent = argparser.add_argument_group('TorchRankerAgent')
+        agent.add_argument(
+            '-tc', '--train-candidates', type=str, default='batch',
+            choices=['batch', 'inline', 'fixed', 'vocab'],
+            help='The source of candidates during training')
+        agent.add_argument(
+            '-ec', '--eval-candidates', type=str, default='batch',
```
stephenroller: I suggest the shortname for this be `-ecands`, or no short name at all.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,445 @@
...
+            choices=['batch', 'inline', 'fixed', 'vocab'],
+            help='The source of candidates during training')
+        agent.add_argument(
+            '-cands', '--fixed-candidates-path', type=str,
+            help='A text file of fixed candidates to use for all examples, one '
+                 'candidate per line')
+        agent.add_argument(
+            '-candvecs', '--fixed-candidate-vecs', type=str, default='reuse',
```
stephenroller: I suggest we drop the shortname.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,445 @@
...
+    def save_candidate_vecs(self, vecs, path):
+        print("[ Saving fixed candidate set vectors to {} ]".format(path))
+        with open(path, 'wb') as f:
+            torch.save(vecs, f)
+
+    def vectorize_fixed_candidates(self, cands_batch):
+        """Convert a batch of candidates from text to vectors
+
+        Args:
+            cands_batch: a [batchsize] list of candiadates (strings)
```
stephenroller: Use sphinx docstrings instead of google docstrings please.

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,445 @@
...
+        A child class may choose to overwrite this method to perform vecotrization as
+        well as encoding if so desired.
+        """
+        return [self._vectorize_text(cand, truncate=self.truncate, truncate_left=False)
+                for cand in cands_batch]
+
+    def _warn_once(self, flag, msg):
+        """
+        Args:
+            flag: The name of the flag
```
stephenroller: ditto

parlai/tasks/task_list.py
```diff
@@ -690,7 +690,11 @@
         "id": "ConvAI2_wild_evaluation",
         "display_name": "ConvAI2_wild_evaluation",
         "task": "convai2_wild_evaluation",
-        "tags": ["All",  "ChitChat" ],
```
stephenroller: how did this sneak into this diff though lol. thanks for the flake8

bhancock8: ¯\_(ツ)_/¯ beats me. I just do whatever the almighty flake8 checker tells me to do.

parlai/core/torch_ranker_agent.py
bhancock8: I think that makes sense if these really were warnings, but most of these are really more informational, logging which option you chose. When they print as warnings, they look more like errors, even though nothing went wrong.

bhancock8:
Made the suggested changes. It should be ready to merge now once the new integration test you're writing has been added, @stephenroller.

jaseweston:
shouldn't  /torch_ranker_agent be in parlai/core/torch_agent.py ?
we have all the teachers in the same py file...so not sure why that would be separate

parlai/core/torch_ranker_agent.py
```diff
@@ -0,0 +1,448 @@
...
+                 'value as --candidates if no flag is given)')
+        agent.add_argument(
+            '-fcp', '--fixed-candidates-path', type=str,
+            help='A text file of fixed candidates to use for all examples, one '
+                 'candidate per line')
+        agent.add_argument(
+            '--fixed-candidate-vecs', type=str, default='reuse',
+            help="One of 'reuse', 'replace', or a path to a file with vectors "
+                 "corresponding to the candidates at --fixed-candidates-path. "
+                 "By default, a candidate vector file is generated with the same path "
```
jaseweston: i think this should appear as an extension of the model,
not the candidates path, because several models can
use the same candidates...perhaps
<model_name>.<cands_name>

emilydinan: For what it's worth, I've been using <model_file>.<cands_name> in other models

alexholdenmiller:
sorry one more delay: can I land #1227 first?

alexholdenmiller:
(training tests for memnn)

alexholdenmiller:
I'll fix the merge conflicts

parlai/core/torch_ranker_agent.py
alexholdenmiller: I see what you mean, but do note that Metrics is acting on text (slow but fair / consistent) and mean_rank is being calculated on tensors (subject to the model doing it correctly but much faster to calculate and so better during training)

bhancock8:
Pass tests by updating pip to version with --progress-bar off option so we don't end up with logs that exceed travis' file size limits.

PR #1222, opened 2018-10-16
Multi GPU support for seq2seq

emilydinan:
Adds multi GPU support for the Seq2Seq agent. To this end, we: 
- make changes to how candidate-related variables are passed into the `forward` function so that they properly distribute across multiple GPUs. 
- add `seq_len` to forward function so we can recover the proper size of ouputs after they (might be) truncated when pushed through pack_padded, etc. so that we can properly gather outputs across multiple GPUs
- small change to padded_tensor utility function for additional functionality
- some truncation/extension of inputs and outputs to handle case when (near the end of dataset possibly) batchsize is cut down and may not have an even number of elements; during train, we throw an example out, but during evaluation we accommodate for this case with extra padding

Tested on babi and personachat tasks.


parlai/agents/seq2seq/seq2seq.py
```diff
@@ -226,16 +228,18 @@ def _init_model(self, states=None):
...
             print('Seq2seq: fixing embedding weights.')
             self.model.decoder.lt.weight.requires_grad = False
             self.model.encoder.lt.weight.requires_grad = False
             if opt['lookuptable'] in ['dec_out', 'all']:
                 self.model.decoder.e2s.weight.requires_grad = False
 
+        if self.use_cuda:
+            self.model.cuda()
+            if self.multigpu:
+                self.model = torch.nn.DataParallel(self.model)
```
stephenroller: you can do `self.model.encoder = self.model.module.encoder` (ditto for decoder), if you want to remove the `if multigpu` blocks later on

emilydinan: doesn't quite work because beam_search (where we call model.encoder, etc.) is a static method. we could pass the encoder, etc. as an argument to beam search but i thought the if statements would end up being a bit cleaner

emilydinan: fixed!

emilydinan:
ready for review cc @stephenroller @alexholdenmiller

emilydinan:
**TEST PERFORMANCE:** If your model is "small," for the most part, training across multiple GPUs will be slower because the cost of model synchronization is not worth it. Can see clear benefits when your model size is large: for one, you can have larger models without running out of memory, and second, training *may* be faster (depending on batchsize).

**One epoch on personachat:**

default params / batchsize 128:
- 1 gpu: 41s 
- 2 gpus: 81s (nearly 2x slowdown)

esz 512, hs 512, nl 3 / batchsize 128:
- 1 gpu: 143s
- 2 gpus: 133s 

esz 1024, hs 512, nl 5 / batchsize 128:
- 1 gpu: OOM
- 2 gpus: 216s

emilydinan:
ignoring travis failures for breaking after a binary operator and landing, because in all of the cases mentioned we *are* doing that

PR #1223, opened 2018-10-16
Fixing DB thread explosion

JackUrb:
There was a small glitch in the `maintain_hit_status` thread that was used when the `use_db` flag was provided that caused thread counts to spiral out of control after the last update, which would eventually use up the maximum thread pool.


JackUrb:
Tests are still only failing due to dependancies not downloading on travis, they pass locally

JackUrb:
I should probably refactor `start_task` in a followup. It handles a bit too much between itself and helpers.

PR #1231, opened 2018-10-19
Added basic implementations of tasks from DecaNLP

CharlieChenFB:
More info on DecaNLP tasks can be found here:

https://github.com/salesforce/decaNLP

I've added minimally working implementations of the tasks from DecaNLP. However, there are a few things here and there which I would like to clear up.

The DecaNLP tasks can be run with the command `display_data.py -t decanlp`, and the individual tasks can be run with a colon after e.g. to run the SST task `display_data.py -t decanlp:sst`

The tasks are individually, so if you do not wish to build all the data for the decathalon, it is possible to run them individually e.g. running `display_data.py -t sst` 

There still a few things I'd like to clear up/fix, however:
1. For the SST task, there is a lot of garbage data, for example:
```
[sst]: 've
Is this sentence positive or negative?
[labels: negative]
[label_candidates: positive|negative]
``` 
and I'm not sure how the original DecaNLP tasks handled them. Furthermore, I've set the cutoff for a positive phrase to be >0.666 and negative to be <0.333 while omitting the other, neutral labels. 

2. How should I handle training/validation/test data? The dataset does not always define the partitions. 

3. How are the shuffling handled? For example, in the CNN/DM data_set it currently goes through the CNN data first and then the DM dataset.

4. As of now, the DecaNLP is merely going through all the datasets in order. So it is creating an episode of 10 by taking the first episode from the first task with the first episode from the 2nd task and so on. I should probably implement some sort of inter-task shuffle since not all the tasks have the same number of episodes.


alexholdenmiller:
Thanks Charlie! I'll add a few comments, but we're debating a few more changes as well

parlai/tasks/decanlp/agents.py
```diff
@@ -0,0 +1,111 @@
...
+
+class WikisqlTeacher(wikisql.DefaultTeacher):
+    pass
+
+
+class MwscTeacher(mwsc.DefaultTeacher):
+    pass
+
+
+class DecaNLPTeacher(DialogTeacher):
```
alexholdenmiller: This should actually be formatted like parlai/tasks/babi/agents.py's All10kTeacher so that each sub-teacher preserves its ID

parlai/tasks/decanlp/agents.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def setup_data(self, path):
+        while True:
+            qa_tasks = [next(task)[0] for task in self.task_setup_data]
+            if len(qa_tasks) == 10:
+                yield qa_tasks[0], True
+                for task_example in qa_tasks[1:]:
+                    yield task_example, False
+            else:
+                break
```
alexholdenmiller: this is only doing 10 exs and considering them all the same example, btw--but with the change above, this function can be removed

parlai/tasks/decanlp/build.py
```diff
@@ -0,0 +1,31 @@
...
+
+    if not build_data.built(dpath, version_string=version):
+        print('Building DecaNLP Tasks...')
+        build_tasks = [importlib.__import__('parlai.tasks.{}.build'.format(task),
+                                            fromlist=['build'])
+                       for task in decanlp_tasks]
+        for task in build_tasks:
+            task.build(opt)
+        build_data.make_dir(dpath)
+        build_data.mark_done(dpath, version_string=version)
```
alexholdenmiller: this whole file might be removable with the multitaskteacher code change I suggested above

parlai/tasks/woz/build.py
```diff
@@ -0,0 +1,35 @@
...
+            # An older version exists, so remove these outdated files.
+            build_data.remove_dir(dpath)
+        build_data.make_dir(dpath)
+
+        # Download the data.
+        fnames = ['woz_test_en.json', 'woz_train_en.json', 'woz_validate_en.json']
+        for fname in fnames:
+            url = 'https://github.com/nmrksic/' \
+                  'neural-belief-tracker/raw/master/data/woz/'
+            + fname
```
alexholdenmiller:
```
url = ('https://github.com/nmrksic/'
       'neural-belief-tracker/raw/master/data/woz/' +
       fname)
```

CharlieChenFB: Aha, I didn't see the multitask teacher. 

I've now added the changes you've stated.

parlai/tasks/decanlp/agents.py
CharlieChenFB: Deleted the file

stephenroller:
@CharlieChenFB, note you can use `bash tests/lint_changed.sh` to check what travis is going to complain about. You may have to `pip install flake8` first.

emilydinan:
HI @CharlieChenFB , can you check that SSTTeacher runs? The failure indicates the following error:

```
File "/home/travis/build/facebookresearch/ParlAI/parlai/core/agents.py", line 589, in create_task_agent_from_taskname
    task_agents = teacher_class(opt)
  File "/home/travis/build/facebookresearch/ParlAI/parlai/tasks/sst/agents.py", line 28, in __init__
    super().__init__(opt, shared)
  File "/home/travis/build/facebookresearch/ParlAI/parlai/core/teachers.py", line 461, in __init__
    cands=self.label_candidates(), **kwargs)
  File "/home/travis/build/facebookresearch/ParlAI/parlai/core/teachers.py", line 558, in __init__
    self._load(data_loader, opt['datafile'])
  File "/home/travis/build/facebookresearch/ParlAI/parlai/core/teachers.py", line 646, in _load
    for episode in self._read_episode(data_loader(datafile)):
  File "/home/travis/build/facebookresearch/ParlAI/parlai/core/teachers.py", line 579, in _read_episode
    for entry, new in data_loader:
  File "/home/travis/build/facebookresearch/ParlAI/parlai/tasks/sst/agents.py", line 71, in setup_data
    new_episode = self.c % 3 == 2
AttributeError: 'DefaultTeacher' object has no attribute 'c'
```

It seems like in the class SSTTeacher, the parent class is initialized before setting the variable "self.c=0", and when the parent class is initialized, it calls `setup_data` in SSTTeacher before this variable is initialized.

parlai/tasks/sst/agents.py
```diff
@@ -0,0 +1,81 @@
...
+        self.dt = opt['datatype'].split(':')[0]
+
+        # store identifier for the teacher in the dialog
+        self.id = 'sst'
+
+        self.SST_LABELS = ['negative', 'positive']
+
+        self.label_path, opt['datafile'] = self._path(opt)
+
+        super().__init__(opt, shared)
```
emilydinan: these are the lines i'm referring to in the above comment fyi

CharlieChenFB: Thank you, Emily, I've fixed it now.

alexholdenmiller:
@emilydinan Travis got a memory error trying out this dataset--maybe we could switch the test to :stream mode?

emilydinan:
@alexholdenmiller I'll put out a fix

CharlieChenFB:
Hi, I think there are some discrepancies between my local run of flake and the one run here:

`
parlai/tasks/multinli/agents.py:44:30: W504 line break after binary operator
parlai/tasks/mwsc/agents.py:39:14: W605 invalid escape sequence '\]'
parlai/tasks/mwsc/agents.py:39:19: W605 invalid escape sequence '\['
parlai/tasks/woz/build.py:30:20: W504 line break after binary operator
`

These warnings don't appear when I run it locally.

CharlieChenFB:
Travis is also saying:

`(No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.)`

Is this just because this is a much bigger task than its used to?

alexholdenmiller:
Yeah `(No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.)` can just mean it's taking too long, if the test is passing locally then you're good!

alexholdenmiller:
hm a different test failed but I don't think it's related, still running the new tests again

alexholdenmiller:
Got `RecursionError: maximum recursion depth exceeded while calling a Python object` during DecaNLP task so there appears to be a bug still, I can try to debug a bit later

stephenroller:
Superseded by #1546.

PR #1237, opened 2018-10-22
Pytorch data teacher args.

stephenroller:
This is jason's view. I won't merge because I'm sure it will conflict with #1210.


stephenroller:
Abandoning; Kurt is going to do this in #1210.

PR #1240, opened 2018-10-22
switch pickle to json for opt files

alexholdenmiller:


alexholdenmiller:
travis seems confused

parlai/core/torch_agent.py
```diff
@@ -807,11 +807,10 @@ def save(self, path=None):
                     torch.save(states, write)
 
                 # save opt file
-                with open(path + ".opt", 'wb') as handle:
+                with open(path + '.opt', 'w') as handle:
```
jaseweston: would be better as a function save_opt() perhaps?   same for load

jaseweston:
would be better with save_opt and load_opt functions?

PR #1241, opened 2018-10-22
starspace: input feature dropout

jaseweston:


parlai/agents/starspace/starspace.py
```diff
@@ -291,6 +293,24 @@ def compute_metrics(self, loss, scores):
...
+    def input_dropout(self, xs, ys, negs):
+        def dropout(x, rate):
+            xd = []
+            for i in x[0]:
+                if random.uniform(0, 1) > rate:
+                    xd.append(i)
+            if len(xd) == 0:
+                # pick one random thing to put in xd
+                xd.append(x[0][random.randint(0, x.size(1)-1)])
+            return torch.LongTensor(xd).unsqueeze(0)
```
alexholdenmiller: could also do as UnknownDropout like we do in seq2seq, which makes it more robust to unknown tokens during eval
```
mask = input.new(input.size()).float().uniform_(0, 1) < self.prob
input.masked_fill_(mask, self.unknown_idx)
```
https://github.com/facebookresearch/ParlAI/blob/1cb7d600fa57c125ec440f9eefaf4aa1269b5f44/parlai/agents/seq2seq/modules.py#L290-L313

jaseweston:
ah, i see by input i mean input features of all kinds ..  "features" is also a bit ambiguous that people sometimes take it to mean 'learnt features' (so layers inside the net)

PR #1247, opened 2018-10-23
Add gatekeeper options.

stephenroller:
Adds gatekeeper options.

These options are similar to hidden, in that `--show-advanced-args` causes them to display, but they may also be triggered if a "gatekeeper" flag is set to a non-null value.

This is a good way to do grouped advanced options, as an alternative to what was done in #1237 or #1246 (cc @klshuster)


stephenroller:
This didn't seem to scratch the itch I had hoped for. Abandoning.

PR #1248, opened 2018-10-25
Fix for `max_hits_per_worker`

emilydinan:
`max_hits_per_worker` always defaulted to 1 if it was set because you need `unique_qual_name` to use `max_hits_per_worker`


JackUrb:
(The travis build will continue to fail because it isn't installing botocore)

PR #1249, opened 2018-10-25
Personality Captions

klshuster:
Adds the Personality-Captions Dataset

Some notes:
- There is a provided `download_images.py` script that will run when building the data. This script will require the user to confirm that they (1) have permission from Yahoo! to use the data, and (2) they would like to run the download which can take a _**LONG**_ time.
- Because of this very long download, a user can specify the `--yfcc-path`, which has the images saved already (as `<hash>.jpg`)
- I will update the NEWS url with the appropriate arxiv url when it is on there


jaseweston:
can you add to
https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/task_list.py

as well? thanks!

klshuster:
thank you for reminding me - added in recent commit

klshuster:
Note: this fails the travis tests because it requires input from the user to download images. I have checked the lint tests to confirm that no linting errors occur.

PR #1250, opened 2018-10-25
Fix for params.py

emilydinan:
Command line bug around advanced options when `add_parlai_args`  is `False`


stephenroller:
thx for fixing my bugs

PR #1252, opened 2018-10-25
PMT Frontend webapp

JackUrb:
This PR lays the groundwork for the frontend webapp that will eventually be used to test UIs, launch tasks, manage qualifications, review work, manage workers, and more.

Right now it hooks up to the DB that is created when using the optional `use_db` flag. It can be used by running `npm install; npm run dev; python server.py` from inside the `webapp` directory.

Current features:
Provides views into runs and workers (currently it acts as a passive way to view metadata about runs).

Next up I'll be adding the reviewer interface, which will follow a change to how PMT stores data.

### DIFF
--- a/parlai/mturk/core/mturk_data_handler.py
+++ b/parlai/mturk/core/mturk_data_handler.py
@@ -105,9 +105,9 @@ class MTurkDataHandler():
         self.create_default_tables()
 
     def _get_connection(self):
-        '''Returns a singular database connection to be shared amongst all
+        """Returns a singular database connection to be shared amongst all
         calls
-        '''
+        """
         curr_thread = threading.get_ident()
         if curr_thread not in self.conn or self.conn[curr_thread] is None:
             try:
@@ -123,13 +123,14 @@ class MTurkDataHandler():
         return self.conn[curr_thread]
 
     def _force_task_group_id(self, task_group_id):
-        '''Throw an error if a task group id is neither provided nor stored'''
+        """Throw an error if a task group id is neither provided nor stored"""
         if task_group_id is None:
             task_group_id = self.task_group_id
         assert task_group_id is not None, 'Default task_group_id not set'
         return task_group_id
 
     def create_default_tables(self):
+        """Prepares the default tables in the database if they don't exist"""
         with self.table_access_condition:
             conn = self._get_connection()
             c = conn.cursor()
@@ -141,6 +142,7 @@ class MTurkDataHandler():
             conn.commit()
 
     def log_new_run(self, target_hits, task_group_id=None):
+        """Add a new run to the runs table"""
         with self.table_access_condition:
             task_group_id = self._force_task_group_id(task_group_id)
             conn = self._get_connection()
@@ -150,7 +152,7 @@ class MTurkDataHandler():
             conn.commit()
 
     def log_hit_status(self, mturk_hit_creation_response, task_group_id=None):
-        '''Create or update an entry in the hit status table'''
+        """Create or update an entry in the hit status table"""
         task_group_id = self._force_task_group_id(task_group_id)
 
         hit_details = mturk_hit_creation_response['HIT']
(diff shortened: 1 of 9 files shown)


parlai/mturk/core/mturk_data_handler.py
```diff
@@ -443,6 +470,16 @@ def get_worker_assignment_pairing(self, worker_id, assignment_id):
             results = c.fetchone()
             return results
 
+    def get_all_run_data(self, start=0, count=30):
+        '''get all the run data for all task_group_ids.'''
```
emilydinan: nit: for all of our docstrings elsewhere we use double quotation marks

parlai/mturk/webapp/server.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def get(self):
+        try:
+            req = tornado.escape.json_decode(
+                tornado.escape.to_basestring(self.request.body)
+            )
+        except Exception:
+            req = {}
+
+        # self.render(
```
emilydinan: left in by accident?

parlai/mturk/webapp/server.py
```diff
@@ -0,0 +1,375 @@
...
+                        help='host to run the server on.')
+    parser.add_argument('-db_file', metavar='db_file', type=str,
+                        default=DEFAULT_DB_FILE,
+                        help='name of database to use (in core/run_data)')
+    parser.add_argument('-logging_level', metavar='logger_level', default='INFO',
+                        help='logging level (default = INFO). Can take logging '
+                             'level name or int (example: 20)')
+    FLAGS = parser.parse_args()
+
+    try:
```
emilydinan: i'm pretty sure a call to _checkLevel already typechecks for you, so you can just always call that (and you don't neeed this try/except block

parlai/mturk/webapp/server.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+def get_rand_id():
+    return str(hex(int(time.time() * 10000000))[2:])
+
+
+def ensure_dir_exists(path):
+    """Make sure the parent dir exists for path so we can write a file."""
+    try:
+        os.makedirs(os.path.dirname(path))
```
klshuster: are you going for `os.makedirs(os.path.dirname(path), exists_ok=True)`?

parlai/mturk/webapp/server.py
```diff
@@ -0,0 +1,375 @@
...
+        self.subs = app.subs
+        self.sources = app.sources
+        self.port = app.port
+        self.data_handler = app.data_handler
+
+    def get(self):
+        results = self.data_handler.get_all_worker_data()
+        processed_results = []
+        for res in results:
+            processed_results.append(dict(zip(res.keys(), res)))
```
klshuster: looks like you could use your `row_to_dict` function from above for `res`

parlai/mturk/core/mturk_data_handler.py
JackUrb: Oops, good catch. Fixing!

parlai/mturk/webapp/server.py
JackUrb: Yup - fixed

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -473,6 +519,15 @@ def get_pairings_for_assignment(self, assignment_id):
             results = c.fetchall()
             return results
 
+    def get_pairings_for_run(self, task_group_id):
```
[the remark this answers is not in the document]
JackUrb: Unintentional, adding docstrings

parlai/mturk/webapp/dev/app.jsx
```diff
@@ -0,0 +1,805 @@
...
+      case 'failed':
+      default:
+        return <span>{props.value}</span>;
+    }
+  }
+
+  getColumnValue(header_name, item) {
+    switch(header_name) {
+      case 'run_id': return item.run_id;
+      case 'run_status': return 'complete';
```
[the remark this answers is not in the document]
JackUrb: I should put a TODO here - I haven't been able to write the part that actually computes this value for a HIT yet

parlai/mturk/webapp/server.py
JackUrb: Oh that works a lot better than this. Thanks for the note!

PR #1253, opened 2018-10-26
[MTurk] Universal task saving 'format'

JackUrb:
In order to standardize the way that MTurk files are saved/loaded/etc for the reviewal frontend, the raw outputs for the data need to be put in a location that follows some protocol.

This updates the `qa_data_collection` task to follow such a protocol. The world implements a `get_custom_task_data()` method that packages the task data in a format that will be used later. The workers full chat logs are saved so that they can be loaded later for reviewing. Directories are created in the `run_data` folder to ensure data is saved to locations that can later be retrieved using what is logged in the database.

Enabling automatic data saving is as simple as writing `return world.prep_save_data(workers)` in any task_function or onboarding_function. Data returned by a task or onboarding function is assumed to be of the format returned by `prep_save_data` and is then logged to disk.

### DIFF
--- a/parlai/mturk/core/mturk_data_handler.py
+++ b/parlai/mturk/core/mturk_data_handler.py
@@ -114,7 +114,7 @@ class MTurkDataHandler():
 
     @staticmethod
     def save_world_data(prepped_save_data, task_group_id,
-                        conversation_id, sandbox=False, onboarding=False):
+                        conversation_id, sandbox=False):
         target = 'sandbox' if sandbox else 'live'
         target_dir = os.path.join(
             data_dir, target, task_group_id, conversation_id)
--- a/parlai/mturk/core/mturk_data_handler.py
+++ b/parlai/mturk/core/mturk_data_handler.py
@@ -86,7 +86,6 @@ CREATE_PAIRING_DATA_SQL_TABLE = (
         task_start int,
         task_end int,
         conversation_id string,
-        onboarding_id string,
         bonus_amount int,
         bonus_text string,
         bonus_paid boolean,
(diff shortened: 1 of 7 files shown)


jaseweston:
cool... is it connected to parlaidialogformat also?

JackUrb:
@jaseweston not just yet, this just brings the data to one place. The review scripts I'm building into the frontend will be responsible for converting what is accepted to the right formatting.

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -104,6 +111,28 @@ def __init__(self, task_group_id=None, file_name='pmt_data.db'):
         self.table_access_condition = threading.Condition()
         self.create_default_tables()
 
+    @staticmethod
+    def save_world_data(prepped_save_data, task_group_id,
```
klshuster: just for my understanding, is this the structure of the directory:
run_data/
    sandbox/
    live/
        task_id(of this run?)/
            conversation_id/
                custom_data.json
                worker{id}.json

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -104,6 +111,28 @@ def __init__(self, task_group_id=None, file_name='pmt_data.db'):
...
+                        conversation_id, sandbox=False, onboarding=False):
```
klshuster: Also, I noticed you don't actually use the `onboarding` flag here

JackUrb:
Many tasks will not need to use `get_custom_task_data`, as once I set up the rest of the pipeline, chat data from a conversation will be converted to the ParlAI data format once both halves of a conversation are accepted. If a task defines itself as needing its own 'compiler', it could be useful to ensure the correct data is stored in the episode somehow.

parlai/mturk/core/mturk_data_handler.py
JackUrb: - run_data
  |- sandbox (interior structure same as live)
  |- live
      |- {task_id}
          |- {conversation_id}
               |- custom
                   |- data.json
                   |- data.pickle
               |- workers
                   |- {worker_id}.json

JackUrb: Just realized you had it essentially correct but github dropped out the spaces.

PR #1254, opened 2018-10-28
wmt: display data error

blackredscarf:
There was an error when display wmt data. Because in `scripts/display_data.py`, opt['datatype'] is `train:stream`, file `en_de_train:stream.txt` not found.


alexholdenmiller:
this actually requires a little more subtle fix, but thank you for catching I'll upload a PR in one sec!

alexholdenmiller:
#1259

PR #1256, opened 2018-10-29
verify_data script should stream

emilydinan:
Default to streaming data so that new unit test doesn't have memory issues.


alexholdenmiller:
(ignore build failure unless you want to fix, looks like mturk webapp forgot bookkeeping)

alexholdenmiller:
merging this

PR #1258, opened 2018-10-30
Add Online Static Indicator to ParlAI-MTurk worker view

vtantia:
Objective: Workers using ParlAI-MTurk often send emails when they disconnect because they don't understand why the disconnection has occurred. Being able to provide a simple display location for the current connection status would be useful.

I have done the following for it:
- Created a new measuring variable heartbeat_sends_without_pong to track the number of times the frontend client has sent a message to the router and not received a response back.
- A colored button has been added to be fixed to the top-right of this pane that will display the connection status.
- The button added in 2 displays "connected" (in green) when both measures are fairly low (< 3), "reconnecting to router" (in orange) if heartbeat sends without pong exceeds 3, "reconnecting to server" (in orange) if pongs without heartbeat exceeds 3, and "disconnected" (in red) if the socket is closed.

These are the screenshots showing different status messages.
screen shot 2018-10-30 at 10 31 14 am
screen shot 2018-10-30 at 10 37 04 am
screen shot 2018-10-30 at 10 40 54 am

I have tested the lint and run the tests. 1 test has an error in both the current master and my code. All other tests pass.

### DIFF
--- a/parlai/mturk/core/server/html/core.html
+++ b/parlai/mturk/core/server/html/core.html
@@ -29,6 +29,7 @@
     <div id="right-top-pane" style="width: 100%; height: 570px; padding-top: 60px; padding-left: 20px; padding-right: 20px; padding-bottom: 20px; overflow:scroll; ">
         <div id="message_thread" style="width: 100%">
         </div>
+        <button id="connected-button" class="btn btn-lg" style="position: absolute; top: 2px; right: 2px; opacity: 1; font-size: 11px; color: white;" disabled>connected</button>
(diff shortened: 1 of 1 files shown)


parlai/mturk/core/server/html/core.html
```diff
@@ -523,10 +525,20 @@ <h1>Live Chat</h1>
       // Heartbeats ensure we're not disconnected from the server
       log('received heartbeat: ' + msg.id, 4);
       pongs_without_heartbeat = 0;
+      heartbeats_without_pong += 1;
```
JackUrb: Ah I see the confusion here - heartbeats without pong should refer to _sent_ heartbeats without receiving a pong back.

PR #1259, opened 2018-10-30
fix wmt datatype checking

alexholdenmiller:


alexholdenmiller:
"transient" error in travis

PR #1263, opened 2018-11-01
Torchagent flags

abisee:
Add two flags to TorchAgent to enable access to existing features: 
--use-reply to use either labels or model's own previous replies; 
--add-p1-after-newln to move the __p1__ token to the last newline


parlai/core/torch_agent.py
```diff
@@ -183,6 +183,17 @@ def add_cmdline_args(cls, argparser):
                  'text and __p2__ in front of past labels when available or '
                  'past utterances generated by the model. these are added to '
                  'the dictionary during initialization.')
+        agent.add_argument(
+            '--use-reply', default='label',
+            choices = ['label', 'model'],
+            help='Which previous replies to use as history. If label, use '
+            'gold dataset replies. If model, use model\'s own replies.')
+        agent.add_argument(
+            '--add-p1-after-newln', type='bool', default=False,
```
jaseweston: be good if we can generalize the speaker token to use the id as well if you want (every agent has an id)

alexholdenmiller:
So `--add-p1-after-newln` I left out of the commandline params (to reduce pollution) because I didn't think it was one that you would want to *toggle*: that is, you probably just want to fix it to a specific setting for your model.

The way to do this in that case is to subclass the model, override the method:
```
def get_dialog_history(self, *args, **kwargs):
    kwargs['add_p1_after_newln'] = True
    return super().get_dialog_history(*args, **kwargs)
```

However, if you have a use case where this isn't valid than we can add to the args.


For the replies it seemed like we basically always were using 'model_else_label' before so I just made that the behavior. You want to be able to manually disable using the label from the commandline? Then yeah we can add this.

parlai/core/torch_agent.py
alexholdenmiller: we'll need to make sure we add those ids to our dictionary

alexholdenmiller:
It would be great if you could fix the errors flagged by `./tests/lint_changed.sh`, but the other ones are not your fault--sorry!

stephenroller:
A rebase should make most of the errors go away.

abisee:
@alexholdenmiller:

- For --use-reply, I do think it's useful to have as a toggle. I'm using it right now to run eval_wordstats.py for both use_reply=label and use_reply=model. I'm tracking things like amount of repetition, and my dialog agent does way more repetition when I run use_reply=model (which is much more like the human eval setting). I think it would be useful to have this as a toggle-able flag.
- For --add-p1-after-newln: if you think it's better for people to override the method than that's fine with me. Is there some way we can make it default behaviour for this flag to be on for things like PersonaChat though?

alexholdenmiller:
Hm well okay we can keep them both as flags, but can you mark them as `hidden` flags (just do `hidden=True` in the `add_argument` function call.

One last lint error:
`parlai/core/torch_agent.py:30:1: F841 local variable 'e' is assigned to but never used`

alexholdenmiller:
when you're ready you can click "squash and merge" @abisee!

PR #1265, opened 2018-11-01
Copyright and __init__.py updates.

stephenroller:


stephenroller:
Landed without tests; these were because some dependencies weren't installed. @JackUrb should they be added to the .travis.yml pip install?

JackUrb:
I've been wondering about the fix for this as I haven't looked enough into Travis setup. These deps haven't been required for using ParlAI as a whole but they are required for PMT, and so travis is failing those tests: `pip install boto3 botocore joblib websocket sh`.

JackUrb:
I put the new pip installs in 38e200b8777f83c8715e99ac15221ea4e9c71d02. Thanks for pointing out the location!

PR #1266, opened 2018-11-02
Image-Chat

klshuster:
Adds Image-Chat dataset (https://klshuster.github.io/image_chat/)


klshuster:
Travis fails because of necessary user input (for confirming download of images); merging regardless

PR #1268, opened 2018-11-03
parlai dialog should return copy of its messages, not the (modifiable) original

alexholdenmiller:
torch_agent modifies the message dict that it receives, which for parlai dialog teacher (e.g. fromfile tasks), the original message is modified, becoming corrupted over time e.g. by p1/p2 tokens piling up within the messages


stephenroller:
Will this get in the way of pytorch data teacher?

alexholdenmiller:
The FbDialogTeacher is already creating a new dict for every example return, so I don't think so, but we'll have Kurt review too :)

PR #1270, opened 2018-11-05
Wizard of Wikipedia Task

klshuster:
Adds the Wizard of Wikipedia dataset to ParlAI.

There are 5 teachers in the agents file, each with their own function:
1. `WizardOfWikipediaTeacher` -> returns raw data, is subclassed by all other teachers
2. `BasicDialogTeacher` -> returns _just_ dialog between wizard and apprentice (can specify whether to use dialog from just the wizard's perspective).
3. **`WizardDialogKnowledgeTeacher`** -> returns data for the full task (this is the `DefaultTeacher` for the class)
4. `GeneratorTeacher` -> an extension of the `WizardDialogKnowledgeTeacher` in which the amount/type of knowledge sent in the `action` dict is controllable (this is a consolidated version of the generator teachers done by @stephenroller)
5. `DocReaderTeacher` -> a teacher useful for training a document reader; there are five options, each detailed in the docstring 

I'm including a test to make sure that all the data is loaded properly for each teacher variant.

The link in the `NEWS.md` file can be replaced with the arxiv one once it is online.

### DIFF
--- a/parlai/tasks/wizard_of_wikipedia/agents.py
+++ b/parlai/tasks/wizard_of_wikipedia/agents.py
@@ -6,7 +6,7 @@
 # LICENSE file in the root directory of this source tree. An additional grant
 # of patent rights can be found in the PATENTS file in the same directory.
 """
-    Dialogs from the Wizard of PerZOna Mturk task!
+    Dialogs from the Wizard of Wikipedia dataset
 
     Each episode in the json file has the following fields:
         'wizard_eval': <evaluation of wizard>,
@@ -37,8 +37,7 @@
         'checked_passage': <checked_passage if wizard, else None>
     }
 
-    The 'shown_passages' are lists of 1 entry dicts,
-    mapping a topic to the sentences
+    The 'passages' are lists of 1 entry dicts, mapping a topic to the sentences
 """
 
 from parlai.core.teachers import FixedDialogTeacher
@@ -113,16 +112,16 @@ class WizardOfWikipediaTeacher(FixedDialogTeacher):
     """Gives dialogs where the Wizard has good word overlap
 
         Specify the valid/test split after the last colon in the task, e.g.
-        wizard_of_perZOna:<teacher>:random_split
+        wizard_of_wikipedia:<teacher>:random_split
     """
     def __init__(self, opt, shared=None):
         super().__init__(opt, shared)
         self.opt = opt
         task = opt.get('task',
-                       'wizard_of_perzona:WizardOfPerzona:random_split')
+                       'wizard_of_wikipedia:WizardOfWikipedia:random_split')
         split = task.split(':')
         split = split[2] if len(split) == 3 else 'random_split'
-        opt['task'] = 'wizard_of_perZOna'
+        opt['task'] = 'wizard_of_wikipedia'
         if shared and 'data' in shared:
             self.data = shared['data']
         else:
@@ -631,7 +630,7 @@ class DocReaderTeacher(WizardOfWikipediaTeacher):
         }
 
         if self.teacher_type == 'docs':
-            action['text'] = passage + '\n' + text
+            action['text'] = '{}\n{}'.format(passage, text)
         elif self.teacher_type == 'docs_sentence':
             action['text'] = text
             action['label_candidates'] = self.sent_tok.tokenize(passage)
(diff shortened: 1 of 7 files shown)


jaseweston:
I think this naming is ok after all...  it just means you have to specify
 "-t hello_there" to get HelloThere from the commandline i think..

klshuster:
I'm not sure I understand Jason's comment, but @emilydinan the names would look like:

```
WizardofwikipediaTeacher
WizarddialogknowledgeTeacher
BasicdialogTeacher
GeneratorTeacher
DocreaderTeacher
```

The first two look the most awkward out of those 5. On the plus side, most people will most likely never specify those two teachers on the command line (the first returns raw data, the second is the default teacher and thus can be used with `-t wizard_of_wikipedia`), so it might be worth it to at least change to `Basicdialog` and `Docreader`? Does that make sense?

klshuster:
Per discussion with @emilydinan, I have changed `BasicDialogTeacher` and `DocReaderTeacher` to `BasicdialogTeacher` and `DocreaderTeacher`

parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -0,0 +1,734 @@
...
+            if 'train' in self.datatype:
+                label_cands = []
+            else:
+                label_cands = wizard_entry.get('candidate_responses', [])
+
+        action = {
+            'id': 'WizardDialogKnowledgeTeacher',
+            'text': text,
+            'labels': labels,
+            'chosen_topic': chosen_topic,
```
g-karthik: Is this field actually being used anywhere for training right now?

klshuster: The learning models are not using it, but we did use it in the retrieval system and it's also in the first line of the dialogue. Note that the dataset is out but the full models are not out yet - we are working on getting them into the repository.

g-karthik: When are the full models likely to be out?

Will the code reading the 'knowledge' key in the action dict also be coupled with this release and will there be framework-level support (such as changes to train_step and eval_step in TorchGeneratorAgent or in a new Agent) added for training and consuming separate knowledge/fact encoders?

stephenroller:
Hi Karthik,

We do anticipate releasing the models at some point; presently the code is still research quality, and we prefer to have things cleaned up before a public release.

As far as the coupling, we will likely leave it as project specific for the time being; currently the models support it by overriding parent behavior. We will likely leave this behavior -- I don't think we want this behavior moving into a superclass like TorchAgent at the moment. We are currently discussing paths to a more general way of providing additional context in conversations for more general model support, but we haven't settled on our API yet, so keep an eye on that in the future.

g-karthik:
Thanks for addressing the need for general support for additional conversational context, I will keep an eye out for relevant changes.

Sure, I would also prefer a cleaned up version released, but it would be nice to have it out soon (even the messy version, perhaps in a different branch) in the interest of quick reproducibility of the reported numbers in your paper.

bhedayat:
Is there any update on releasing code for these models? Thank you.

emilydinan:
Hi @bhedayat -- some code and pretrained models have been added here:  https://github.com/facebookresearch/ParlAI/tree/master/projects/wizard_of_wikipedia.

g-karthik:
Thanks Emily, we'll check it out!

bhedayat:
Thank you Emily. I just want to confirm something with you so the EndtoEnd model (https://github.com/facebookresearch/ParlAI/blob/a43f2880719c5a048fdf3d0aa5d5b25eeb9a1a41/projects/wizard_of_wikipedia/generator/modules.py#L39) is inheriting from the TransformerGeneratorModel(https://github.com/facebookresearch/ParlAI/blob/760c638442eaf0ee92f3a9f6adddad5549984837/parlai/agents/transformer/modules.py) which is inheriting from the TorchGeneratorModel (https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/torch_generator_agent.py#L35).

Which eventually calls eval_step to compute the loss (https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/torch_generator_agent.py#L592). And the loss is computed from calling decode_forced (https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/torch_generator_agent.py#L107).

And this loss will be converted to perplexity here (https://github.com/facebookresearch/ParlAI/blob/06015e7dd4c01c4a7cbc43a5b1a01def0222ea77/parlai/core/torch_generator_agent.py#L49)

So my question is are the perplexity numbers reported in your table here (https://github.com/facebookresearch/ParlAI/tree/master/projects/wizard_of_wikipedia), the same one that gets computed from the process i mentioned above?

The reason why im asking is because decode forced takes in the target labels and feeds that that into the model. unless im wrong, in which case please correct me.

stephenroller:
That's correct. Perplexity is computed from forced decoding.

g-karthik:
@emilydinan In the README.md file, the provided command for evaluating your pre-trained end-to-end generative models won't work because there is no End2EndGeneratorTeacher in the task definition. Can you please look into this? Thanks!

emilydinan:
Hey @g-karthik -- thanks for point this out. I'll leave to @stephenroller to fix when he returns from vacation next week.

g-karthik:
Hi @emilydinan, if it is possible to fix it partially before Stephen returns, that would be great!

stephenroller:
@g-karthik Updated in #1580

g-karthik:
Thanks @stephenroller!

PR #1272, opened 2018-11-06
add function for retrieving past model replies in episode

alexholdenmiller:
makes text of model replies for full episode history available to model via torch_agent function

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -791,6 +791,41 @@ class TorchAgent(Agent):
             return batch_reply[self.batch_idx].get('text')
         return None
 
+    def _save_history(self, observations, replies):
+        """Save the model replies to the history."""
+        # make sure data structure is set up
+        if 'predictions' not in self.replies:
+            self.replies['predictions'] = {}
+        if 'episode_ends' not in self.replies:
+            self.replies['episode_ends'] = {}
+        # shorthand
+        preds = self.replies['predictions']
+        ends = self.replies['episode_ends']
+        for i, obs in enumerate(observations):
+            # iterate through batch, saving replies
+            if i not in preds:
+                preds[i] = []
+            if ends.get(i):
+                # check whether *last* example was the end of an episode
+                preds[i].clear()
+            ends[i] = obs.get('episode_done', True)
+            preds[i].append(replies[i].get('text'))
+
+    def reply_history(self):
+        """Get the model's predicted reply history within this episode.
+
+        :param batch: (default False) return the reply history for every
+                      row in the batch, otherwise will return just for this
+                      example.
+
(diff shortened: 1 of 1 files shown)


parlai/core/torch_agent.py
```diff
@@ -791,6 +791,41 @@ def last_reply(self, use_label=True):
...
+        :param batch: (default False) return the reply history for every
+                      row in the batch, otherwise will return just for this
+                      example.
+
+        :return: list of lists of strings, each of the past model replies in
+                 in the current episode. will be None wherever model did not
+                 reply.
+        """
+        # make sure in batch order
+        preds = sorted((b, p) for b, p in self.replies['predictions'].items())
```
uralik: can we make key explicit here? (should import itemgetter from operator or use other some other lambda)
```suggestion
        preds = sorted((b, p) for b, p in self.replies['predictions'].items(), key=itemgetter(0))
```

abisee:
Looks good to me, but I'm not sure what's the use-case where this is useful?

If there's an easy way to get full self.history in the batched case the same as I would in the batch-size 1 case, then that would be super useful for me. AFAIK that doesn't seem to be something you can do with this code. 

What if _save_history didn't just save self.replies, it also saved self.history for the whole batch?

alexholdenmiller:
@abisee what is in self.history that isn't returned from get_dialog_history? that is, batch.text_vecs has the vectorized version of the history already from the batch view. is there something else that would be useful?

but this function is adding the ability to use the true labels in get_dialog_history, but also retrieve the responses the model predicted (it does work in the batch setting)

abisee:
My understanding is: `get_dialog_history` returns an observation with the full history in `observation['text']` (separated by \n), and it also saves the same list of utterances in `self.history`. 

I essentially want to access `self.history`, including the line breaks/utterance separation (for each example in the batch) from my Seq2seqAgent when I'm performing beam search, but that's not possible when batch_size>1. As far as I can tell, the best way around this is to add some extra code to the end of my agent's `batchify` function: I grab `ex['text']` for each `ex` in the `exs` in`batchify`, separate into history (i.e. a list of strings for each ex), and add that to the `Batch` object. 

My [code](https://github.com/fairinternal/ParlAI-Internal/blob/master/projects/seq2plan2seq/controlled_seq2seq/seq2seq.py#L409-L442) to do this is a bit messy though. Perhaps I'm missing something?

uralik:
@abisee , I think as you do with batchify is actually how it should be done if you inherit from TorchAgent and need more functionality. For this change - before there was no way to explicitly get model replies from the whole episode, only the last one, so this will add more flexibility for doing that. E.g. you can change get_dialogue_history to randomly put some model replies on corresponding places during the training etc.

Issue #1271, opened 2018-11-06
Error running interactive.py in quick start tutorial


dalton:
**Describe the bug**

I am working through the [quick tutorial](http://www.parl.ai/static/docs/tutorial_quick.html) and having issues "asking the model a question" myself with this command

```
python examples/interactive.py -mf /tmp/babi_memnn
...
Enter your message: John went to the hallway.\n Where is John?
```

This is the output I get

> ➜  ParlAI git:(master) python examples/interactive.py -mf /tmp/babi_memnn
> 
> [ optional arguments: ] 
> [  display_examples: False ]
> [  display_ignore_fields: label_candidates,text_candidates ]
> [  display_prettify: False ]
> [ Main ParlAI Arguments: ] 
> [  datapath: ./ParlAI/data ]
> [  datatype: train ]
> [  download_path: ./ParlAI/downloads ]
> [  hide_labels: False ]
> [  image_mode: raw ]
> [  numthreads: 1 ]
> [  pytorch_teacher_dataset: None ]
> [  pytorch_teacher_task: None ]
> [  show_advanced_args: False ]
> [  task: None ]
> [ Batching Arguments: ] 
> [  batchsize: 1 ]
> [  include_labels: True ]
> [ PytorchData Arguments: ] 
> [  batch_length_range: 5 ]
> [  batch_sort_cache: none ]
> [  numworkers: 4 ]
> [  pytorch_datafile: None ]
> [  pytorch_preprocess: False ]
> [  shuffle: False ]
> [ ParlAI Model Arguments: ] 
> [  dict_class: parlai.core.dict:DictionaryAgent ]
> [  model: None ]
> [  model_file: /tmp/babi_memnn ]
> [ Local Human Arguments: ] 
> [  local_human_candidates_file: None ]
> [  single_turn: False ]
> [ ParlAI Image Preprocessing Arguments: ] 
> [  image_cropsize: 224 ]
> [  image_size: 256 ]
> [ MemNN Arguments: ] 
> [  embedding_size: 128 ]
> [  hops: 3 ]
> [  init_model: None ]
> [  memsize: 32 ]
> [  position_encoding: False ]
> [  time_features: True ]
> [ TorchAgent Arguments: ] 
> [  add_p1_after_newln: False ]
> [  embedding_projection: random ]
> [  embedding_type: random ]
> [  gpu: -1 ]
> [  gradient_clip: 0.1 ]
> [  history_size: -1 ]
[...]

PR #1274, opened 2018-11-07
during eval revert to vocab when no cands in ranker

alexholdenmiller:
error flagged in #1271, during interactive mode no candidates are available so model crashes

this way model can at least send one-token replies during eval to attempt to answer the question

during training, if a batch happens to have no labels, skip it (some datasets have a final text-only example at the end of episodes with no label)

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -546,7 +546,6 @@ class TorchAgent(Agent):
                 for i, c in enumerate(vecs):
                     vecs[i] = self._check_truncate(c, truncate)
         elif self.rank_candidates and obs.get('label_candidates'):
-            # import pdb; pdb.set_trace()
             obs['label_candidates'] = list(obs['label_candidates'])
             obs['label_candidates_vecs'] = [
                 self._vectorize_text(c, add_start, add_end, truncate, False)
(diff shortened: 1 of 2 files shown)


parlai/core/torch_agent.py
```diff
@@ -545,7 +545,8 @@ def _set_label_cands_vec(self, obs, add_start, add_end, truncate):
                 vecs = obs['label_candidates_vecs']
                 for i, c in enumerate(vecs):
                     vecs[i] = self._check_truncate(c, truncate)
-        elif self.rank_candidates and 'label_candidates' in obs:
+        elif self.rank_candidates and obs.get('label_candidates'):
+            # import pdb; pdb.set_trace()
```
stephenroller: drop

alexholdenmiller:
well, you might only want to use vocab during interactive

stephenroller:
ya but you can still give the -ecands to interactive.py, no?

alexholdenmiller:
Oh I didn't see the ecands arg as separate from cands... yeah thanks

bhancock8:
Yeah, I think I would also lean toward the hard error that lets them know that they need to provide candidates for interactive mode. 'vocab' lets it output something, but it's almost never the best candidate set (e.g., even for babi, they'd rather have the six possible answers, not the 40-something words in the vocab). If we back off to 'vocab' and move along, it's not too hard to imagine someone not paying attention to the one line in the log that says what happened and just being confused why their model works but is outputting short messages.

alexholdenmiller:
how does #1277 look?

PR #1277, opened 2018-11-07
ranker changes

alexholdenmiller:
- still small change in torch_agent
- error message in interactive
- allow skipping batch during training if no labels, with warning printed *every time*

### DIFF
--- a/parlai/scripts/interactive.py
+++ b/parlai/scripts/interactive.py
@@ -52,6 +52,10 @@ def interactive(opt, print_parser=None):
         opt = opt.parse_args()
     opt['task'] = 'parlai.agents.local_human.local_human:LocalHumanAgent'
 
+    if opt.get('eval_candidates', '') in ['inline', 'batch']:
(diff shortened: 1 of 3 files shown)


parlai/scripts/interactive.py
```diff
@@ -52,6 +52,10 @@ def interactive(opt, print_parser=None):
         opt = opt.parse_args()
     opt['task'] = 'parlai.agents.local_human.local_human:LocalHumanAgent'
 
+    if opt.get('eval_candidates', '') in ['inline', 'batch']:
+        print('[ Warning: using candidate type unavailable during interactive.'
+              ' Consider switching to "-ecands vocab" or "-ecands fixed". ]')
```
stephenroller: nit i prefer spelling out long options in error messages `--eval-candidates` .

It seems weird to have this exception for rankers in interactive. should we not have the error message in TorchRankerAgent?

alexholdenmiller:
Adding an interactive flag seems pretty reasonable to me instead as well.

PR #1279, opened 2018-11-08
Fix the CUDA/Hogwild test_seq2seq bug.

stephenroller:
Alternative to #1278, which makes the test pass locally here.


alexholdenmiller:
Wow

PR #1280, opened 2018-11-08
update ranking error message

alexholdenmiller:
the error message that printed before didn't work because changing -cands doesn't change ecands for a loaded model, even if ecands wasn't manually set originally. now it's more obvious to use ecands to change candidate set.


superchargez:
Well the issue #1271 is still creating problems and same results. (I am running windows)

PR #1284, opened 2018-11-14
remove lua torch from readme paragraph

alexholdenmiller:


alexholdenmiller:
ha well we need review to merge so just was waiting for that

PR #1291, opened 2018-11-21
Image Chat MTurk task

klshuster:
Image Chat Mturk task. Pretty similar setup to Personality-Captions. Allows for collecting both the 2nd and 3rd rounds of dialog (specify `--second-response false` for second round of dialog, `true` for third round). 

cc @jaseweston

### DIFF
--- a/parlai/mturk/tasks/image_chat/image_chat_collection/worlds.py
+++ b/parlai/mturk/tasks/image_chat/image_chat_collection/worlds.py
@@ -154,7 +154,7 @@ class ExampleGenerator(object):
 
 
 class RoleOnboardWorld(MTurkOnboardWorld):
-    '''A world that provides the appropriate instructions during onboarding'''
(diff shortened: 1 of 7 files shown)


parlai/mturk/tasks/image_chat/image_chat_collection/worlds.py
```diff
@@ -0,0 +1,420 @@
...
+    def push_example(self, idx):
+        self.idx_stack.append(idx)
+
+    def save_idx_stack(self):
+        with open(self.examples_idx_stack_path, 'wb') as handle:
+            pickle.dump(self.idx_stack, handle)
+
+
+class RoleOnboardWorld(MTurkOnboardWorld):
+    '''A world that provides the appropriate instructions during onboarding'''
```
emilydinan: double quotation

PR #1295, opened 2018-11-27
Random index while streaming PytorchDataTeacher

klshuster:
Allow for random indexing while using `-dt train:stream` with `PytorchDataTeacher` by specifying `--shuffle true`.

Also, adds a test for the `PytorchDataTeacher`. Currently, there is only one test to ensure that the appropriate sampler is used with the various combinations of datatypes and shuffle options. Note that this test file will be expanded in the future.

Here's a table that clarifies what the test is looking for (note that it is always sequential for valid and test):

image

For speed comparison, I "trained" a `repeat_label` agent on the BookTest task, which has ~14 million training examples, for 0.01 epochs (command: `python examples/train_model.py -pyt booktest -dt train:stream -eps .01 -m repeat_label`). Speed results:
- `--shuffle false`: 98.6s
- `--shuffle true`: 44.2s


parlai/core/pytorch_data_teacher.py
```diff
@@ -426,15 +426,32 @@ def __init__(self, opt):
...
     def __getitem__(self, index):
-        while True:
-            idx, ep = next(self.data_gen)
-            if idx == index:
-                return (index, ep)
+        if self.ordered or not self.training:
+            while True:
+                idx, ep = next(self.data_gen)
+                if idx == index:
+                    return (index, ep)
```
alexholdenmiller: when would idx != index?

parlai/core/pytorch_data_teacher.py
```diff
@@ -426,15 +426,32 @@ def __init__(self, opt):
...
+            episode_done = False
+            with open(self.datafile) as f:
+                while not episode_done:
+                    ex_offset = self.char_index[current_idx]
+                    f.seek(ex_offset)
+                    example = json.loads(f.readline())
+                    episode.append(example)
+                    episode_done = example['episode_done']
+                    current_idx += 1
+            return (index, episode)
```
alexholdenmiller: how much slower is this to load a whole epoch? (e.g. with display_data?)

stephenroller: Just to illuminate the use case here: this allows for randomized iteration through a large dataset without loading it into memory. This is helpful for me, as I can pre-BPE tokenize/vectorize a very large corpus, without having to hold the entire thing in memory.

alexholdenmiller: excellent!

parlai/core/pytorch_data_teacher.py
```diff
@@ -23,7 +23,7 @@
 from functools import lru_cache
 try:
     import torch  # noqa: F401
-except Exception as e:
+except Exception:
```
stephenroller: change to ImportError

parlai/core/pytorch_data_teacher.py
```diff
@@ -426,15 +426,32 @@ def __init__(self, opt):
...
+                if idx == index:
+                    return (index, ep)
+        else:
+            episode = []
+            current_idx = index
+            episode_done = False
+            with open(self.datafile) as f:
+                while not episode_done:
+                    ex_offset = self.char_index[current_idx]
+                    f.seek(ex_offset)
```
stephenroller: the `f.seek` call should only be necessary for the first example in the episode, right? Otherwise they're sequential, so no seek is necessary? That would slightly speed it up.

klshuster: yes, that is a good point - I rewrote to simplify the code, but it did not actually result in any speedup for me

parlai/core/pytorch_data_teacher.py
klshuster: will do

PR #1297, opened 2018-11-27
Sharing task react components

JackUrb:
This PR establishes a more reusable way to handle react component management. It comes in two forms, displayed in the tasks `react_custom_no_extra_deps` and `react_custom_with_deps`. These two folders are mostly just the old `react_task_demo` task duplicated in two forms to display how tasks can be configured in these formats. Here's how they work:

#### `react_custom_no_extra_deps`
- Requires creating a `frontend/components` folder in the task directory. Custom components are stored in a `custom.jsx` file.
- Upon running, the `custom.jsx` file replaces the default (empty) `custom.jsx` file in a copy of the `react_server/dev/components` folder that is being used to make the heroku server setup. 
- `custom.jsx` is compiled by webpack with the rest of the frontend
- `CustomComponents` are loaded out of the `custom.jsx` file.
- The frontend uses `CustomComponents` to render content

#### `react_custom_with_deps`
- Requires creating a `frontend/package.json` file in the task directory. Custom components should then be built into a `custom.jsx` file in `frontend/components`. I should provide templates for the `.babelrc` and `package.json` and `webpack.config.js` files in the future, but for now they can be copied from the `react_custom_with_deps` task folder.
- Custom javascript requirements can be installed using `npm install <package>` from within the frontend directory. They can then be referred to in `custom.jsx`
- Upon running, the `frontend` folder replaces the dummy `built_custom_components` folder in a copy of the `react_server/dev/components` folder that is being used to make the heroku server setup.
- `npm install` and `npm run dev` is called on the `built_custom_components` folder to generate the output javascript packed with the special dependencies. The final packed component is expected to be sent to `built_custom_components/dist/custom.jsx` (configured in `webpack.config.js`). This creates a dependency in the main frontend on a package titled `custom_built_frontend` which points to the custom javascript.
- `built_custom_components/dist/custom.jsx` is compiled by webpack into the rest of the frontend
- `BuiltCustomComponents` are loaded out of the compiled package.
- The frontend uses `BuiltCustomComponents` to render content.

###  Summary of changes to achieve the above behavior
- `react_task_demo` split into two files. Actual task content hasn't changed.
- `MTurkManager` now checks for existence of a `package.json` file when compiling `task_files_to_copy` in order to set a flag for custom built frontend components.
- added a dummy `built_custom_components` folder and package to the default `react_server` so that the frontend has _something_ to import even when the task doesn't define any built custom components.
- Moved actual display components out from the main `react_server/dev/app.jsx` file into a `core_components.jsx` file so that they can be imported in the parlai frontend webapp in the future. Also cleans up the logic for `app.jsx`.
- Updated the logic by which `core_components` are swapped out for `UseCustomComponents` (defined as in one of the two above flows, built components take preference). New flow allows two components on the same page to render different views depending on agent id. This is achieved by setting `XComponentName` inside individual components based on `agent_id` rather than globally.
- updated `server_utils.py` to properly handle the above two build scenarios separately.
- added node build files to the `react_custom_with_deps` task to demonstrate the flow (and to eventually template out from).

### Other changes that snuck into this branch
- Frontend `ChatMessage` objects are now provided the `context` field from any `act`'s `context` field.
- `duration` is now recorded for every blocking act message, and is calculated as the time from when the message was requested to when it was sent.
- `handleQueueMessage` now takes `is_system` argument to be able to send messages from the system that won't spoil the `duration` tag.
- Bumped `jQuery` versions.
- Working time logs are no longer checked when running on sandbox.


parlai/mturk/tasks/react_task_demo/react_custom_with_deps/worlds.py
```diff
@@ -0,0 +1,163 @@
...
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+from parlai.mturk.core.worlds import MTurkOnboardWorld, MTurkTaskWorld
+import threading
+
+
+class AskerOnboardingWorld(MTurkOnboardWorld):
+    '''Example onboarding world. Sends a message from the world to the
```
emilydinan: nit: elsewhere we use double quotation marks for doc strings

parlai/mturk/tasks/react_task_demo/react_custom_no_extra_deps/run.py
```diff
@@ -0,0 +1,147 @@
...
+                    use_workers.append(worker)
+                    filled_roles.append(worker.demo_role)
+            return use_workers
+
+        eligibility_function = {
+            'func': check_workers_eligibility,
+            'multiple': True,
+        }
+
+        # Assign worker roles is used to determine what the role each worker
```
emilydinan: all of these comments are really helpful!

JackUrb:
Custom builds will _probably_ prefer to just include all components. There's a way to share react and others, but it's somewhat complicated to configure and I feel this build system is probably complicated enough as is so I don't use that behavior by default.

JackUrb:
Merging with Travis errors as I still haven't fixed travis runs of MTurk tests, but they pass locally.

PR #1298, opened 2018-11-29
[MTurk] Making notification sound on COMMAND_SEND_MESSAGE

JackUrb:
Includes new functionality for the react frontend that pings when the user is supposed to respond to a message. Hopefully this will prevent having distracted workers for too long, leaving their partners hanging. 

Also includes a volume control for the ping sound.

### Screenshots
screen shot 2018-11-28 at 8 47 39 pm
screen shot 2018-11-28 at 8 47 46 pm
screen shot 2018-11-28 at 8 47 53 pm

(They're kinda boring because you can't hear the sound...)

### Bugs to solve before commit:
- [x] volume control only properly fits in the top with the status indicator when "connected". I need to put the two into the same fixed position div rather than having two separate ones.
- [x] Custom components that are calling focus on the input call them on every update, including the state changes induced by changing the volume slider. I need to make focusing more deliberate.


parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -82,10 +82,17 @@ class MainApp extends React.Component {
       task_done: false,
       messages: [],
       agent_id: 'NewWorker',
-      context: {}
+      context: {},
+      volume: 1
```
klshuster: is 1 the lowest or highest volume?

JackUrb:
@stephenroller normally I'd agree that the convenience of a mute button is nice, but in this case I'd rather it be a more involved process for one worker to shut out another worker.

parlai/mturk/core/react_server/dev/app.jsx
JackUrb: 1 is max volume

JackUrb:
New look with a navbar style to put components together. Could be useful in the future for any additional functionality (such as off-task chatting).
screen shot 2018-11-29 at 4 43 04 pm

PR #1299, opened 2018-11-30
View custom components in review frontend

JackUrb:
Changes the architecture for the frontend server to handle the build system designed in #1297. This lets the frontend render components custom to tasks, regardless of if those components required an extra build step. *This is the precursor to being able to demo a task in this frontend.*

Existing flows still work:
screen shot 2018-11-30 at 3 45 52 pm

#### Change list
- Pull task description from custom components, allowing it to be anything a task specifies
- Search task directories recursively to allow for cases where tasks are subtasks of some other folder (ie: `react_task_demo/react_custom_no_deps`). Use this to find the task config later
- Build task components (when building is required) and copy them to the server directory to build the full server
- Add duration view to messages whenever they have the `is_review` prop.

#### Tests
- [x] Launch the frontend for tasks with and without custom frontends
- [x] launch mturk tasks with and without custom frontends


parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -35,6 +35,16 @@ class ChatMessage extends React.Component {
...
+    let duration = null;
+    if (this.props.duration !== undefined) {
+      let duration_seconds = Math.floor(this.props.duration / 1000) % 60;
+      let duration_minutes = Math.floor(this.props.duration / 60000);
+      let min_text = duration_minutes > 0 ? duration_minutes + ' min' : '';
+      let sec_text = duration_seconds > 0 ? duration_seconds + ' sec' : '';
+      duration = <small>
+        <br /><i>Duration: </i>{min_text + ' ' + sec_text}
+      </small>;
+    }
```
klshuster: so this shows the amount of time a turker took to write a message?

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -522,6 +526,26 @@ class LeftPane extends React.Component {
   }
 }
 
+class LeftPane extends React.Component {
+  render () {
+    let v_id = this.props.v_id;
```
klshuster: i notice that you call this `v_id` but elsewhere it's `agent_id`, does this mean something like `viewer_id`?

parlai/mturk/webapp/server.py
```diff
@@ -482,6 +496,85 @@ def start_server(port=DEFAULT_PORT, hostname=DEFAULT_HOSTNAME,
...
+    copy_dirs = {}
+    parlai_task_dir = os.path.join(parlai_path, 'mturk', 'tasks')
+    tasks.update(crawl_dir_for_tasks(parlai_task_dir))
+
+    if parlai_int_path is not None:
+        parlai_internal_task_dir = \
+            os.path.join(parlai_int_path, 'mturk', 'tasks')
+        tasks.update(crawl_dir_for_tasks(parlai_internal_task_dir))
+
+    print (tasks)
```
klshuster: nit: necessary print statement?

parlai/mturk/core/react_server/dev/components/core_components.jsx
JackUrb: Yup, but only in review mode. Turkers can't see this

parlai/mturk/core/react_server/dev/components/core_components.jsx
JackUrb: Yes - `v_id` is like the MTurk equivalent of viewer context. When on MTurk it will always be agent_id, but in the react frontend in order to demo tasks you need to be able to swap between multiple `v_id`'s

parlai/mturk/webapp/server.py
JackUrb: Necessary for my debugging in the next step, but on the chopping block to be removed after

PR #1301, opened 2018-12-03
look for local dict file, this was set wrong

jaseweston:


jaseweston:
is torch agent etc. doing this too?  all models should be doing this..

stephenroller:
Uh travis said this had syntax errors....

PR #1307, opened 2018-12-07
fix for batched executable worlds

ashot:
We're using the wrong index here, fails if batch size > num agents.


ashot:
Sorry got my branches crossed.  Fixed now, it's a one-liner.

PR #1308, opened 2018-12-08
[MTurk] Demo react tasks in webapp frontend

JackUrb:
## Motivation
As of now it takes time to restart the PMT sandbox when attempting to iterate on a HIT's design. Any changes to the world or to the frontend require ~2 mins of build and launch time, and require copying links and opening tabs on MTurk sandbox. This PR adds functionality to be able to launch 'HITs' in the ParlAI webapp to demo the functionality of the task. The key goal of this was to be able to run these hits without a user having to have multiple `run.py` files or any special flags at all for that matter. Tasks can now be demo and worked on from within the webapp, and all of it is done in one browser tab.

### Screenshots
screen shot 2018-12-07 at 6 52 10 pm
screen shot 2018-12-07 at 6 53 42 pm
screen shot 2018-12-07 at 6 53 51 pm

### Implementation Overview
- `MockTurkManager` and `MockTurkAgent` classes were created to mimic the interface/api of the regular `MTurkManager` and `MTurkAgent` classes while using an almost completely different local backend. Onboarding/task/pairing functionality is equivalent, but the rest has been entirely modified.
- Two new endpoints are added to the webapp server: `TaskSocketHandler` for instantiating the messaging socket, `TaskRunHandler` for launching the run of a task.
- The server launch was updated to first copy over all of the frontend content into local directories, and then the frontend is built. As of now this build step is now the slowest part of the process for running demo tasks, but I have plans to revisit the build to improve this in the future.

### Additional bug fixes
- Adds `jquery` import to all of the react frontend components, as this script should be included in the build rather than in the html.
- Changes default duration from `null` to `undefined` to preserve behavior of dropping duration from messages when it doesn't exist or when `is_review` is false.
- Changes imports of `task_config` to use the full module path import, as these `run` files are now being imported by the server on demand, and the relative module paths no longer work.

### Known issues I'm currently solving
- [x] Autoscroll doesn't happen on the webapp frontend when receiving messages.
- [x] Only way to get to a task demo page is directly by entering the url, there's no way to get there yet otherwise.
- [x] The system does not handle task_done state, so when the task is finished the system hangs
- [ ] ~~Only way to rebuild a particular frontend is to restart the whole server. It also rebuilds everything rather than only files and folders that have actually been changed, making the estimated feedback loop build time around 15 seconds, and this will only grow with more tasks.~~  (This will be a followup PR sometime in the future)

### DIFF
--- a/parlai/mturk/tasks/react_task_demo/react_custom_with_deps/run.py
+++ b/parlai/mturk/tasks/react_task_demo/react_custom_with_deps/run.py
@@ -10,7 +10,8 @@ from parlai.mturk.tasks.react_task_demo.react_custom_with_deps.worlds import \
     AskerOnboardingWorld, AnswererOnboardingWorld, EvaluatorOnboardingWorld, \
     MultiRoleAgentWorld
 from parlai.mturk.core.mturk_manager import MTurkManager
-from task_config import task_config
+from parlai.mturk.tasks.react_task_demo.react_custom_with_deps.task_config \
+    import task_config
 import os
 
 
--- a/parlai/mturk/webapp/dev/app.jsx
+++ b/parlai/mturk/webapp/dev/app.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {ToggleButtonGroup, ToggleButton, Button, FormControl,
-  ButtonGroup, ButtonToolbar, Panel, Table, Modal, InputGroup} from 'react-bootstrap';
-import {getCorrectComponent, setCustomComponents} from './task_components/core_components.jsx';
+  ButtonGroup, ButtonToolbar, Panel, Table, Modal, InputGroup,
+  Nav, NavItem} from 'react-bootstrap';
+import {BaseFrontend, getCorrectComponent, setCustomComponents} from './task_components/core_components.jsx';
 import ReactTable from "react-table";
 import 'react-table/react-table.css';
 import 'fetch';
@@ -11,7 +12,7 @@ import 'fetch';
 setCustomComponents({});
 
 var AppURLStates = Object.freeze({
-  init:0, tasks:1, unsupported:2, runs:3, workers:4, assignments:5,
+  init:0, home:1, unsupported:2, runs:3, workers:4, assignments:5, tasks: 6,
 });
 
 function convert_time(timestamp){
@@ -40,6 +41,7 @@ function resolveState(state_string) {
 }
 
 function postData(url = ``, data = {}) {
+  console.log('posting', data, url);
   return fetch(url, {
       method: "POST",
       mode: "cors",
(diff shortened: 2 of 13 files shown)


JackUrb:
screen shot 2018-12-10 at 6 13 03 pm
View for tasks on the home page

parlai/mturk/webapp/run_mocks/mock_turk_agent.py
```diff
@@ -0,0 +1,311 @@
...
+class MockTurkAgent(Agent):
+    """Mock turk agent that can act in a parlai mturk world"""
+
+    # MTurkAgent Possible Statuses
+    ASSIGNMENT_NOT_DONE = 'NotDone'
+    ASSIGNMENT_DONE = 'Submitted'
+    ASSIGNMENT_APPROVED = 'Approved'
+    ASSIGNMENT_REJECTED = 'Rejected'
+
+    MTURK_DISCONNECT_MESSAGE = MTURK_DISCONNECT_MESSAGE
```
klshuster: is there a reason to redefine these here?

parlai/mturk/tasks/react_task_demo/react_custom_with_deps/run.py
```diff
@@ -10,7 +10,8 @@
     AskerOnboardingWorld, AnswererOnboardingWorld, EvaluatorOnboardingWorld, \
     MultiRoleAgentWorld
 from parlai.mturk.core.mturk_manager import MTurkManager
-from task_config import task_config
+from parlai.mturk.tasks.react_task_demo.react_custom_with_deps.task_config \
```
emilydinan: since this is a demo, maybe you could leave a comment here noting why we have to use the full path to import the task_config

parlai/mturk/webapp/run_mocks/mock_turk_agent.py
JackUrb: It's so that external files that haven't imported the `MTurkAgent` file can pull them from the agent itself. I should really be pulling these from said `MTurkAgent` file though!

parlai/mturk/webapp/dev/app.jsx
```diff
@@ -1525,6 +1683,265 @@ class WorkerListPanel extends React.Component {
...
+      }
+      let curr_worker = curr_worker_data[w.worker_id];
+      curr_worker.messages = curr_worker.messages.concat(w.new_messages);
+      curr_worker.task_done = w.task_done;
+      curr_worker.done_text = w.done_text;
+      curr_worker.world_state = w.status;
+      curr_worker.chat_state = chat_state;
+      curr_worker.agent_id = w.agent_id;
+      if (w.all_messages.length > curr_worker.messages.length) {
+        curr_worker.messages = w.all_messages;
```
emilydinan: can you explain what this is doing here?

parlai/mturk/webapp/run_mocks/mock_turk_agent.py
```diff
@@ -0,0 +1,311 @@
...
+    def reduce_state(self):
+        """Cleans up resources related to maintaining complete state"""
+        self.flush_msg_queue()
+        self.msg_queue = None
+
+    def get_new_act_message(self):
+        """Get a new act message if one exists, return None otherwise"""
+        # See if any agent has disconnected
+        if self.disconnected or self.some_agent_disconnected:
+            msg = {
```
emilydinan: this format is used a lot, you could just have some constant "MSG" and edit the text, returning as needed

parlai/mturk/tasks/react_task_demo/react_custom_with_deps/run.py
JackUrb: Ultimately we import the rest of the things from the full path, and the fact that we haven't been doing the same for the task_config was an artifact of copypasting one of the first iterations this far. I'll note the reasoning in the docs, but I'm going to end up migrating all of our tasks over.

parlai/mturk/webapp/dev/app.jsx
JackUrb: Added comment. Basic idea is that if there are more messages in the thread than we think there are on the frontend, we sync to the full thread. We only choose to not use it all the time because sending messages wouldn't be able to be added to the thread until the send is received, which can sometimes take a second or so and looks clunky.

parlai/mturk/webapp/run_mocks/mock_turk_agent.py
JackUrb: I'm going to subclass the real MTurkAgent and standardize this because it can be cleaner for sure.

PR #1311, opened 2018-12-11
Talk the Walk Datasets

ashot:
Putting up a very early version for comments.

Refactors the ttw data loading to fit ParlAI pattern. 
Adds teacher, world, tourist agent and guide agent.

Data loader feeds in an initial observation to start episodes (neighborhood, start_location, landmarks), and then each individual piece of the dialog as an example.  We rely on a custom World to pull out landmark features, move the agent, and generate observation features, and on episodic memory in the agent to generate trainable examples.  The goal is to move as much of the logic out of the loader/teacher as possible.  This is kind of an extreme setup though so perhaps a middle ground would be better, I'm not sure.

One consequence of this is that we have to gate some of the training until we have built up some memory.  More generally the agents have to be a bit more robust and can assume less about the examples.

Guide and Tourist models (imported from ttw repo) train loops run, as well as the built in seq2seq model.

To run:
python examples/train_model.py -t talkthewalk -m parlai.tasks.talkthewalk.agents:TouristAgent -nw 0 -bs 128
python examples/train_model.py -t talkthewalk -m parlai.tasks.talkthewalk.agents:GuideAgent -nw 0 -bs 128
python examples/train_model.py -t talkthewalk -m seq2seq -nw 0 -bs 128


ashot:
moving away from custom models, now piping observation/map data into the text so we can use built-in parlai models to train.

train tourist (including actions):
`python examples/train_model.py -t talkthewalk:tourist -m seq2seq -nw 0 -bs 200 -mf /tmp/tourist --train-actions 1 -eps 1`

train guide (including actions):
`python examples/train_model.py -t talkthewalk:guide -m seq2seq -nw 0 -bs 200 -mf /tmp/guide --train-actions 1 -eps 1`

have tourist and guide models play each other in a simulation world: (no training here yet)
`python parlai/tasks/talkthewalk/run.py -tmf /tmp/tourist -gmf /tmp/guide`

alexholdenmiller:
@klshuster can you take a quick look as well with
```
git checkout -b ashot-talk master
git pull git://github.com/ashot/ParlAI.git talk

py examples/display_data.py -t talkthewalk:tourist --train-actions true
py examples/display_data.py -t talkthewalk:guide
py examples/display_data.py -t talkthewalk:guide_localize
```

parlai/tasks/talkthewalk/worlds.py
```diff
@@ -0,0 +1,259 @@
...
+
+        if agents:
+            self.send_map(self.guide)
+            self.send_location(self.tourist)
+
+    def send_map(self, agent):
+        agent.observe({'text': self.sim.get_text_map()})
+
+    def send_location(self, agent):
+        agent.observe({'text': self.sim.get_current_view()})
```
klshuster: should this be `self.sim.get_agent_location` or did you mean to send the view that the agent sees?

parlai/tasks/talkthewalk/worlds.py
```diff
@@ -0,0 +1,259 @@
...
+    def observe(self, agent, act):
+        self.sim.add_view_to_text(act)
+        return act
+
+
+class Map(object):
+    """Map with landmarks"""
+
+    def __init__(self, data_dir, include_empty_corners=True):
+        super(Map, self).__init__()
```
klshuster: nit: `super().__init__()`

klshuster:
I think the display data stuff looks good

parlai/tasks/talkthewalk/agents.py
```diff
@@ -0,0 +1,155 @@
...
+        self.episodes = json.load(open(datafile))
+        self.data = []
+        self.examples_count = 0
+
+        for episode in self.episodes:
+            init = {x:y for x,y in episode.items() if x in ['start_location',
+                'neighborhood', 'boundaries', 'target_location']}
+            self.sim.init_sim(**init)
+            if episode:
+                episode = self._setup_episode(episode)
```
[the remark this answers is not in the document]
ashot: its an abstract base class for the other two. added a function stub.

parlai/tasks/talkthewalk/agents.py
```diff
@@ -0,0 +1,157 @@
...
+    return opt['ttw_data'], os.path.join(opt['ttw_data'], 'talkthewalk.' + dt + '.json')
+
+
+class TTWTeacher(FixedDialogTeacher):
+
+    @staticmethod
+    def add_cmdline_args(argparser):
+        agent = argparser.add_argument_group('Talk the Walk Teacher Arguments')
+        agent.add_argument('--train-actions',
+                type=bool, default=False, help='Train model to take actions')
```
alexholdenmiller: 'bool'

alexholdenmiller:
can you fix the errors in `./tests/lint_changed.sh`, add the copyright notice to the files, and change the arg type to 'bool' (a string) and then we'll merge!

ashot:
sounds good

alexholdenmiller:
we can merge past a "time limit exceeded error" if they pass on your machine but should fix the lints again :)

parlai/core/worlds.py
```diff
@@ -600,14 +600,14 @@ def batch_observe(self, index, batch_actions, index_acting):
...
+            if index == index_acting:
+                return None  # don't observe yourself talking
             if hasattr(w, 'observe'):
                 # The world has its own observe function, which the action
                 # first goes through (agents receive messages via the world,
                 # not from each other).
                 observation = w.observe(agents[index], validate(batch_actions[i]))
             else:
-                if index == index_acting:
-                    return None  # don't observe yourself talking
```
alexholdenmiller: I think actually we don't want this code moved in case an ExecutableWorld wants to send you something based on your action (it could choose not to by doing this check there)

alexholdenmiller:
but I think that's the last thing!

PR #1312, opened 2018-12-11
Add Daily Dialog task.

stephenroller:


parlai/tasks/dailydialog/parse.py
```diff
@@ -0,0 +1,131 @@
...
+        ("\u2014", "--"),
+        ("\u2013", "--"),
+        ("\u3002", ". "),
+        ("\u2032", "'"),
+        ("\u3001", ", "),
+    ]
+    for before, after in SWITCH_LIST:
+        text = text.replace(before, after)
+
+    # fix some shitty sentence tokenization
```
klshuster: nit: maybe clean up this comment lol

parlai/tasks/dailydialog/agents.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _setup_data(self, fold):
+        self.data = []
+        fpath = os.path.join(self.opt['datapath'], 'dailydialog', fold + '.json')
+        with open(fpath) as f:
+            for line in f:
+                self.data.append(json.loads(line))
+
+    def get(self, episode_idx, entry_idx=0):
+        # we play both sides!
```
klshuster: can you elaborate on how this setup enforces playing both sides?

stephenroller: So the conversation doesn't have a canonical Speaker 1 and Speaker 2, and the evaluation in the original paper seems to generally evaluate responses. So in this Teacher, I've made it that on even episode_idx, we play Speaker 1, and on odd episode_idx, we play Speaker 2, so you learn from both sides of the conversation.

emilydinan: Instead of doing this, shouldn't we effectively double the data by having each episode appear twice (once where we are Speaker1, and once where we are Speaker2)? This is what I did for ConvAI2

stephenroller: I do, I just don't store the data twice.

emilydinan: I see...is the number of episodes odd? Otherwise the speaker ID will remain the same for each episode

stephenroller: No, b/c:

https://github.com/facebookresearch/ParlAI/blob/ff72628bc6f3832704b4f2a74660846e94243e42/parlai/tasks/dailydialog/agents.py#L51

So I "double" the amount of episodes in the data, then alternate between which speaker I play

klshuster: Ahh ok I think I see, it's the `episode_idx // 2` that makes it so you see each example twice right?

PR #1316, opened 2018-12-13
Switch from ProgressLogger to tqdm.

stephenroller:


alexholdenmiller:
travis will only time out, so def rerun unit tests on your own. lintchanged is also failing.

stephenroller:
Local tests are fine, ignoring travis.

PR #1317, opened 2018-12-13
PytorchDataTeacher Integration Tests

klshuster:
Added a bunch of tests for the `PytorchDataTeacher` to ensure that everything (or anything I could think worth testing) is working.

The tests ensure that:
1. The teacher shuffles and does not shuffle data with appropriate `datatype` and `shuffle` flags.
2. A seq2seq model can be trained successfully on `babi:task10k:1`, with and without streaming, with and without random sampling of training episodes.
3. Data built via `PytorchDataTeacher` with `--pytorch-preprocess true -m seq2seq` yields the same examples as a processed observation by the `seq2seq` model with `--pytorch-preprocess false` (preprocessing data yields same data as before)
4. Batch sorting yields all episodes for two straight epochs (the first is loading all episodes into memory; the second is with cached episodes)
5. A seq2seq model can be trained successfully on `babi:task10k:1`, with and without streaming, with and without preprocessing by the `PytorchDataTeacher`, **with batch sorting enabled**
6. When using a PyTorch `Dataset` with the `PytorchDataTeacher`, the acts are the same as when using the default task teacher (in this case, I've used `flickr30k` as an example)
7. Batch sorting _actually_ sorts and returns batches with similarly sized examples, when using both the `text` field and using the `text_vec` field.
8. Multitask functionality works with the `PytorchDataTeacher`

Some things I learned/fixed along the way:
1. The `PytorchDataTeacher` now multitasks with every possible combination of `-pyt`, `-pytd`, and `-t` (actually did not yet check using all three...)
2. The teacher now enforces one of the two ways of batch sorting when using stream data (to ensure that all episodes are seen each epoch)
3. `FixedDialogTeacher.batch_act` is no longer deprecated (the `PytorchDataTeacher` utilizes this during batched training)
4. **Very Important** - the dictionary when building the pytorch data with `--pytorch-preprocess true` was _different_ from the eventual dictionary used by the agent. This is because the episodes are drawn out and repeated (i.e. a 6-example episode becomes 6 separate examples) which messes with the counts of the tokens and ultimately with the `ind2tok`; thus, I now ensure that, if `--pytorch-preprocess` is true, we use the same dictionary that was used when building the data. 

TODO:

- [x] Cleanup some artifacts I introduced in the code from testing all these cases (possible stray comments, pdbs, etc)
- [x] Ensure that the whole test suite runs to completion with no errors 😃


klshuster:
Ready for review

klshuster:
Pulled in changes from #1351, confirmed that nothing breaks with test suite

stephenroller:
How come you made the unit tests an integration test?

klshuster:
Most of the tests in that file are integration tests so it seemed better for me to just move the whole set to `integration/` rather than split up the tests between two files. Basically, I was not sure how to resolve that cleanly

stephenroller:
The main difference right now is what travis runs every time, vs what we run when we think to do so.

klshuster:
Ok, I think what I can do then is move this back to unit tests, and then just skip the long-running tests (specifically the training ones) when Travis is run (the tests take about a half hour to run otherwise)

tests/test_pytorch_data_teacher.py
```diff
@@ -130,7 +130,11 @@ def get_teacher_act(defaults, teacher_processed=False, agent_to=None):
             val1 = agent_processed_observation[key]
             val2 = teacher_processed_act[key]
             if type(val1) is torch.Tensor:
-                self.assertTrue(bool(torch.all(torch.eq(val1, val2))))
+                self.assertTrue(bool(torch.all(torch.eq(val1, val2))),
```
stephenroller: Should this be torch.allclose instead? I don't know what the type of var1, var2 are off the top of my head.

klshuster: Ahh that's a useful function, but in this case val1 and val2 are LongTensors so I don't think it is necessary

klshuster:
Travis, while running my unit tests, caught a bug that my local repo did not catch due to stale data. Re-running integration tests to ensure that everything works, then will merge

PR #1318, opened 2018-12-17
[Messenger] sync with internal change

emilydinan:
Sync with internal change.


JackUrb:
After fixing lint this is good to go!

PR #1319, opened 2018-12-18
Distributed Training

stephenroller:
Support training across many GPUs on many possibly hosts using `torch.distributed`. This works for both one-host-many-gpu and many-host-many-GPU.

Fairseq distributes by sharding the data set. I looked into this and while we can do it, it doesn't seem helpful since we work with sampling with replacement. I've implemented validation/test as depend-on-main-host instead. This slows down validation (which would otherwise be get n-gpu x speedup), but does simplify a lot of our logic here.

TODO:
- [x] ~~warmup LR scheduler (pretty important for these very large batch sizes)~~ Going to do this in another PR
- [x] true distributed training
- [x] ~~a launch script~~ distributed launch script will be in internal. I'll include an example sbatch in the docs.
- [x] combine metrics from all hosts; epochs/exs is particularly wrong right now.
- [x] mute the helper hosts?
- [x] ~~support rankers (right now, the wrapper is only for TorchGeneratorAgent. probably will require some refactoring)~~ fix by just copy-pasting 3 lines, less DRY but...
- [x] ~~adjust epochs by the number of hosts.~~
- [x] handle ctrl-c
- [x] undo the sharding decision
- [x] fix up time_left
- [x] probably need some exception handling
- [x] more testing
- [x] example sbatch in docs.


parlai/core/torch_agent.py
```diff
@@ -402,6 +402,10 @@ def _copy_embeddings(self, weight, emb_type, log=True):
         :param weight:   weights of lookup table (nn.Embedding/nn.EmbeddingBag)
         :param emb_type: pretrained embedding type
         """
+        if self.get('rank', 0) != 0:
```
stephenroller: change to `is_primary_worker`

parlai/core/torch_generator_agent.py
```diff
@@ -325,6 +325,13 @@ def __init__(self, opt, shared=None):
                 optim_states=states.get('optimizer'),
                 saved_optim_type=states.get('optimizer_type'))
 
+        if shared is None and opt.get('distributed_world_size'):
```
stephenroller: change to `is_distributed`

parlai/scripts/distributed_train.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python
```
stephenroller: rename to multiprocessing_train

stephenroller:
Ready for review.

parlai/core/distributed_utils.py
```diff
@@ -0,0 +1,201 @@
...
+    Returns False if we are a secondary worker. Returns True if we are either
+    (1) not in distributed mode (2) or are the primary (rank 0) worker.
+    """
+    return not is_distributed() or dist.get_rank() == 0
+
+
+def override_print(suppress=False, prefix=None):
+    """
+    Overrides the builtin print, to either mute or annotate the output of
+    individual workers. Mutes if suppress is True, and this is not the
```
klshuster: nit: does it suppress if mute is true AND this is not the primary/only worker? or dies it suppress if mute is true OR this is not the primary/only worker?

parlai/core/distributed_utils.py
```diff
@@ -0,0 +1,201 @@
...
+
+    buffer_rank = buffer[rank * max_size: (rank + 1) * max_size]
+    buffer_rank[0] = enc_size // 255  # this encoding works for max_size < 65k
+    buffer_rank[1] = enc_size % 255
+    buffer_rank[2: enc_size + 2] = torch.ByteTensor(list(enc))
+
+    dist.all_reduce(buffer)
+
+    result = []
+    for i in range(world_size):
```
klshuster: I notice in the fairseq implementation, they catch the unpickling error and raise their own exception. Did you intentionally omit that?

parlai/scripts/train_model.py
```diff
@@ -107,32 +110,44 @@ def setup_args(parser=None):
...
+
+    valid_world = create_task(opt, valid_agent)
+    return valid_world
+
+
+def run_eval(valid_world, opt, datatype, max_exs=-1, write_log=False):
+    """
+    Eval on validation/test data.
 
+    :param valid_world: the pre-created validation world.
```
klshuster: love how clean this looks

parlai/core/distributed_utils.py
stephenroller: I will update the docstring correspondingly.

parlai/core/distributed_utils.py
stephenroller: That's newer than the version of the code I was looking at. It's a good idea; I'll add it.

stephenroller:
No, it shouldn't do that. Admittedly ctrl-c sometimes has problems in distributed mode

parlai/core/distributed_utils.py
```diff
@@ -0,0 +1,220 @@
...
+        all_gather_list._buffer.numel() < buffer_size
+    ):
+        all_gather_list._buffer = torch.cuda.ByteTensor(buffer_size)
+
+    buffer = all_gather_list._buffer
+    buffer.zero_()
+
+    enc = pickle.dumps(data)
+    enc_size = len(enc)
+    if enc_size + 2 > max_size:
```
emilydinan: what is the +2 here? can you list enc_size and max_size in the error statement

parlai/scripts/train_model.py
```diff
@@ -107,32 +110,44 @@ def setup_args(parser=None):
...
-    - Agent is the agent to use for the evaluation.
-    - opt is the options that specific the task, eval_task, etc
-    - datatype is the datatype to use, such as "valid" or "test"
-    - write_log specifies to write metrics to file if the model_file is set
-    - max_exs limits the number of examples if max_exs > 0
-    - valid_world can be an existing world which will be reset instead of reinitialized
-    """
-    print('[ running eval: ' + datatype + ' ]')
+def _maybe_load_world(agent, opt, datatype):
+    if not is_primary_worker():
```
emilydinan: this was a bit confusing without the rest of the context. maybe call it `_maybe_load_eval_world`?

parlai/scripts/train_model.py
```diff
@@ -220,17 +235,26 @@ def __init__(self, opt):
...
-            self.agent, opt, 'valid', opt['validation_max_exs'],
-            valid_world=self.valid_world)
+        valid_report = sync_object(run_eval(
+            self.valid_world, opt, 'valid', opt['validation_max_exs'],
+        ))
 
         # logging
-        if opt['tensorboard_log'] is True:
+        if opt['tensorboard_log'] is True and is_primary_worker():
             self.writer.add_metrics('valid', int(
```
emilydinan: casting as an int already takes the floor, you don't need that call

parlai/scripts/train_model.py
```diff
@@ -284,37 +308,100 @@ def validate(self):
...
+                    to_average[k].append(v)
+        for k, vals in to_average.items():
+            finalized[k] = np.mean(vals)
+
+        return finalized
+
+    def _nice_format(self, dictionary):
+        return {k: round_sigfigs(v, 4) for k, v in dictionary.items()}
+
+    def _compute_eta(self, epochs_completed, time_elapsed):
```
emilydinan: nice!

parlai/tasks/convai2/agents.py
```diff
@@ -27,7 +27,6 @@ def _path(opt, persona, use_cands):
     build(opt)
     datatype = opt['datatype'].split(':')[0]
     if datatype == 'test':
-        print("WARNING: Test set not included. Setting datatype to valid.")
```
emilydinan: why did you get rid of this?

parlai/core/torch_generator_agent.py
```diff
@@ -359,15 +367,16 @@ def build_criterion(self):
...
         """Pre-initialize CUDA buffer by doing fake forward pass."""
-        if self.use_cuda and not hasattr(self, 'buffer_initialized'):
+        if self.use_cuda and (force or not hasattr(self, 'buffer_initialized')):
             try:
-                print('preinitializing pytorch cuda buffer')
                 dummy_xs = torch.ones(batchsize, maxlen).long().cuda()
                 dummy_ys = torch.ones(batchsize, 2).long().cuda()
-                scores, _, _ = model(dummy_xs, dummy_ys)
-                loss = criterion(scores.view(-1, scores.size(-1)), dummy_ys.view(-1))
+                scores, _, _ = self.model(dummy_xs, dummy_ys)
```
emilydinan: maybe a silly question, but can we assume that all future torch generator agents will have the same input/output format?

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -154,15 +154,6 @@ def build_model(self, states=None):
 
         if self.use_cuda:
             self.model.cuda()
-            if self.multigpu:
```
emilydinan: this is so much cleaner now 😄

parlai/scripts/train_model.py
stephenroller: (y)

parlai/core/torch_generator_agent.py
stephenroller: I'm going with Cross That Bridge Later

parlai/scripts/train_model.py
stephenroller: (y)

parlai/tasks/convai2/agents.py
stephenroller: Mistake

parlai/agents/transformer/modules.py
```diff
@@ -301,9 +301,9 @@ def __init__(
         self.attention = MultiHeadAttention(
             n_heads, embedding_size, dropout=attention_dropout
         )
-        self.norm1 = nn.LayerNorm([embedding_size])
```
stephenroller: TODO: move transformer changes into another diff.

parlai/core/params.py
```diff
@@ -368,6 +368,20 @@ def add_parlai_args(self, args=None):
 
         self.add_pytorch_datateacher_args()
 
+    def add_distributed_training_args(self):
+        grp = self.add_argument_group('Distributed Training')
+        grp.add_argument(
+            '--distributed-world-size', type=int,
+            help='Number of workers.'
+        )
+        # TODO: use --debug or --verbose instead?
```
stephenroller: Remove this comment

parlai/agents/seq2seq/seq2seq.py
stephenroller: Actually I should double check this one agaib

parlai/scripts/multiprocessing_train.py
```diff
@@ -0,0 +1,115 @@
...
+        multiprocess_train,
+        (opt, port),
+        nprocs=opt['distributed_world_size'],
+        join=False,
+    )
+
+    try:
+        spawncontext.join()
+    except KeyboardInterrupt:
+        print("killing subprocesses.")
```
stephenroller: Better message

parlai/scripts/train_model.py
```diff
@@ -284,37 +311,100 @@ def validate(self):
...
+        if not is_distributed():
+            # nothing special needed
+            return metrics
+
+        all_versions = all_gather_list(metrics)
+        finalized = {}
+        to_average = {}
+        for m in all_versions:
+            for k, v in m.items():
+                if k == 'num_epochs':
```
stephenroller: check if this one is still needed

parlai/scripts/train_model.py
```diff
@@ -284,37 +311,100 @@ def validate(self):
...
+                    # just drop this metric
+                    pass
+                elif k == 'exs' or k == 'total_skipped_batches':
```
stephenroller: check if still needed

parlai/agents/seq2seq/seq2seq.py
stephenroller: yeah it's all good :-O

stephenroller:
Usually ctrl-c seems to do what's intended, though sometimes it just exits fast. Not sure about that one.

PR #1322, opened 2018-12-19
Additionally pass the observations to Batch objects.

stephenroller:
As discussed.

Didn't actually test, letting the unit tests do this.


samhumeau:
Just FMI, the observation here doesn't contain any of the history right?

stephenroller:
Correct

PR #1323, opened 2018-12-19
[MTurk] Evaluate chat partner on 2 person tasks

JackUrb:
## Description/motivation
In order to isolate particular high and low quality examples it's useful to get feedback on individual interactions between workers, rated by the workers themselves rather than via hand review. It remains to be seen how noisy this rating data will be, but chances are it will at least be useful. Workers in the past have asked for this functionality as they don't like having others not put effort into the work as it degrades the platform for those who do.

## Implementation
- A new react component `ReviewButtons` was added to the `core_components.jsx` file. This component handles submitting feedback through the regular message channel over the socket (a system message). It then reports the action to its parent `DoneButton` in order to get the done button to show. It is treated as an extension of the done button for tasks that have enabled `allow_reviews` in their `task_config`.
- On agent, the agent queue is flushed once the HIT is submitted in order to extract potential feedback messages that were sent after the world ended. These are then put into the `agent.feedback` field.
- `MTurkDataWorld` was updated to extract the feedback from an agent, and then put it into the saved data as `given_feedback`. It also strips the command message from the saved messages, and attaches the partner's feedback as `gotten_feedback`
- An `AssignmentFeedback` component was added to the ParlAI MTurk webapp, which serves the purpose of displaying the `ReviewButtons` component with the feedback found in the saved data. 
- Moves the components displayed in the assignment view in the webapp into left and right panes in order to add the feedback panel into the flow in a place that makes sense.
- Adds an `_assert_opts` call to `MTurkManager` to ensure that the options being passed are compatible with one another (as, for example, `allow_reviews` only works with the react frontend). This warning is printed before the user hits enter to confirm task launching.
- Fixes behavior for react socket messaging when `is_system` is true to prevent system messages from being appended into the chat flow.

#### Neutral
screen shot 2018-12-19 at 12 08 42 pm
#### Positive
screen shot 2018-12-19 at 12 09 11 pm
screen shot 2018-12-19 at 12 14 41 pm
#### Negative
screen shot 2018-12-19 at 12 08 52 pm
screen shot 2018-12-19 at 12 09 00 pm
#### Decline feedback
screen shot 2018-12-19 at 12 10 20 pm
#### Submit feedback
screen shot 2018-12-19 at 12 15 03 pm
#### No feedback
screen shot 2018-12-19 at 12 25 16 pm
#### With feedback
screen shot 2018-12-19 at 12 26 48 pm
#### `allow_reviews` on a task where the number of people is not 2
screen shot 2018-12-19 at 12 34 48 pm
#### `allow_reviews` on a task not running react frontend
screen shot 2018-12-19 at 12 35 25 pm

## Additional fixes 
- Frame height is now set in `task_config` for react frontend tasks, defaults to 650 (and stays 650 for tasks without react frontend).
- `block_mobile` is added to the same flow of configuring in `task_config`
- The above two fixes require sending opts to the `create_hit_config` function (as well as `create_hit_with_hit_type`) so that these values can be extracted.
- `onInputResize` and callbacks have been added to input components on the frontend as a method to call when the input size changes, as this allows the chat window to resize as well to stay in `frame_height` without any other changes.

### DIFF
--- a/parlai/mturk/core/react_server/dev/components/core_components.jsx
+++ b/parlai/mturk/core/react_server/dev/components/core_components.jsx
@@ -8,7 +8,7 @@
 
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {FormControl, Button} from 'react-bootstrap';
+import {FormControl, Button, ButtonGroup} from 'react-bootstrap';
 import Slider from 'rc-slider';
 import $ from 'jquery';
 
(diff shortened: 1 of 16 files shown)


parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -312,11 +318,199 @@ class IdleResponse extends React.Component {
...
+    'Other',
+  ]
+
+  BAD_REASONS = [
+    'Not specified',
+    "Didn't understand task",
+    "Bad grammar/spelling",
+    "Total nonsense",
+    "Slow responder",
+    'Other',
```
emilydinan: nit: you switch between single and double quotes here (also using single quotes in `GOOD_REASONS`)

parlai/mturk/core/worlds.py
```diff
@@ -20,13 +20,26 @@ def prep_save_data(self, workers):
...
+                'messages': save_messages,
+                'given_feedback': agent.feedback,
             }
 
+        # In simple pairing case, attach the feedback right here
+        if len(workers) == 2:
+            data = save_data['worker_data']
+            a_0 = workers[0]
+            a_1 = workers[1]
+            data[a_0.worker_id]['gotten_feedback'] = a_1.feedback
```
emilydinan: nit: slightly awkward phrasing. maybe "feedback_received"

PR #1326, opened 2018-12-20
Bunch of bugs/corner cases fixes for TRA

samhumeau:
- Do not define optimizer for every agent
- handle case where NULL_IDX is not 0 
- clip is not really defined anywhere is it? added the case where it's not define
- In some weird cases you can expect the text vec to have been over padded. The text + pad can end up being longer than candidates.

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -10,7 +10,7 @@ import os
 
 import torch
 from torch import nn
-import pdb
+
 from parlai.core.torch_agent import TorchAgent, Output
 from parlai.core.thread_utils import SharedTable
 from parlai.core.utils import round_sigfigs, padded_3d, warn_once
@@ -73,8 +73,9 @@ class TorchRankerAgent(TorchAgent):
             self.model.cuda()
             self.rank_loss.cuda()
 
-        optim_params = [p for p in self.model.parameters() if p.requires_grad]
-        self.init_optim(optim_params)
+        if not shared:
+            optim_params = [p for p in self.model.parameters() if p.requires_grad]
+            self.init_optim(optim_params)
 
     def score_candidates(self, batch, cand_vecs):
         """Given a batch and candidate set, return scores (for ranking)"""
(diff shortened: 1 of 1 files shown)


parlai/core/torch_ranker_agent.py
```diff
@@ -224,11 +225,15 @@ def _build_candidates(self, batch, source, mode):
...
 
             cands = batch.candidates
-            cand_vecs = padded_3d(batch.candidate_vecs, use_cuda=self.use_cuda)
+            cand_vecs = padded_3d(batch.candidate_vecs, self.NULL_IDX,
+                                  use_cuda=self.use_cuda)
             if label_vecs is not None:
                 label_inds = label_vecs.new_empty((batchsize))
                 for i, label_vec in enumerate(label_vecs):
-                    label_vec_pad = label_vec.new_zeros(cand_vecs[i].size(1))
+                    label_vec_pad = label_vec.new_zeros(cand_vecs[i].size(1)) \
```
bhancock8: I think elsewhere in the code we usually use parenthesis to group multi-line statements rather than backslashes.

parlai/core/torch_ranker_agent.py
```diff
@@ -285,7 +290,7 @@ def share(self):
 
     def update_params(self):
         """Do optim step and clip gradients if needed."""
-        if self.clip > 0:
+        if hasattr(self, 'clip') and self.clip > 0:
```
bhancock8: simpler: `if self.get('clip', 0) > 0:`

stephenroller: Should be `self.get('gradient_clip', 0)` I believe.

parlai/core/torch_ranker_agent.py
```diff
@@ -224,11 +225,15 @@ def _build_candidates(self, batch, source, mode):
                     "".format(m='candidates' if mode == 'train' else 'eval-candidates'))
 
             cands = batch.candidates
-            cand_vecs = padded_3d(batch.candidate_vecs, use_cuda=self.use_cuda)
+            cand_vecs = padded_3d(batch.candidate_vecs, self.NULL_IDX,
```
stephenroller: oh good, thx

parlai/core/torch_ranker_agent.py
```diff
@@ -73,8 +73,9 @@ def __init__(self, opt, shared=None):
             self.model.cuda()
             self.rank_loss.cuda()
 
-        optim_params = [p for p in self.model.parameters() if p.requires_grad]
-        self.init_optim(optim_params)
+        if not shared:
+            optim_params = [p for p in self.model.parameters() if p.requires_grad]
```
stephenroller: :pikachu_meme:

parlai/core/torch_ranker_agent.py
```diff
@@ -224,11 +225,15 @@ def _build_candidates(self, batch, source, mode):
...
             cands = batch.candidates
-            cand_vecs = padded_3d(batch.candidate_vecs, use_cuda=self.use_cuda)
+            cand_vecs = padded_3d(batch.candidate_vecs, self.NULL_IDX,
+                                  use_cuda=self.use_cuda)
             if label_vecs is not None:
                 label_inds = label_vecs.new_empty((batchsize))
                 for i, label_vec in enumerate(label_vecs):
-                    label_vec_pad = label_vec.new_zeros(cand_vecs[i].size(1))
+                    label_vec_pad = label_vec.new_zeros(cand_vecs[i].size(1)) \
+                                    + self.NULL_IDX
```
stephenroller: is .fill_(NULL_IDX) perhaps better?

parlai/core/torch_ranker_agent.py
stephenroller: I believe optimizer needs to come from shared or something though, hence why it's breaking unit tests

parlai/core/torch_ranker_agent.py
samhumeau: Not sure where this self.get comes from, TRA is not a dict ?

PR #1327, opened 2018-12-20
[wip] Adding structure to act context

JackUrb:
## Precursor 
While this is a WIP, I'm very open to feedback. I want to get some code down to start seeing how this shapes up but I get that a lot of it is up to change!

## Motivation
Using context right now in tasks has mostly amounted to string parsing. Rather than require models to directly parse strings, the thought here is to make it so that acts can be parsed into objects that can have utility methods and that do have a defined structure. More motivations are in the context quip, which I've linked in the main doc.

## Implementation
So far the thought is that acts will contain `context_data` in a container class filled with children of the `BaseContext` object. This will allow us to directly convert the content of `context_data` into a string when needed by a model that doesn't parse the context on its own. With this we'll be able to make

## Testing/Semantics
```
Python 3.6.5 |Anaconda, Inc.| (default, Mar 29 2018, 13:14:23) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from parlai.core.context import ContextBinder
>>> from parlai.context.persona_context import PersonaContext
>>> my_persona = PersonaContext(['I am a dog', 'I like bones', 'I hate mail people'])
>>> act = {'id': 'partner', 'text': 'Hey there buddy are you hungry?', 'context_data': ContextBinder([my_persona])}
>>> act
{'id': 'partner', 'text': 'Hey there buddy are you hungry?', 'context_data': {<ContextBinder>: {'persona': [['I am a dog', 'I like bones', 'I hate mail people']]}}}
>>> str(act)
"{'id': 'partner', 'text': 'Hey there buddy are you hungry?', 'context_data': {<ContextBinder>: {'persona': [['I am a dog', 'I like bones', 'I hate mail people']]}}}"
>>> str(act['context_data'])
'Your persona: I am a dog\nYour persona: I like bones\nYour persona: I hate mail people'
```

## TODO
- actually make teachers automatically put the content in context_data into a string context field?
- make the context binder automatically convert everything to data form if an act is trying to be converted to JSON.
- make the context binder able to be re-instanced from just the JSON if possible.


parlai/core/context.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
```
stephenroller: Add some docs here, argument doc strings if possible. Makes sense for us to finalize first.

parlai/core/context.py
```diff
@@ -0,0 +1,118 @@
...
+    @staticmethod
+    def from_data(input_data):
+        """This function should take an input data object and produce the
+        context representation for this data. It is expected to be used by
+        tasks that are constructing context, and should be the inverse of
+        to_data_form
+        """
+        raise NotImplementedError
+
+    def to_data_form(self):
```
stephenroller: What do you think about serialize/deserialize?

parlai/core/context.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        raise NotImplementedError
+
+    def to_display_string(self):
+        """This function will be invoked by the context binder whenever an
+        act with a context binder is printed
+        """
+        return self.to_data_form()
+
+    def get_type(self):
```
stephenroller: Why wouldn't we want to use something like `isinstance` instead?

parlai/core/context.py
```diff
@@ -0,0 +1,118 @@
...
+        if context_type not in self.content:
+            self.content[context_type] = []
+        self.content[context_type].append(context_object)
+
+    def get_context(self, context_type):
+        """Return all of the stored context of a context type"""
+        return self.content.get(context_type, [])
+
+    def __repr__(self):
+        return '{{<ContextBinder>: {}}}'.format(self.to_display_string())
```
stephenroller: (y)

parlai/core/context.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __str__(self):
+        return self.to_context_string()
```
stephenroller: (y)

parlai/core/context.py
```diff
@@ -0,0 +1,118 @@
...
+        raise NotImplementedError
+
+    def to_data_form(self):
+        """This function should return an object that is a less structured
+        representation of the data contained in this object (generally in
+        the form of a dict or an array)
+        """
+        raise NotImplementedError
+
+    def to_context_string(self):
```
stephenroller: Would be more pythonic to just require the user implement __str__ and __repr__ directly? Even __repr__ could have a default implementation of `str(type(self)) + ":{" + str(self) + '}'` or similar.

parlai/core/context.py
JackUrb: I'll change to those - they tend to be more standard and definitionally correct.

parlai/core/context.py
JackUrb: Yeah I can see that as being valid, especially considering that's exactly how I'm using them in the other example

parlai/core/context.py
JackUrb: Initially I had concerns about using Instance names because they aren't directly serializable. I can still reflect on this though

parlai/core/context.py
```diff
@@ -0,0 +1,87 @@
...
+    def serialize(self):
+        """Makes a best effort to serialize the data here so that it can be
+        sent over wire if necessary
+        """
+        return {(k, v.serialize() if hasattr(v, 'serialize') else v)
+                for (k, v) in self.items()}
+
+    def __setitem__(self, key, val):
+        if self.locked:
+            raise Exception(
```
stephenroller: RuntimeError

parlai/core/context.py
```diff
@@ -0,0 +1,87 @@
...
+                'Cannot update the value of an immutable TaskData object')
```
alexholdenmiller: maybe include how to unlock it in case that's what they want

parlai/core/context.py
```diff
@@ -0,0 +1,87 @@
...
+        else:
+            dict.__setitem__(self, key, val)
```
alexholdenmiller: maybe better to use `super()` instead of `dict` in each of these spots?

stephenroller:
Calling this abandoned.

JackUrb:
Sad, but true. Maybe to be revisited someday when context causes us problems again.

PR #1332, opened 2019-01-07
Starspace inherits from Torch Agent

emilydinan:
Starspace inheriting from Torch Ranker Agent. Main changes to model are that it is now compatible with batchsize > 1.

Also included changes to Torch Ranker Agent that allows for a "custom set of candidates". Requires implementing a function `build_custom_candidates` that will replace inline candidates. 

Note: will not be totally backwards compatible with original starspace agent as `get_dialog_history` differs *slightly* from previous way dialog history was maintained (previously joined history with spaces, now joins with newlines, depending on your tokenizer this may make a difference).

Original starspace has been moved to a legacy agent


stephenroller:
Can you clarify "it is not compatible with batchsize > 1"? My understanding was this was one of your major improvements.

emilydinan:
> Can you clarify "it is not compatible with batchsize > 1"? My understanding was this was one of your major improvements.

haha.... i meant to write "now" instead of "not". will update! thanks for catching

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+import torch.nn as nn
+import os
+import random
+
+
+class StarspaceTorchAgent(TorchRankerAgent):
+    """Simple implementation of the starspace algorithm: https://arxiv.org/abs/1709.03856
+    """
+
+    OPTIM_OPTS = {
```
stephenroller: Do we really need LBFGS and all of these?

TBH I feel like this should really be happening in TorchAgent

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+        'adamax': optim.Adamax,
+        'asgd': optim.ASGD,
+        'lbfgs': optim.LBFGS,
+        'rmsprop': optim.RMSprop,
+        'rprop': optim.Rprop,
+        'sgd': optim.SGD,
+    }
+
+    @staticmethod
+    def dictionary_class():
```
stephenroller: Since this doesn't override, let's just drop it

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+        TorchRankerAgent.add_cmdline_args(argparser)
+        agent = argparser.add_argument_group('StarSpace Arguments')
+        agent.add_argument(
+            '-esz', '--embeddingsize', type=int, default=128,
+            help='size of the token embeddings')
+        agent.add_argument(
+            '-enorm', '--embeddingnorm', type=float, default=10,
+            help='max norm of word embeddings')
+        agent.add_argument(
+            '-shareEmb', '--share-embeddings', type='bool', default=True,
```
stephenroller: `-shareemb` for consistency?

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+            help='whether LHS and RHS share embeddings')
+        agent.add_argument(
+            '--lins', default=0, type=int,
+            help='If set to 1, add a linear layer between lhs and rhs.')
+        agent.add_argument(
+            '-lr', '--learningrate', type=float, default=0.1,
```
stephenroller: Double check what's covered in TA already:
https://github.com/facebookresearch/ParlAI/blob/9e113105d1690deda0b372f6aeaf933ab58e3a72/parlai/core/torch_agent.py#L161

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+            help='fraction of input/output features to dropout during training')
+        agent.add_argument(
+            '-opt', '--optimizer', default='sgd',
+            choices=StarspaceTorchAgent.OPTIM_OPTS.keys(),
+            help='Choose between pytorch optimizers. '
+                 'Any member of torch.optim is valid and will '
+                 'be used with default params except learning '
+                 'rate (as specified by -lr).')
+        agent.add_argument(
+            '-tr', '--truncate', default=1000, type=int,
```
stephenroller: drop

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+            '--parrot-neg', type=bool, default=False,
+            help='include query as a negative')
+        agent.add_argument(
+            '--tfidf', type='bool', default=False,
+            help='Use frequency based normalization for embeddings.')
+        agent.add_argument(
+            '-cs', '--cache-size', type=int, default=1000,
+            help='size of negative sample cache to draw from')
+        agent.add_argument(
+            '-cands', '--candidates', type=str, default='custom',
```
stephenroller: These are provided by TRA, yes? No need to override them here.

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+            )
+        else:
+            xe, ye = self.model(
+                batch.text_vec,
+                ys=batch.label_vec,
+                cands=cand_vecs
+            )
+        xe = torch.cat([xe for _ in range(ye.size(1))], dim=1)
+
+        if self.use_cuda:
```
stephenroller: Define y and y_0, then move them to cuda separately. Slightly DRYER. You could also do `.to(xe.device)` like suggested above.

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+
+        # Update metrics
+        self.metrics['loss'] += loss.item()
+        self.metrics['examples'] += batchsize
+        _, ranks = scores.sort(1, descending=True)
+        for b in range(batchsize):
+            rank = (ranks[b] == label_inds[b]).nonzero().item()
+            self.metrics['rank'] += 1 + rank
+
+        return Output()
```
stephenroller: Stuff needed here I believe (you compute the ranks, so might as well return something)

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+            if cand_vecs.dim() == 2:
+                cand_list = cands
+            elif cand_vecs.dim() == 3:
+                cand_list = cands[i]
+            cand_preds.append([cand_list[rank] for rank in ordering])
+        preds = [cand_preds[i][0] for i in range(batchsize)]
+        return Output(preds, cand_preds)
+
+    def observe(self, observation):
+        labels = observation.get('labels')
```
stephenroller: This part could use some comments, because it's not obvious what's going on to me.

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+    def _get_custom_candidates(self, obs):
+        """Sets custom candidates for the observation."""
+        if not obs.get('eval_labels'):
+            # only use custom candidates during training
+            cands = []
+            cache_sz = len(self.ys_cache) - 1
+            if cache_sz < 1:
+                return cands
+            k = self.opt['neg_samples']
+            for i in range(1, k * 3):
```
stephenroller: Why 3?

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+                    if len(cands) >= k:
+                        break
+            if self.opt['parrot_neg']:
+                query = self.history.split('\n')[-1]
+                cands.append(query)
+                return cands
+        # if we aren't training, return label candidates if available
+        return obs.get('label_candidates')
+
+    def _init_embeddings(self, log=True):
```
stephenroller: I think you can delete all this and just use `TorchAgent. _copy_embeddings`

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+                  ''.format(cnt, round(cnt * 100 / len(self.dict), 1), name))
+
+    def same(self, labels, cand):
+        if labels:
+            for label in labels:
+                if label == cand:
+                    return True
+        return False
+
+    def input_dropout(self, xs, ys, negs):
```
stephenroller: Hmm, this looks like a variant on UnknownDropout in Seq2Seq. I wonder if it's worth refactoring both of these. https://github.com/facebookresearch/ParlAI/blob/9e113105d1690deda0b372f6aeaf933ab58e3a72/parlai/agents/seq2seq/modules.py#L149-L172

parlai/agents/starspace_torch/starspace_torch.py
```diff
@@ -0,0 +1,367 @@
...
+    def reset(self):
+        """Reset observation and episode_done."""
+        self.observation = {}
+        self.history.clear()
+        self.replies.clear()
+        self.reset_metrics()
+        # reset optimizer
+        self.optimizer_reset()
+
+    def share(self):
```
stephenroller: Both dict and model are set in TorchAgent and TorchRankerAgent respectively, so I don't think you need this.

parlai/core/torch_ranker_agent.py
```diff
@@ -28,7 +28,7 @@ def add_cmdline_args(argparser):
                  '(see TorchRankerAgent._build_candidates() for details).')
         agent.add_argument(
             '-ecands', '--eval-candidates', type=str,
-            choices=['batch', 'inline', 'fixed', 'vocab'],
+            choices=['batch', 'inline', 'fixed', 'vocab', 'custom'],
```
stephenroller: I'm still confused w.r.t. all of this, and mostly wondering what's the logic here, and whether all this extra logic should be given a name and moved into TRA.

parlai/agents/starspace_torch/starspace_torch.py
emilydinan: that was what was in original starspace

parlai/core/torch_ranker_agent.py
emilydinan: the logic is in TRA for custom candidates

parlai/agents/starspace_torch/starspace_torch.py
emilydinan: i'm guessing it was just a "reasonable" number of examples to search through to get k negatives that were not equal to the label

parlai/core/torch_ranker_agent.py
stephenroller: Ahh, mostly I just don't understand what Starspace is trying to do w.r.t negative sampling, and whether it should be supported outside of starspace.

parlai/agents/starspace_torch/modules.py
```diff
@@ -0,0 +1,90 @@
...
+        xs_emb = self.lt(xs)
+        if self.freqs is not None:
+            # tfidf embeddings
+            bsz = xs.size(0)
+            len_x = xs.size(1)
+            if xs_emb.is_cuda:
+                x_scale = torch.Tensor(bsz, len_x).cuda()
+            else:
+                x_scale = torch.Tensor(bsz, len_x)
+            for i in range(len_x):
```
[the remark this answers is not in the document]
emilydinan: unfortunately i don't believe we can parallelize this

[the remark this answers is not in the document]
emilydinan: and yes, it's very slow 😢

parlai/agents/starspace_torch/starspace_torch.py
emilydinan: i added learning rate as an arg here to change the default to 1 from 0.1 (in TA)

parlai/agents/starspace_torch/starspace_torch.py
emilydinan: overridden because the default here is 'custom'

parlai/agents/starspace_torch/starspace_torch.py
emilydinan: doesn't really make sense to returning anything here, as during training we are ranking a set of custom candidates and not the candidates provided by the task

parlai/agents/starspace_torch/starspace_torch.py
emilydinan: yep, it's a bit different. namely, in starspace we are getting rid of random tokens in the sentence at some rate and here we are replacing them with unknown tokens (the latter is a bit cleaner).. i'll think about ways to make this better

parlai/agents/starspace_torch/modules.py
```diff
@@ -21,12 +21,12 @@ def __init__(self, opt, num_features, dict):
                                max_norm=opt['embeddingnorm'])
         if not opt['tfidf']:
             dict = None
-        self.encoder = Encoder(self.lt, dict)
+        self.encoder = Encoder(self.lt, dict, null_idx=self.null_idx)
```
stephenroller: nice catch

parlai/agents/starspace_torch/starspace_torch.py
stephenroller: (y)

parlai/core/torch_ranker_agent.py
```diff
@@ -267,7 +287,8 @@ def _build_candidates(self, batch, source, mode):
 
     @staticmethod
     def _find_match(cand_vecs, label_vec):
-        return ((cand_vecs == label_vec).sum(1) == cand_vecs.size(1)).nonzero()[0]
+        matches = ((cand_vecs == label_vec).sum(1) == cand_vecs.size(1)).nonzero()
+        return matches[0] if (len(matches.tolist()) > 0) else None
```
stephenroller: do you really need the call to tolist()? len should work on the tensor directly and be faster, no?

parlai/core/torch_ranker_agent.py
```diff
@@ -230,9 +230,29 @@ def _build_candidates(self, batch, source, mode):
             if label_vecs is not None:
                 label_inds = label_vecs.new_empty((batchsize))
                 for i, label_vec in enumerate(label_vecs):
-                    label_vec_pad = label_vec.new_zeros(cand_vecs[i].size(1))
-                    label_vec_pad[0:label_vec.size(0)] = label_vec
-                    label_inds[i] = self._find_match(cand_vecs[i], label_vec_pad)
+                    if cand_vecs[i].size(1) < label_vec.size(0):
```
stephenroller: I don't get this one, can you add a small comment explaining?

parlai/agents/starspace_torch/modules.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, opt, num_features, dict):
         self.lins = opt.get('lins', 0)
 
     def forward(self, xs, ys=None, cands=None):
-        if not ys and not cands:
+        if ys is None and cands is None:
```
stephenroller: awesome, thank you

emilydinan:
discussed some other possibilities with @jaseweston , like allowing the batch to share candidates... i think we can have this in a later diff possibly? otherwise, ready for review (cc @stephenroller )

parlai/core/torch_agent.py
```diff
@@ -403,6 +403,8 @@ def _copy_embeddings(self, weight, emb_type, log=True):
         :param weight:   weights of lookup table (nn.Embedding/nn.EmbeddingBag)
         :param emb_type: pretrained embedding type
         """
+        if emb_type == 'random':
```
stephenroller: I don't think this is supposed to be in

parlai/agents/starspace/README.md
```diff
@@ -7,5 +7,5 @@ This agent contains a simple implementation of the *starspace* algorithm, slight
 
 Train a starspace model on the "sentence SQuAD" [task](https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/squad/agents.py).
 ```bash
-python examples/train_model.py --task squad:sentence --model starspace -lr 0.01 -esz 512 -k 10 -mf /tmp/starspacesquad
+python examples/train_model.py --task squad:sentence --model starspace_torch -lr 0.01 -esz 512 -k 10 -mf /tmp/starspacesquad
```
stephenroller: Isn't the new name `starspace`?

parlai/agents/legacy_agents/starspace/starspace_v0.py
```diff
@@ -0,0 +1,493 @@
...
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+#
+# Simple implementation of the starspace algorithm, slightly adapted for dialogue.
+# See: https://arxiv.org/abs/1709.03856
+# TODO: move this over to TorchRankerAgent when it is ready.
+
+from parlai.core.agents import Agent
+from parlai.core.dict import DictionaryAgent
+from parlai.core.utils import maintain_dialog_history, load_cands
```
alexholdenmiller: can use legacy copypasta from legacy/seq2seq folder

parlai/agents/starspace/starspace.py
```diff
@@ -8,486 +8,324 @@
...
-            )
+        # During observe, we add each of the labels to the ys_cache as a set
+        # of possible negative candidates to use for the future
+        labels = observation.get('labels')
+        if not labels:
+            labels = observation.get('eval_labels')
+        if labels:
+            for label in labels:
+                self.add_to_ys_cache(label)
+        return super(StarspaceAgent, self).observe(observation)
```
stephenroller: here I believe you can just do plain `super().observe(observation)`.

parlai/agents/starspace/starspace.py
```diff
@@ -8,486 +8,324 @@
...
+            help='include query as a negative')
+        agent.add_argument(
+            '--tfidf', type='bool', default=False,
+            help='Use frequency based normalization for embeddings.')
+        agent.add_argument(
+            '-cs', '--cache-size', type=int, default=1000,
+            help='size of negative sample cache to draw from')
+        agent.add_argument(
+            '-cands', '--candidates', type=str, default='custom',
+            choices=['batch', 'inline', 'fixed', 'vocab', 'custom'],
```
stephenroller: Does it make sense to move the 'custom' option into TRA?

parlai/agents/starspace/starspace.py
```diff
@@ -92,29 +93,41 @@ def __init__(self, opt, shared=None):
...
-    def model_version():
-        """Return current version of this model, counting up from 0.
-        Models may not be backwards-compatible with older versions.
-        Version 1 split from version 0 on Jan 15, 2019.
-        To use version 0, use --model legacy:starspace:0
-        (legacy agent code is located in parlai/agents/legacy_agents).
-        """
-        return 1
+    # @staticmethod
+    # def model_version():
```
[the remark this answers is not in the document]
emilydinan: didn't mean to push that oops

PR #1335, opened 2019-01-09
Update vqa_v1 URL.

stephenroller:
URL seemed to move, just updating.


stephenroller:
Ignoring Travis because it passes locally.

PR #1337, opened 2019-01-10
Small Documentation updates

stephenroller:
- Delete an old class (had two copies of Beam)
- Small cosmetic changes (change color, copyright)
- Reorganize TOC a little
- Small paragraph on the point of TorchAgent etc.
- Make sure docstrings for Batch and Output look good in autodocs.


stephenroller:
Also deleted docs for the "reference" models that seemed weird to me, and a unit test to make sure the docs compile.

PR #1340, opened 2019-01-10
[MTurk] Small MTurk fixes

JackUrb:
Ran into two bugs while working with the new react frontend and peer feedback systems. 

The first is that if somebody mashes enter or the send button pretty fast (or has a really slow connection) the react frontend would allow doublesends. This is fixed by giving all inputs a message sending state that disables sending new messages.
The second is related to a world saving data - if any messages didn't have a 'text' field, the saving code would break while trying to filter out peer feedback messages. This is fixed by using `get` rather than a direct access.

Tested by running `qa_data_collection`, `react_custom_no_extra_deps`, and an internal task.

### DIFF
--- a/parlai/mturk/tasks/react_task_demo/react_custom_no_extra_deps/frontend/components/custom.jsx
+++ b/parlai/mturk/tasks/react_task_demo/react_custom_no_extra_deps/frontend/components/custom.jsx
@@ -38,19 +38,25 @@ class EvaluatorIdleResponse extends React.Component {
 class NumericResponse extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {'textval': ''};
+    this.state = {'textval': '', 'sending': false};
   }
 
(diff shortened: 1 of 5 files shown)


parlai/mturk/tasks/react_task_demo/react_custom_no_extra_deps/frontend/components/custom.jsx
```diff
@@ -38,19 +38,25 @@ class EvaluatorIdleResponse extends React.Component {
 class NumericResponse extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {'textval': ''};
+    this.state = {'textval': '', 'sending': false};
   }
 
   componentDidUpdate(prevProps, prevState, snapshot) {
+    // Only change in the active status of this component should cause a
```
emilydinan: thanks for the commentary! this will make it easier for when people write their own custom components

PR #1342, opened 2019-01-10
Small updates to transformer.

stephenroller:
Moving things from the disgpu diff into here.

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -301,9 +301,9 @@ class TransformerEncoderLayer(nn.Module):
         self.attention = MultiHeadAttention(
             n_heads, embedding_size, dropout=attention_dropout
         )
-        self.norm1 = nn.LayerNorm([embedding_size])
+        self.norm1 = nn.LayerNorm(embedding_size)
         self.ffn = TransformerFFN(embedding_size, ffn_size, dropout=relu_dropout)
-        self.norm2 = nn.LayerNorm([embedding_size])
+        self.norm2 = nn.LayerNorm(embedding_size)
 
     def forward(self, tensor, mask):
         tensor = tensor + self.attention(tensor, mask=mask)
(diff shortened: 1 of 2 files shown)


parlai/agents/transformer/modules.py
```diff
@@ -393,16 +393,15 @@ def __init__(
         self.self_attention = MultiHeadAttention(
             n_heads, embedding_size, dropout=attention_dropout
         )
-        self.selfattn_norm = nn.LayerNorm([embedding_size])
```
stephenroller: This is the really really really important one. We have an unused parameter!

parlai/agents/transformer/transformer.py
```diff
@@ -25,12 +25,12 @@
...
                            help='Size of all embedding layers')
-    argparser.add_argument('--n-layers', type=int, default=2)
-    argparser.add_argument('--ffn-size', type=int, default=300,
+    argparser.add_argument('-nl', '--n-layers', type=int, default=2)
+    argparser.add_argument('-hid', '--ffn-size', type=int, default=300,
                            help='Hidden size of the FFN layers')
     argparser.add_argument('--attention-dropout', type=float, default=0.0)
     argparser.add_argument('--relu-dropout', type=float, default=0.0)
-    argparser.add_argument('--n-heads', type=int, default=3)
+    argparser.add_argument('--n-heads', type=int, default=2)
```
samhumeau: Although I am pretty sure here it doesn't change anything, changing default parameters seem like a bad practice since some people may have run experiments while letting the parameter blank. Saved somewhere the command line to reproduce and now it doesn't.

stephenroller: It does have a BETA warning at the top. Saved models will still produce the exact same results.

If you ask, I will move it back.

samhumeau: Right. Didn't see the beta. Seems legit.

stephenroller:
Heads up @samhumeau , this merged. @emilydinan  and @klshuster  should watch out too in case you were using unused defaults.

It also means that saves you might be broken now.

PR #1345, opened 2019-01-11
Open subtitles changed URLs, compression

stephenroller:
OpenSubtitles 2018 was broken by external changes.

Still running tests locally, will confirm here when they pass.


stephenroller:
After discussing with @alexholdenmiller, seems like all issues with this dataset are resolved.

stephenroller:
The change in the # of movies was a mistake by me.

PR #1348, opened 2019-01-11
adding an option to pad left in the torch agent

samhumeau:
Couldn't find this option so added it. 
Note that it doesn't concern the padding of the candidates in the torch ranker agent


stephenroller:
I would rather this be something that an agent overrides, rather than a command line argument.

PR #1359, opened 2019-01-16
Relicense ParlAI from BSD+Patents to MIT

stephenroller:
codemod.


stephenroller:
Tests are never going to pass:

(1) would have to lint projects/ which is not happening
(2) Downloads of all those data sets isn't going to happen either.

Regular short tests pass locally.

PR #1361, opened 2019-01-16
Fix Image Chat Download Images

klshuster:
The image download script for Image-Chat/Personality-Captions had the incorrect save location for the images; this PR fixes.


klshuster:
note: travis fails due to requiring user input

samhumeau:
Thanks! 
Is there some mechanism that prevents re-downloading it ? (in particular in the case where one builds persona-caption and then builds image-chat?)

klshuster:
@samhumeau yes - the images for each are downloaded to the same folder, which is marked as built by ParlAI (so, e.g., if you download via `-t personality_captions` and then run `-t image_chat`, it'll check the same images directory before redownloading)

PR #1364, opened 2019-01-17
Nicer syntax

stephenroller:
Nicer syntax suggested by @emilydinan in #1358, but previously disregarded to land faster.

### DIFF
--- a/parlai/scripts/train_model.py
+++ b/parlai/scripts/train_model.py
@@ -319,8 +319,7 @@ class TrainLoop():
         to_reduce = {}
         for d in all_versions:
             for k, v in d.items():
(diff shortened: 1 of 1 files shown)


parlai/scripts/train_model.py
```diff
@@ -319,8 +319,7 @@ def _average_dicts(self, all_versions):
         to_reduce = {}
         for d in all_versions:
             for k, v in d.items():
-                if k not in to_reduce:
-                    to_reduce[k] = []
+                to_reduce.setdefault(k, [])
```
emilydinan: you can even do it in one line `to_reduce.setdefault(k, []).append(v)`

PR #1367, opened 2019-01-17
Adding context support to the MTurk ecosystem.

JackUrb:
## Summary

In anticipation of #1327, and as one of the prime motivations for me to be working on that in the first place, I've added a more mature form of context handling to the ParlAI MTurk client.

## Implementation

The basic idea here is to leverage the component swapping nature of the PMT react frontend to allow tasks to define a context-rendering component, and then for an agent to observe context in `act['task_data']` and have it magically fed to the context component. This `task_data` is later stored with the message, and can be retrieved both when reviewing tasks and when saving the task into a dataset. As a baseline implementation, people can put raw HTML into `act['task_data']['html']` to have that content rendered without needing to create a `custom.jsx` file at all, but when I migrate some of our existing tasks to frontend v1.0 with context, I'll be demonstrating custom components. This version only allows for the 2-panel context approach, where the left panel is given tabs for task instructions and context. Down the road I'll support 3-panel context for those that want task instructions up at all times.

### Major contributions
- Added `onClickMessage` prop to `MessageList` to handle updating the 'currently selected message index', which is used by the webapp to display relevant context.
- Creates `ContextView` that renders arbitrary html if it is provided.
- Adds `ContextView` into a tab of the left pane if `task_data` is updated. This is managed by tracking the `last_update` time of the context window. `last_update` is also used to force select the context tab upon receiving new context as a method for denoting that context has changed. (Maybe an animation flash or something might be more appropriate for that? will re-investigate down the road). If there is no `task_data` the tabulations don't change at all.
- Creates `AssignmentContext` view for the ParlAI webapp task review window. This pulls `task_data` context from the main webapp.
- Adds `task_data` management to the mock agents in the webapp, building context up from the list of messages received by the agent.
### Minor fixes
- Consolidates all data objects containing context into `task_data` throughout the PMT frontend for consistency with #1327 
- Updates `SocketHandler` function props to better follow the react style of updating state at the highest level rather than having subcomponents alter state of an above component. Better to follow react best practices!
- Changes the display type of the left panel to flex and updates children to better 
- Fixed some truthiness checks in the feedback display panel that would break it if one person didn't give feedback but the other did.
- Wraps `MockTurkManager`'s main evaluation loop in error checking code to print useful stack traces when something goes wrong rather than just failing, which makes sense as the webapp is used as a debug and development environment.
- Fixes a null dereference bug in the webapp server caused by the `MockTurkManager` taking more time than expected to start up and attach.

### Chat window with context tab
(tested both in ParlAI Webapp and MTurk sandbox)
screen shot 2019-01-17 at 2 39 43 pm
### Review panel in webapp
(clicking a message updates the context window to display the context the worker had when sending that message)
screen shot 2019-01-17 at 3 08 29 pm

### DIFF
--- a/parlai/mturk/core/react_server/dev/components/core_components.jsx
+++ b/parlai/mturk/core/react_server/dev/components/core_components.jsx
@@ -73,12 +73,12 @@ class MessageList extends React.Component {
     // the m.id field.
     let XChatMessage = getCorrectComponent('XChatMessage', this.props.v_id);
     let onClickMessage = this.props.onClickMessage;
-    if (onClickMessage === undefined) {
+    if (typeof onClickMessage !== 'function') {
       onClickMessage = (idx) => {};
     }
     return messages.map(
       (m, idx) =>
-        <div key={m.message_id} onClick={() => this.props.onClickMessage(idx)}>
+        <div key={m.message_id} onClick={() => onClickMessage(idx)}>
           <XChatMessage
             is_self={m.id == agent_id}
             agent_id={m.id}
--- a/parlai/mturk/core/react_server/dev/components/core_components.jsx
+++ b/parlai/mturk/core/react_server/dev/components/core_components.jsx
@@ -9,7 +9,8 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import {
-  FormControl, Button, ButtonGroup, InputGroup, MenuItem, DropdownButton
+  FormControl, Button, ButtonGroup, InputGroup, MenuItem, DropdownButton,
+  Nav, NavItem
 } from 'react-bootstrap';
 import Slider from 'rc-slider';
 import $ from 'jquery';
@@ -77,7 +78,7 @@ class MessageList extends React.Component {
         is_self={m.id == agent_id}
         agent_id={m.id}
         message={m.text}
-        context={m.context}
+        task_data={m.task_data}
         message_id={m.message_id}
         duration={this.props.is_review ? m.duration : undefined}/>
     );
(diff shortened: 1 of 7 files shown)


parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -768,37 +775,145 @@ class TaskDescription extends React.Component {
...
+        </NavItem>,
+        <NavItem
+          eventKey={'context'}
+          key={'context-selector'}
+          title={'Context'}>
+          {'Context'}
+        </NavItem>
+      ];
+      let display_instruction = {
+        'backgroundColor': '#dff0d8',
```
klshuster: might be good to standardize this color somewhere since it appears a few times in these files

parlai/mturk/webapp/dev/app.jsx
```diff
@@ -758,7 +760,7 @@ class AssignmentFeedback extends React.Component {
     var bsStyle = null;
     let given_feedback = null;
     let received_feedback = null;
-    if (review_data !== undefined) {
+    if (!!review_data) {
```
klshuster: TIL `!!` converts to `true`

parlai/mturk/core/react_server/dev/components/core_components.jsx
JackUrb: Yup - I'm planning to migrate this to a css file once everything is properly supported. Should make things better to have style separate from everything else.

parlai/mturk/webapp/dev/app.jsx
JackUrb: Right, it's a truthiness check. Technically I could just leave it as `if (review_data) {` but I like to make the check explicit.

PR #1371, opened 2019-01-18
Show default argparse values in console

ksurya:


stephenroller:
For posterity, here's an example output:

```
Dictionary Arguments:
  -df DICT_FILE, --dict-file DICT_FILE
                        path to dictionary file. defaults to [model_file].dict
                        if not set and model_file is set. (default: None)
  --dict-minfreq DICT_MINFREQ
                        minimum frequency of words to include them in sorted
                        dict or minimum frequency of bpe codecs (default: 0)
  --dict-maxtokens DICT_MAXTOKENS
                        max number of tokens to include in dictionary or bpe
                        codecs (default: -1)
  -tok DICT_TOKENIZER, --dict-tokenizer DICT_TOKENIZER
                        Which tokenizer to use. Defaults to "split", which
                        splits on whitespace as well as recognizing basic
                        punctuation. Other options include nltk and spacy.
                        (default: re)
  --dict-lower DICT_LOWER
                        Whether or not to lowercase all text seen. (default:
                        False)
```

PR #1372, opened 2019-01-18
Lint

stephenroller:
We've diverged a little from full linting of parlai/, fixing this.

@emilydinan there's still some small complaints about starspace, but I ignored them since you're redoing it anyway.


projects/convai2/baselines/language_model/eval_ppl.py
```diff
@@ -50,35 +50,40 @@ def next_word_probability(self, partial_out):
 
         self.model.eval()
 
-        if obs['episode_done'] == True:
+        if obs['episode_done'] is True:
```
alexholdenmiller: I guess just `if obs['episode_done']:`

PR #1374, opened 2019-01-22
Fix for prebuilt dict for PytorchDataTeacher

klshuster:
Does not build a new dictionary if one is already built for specified task when using `PytorchDataTeacher`


stephenroller:
Thanks

PR #1377, opened 2019-01-22
Add additional optimizers and optimizer params to torchagent

alexholdenmiller:
is there a better way to set up the imports for OPTIM_OPTS?
we made that a class field so that subclasses can just add more if they want, but maybe there's a cleaner way

this adds --nus, --betas, --nesterov, and the qhm and qhadam optimizers from https://openreview.net/pdf?id=S1fUpoR5FQ

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -49,6 +49,7 @@ def str2bool(value):
     else:
         raise argparse.ArgumentTypeError('Boolean value expected.')
 
+
 def str2floats(s):
     """Look for single float or comma-separated floats."""
     return tuple(float(f) for f in s.split(','))
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -49,6 +49,10 @@ def str2bool(value):
     else:
         raise argparse.ArgumentTypeError('Boolean value expected.')
 
+def str2floats(s):
+    """Look for single float or comma-separated floats."""
+    return tuple(float(f) for f in s.split(','))
+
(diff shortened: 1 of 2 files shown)


alexholdenmiller:
so far qhm/qhadam don't seem to be performing much different from sgd/sgd-nesterov or adam as long as you set the learning rate right

alexholdenmiller:
I'll upload a spreadsheet later

stephenroller:
Would you mind trying with transformer/generator? That's one of the models the paper says it improves stability on.

alexholdenmiller:
haven't run the transformer yet, but updated results for seq2seq.

example command from sweep: `python -u examples/train_model.py -t convai2:self -bsrt True --max-train-time 13680 --model seq2seq -pt False -hs 1024 -esz 300 -att general -nl 2 -rnn lstm -lr 54 -dr 0.1 -clip 0.1 -lt enc_dec -opt sgd -mom 0.99 -emb glove --warmup-updates 1000 -bs 128 --validation-every-n-epochs 0.5 --validation-metric ppl --validation-metric-mode min --validation-patience 6 --save-after-valid False --load-from-checkpoint False --dict-lower True --dict-maxexs -1 --dict-include-valid True --log-every-n-secs 10 -mf /tmp/model`

Sweep results that got below 39 ppl on valid (last two columns are warmup-updates and best valid ppl):
screenshot 2019-01-28 18 04 57

PR #1380, opened 2019-01-23
Fix for TFIDF Retriever with Wikipedia

klshuster:
Had to make sure wikipedia teacher provided a labels field, otherwise one cannot train a tfidif retriever with wikipedia


klshuster:
I must've trained it prior to changing the logic of the agent; using the models in the model zoo still work for interactive

klshuster:
Tests pass locally, merging

PR #1385, opened 2019-01-25
Fixes with pyteacher and distributed training.

stephenroller:
- Don't use `-pyt` at validation time
- Get rid of a the _data_generator field. This didn't seem to actually be used, and wasn't pickleable. This made for problems with respect to combining pyt with distributed training in some cases.

### DIFF
--- a/parlai/core/pytorch_data_teacher.py
+++ b/parlai/core/pytorch_data_teacher.py
@@ -438,7 +438,6 @@ def process(ex_or_batch):
     ParlAI Implementations of Pytorch Datasets
 """
 
-
 class StreamDataset(Dataset):
     """A Pytorch Dataset utilizing streaming"""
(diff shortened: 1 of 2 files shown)


acha21:
Can I know the reason why you did disable to use -pyt at validation time?
I want to use pytorch_teacher_task pre-processed data while validation time, as it takes so long time in validation task if I use non preprocessed data.
Is there any problem if I enable it?

stephenroller:
According to my chat logs with @klshuster, there was a very bad mix with distributed training at validation time, but I confess I don’t remember the specifics.

You’re most welcome to undo this change and test it. Perhaps confirm you get the same results on a subset to be safe. Please report back what you find in this PR.

It’s worth noting that over the next few months I expect we will implement changes to mitigate the downsides of using a regular teacher.

PR #1386, opened 2019-01-25
Safeguards around using invsqrt lr scheduling without warmup.

stephenroller:
if someone tried to use `--lr-scheduler invsqrt` but didn't provide a `--warmup-updates` argument, they'd have a bad time.


stephenroller:
Note that we already have one guard:

https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/torch_agent.py#L414-L416

But I still managed to trigger a divide by zero in some obscure setup, so might as well make it safe.

PR #1388, opened 2019-01-25
Fix issue with subclassing classmethod

EricMichaelSmith:
To avoid a `TypeError: super(type, obj): obj must be an instance or subtype of type` error when subclassing `TransformerRankerAgent`

For reference:
- https://stackoverflow.com/questions/1269217/calling-a-base-classs-classmethod-in-python
- https://stackoverflow.com/questions/15291302/how-to-call-a-parent-classs-classmethod-from-an-overridden-classmethod-in-pyt
- https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/torch_generator_agent.py#L257

### DIFF
--- a/parlai/agents/seq2seq/seq2seq.py
+++ b/parlai/agents/seq2seq/seq2seq.py
@@ -88,8 +88,7 @@ class Seq2seqAgent(TorchGeneratorAgent):
         agent.add_argument('-idr', '--input-dropout', type=float, default=0.0,
                            help='Probability of replacing tokens with UNK in training.')
 
-        super(cls, Seq2seqAgent).add_cmdline_args(argparser)
-        Seq2seqAgent.dictionary_class().add_cmdline_args(argparser)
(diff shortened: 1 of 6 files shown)


stephenroller:
Added the other cases, want @emilydinan 's approval too since she found the same bug.

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -88,8 +88,7 @@ def add_cmdline_args(cls, argparser):
         agent.add_argument('-idr', '--input-dropout', type=float, default=0.0,
                            help='Probability of replacing tokens with UNK in training.')
 
-        super(cls, Seq2seqAgent).add_cmdline_args(argparser)
-        Seq2seqAgent.dictionary_class().add_cmdline_args(argparser)
```
emilydinan: did you mean to delete this line

alexholdenmiller:
thanks! I got this bug too actually in a local branch with a weird init, thanks for finding the fix--looks good after the fix emily pointed out

PR #1389, opened 2019-01-25
Broken pyt integration test

stephenroller:
Broken for two reasons:
- Edge case when there's no `--dict-file`, which causes a call to `os.path.exists(None)` and crashes
- I messed up merges for #1316 and #1335 which caused one to undo the other.


klshuster:
I take care of the first fix (the dict part) in #1392

PR #1392, opened 2019-01-25
Pyt Dict File

klshuster:
Ensures that if a model file is specified, we save the dictionary there (rather than the default PytorchDataTeacher file).

Integration tests pass on local machine.

### DIFF
--- a/parlai/scripts/build_pytorch_data.py
+++ b/parlai/scripts/build_pytorch_data.py
@@ -28,7 +28,7 @@ from collections import deque
 def get_pyt_dict_file(opt):
     if opt.get('dict_file') and os.path.exists(opt.get('dict_file')):
         return opt['dict_file']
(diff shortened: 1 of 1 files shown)


parlai/scripts/build_pytorch_data.py
```diff
@@ -26,8 +26,10 @@
 
 
 def get_pyt_dict_file(opt):
-    if os.path.exists(opt.get('dict_file', '')):
+    if opt.get('dict_file') and os.path.exists(opt.get('dict_file')):
         return opt['dict_file']
+    if opt.get('dict_file', None) is None and opt.get('model_file'):
```
stephenroller: maybe just `opt.get('dict_file') is None and opt.get('model_file')`; the None is the default argument for `opt.get`

klshuster: right, i'll change that

PR #1401, opened 2019-01-29
Shortcut for using agents from the projects folder

emilydinan:
Similar to the way we use `internal:` to get agents in internal, look for agents in a projects folder by specifying `projects:` and the name of the projects folder.

This is useful for using `projects:personachat:kvmemnn` or for when we add project specific model code (like we are doing with wizard).

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -405,8 +405,8 @@ def get_agent_module(dir_name):
                                ''.format(dir_name))
         folder_name = s[1]
         model_name = s[2]
-        module_name = 'projects.{v}.{m}.{m}'.format(
-            m=model_name, v=folder_name)
+        module_name = 'projects.{p}.{m}.{m}'.format(
+            m=model_name, p=folder_name)
         class_name = name_to_agent_class(model_name)
     elif ':' in dir_name:
         # e.g. -m "parlai.agents.seq2seq.seq2seq:Seq2seqAgent"
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -565,6 +565,7 @@ def get_task_module(taskname):
     teacher_class = getattr(my_module, teacher)
     return teacher_class
 
+
 def add_task_flags_to_agent_opt(agent, opt, flags):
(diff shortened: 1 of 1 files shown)


parlai/core/agents.py
```diff
@@ -392,6 +396,18 @@ def get_agent_module(dir_name):
...
+    elif dir_name.startswith('projects:'):
+        # e.g. -m projects:personachat:kvmemnn
+        s = dir_name.split(':')
+        if len(s) != 3:
+            raise RuntimeError('projects paths should follow pattern '
+                               'projects:folder:model; you used {}'
+                               ''.format(dir_name))
+        folder_name = s[1]
+        model_name = s[2]
+        module_name = 'projects.{v}.{m}.{m}'.format(
```
stephenroller: let's make it `{p}` instead of `{v}`.

parlai/core/agents.py
```diff
@@ -392,6 +396,18 @@ def get_agent_module(dir_name):
         module_name = 'parlai.agents.legacy_agents.{m}.{m}_v{v}'.format(
             m=model_name, v=s[2])
         class_name = name_to_agent_class(model_name)
+    elif dir_name.startswith('projects:'):
+        # e.g. -m projects:personachat:kvmemnn
+        s = dir_name.split(':')
+        if len(s) != 3:
+            raise RuntimeError('projects paths should follow pattern '
+                               'projects:folder:model; you used {}'
+                               ''.format(dir_name))
```
stephenroller: Maybe we should say that this maps to `parlai.projects.folder.model.model`?

parlai/core/agents.py
stephenroller: Just wondering: does this automatically insert the `parlai.`? It probably should?

Also wondering how this deals with capitalizations

parlai/core/agents.py
emilydinan: it doesn't, projects is outside of parlai

stephenroller:
Feel free to merge without travis because this didn't incorporate #1407

PR #1410, opened 2019-01-30
[Wizard of Wikipedia] Ranker Agent

emilydinan:
Agent that inherits from Transformer Ranker Agent that allows us to incorporate and condition on knowledge, as in the wizard of wikipedia task.

Other small changes:
- add init model flag in torch ranker agent
- if we specifically specify a dictionary we should be able to load that one (cc @stephenroller and @klshuster  -- can you check this change?)


projects/wizard/wizard_transformer_ranker/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) 2017-present, Facebook, Inc.
```
stephenroller: We should change the filename to projects/wizard_of_wikipedia/...

emilydinan:
@stephenroller @klshuster ready for review

projects/wizard_of_wikipedia/wizard_transformer_ranker/wizard_transformer_ranker.py
```diff
@@ -0,0 +1,123 @@
...
+
+import numpy as np
+import torch
+
+
+class WizardTransformerRankerAgent(TransformerRankerAgent):
+    @classmethod
+    def add_cmdline_args(cls, argparser):
+        """Add command-line arguments specifically for this agent."""
+        super(WizardTransformerRankerAgent, cls).add_cmdline_args(argparser)
```
klshuster: nit: `super().add_cmdline_args` works here

stephenroller: does it??!

klshuster: oops, no you're right nevermind

emilydinan:
Made changes requested by @klshuster and @stephenroller, plus a script for evaluating a pre-trained retrieval model 😄 

Later will add interactive scripts.

projects/wizard_of_wikipedia/scripts/eval_retrieval_model.py
```diff
@@ -0,0 +1,40 @@
...
+from parlai.core.params import ParlaiParser
+from parlai.scripts.eval_model import eval_model
+from parlai.zoo.wizard_of_wikipedia.full_dialogue_retrieval import download
+from projects.wizard_of_wikipedia.wizard_transformer_ranker\
+    .wizard_transformer_ranker import WizardTransformerRankerAgent
+
+"""Evaluate pre-trained retrieval model on the full Wizard Dialogue task.
+
+NOTE: Metrics here differ slightly to those reported in the paper as a result
+of code changes.
```
klshuster: do you maybe want to add what scores we get with the new model?

projects/wizard_of_wikipedia/wizard_transformer_ranker/wizard_dict.py
```diff
@@ -0,0 +1,83 @@
...
+
+class WizardDictAgent(DictionaryAgent):
+    def __init__(self, opt, shared=None):
+        # initialize fields
+        self.opt = copy.deepcopy(opt)
+        self.minfreq = opt.get('dict_minfreq', DictionaryAgent.default_minfreq)
+        self.null_token = '__PAD__'
+        self.end_token = '__SOC__'
+        self.unk_token = '__UNK__'
+        self.start_token = '__SOC__'
```
klshuster: start and end are same token? interesting, didn't realize before

projects/wizard_of_wikipedia/wizard_transformer_ranker/wizard_transformer_ranker.py
```diff
@@ -0,0 +1,144 @@
...
+            '--legacy', type='bool', default=False,
+            help='legacy model'
+        )
+        argparser.set_defaults(
+            learningrate=0.0008,
+            eval_candidates='inline',
+            candidates='batch',
+            lr_factor=1,
+            add_p1_after_newln=False,
+            delimiter=' '
```
klshuster: above we set `delimiter` to `__SOC__` - why is this one different?

projects/wizard_of_wikipedia/wizard_transformer_ranker/wizard_dict.py
emilydinan: actually i think there shouldn't be an end token at all

emilydinan: nice catch

projects/wizard_of_wikipedia/wizard_transformer_ranker/wizard_transformer_ranker.py
emilydinan: `__SOC__` was only used in legacy models, because we were finetuning on top of a model we trained this way. Since this isn't standard in ParlAI I think we should only do this in the case of legacy models.

PR #1411, opened 2019-01-30
[wip] add context handling to torch agent (default merges into text)

alexholdenmiller:
- for memnn, set split_lines = True on context, like we do for text, so that our memories are populated from both the context and text fields
- [undocumented and kind of hacky] add ability to push message dicts to the DialogData class. most of the line changes there are indentation changes.
- add context_vec and context_lens to Batch
- add context history to torch agent
- by default (and when -ctx False), merge context into text field (prepend with newline) and delete context field
- [need to undo this or add flag to turn it on] babi puts info sentences into context and questions in text


parlai/core/teachers.py
```diff
@@ -691,22 +694,25 @@ def build_table(self, entry):
...
-                    if entry[3] is not None:
-                        table['label_candidates'] = entry[3]
-                    if len(entry) > 4 and entry[4] is not None:
-                        img = self.image_loader.load(entry[4])
-                        if img is not None:
-                            table['image'] = img
+        if isinstance(entry, dict):
+            table = entry
+        else:
+            table = {}
```
stephenroller: I'm pretty sure that you don't need the cascading ifs right?

```python
if len(entry) > 0 and entry[0] is not None:
    table['text'] = entry[0]
if len(entry) > 1 and entry[1] is not None:
     table['labels'] = entry[1]
if len(entry) > 2 and entry[2] is not None:
    table['reward'] = entry[2]
# ...
```

really I think you can:

```python
for key, value in zip(['text', 'labels', 'reward', 'label_candidates'], entry):
    if value is not None:
        table[key] = value
# special case on handling images
if len(entry) > 4 and entry[4] is not None:
    table['image'] = self.image_loader.load(entry[4])
```
since zip only iterates through the shortest list:

```
>>> list(zip("abcd", [1, 2, 3]))
[('a', 1), ('b', 2), ('c', 3)]
```

parlai/core/torch_agent.py
```diff
@@ -435,9 +454,7 @@ def report(self):
         return metrics
 
     def _is_lr_warming_up(self):
-        """
-        Checks if we're warming up the learning rate.
-        """
+        """Checks if we're warming up the learning rate."""
```
stephenroller: :'(

parlai/core/torch_agent.py
```diff
@@ -993,10 +1059,16 @@ def observe(self, observation):
 
         This includes remembering the past history of the conversation.
         """
+        obs = observation.copy()
+
+        if 'context' in obs and not self.opt.get('parse_context', False):
+            # default to merging context into text, not parsing it separately
+            obs['text'] = obs.pop('context') + '\n' + obs['text']
```
stephenroller: brain dump:

Should we do the .copy here instead?

As I recall, one advantage of NOT copying the data here is that the vectorized text is saved in case we visit this observation again. (We actually go through great lengths in `_set_text_vec` etc to do this). In fact, I'm not even sure that pytorch data teacher doesn't rely on that (admittedly devious) behavior -- I'd have to check.

It gets real weird because we're also going to start adding p1 after newlines, which might get into the context? Though I guess we're already doing that, aren't we?

parlai/tasks/babi/agents.py
```diff
@@ -64,12 +64,24 @@ def __init__(self, opt, shared=None):
         self.task_num = task.split(':')[2]
         opt['datafile'] = _path('-10k', self.task_num, opt)
         opt['cands_datafile'] = _path('-10k', task.split(':')[2], opt, 'train')
+        # TODO: change to real check
```
stephenroller: Just making sure this TODO doesn't get forgotten.

parlai/core/teachers.py
```diff
@@ -589,55 +589,58 @@ def _read_episode(self, data_loader):
...
-                    # process labels if available
-                    if entry[1] is None:
-                        new_entry.append(None)
-                    elif hasattr(entry[1], '__iter__') and type(entry[1]) is not str:
-                        # make sure iterable over labels, not single string
-                        new_entry.append(tuple(sys.intern(e) for e in entry[1]))
+            if isinstance(entry, dict):
+                episode.append(entry)
+            else:
+                # intern all strings so we don't store them more than once
```
stephenroller: I didn't realize there was a call to sys.intern here -- I actually wasn't familiar with that call. I wonder cheap that one time cost is and how much it gains us later on.

parlai/core/teachers.py
```diff
@@ -589,55 +589,58 @@ def _read_episode(self, data_loader):
...
-                                new_entry.append(tuple(
-                                    sys.intern(e) for e in entry[3]))
+                        elif hasattr(entry[1], '__iter__') and type(entry[1]) is not str:
+                            # make sure iterable over labels, not single string
+                            new_entry.append(tuple(sys.intern(e) for e in entry[1]))
+                        else:
+                            raise TypeError(
+                                'Must provide iterable over labels, not a single string.'
+                            )
+                        if len(entry) > 2:
```
stephenroller: You can use this chance to unindent some of these right? no need to put the if X > 3 inside X > 2?

emilydinan:
> Started looking at this because I'm worried that it could conflict with complicate something @emilydinan was considering doing (I think it would...)
> 
> She was considering storing history as pre-vectorized (and then concating) rather than storing strings and re-vectorizing. As it is, we're making an order of magnitude more calls to `DictionaryAgent.__getitem__` compared to some models that were doing it smarter, and we're paying a very heavy penalty for this.

After testing the hypothesis that re-tokenizing the history was causing a major slow down for us, I feel very confident that we should do this. Taking a quick look at this code, I don't see why it should conflict with what I'm doing, as context history is handled separately from dialog history.

parlai/core/teachers.py
alexholdenmiller: brill

alexholdenmiller:
Still cleaning this up and trying some experiments with it, but here's an interesting one...

After cleaning up the DialogData class to nicer code and trying to just use `dict` instead of `tuple` for storage and removing the sys.intern calls, I got much faster loading but much more memory usage.

Here's the test I ran:
`time python examples/display_data.py -t opensubtitles -dt train:ordered`

Timing results to display ten examples after loading the whole dataset and memory usage at the end:
```
master  --  36m, 30.5gb RAM
master with sys.intern removed  --  34m, 35.4gb RAM
this branch with intern added back  --  22m, 53.8gb RAM
this branch  --  18m, 58.46gb RAM
```

So sys.intern on opensubtitles 2018 seems to save about 5GB of RAM! But does take a significant amount of time to run.

I believe the remaining differences is storing the examples in tuples vs dicts but I didn't expect it to be that huge and haven't had the chance to dig deeper yet.

alexholdenmiller:
abandoning for now

PR #1413, opened 2019-01-31
[Torch Ranker] faster metrics calculations

emilydinan:
Option to not calculate metrics at all during train time. If training with batch candidates, added a faster metrics calculation (resulting in a speedup ~1.4x).

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -43,9 +43,9 @@ class TorchRankerAgent(TorchAgent):
                  "the flag --fixed-candidates-path. By default, this file is created "
                  "once and reused. To replace it, use the 'replace' option.")
         agent.add_argument(
-            '--train-predict', type='bool', default=True,
+            '--train-predict', type='bool', default=False,
             help='Get predictions and calculate mean rank during the train '
-                 'step. Turning this off will speed up training.'
+                 'step. Turning this on may slow down training.'
         )
 
     def __init__(self, opt, shared=None):
@@ -111,20 +111,20 @@ class TorchRankerAgent(TorchAgent):
             'Abstract class: user must implement build_model()')
 
 
-    def get_train_preds(self, scores):
-        if self.opt['candidates'] == 'batch':
-            batchsize = scores.size(0)
-            # get accuracy
-            targets = scores.new_empty(batchsize).long()
-            targets = torch.arange(batchsize, out=targets)
-            nb_ok = (scores.max(dim=1)[1] == targets).float().sum().item()
-            self.metrics['train_accuracy'] += nb_ok
-            # calculate mean rank
-            above_dot_prods = scores - scores.diag().view(-1, 1)
-            rank = (above_dot_prods > 0).float().sum().item()
-            self.metrics['rank'] += rank
-            return Output()
+    def get_batch_train_metrics(self, scores):
+        batchsize = scores.size(0)
+        # get accuracy
+        targets = scores.new_empty(batchsize).long()
+        targets = torch.arange(batchsize, out=targets)
+        nb_ok = (scores.max(dim=1)[1] == targets).float().sum().item()
+        self.metrics['train_accuracy'] += nb_ok
+        # calculate mean rank
+        above_dot_prods = scores - scores.diag().view(-1, 1)
+        rank = (above_dot_prods > 0).float().sum().item()
+        self.metrics['rank'] += rank
 
+
+    def get_train_preds(self, scores):
         # TODO: speed these calculations up
         _, ranks = scores.sort(1, descending=True)
         for b in range(batchsize):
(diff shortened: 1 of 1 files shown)


parlai/core/torch_ranker_agent.py
```diff
@@ -117,23 +151,16 @@ def train_step(self, batch):
...
 
-        # Get predictions but not full rankings for the sake of speed
-        if cand_vecs.dim() == 2:
-            preds = [cands[ordering[0]] for ordering in ranks]
-        elif cand_vecs.dim() == 3:
-            preds = [cands[i][ordering[0]] for i, ordering in enumerate(ranks)]
-        return Output(preds)
+        # Get train predictions
+        if self.opt.get('train_predict', True):
+            return self.get_train_preds(scores)
```
stephenroller: in the else case, can we throw a warn_once that some training metrics are omitted for training speed?

Is there a middle ground where we produce partial metrics or something?

parlai/core/torch_ranker_agent.py
```diff
@@ -323,6 +351,8 @@ def report(self):
             m['loss'] = self.metrics['loss']
             m['mean_loss'] = self.metrics['loss'] / examples
             m['mean_rank'] = self.metrics['rank'] / examples
+            if self.metrics['train_accuracy'] > 0:
+                m['train_accuracy'] = self.metrics['train_accuracy'] / examples
```
stephenroller: does it make more sense to guard this with self.opt rather than is 0? What if I literally get 0 accuracy lol.

parlai/core/torch_ranker_agent.py
```diff
@@ -104,6 +110,34 @@ def build_model(self):
         raise NotImplementedError(
             'Abstract class: user must implement build_model()')
 
+
+    def get_train_preds(self, scores):
+        _, ranks = scores.sort(1, descending=True)
```
stephenroller: Let's have a  TODO here saying we should optimize this some. The `sort` is a huge red flag, and probably the bottom for loops could be vectorized.

parlai/core/torch_ranker_agent.py
emilydinan: good point lol will update

parlai/core/torch_ranker_agent.py
emilydinan: oh i should move that after the if statement, we don't actually use ranks if cands=batch

emilydinan: will also add todo

parlai/core/torch_ranker_agent.py
emilydinan: We still calculate train loss. We just omit accuracy and rank.

emilydinan: and also f1 i guess

parlai/core/torch_ranker_agent.py
```diff
@@ -42,6 +42,11 @@ def add_cmdline_args(cls, argparser):
                  "<cands_name> is the name of the file (not the full path) passed by "
                  "the flag --fixed-candidates-path. By default, this file is created "
                  "once and reused. To replace it, use the 'replace' option.")
+        agent.add_argument(
+            '--train-predict', type='bool', default=True,
+            help='Get predictions and calculate mean rank during the train '
+                 'step. Turning this off will speed up training.'
+        )
```
alexholdenmiller: tbh I think defaulting to False might even be better, you do have the warning for anyone who might be surprised and then we won't have to remember to set it to False every time we are training (we expect this to be the best experience for the user)

parlai/core/torch_ranker_agent.py
```diff
@@ -104,6 +110,35 @@ def build_model(self):
...
+            # get accuracy
+            targets = scores.new_empty(batchsize).long()
+            targets = torch.arange(batchsize, out=targets)
+            nb_ok = (scores.max(dim=1)[1] == targets).float().sum().item()
+            self.metrics['train_accuracy'] += nb_ok
+            # calculate mean rank
+            above_dot_prods = scores - scores.diag().view(-1, 1)
+            rank = (above_dot_prods > 0).float().sum().item()
+            self.metrics['rank'] += rank
+            return Output()
```
alexholdenmiller: this section looks pretty cheap to compute. if that's true (@emilydinan can you verify?), then I think we should leave it on by default always (no need to gate a cheap metric behind a flag, better to have more metrics available without having to rerun your script with a different flag to see them)

emilydinan: verified, will update

parlai/core/torch_ranker_agent.py
```diff
@@ -117,23 +150,23 @@ def train_step(self, batch):
...
-        elif cand_vecs.dim() == 3:
-            preds = [cands[i][ordering[0]] for i, ordering in enumerate(ranks)]
-        return Output(preds)
+        # Get train predictions
+        if self.opt['candidates'] == 'batch':
+            self.get_batch_train_metrics(scores)
+            return Output()
+        if not self.opt.get('train_predict', False):
+            warn_once(
+                "Some training metrics are ommitted for speed. Set the flag "
```

PR #1416, opened 2019-02-01
[WIP] for rankers distributed accross gpus, allow to use candidates on the other GPUs as negatives.

samhumeau:
The core of it is DistributionConcatenation2D (see comments on it) that allow to concatenate several vectors from various GPU into one big (let's call it 'y') and then use y on all the GPUs. Gradients are back-propagated.

At the core of it there are 2 all_gather operations, one for forward pass and one for backward.

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -10,6 +10,7 @@ import math
 import numpy as np
 
 from parlai.core.torch_generator_agent import TorchGeneratorModel
+from parlai.core.distributed_utils import DistributionConcatenation2D
 
 
 def _normalize(tensor, norm_layer):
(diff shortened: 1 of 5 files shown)


parlai/agents/transformer/modules.py
```diff
@@ -150,6 +155,12 @@ def forward(self, xs, mems, cands):
             context_h = context_h / context_h.norm(2, dim=1, keepdim=True)
             cands_h = cands_h / cands_h.norm(2, dim=1, keepdim=True)
 
+        if self.opt["gather_from_other_nodes"]:
```
stephenroller: Might be some conflict with #1410

stephenroller:
Closing since #1596 solves this problem simpler and it doesn't seem like we will complete this.

PR #1418, opened 2019-02-01
Delay import of torch text.

stephenroller:
I misunderstood why travis broke when I okayed #1414 merge.


stephenroller:
Seq2seq didn't quite hit 0.95 BLEU so travis is sad, but the main breaker is fixed.

PR #1419, opened 2019-02-01
Training support for preemption.

stephenroller:
Would appreciate some extra brain cycles from y'all on this.

The scenario we're trying to capture is:

1. Ctrl-C is pressed.
2. The trainer has 30-60 seconds to shutdown as cleanly as possible. Running a full validation and saving is risky, so ideally we'd like to *be able to restore anytime.
3. You're forcefully quit (kill -9).
4. The *exact same command* is run again. The trainer should attempt to restore any state needed.

These events can happen *at any moment*. To be careful, I catch keyboard interrupts around model.save.

In an effort to do this, I now store a `.trainstats` file. This stores the current training time, number of epochs processed, and the validation patience. That way we can load back up from where we are. There's some nuance around loading from checkpoint vs loading from validation, but I think I cover both those cases. If anyone could think about whether I cover anything.

The downside is that SLURM sends the exact same message whether you're being killed or whether you're being preempted; as such, the launch script will requeue your job even if it runs out of SLURM time. As such, the current default parameter settings for trainloop will cause an infinite loop! @jaseweston points out we should probably change *some* default. His suggestion was `--max-training-time` could be 24 hours by default.

### DIFF
--- a/parlai/scripts/train_model.py
+++ b/parlai/scripts/train_model.py
@@ -23,6 +23,11 @@ Examples
 # TODO List:
 # * More logging (e.g. to files), make things prettier.
 
+import numpy as np
+import os
+import signal
+import json
+
 from parlai.core.agents import create_agent, create_agent_from_shared
 from parlai.core.worlds import create_task
 from parlai.core.params import ParlaiParser
@@ -32,9 +37,7 @@ from parlai.scripts.build_dict import build_dict, setup_args as setup_dict_args
 from parlai.core.distributed_utils import (
     sync_object, is_primary_worker, all_gather_list, is_distributed, num_workers
 )
-import numpy as np
 from parlai.scripts.build_pytorch_data import get_pyt_dict_file
-import os
 
 
 def setup_args(parser=None):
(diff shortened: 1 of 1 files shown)


parlai/scripts/train_model.py
```diff
@@ -242,9 +253,59 @@ def __init__(self, opt):
         self.saved = False
         self.valid_world = None
         self.opt = opt
+
+        # we may have been preempted, make sure we note that amount
+        self._preempted_epochs = 0.0
+        if (
+            opt.get('model_file') and
+            os.path.isfile(opt['model_file'] + trainstats_suffix)
```
klshuster: so is it _necessarily_ the case that we were preempted if there's a trainstats file? it looks like this is saved after every training run

stephenroller: No, but if you finished training because of `--max-training-time` and then ran train_model again, you would immediately exit due to exceeding time.

klshuster: Ok, and I assume that is similar behavior to specifying `--num-epochs`

PR #1420, opened 2019-02-02
Add empathetic dialogue data

EricMichaelSmith:
Add data task for the empathetic-dialogue paper of Rashkin, et al. (https://arxiv.org/abs/1811.00207). Code was written by Hannah Rashkin and revised by Margaret Li and myself.


EricMichaelSmith:
@stephenroller I just added this - I'll also add the license file itself to the data .tar file on the S3 bucket

PR #1421, opened 2019-02-04
[Torch Agent] add a history class

emilydinan:
[NOTE: work in progress, will likely undergo major revisions]

Add a History class to track history. 

Some benefits to a history class:
- allows us to track the history of any arbitrary field
- keeps tokenized version of history cached as well as the string version
- easily subclassed if an agent needs some special casing 

Some changes to the Torch Agent set up:
- History is passed into `vectorize`. This will allow people to easily override how the history is used in 
`set_text_vec`
- Added `use_memories` as a flag


stephenroller:
apparently made these both those comments and then didn't submit the review, so I put them there for posterity. they might not be relevant anymore.

On another note, I merged #1432, which should help give signal on stability here.

parlai/core/torch_agent.py
```diff
@@ -731,31 +884,44 @@ def _check_truncate(self, vec, truncate, truncate_left=False):
         else:
             return vec[:truncate]
 
-    def _set_text_vec(self, obs, truncate, split_lines):
+    def _make_long_tensor(self, vec):
+        if 'torch' not in str(type(vec)):
```
stephenroller: isinstance

emilydinan:
Some newer significant updates:
- get rid of memories in torch agent, move handling of them to MemNN. In general, agents will have to decide how they want to implement this. (See how I handled for Transformer Ranker for instance).
- update torch agent unit tests to work with this new setup

parlai/agents/transformer/transformer.py
```diff
@@ -117,17 +115,31 @@ def build_model(self, states=None):
...
+            valid_inds, exs = zip(*valid_obs)
+            mems = None
+            if any('memory_vecs' in ex for ex in exs):
+                mems = [ex.get('memory_vecs', None) for ex in exs]
+            batch.memory_vecs = mems
+        return batch
+
+    def _vectorize_memories(self, obs):
+        raise NotImplementedError(
+            'Abstract class: user must implement this function to use memories'
```
stephenroller: Will this change? I don't think that Transformer is an abstract class...

parlai/core/torch_agent.py
```diff
@@ -135,6 +128,151 @@ def __init__(self, text=None, text_candidates=None, **kwargs):
...
+        self.p2_token = p2_token
+
+        # tracking when to clear history
+        self.clear_on_next_update = False
+
+    def parse(self, text):
+        """Tokenize text with the given dictionary."""
+        return self.dict.txt2vec(text)
+
+    def clear(self):
```
stephenroller: should we use reset? It'd be more consistent.

parlai/core/torch_agent.py
```diff
@@ -191,6 +329,14 @@ def dictionary_class():
         """
         return DictionaryAgent
 
+    @staticmethod
```
stephenroller: classmethod please

parlai/core/torch_agent.py
```diff
@@ -321,14 +476,15 @@ def __init__(self, opt, shared=None):
         """Initialize agent."""
         super().__init__(opt, shared)
         opt = self.opt
-
         if not shared:
             # intitialize any important structures from scratch
             self.replies = {}  # past replies
             self.dict = self.dictionary_class()(opt)
             if opt.get('person_tokens'):
-                self.dict[self.P1_TOKEN] = 999999999
```
stephenroller: ???? This originally had the side effect of ADDING the tokens to the dictionary. Why add the guard?

parlai/core/torch_agent.py
```diff
@@ -367,22 +523,28 @@ def __init__(self, opt, shared=None):
...
+            p2_token=self.P2_TOKEN,
+            dict_agent=self.dict,
+        )
 
         self.rank_candidates = opt['rank_candidates']
         self.add_person_tokens = opt.get('person_tokens', False)
 
     def init_optim(self, params, optim_states=None, saved_optim_type=None):
         """Initialize optimizer with model parameters.
-
```
stephenroller: reinsert

parlai/core/torch_agent.py
```diff
@@ -821,7 +972,7 @@ def _set_label_cands_vec(self, obs, add_start, add_end, truncate):
                 for c in obs['label_candidates']]
         return obs
 
-    def vectorize(self, obs, add_start=True, add_end=True, split_lines=False,
+    def vectorize(self, obs, history, add_start=True, add_end=True,
```
stephenroller: Why have the history passed in as a parameter, instead of just using self.history?

tests/test_torch_agent.py
```diff
@@ -49,6 +49,9 @@ def __getitem__(self, key):
             self.idx += 1
             return self.idx
 
+    def get(self, key):
```
stephenroller: This shouldn't be here, DictAgent doesn't have a `get`, so the mock dictagent shouldn't have one.

parlai/core/torch_agent.py
```diff
@@ -745,31 +897,30 @@ def _check_truncate(self, vec, truncate, truncate_left=False):
         else:
             return vec[:truncate]
 
-    def _set_text_vec(self, obs, truncate, split_lines):
+    def _make_long_tensor(self, vec):
```
stephenroller: We can just use calls to `torch.LongTensor()` directly. If it's already a long tensor, it just returns itself.

parlai/core/torch_agent.py
```diff
@@ -712,26 +881,9 @@ def _vectorize_text(self, text, add_start=False, add_end=False,
...
-            vec = vec[len(vec) - truncate:]
-        else:
-            # truncate from the right side
-            # don't check add_end, we know we are truncating it
-            vec = vec[:truncate - add_start]
-            if add_start:
-                # always keep the start token if it's there
-                vec.insert(0, self.START_IDX)
-        tensor = torch.LongTensor(vec)
+        vec = self._add_start_end_tokens(vec, add_start, add_end)
```
stephenroller: I'm tempted to propose we have `add_start` and `add_end` as arguments to txt2vec instead. It would make memoization a lot easier.

parlai/core/torch_agent.py
emilydinan: to make the test_torch_agent test pass

emilydinan: could refactor `MockDictAgent` instead

emilydinan:
> should history.clear() be in reset or something?

it is in reset. i'll name it to `reset` to be more clear

parlai/agents/transformer/transformer.py
emilydinan: discussed offline, Transformer is not an abstract class, but in order to use the memnet aspect we will need to subclass it. the reason i made this method abstract was because the way we use memories is completely task dependent

parlai/core/torch_agent.py
emilydinan: I guess the same reason we pass in observation instead of using self.observation (easier abstraction???)? thoughts @alexholdenmiller?

parlai/core/torch_agent.py
stephenroller: That seems more reasonable. The tests shouldn't dictate changes to the code.

parlai/agents/transformer/transformer.py
stephenroller: I agree with Emily's logic.

Should this actually be in torch ranker agent.

If it stays in  transformer, imho we should leave a more descriptive explanation here.

stephenroller: Also does this necessitate fixing MemNet? Not sure we have a good test for it yet.

parlai/core/torch_agent.py
emilydinan: yep fixing

parlai/core/torch_agent.py
emilydinan: like in `DictionaryAgent`?

tests/test_torch_agent.py
emilydinan: fixed

parlai/agents/transformer/transformer.py
emilydinan: mems will always be None so it won't break anything

emilydinan: as far as moving to torch ranker agent, seems reasonable?

stephenroller: Let's just add a comment TODO: bubble this up to TRA? and unblock this PR

parlai/core/torch_agent.py
stephenroller: Good enough for me

parlai/core/torch_agent.py
stephenroller: Yeah, not in this PR though.

tests/test_torch_agent.py
```diff
@@ -535,126 +535,123 @@ def test__add_person_tokens(self):
...
-        out = agent._add_person_tokens(text, prefix, add_after_newln=False)
+        out = agent.history._add_person_tokens(text, prefix, add_after_newln=False)
         self.assertEqual(out, prefix + ' ' + text)
-        out = agent._add_person_tokens(text, prefix, add_after_newln=True)
+        out = agent.history._add_person_tokens(text, prefix, add_after_newln=True)
         idx = text.rfind('\n') + 1
         self.assertEqual(out, text[:idx] + prefix + ' ' + text[idx:])
 
-    def test_get_dialog_history(self):
+    def test_dialog_history(self):
```
stephenroller: Does it make more sense to divorce this test from the agent and have it only test the History object?

emilydinan: I think this test is good, but I can add more tests to make it a more comprehensive History object test

PR #1422, opened 2019-02-04
[TorchAgent] Separate truncate options; skip vectorizing candidates unnecessarily

emilydinan:
- Don't vectorize label candidates if we don't have to
- Remove truncate from vectorize args (@stephenroller, @alexholdenmiller is this ok? i checked that this won't affect anything that subclasses TorchAgent but i'm not sure why we had it as an arg in the first place)
- Allow us to set separate truncation lengths for LHS and RHS

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -807,7 +807,8 @@ class TorchAgent(Agent):
                 for c in obs['label_candidates']]
         return obs
 
-    def vectorize(self, obs, add_start=True, add_end=True, split_lines=False):
+    def vectorize(self, obs, add_start=True, add_end=True, split_lines=False,
+                  text_truncate=None, label_truncate=None):
         """Make vectors out of observation fields and store in the observation.
 
         In particular, the 'text' and 'labels'/'eval_labels' fields are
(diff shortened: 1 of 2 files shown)


parlai/core/torch_agent.py
```diff
@@ -263,6 +263,16 @@ def add_cmdline_args(cls, argparser):
         agent.add_argument(
             '-tr', '--truncate', default=-1, type=int,
             help='Truncate input lengths to increase speed / use less memory.')
+        agent.add_argument(
+            '--left-truncate', type=int,
```
stephenroller: nit: can we call them text-truncate and cand-truncate? when I hear "left-truncate", I think of which side I'm cutting off. You could imagine left-truncating the label so only its final words are showing.

parlai/core/torch_agent.py
```diff
@@ -351,6 +361,11 @@ def __init__(self, opt, shared=None):
         self.history = deque(maxlen=self.histsz)
         # truncate == 0 might give funny behavior
         self.truncate = opt['truncate'] if opt['truncate'] >= 0 else None
+        self.left_truncate = (self.truncate if not opt.get('left_truncate')
```
stephenroller: Could do `self.left_truncate = self.get('left_truncate', self.truncate)`.

Maybe we should add a guard about negative numbers.

emilydinan: oh duh, oops

PR #1425, opened 2019-02-05
[Transformer Ranker] Add option to use DataParallel

emilydinan:
Add option to wrap your transformer ranker model with nn.DataParallel. To this end, we needed to move the scoring of candidates outside of the forward funciton.

Smaller changes included:
- `n-positions` is a function of `truncate`
-  add option to freeze embeddings

cc @EricMichaelSmith @samhumeau

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -42,6 +42,7 @@ def _build_encoder(opt, dictionary, embedding=None, padding_idx=None, reduction=
         learn_positional_embeddings=opt.get('learn_positional_embeddings', False),
         embeddings_scale=opt['embeddings_scale'],
         reduction=reduction,
+        n_positions=opt.get('truncate', 1024)
     )
 
 
@@ -67,12 +68,14 @@ class TransformerMemNetModel(nn.Module):
         super().__init__()
         self.opt = opt
         self.pad_idx = dictionary[dictionary.null_token]
-        self.scores_norm = opt['scores_norm']
 
         # set up embeddings
         self.embeddings = _create_embeddings(
             dictionary, opt['embedding_size'], self.pad_idx
         )
+
+        if not opt.get('learn_embeddings'):
+            self.embeddings.weight.requires_grad = False
 
         self.context_encoder = _build_encoder(
             opt, dictionary, self.embeddings, self.pad_idx
@@ -132,16 +135,6 @@ class TransformerMemNetModel(nn.Module):
 
         return weights, context_h
 
-    def _score(self, output, cands):
-        if cands.dim() == 2:
-            return torch.matmul(output, cands.t())
-        elif cands.dim() == 3:
-            return torch.bmm(output.unsqueeze(1),
-                             cands.transpose(1, 2)).squeeze(1)
-        else:
-            raise RuntimeError('Unexpected candidate dimensions {}'
-                               ''.format(cands.dim()))
-
     def forward(self, xs, mems, cands):
         weights, context_h = self.encode_context_memory(xs, mems)
         cands_h = self.encode_cand(cands)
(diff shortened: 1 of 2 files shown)


parlai/agents/transformer/modules.py
```diff
@@ -42,6 +42,7 @@ def _build_encoder(opt, dictionary, embedding=None, padding_idx=None, reduction=
         learn_positional_embeddings=opt.get('learn_positional_embeddings', False),
         embeddings_scale=opt['embeddings_scale'],
         reduction=reduction,
+        n_positions=opt.get('truncate', 1024)
```
EricMichaelSmith: Hmm - will this fail if token vectors are not truncated and one vector has length > 1024?

EricMichaelSmith: Also - should we un-hardcode this in the decoder (line 324) as well?

parlai/agents/transformer/transformer.py
```diff
@@ -61,20 +65,44 @@ def add_cmdline_args(cls, argparser):
...
         # model specific arguments
         agent.add_argument('--normalize-sent-emb', type='bool', default=False)
-        agent.add_argument('--share-encoders', type='bool', default=False)
+        agent.add_argument('--share-encoders', type='bool', default=True)
         agent.add_argument('--has-memories', type='bool', default=False,
                            help='If true, text contains newline separated memories '
                                 'before the actual text')
         agent.add_argument('--use-memories', type='bool', default=False,
                            help='If true, use the memories to help with predictions')
         agent.add_argument('--scores-norm', choices={'dot', 'sqrt', 'dim'},
```
stephenroller: Let's drop this option while we're at it. Braden and I found dot was best.

parlai/agents/transformer/transformer.py
```diff
@@ -103,12 +131,25 @@ def score_candidates(self, batch, cand_vecs):
...
-        return self.model(
+        context_h, cands_h = self.model(
             xs=batch.text_vec,
             mems=mems,
             cands=cand_vecs,
         )
 
+        scores = self._score(context_h, cands_h)
+
+        if self.opt['scores_norm'] == 'dot':
```
stephenroller: Let's drop this option.

parlai/agents/transformer/transformer.py
stephenroller: Interestingly --embeddings-scale we seemed to get mixed results on though. I've consistently found it is good.

parlai/agents/transformer/transformer.py
```diff
@@ -61,20 +65,44 @@ def add_cmdline_args(cls, argparser):
...
         cls.dictionary_class().add_cmdline_args(argparser)
 
-        super(TransformerRankerAgent, cls).add_cmdline_args(argparser)
         return agent
 
+    def __init__(self, opt, shared=None):
+        super().__init__(opt, shared)
+        self.data_parallel = opt.get('data_parallel') and self.use_cuda
+        if self.data_parallel:
+            self.model = torch.nn.DataParallel(self.model)
```
stephenroller: If you wanted you could

```
from parlai.core.distributed_utils import is_distributed
if is_distributed():
    raise ValueError('Cannot combine --data-parallel and distributed mode')
```

parlai/agents/transformer/transformer.py
```diff
@@ -103,12 +131,25 @@ def score_candidates(self, batch, cand_vecs):
         else:
             mems = None
 
-        return self.model(
+        context_h, cands_h = self.model(
```
samhumeau: Not sure what motivated moving those 2 guys outside self.model() but that's interesting for this project:
https://github.com/facebookresearch/ParlAI/pull/1416

parlai/agents/transformer/modules.py
emilydinan: Token vectors should be truncated if you specify `truncate`. Will update so it's not hardcoded in the decoder either

parlai/agents/transformer/modules.py
```diff
@@ -67,13 +68,15 @@ def __init__(self, opt, dictionary):
...
         self.pad_idx = dictionary[dictionary.null_token]
-        self.scores_norm = opt['scores_norm']
 
         # set up embeddings
         self.embeddings = _create_embeddings(
             dictionary, opt['embedding_size'], self.pad_idx
         )
 
+        if not opt.get('learn_embeddings'):
+            self.embeddings.weight.requires_grad = False
```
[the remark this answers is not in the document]
emilydinan: No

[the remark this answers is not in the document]
emilydinan: Copy embeddings just sets the embeddings, this prevents us from backward-ing through them

parlai/agents/transformer/transformer.py
emilydinan: We needed to do this to calculate the scores in data parallel mode

PR #1429, opened 2019-02-06
[MTurk] Fixing message sends to properly block send queue

JackUrb:
## Summary
We used to send a flurry of messages for important packet types until they were acknowledged. This flurry was unintentional and sometimes overwhelming for slower connections. This PR turns off the faucet.

It also includes a missing function from the data handler to retrieve a conversation's data rather than a specific worker's data from a conversation.

## Implementation
This is a really long-standing bug that has skirted out from reviews since the design of the messaging system. In order to maintain strict message ordering for blocking messages, they would be reinserted into the send queue with the original send time. This meant no other messages would be able to cut in line in the way that nonblocking messages can be (as you just add the ack time to the send time and put it back into the queue). Unfortunately we didn't actually block the thread to wait for any amount of time, so the sending thread would send a blocking message, put it back in the queue, and then almost immediately send it again as it hadn't been acknowledged. We'd successfully prevent sending _other_ messages, but not prevent sending the same message over and over in a packet-based socket assault.

The implementation for the `get_full_conversation_data` function is merely the inverse of `save_world_data`.

### DIFF
--- a/parlai/mturk/core/mturk_data_handler.py
+++ b/parlai/mturk/core/mturk_data_handler.py
@@ -729,6 +729,7 @@ class MTurkDataHandler():
 
     @staticmethod
     def get_conversation_data(task_group_id, conv_id, worker_id, is_sandbox):
+        '''A poorly named function that gets conversation data for a worker'''
         result = {
             'had_data_dir': False,
             'had_run_dir': False,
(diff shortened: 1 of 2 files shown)


parlai/mturk/core/socket_manager.py
```diff
@@ -322,8 +322,11 @@ def _send_packet(self, packet, connection_id, send_time):
         if packet.requires_ack:
             if packet.blocking:
                 # Put the packet right back into its place to prevent sending
-                # other packets
+                # other packets, then block
                 self._safe_put(connection_id, (send_time, packet))
+                t = time.time() + self.ACK_TIME[packet.type]
+                while time.time() < t and packet.status != Packet.STATUS_ACK:
+                    time.sleep(0.2)
```
klshuster: assuming this is an arbitrary amount of time?

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -729,6 +729,7 @@ def get_all_task_pairings_for_worker(self, worker_id, task_group_id=None):
 
     @staticmethod
     def get_conversation_data(task_group_id, conv_id, worker_id, is_sandbox):
+        '''A poorly named function that gets conversation data for a worker'''
```
klshuster: is, e.g., `get_worker_conversation_data` too long of a name? or would changing this require some amount of nontrivial refactoring elsewhere

parlai/mturk/core/mturk_data_handler.py
```diff
@@ -760,5 +761,37 @@ def get_conversation_data(task_group_id, conv_id, worker_id, is_sandbox):
...
+                custom_data = json.load(infile)
+            pickle_file = os.path.join(target_dir_custom, 'data.pickle')
+            if os.path.exists(pickle_file):
+                with open(pickle_file, 'rb') as infile:
+                    custom_data['needs-pickle'] = pickle.load(infile)
+            return_data['custom_data'] = custom_data
+
+        target_dir_workers = os.path.join(target_dir, 'workers')
+        for w_file in os.listdir(target_dir_workers):
+            w_id = w_file.split('.json')[0]
```
klshuster: So there's a `target_dir_workers/w_id` and `target_dir_workers/w_id.json` file? what's the difference?

parlai/mturk/core/mturk_data_handler.py
JackUrb: Nontrivial, will do in a follow-up

parlai/mturk/core/socket_manager.py
JackUrb: Yes! I should use my arbitrary time constant instead.

parlai/mturk/core/mturk_data_handler.py
emilydinan: nit: we use double quotation marks for docstrings elsewhere

emilydinan: also lol at the description 😆

JackUrb: Agh I have to stop making that mistake!

PR #1434, opened 2019-02-08
Protecting the computation of the metrics against None candidates

samhumeau:
I was doing experiments with open subtitles, and during validation step I observed a bug in the metrics computation because some of the text candidates were None (that happens when you consider the other elements of the batch as candidates).

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -182,9 +182,9 @@ class Metrics(object):
             cnts = {k: 0 for k in self.eval_pr}
             cnt = 0
             for c in text_cands:
-                if c is None:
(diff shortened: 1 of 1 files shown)


alexholdenmiller:
hm should we try to remove them from the batch sooner?

samhumeau:
I obtain this result while doing 
```
python examples/train_model.py -pyt opensubtitles -m transformer/ranker -veps 0.00001 -mf my_os_model -dt train:stream -bs 40 -vme 2019 --candidates batch
```

stephenroller:
Fixed by #1458 .

PR #1437, opened 2019-02-08
Batch & Output objects

alexholdenmiller:
it's fully compatible with any code *not* subclassing batch, I think


stephenroller:
Before this merges I'd like to convert one of the model extensions that subclass batch.

PR #1443, opened 2019-02-11
Add an explicit --n-positions argument

stephenroller:
For backwards compatibility with some models trained with the default 1024. Using --n-positions now completely overrides truncate variants.

If left unspecified, still uses truncate or 1024 as safe defaults as needed.

### DIFF
--- a/parlai/agents/transformer/transformer.py
+++ b/parlai/agents/transformer/transformer.py
@@ -26,13 +26,17 @@ def add_common_cmdline_args(argparser):
     argparser.add_argument('-esz', '--embedding-size', type=int, default=300,
                            help='Size of all embedding layers')
     argparser.add_argument('-nl', '--n-layers', type=int, default=2)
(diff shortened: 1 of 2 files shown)


parlai/agents/transformer/transformer.py
```diff
@@ -26,13 +26,17 @@ def add_common_cmdline_args(argparser):
     argparser.add_argument('-esz', '--embedding-size', type=int, default=300,
                            help='Size of all embedding layers')
     argparser.add_argument('-nl', '--n-layers', type=int, default=2)
-    argparser.add_argument('-hid', '--ffn-size', type=int, default=300,
+    argparser.add_argument('-hs', '--ffn-size', type=int, default=300,
```
emilydinan: why did we change this?

stephenroller: Corresponding to args for seq2seq. I could add both. The full option is the same.

PR #1446, opened 2019-02-11
Multitask task weights 

jaseweston:
Can train with different weights per task (does a parlay with a probability based on the given task weights), e.g.:

python examples/display_data.py  -t personachat,squad -dt train --multitask_weights 1,10

would upweight squad 10:1  versus personachat

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -214,6 +214,7 @@ class MultiTaskTeacher(Teacher):
         return self.tasks[self.task_idx].observe(observation)
 
     def act(self):
+        import pdb; pdb.set_trace()
         if self.new_task:
             self.new_task = False
             if self.random:
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -194,7 +194,7 @@ class MultiTaskTeacher(Teacher):
         self.task_choices = range(len(self.tasks))
         weights = self.opt.get('multitask_weights', [1])
         sum = 0
-        for i, w in enumerate(self.tasks):
+        for i in self.task_choices:
             if len(weights) > i:
                 weight = weights[i]
             else:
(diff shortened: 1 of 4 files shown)


parlai/core/worlds.py
```diff
@@ -440,6 +440,19 @@ def __init__(self, opt, agents=None, shared=None, default_world=None):
...
+        # These come from the 'taskweight' param (if it exists)
+        #  in the teacher from each world (assumed to be agent 0).
+        self.cum_task_weights = [1] * len(self.worlds)
+        self.task_choices = range(len(self.worlds))
+        sum = 0
+        for i, w in enumerate(self.worlds):
+            agents = w.get_agents()
+            weight = 1
+            if len(agents) > 0:
+                weight = float(agents[0].opt.get('taskweight', 1))
```
stephenroller: let's use just `opt`, instead of `agents[0].opt`

Looks like you need to add the command line param to train_model.py. Worth noting that @alexholdenmiller recently added the `--type='floats'` param that pre-parses the options into a tuple of floats.

jaseweston: ok, changed agents[0].opt that seems to work..
but it's actually an agent specific parameter though, not a global paramater of the opts (per task) the way i've done it.. but i could do it other ways.. so not really to do with train_model

parlai/core/agents.py
```diff
@@ -189,6 +189,18 @@ def __init__(self, opt, shared=None):
         self.task_idx = -1
         self.new_task = True
         self.random = opt.get('datatype') == 'train'
+        # Make multi-task task probabilities.
+        self.cum_task_weights = [1] * len(self.tasks)
+        self.task_choices = range(len(self.tasks))
+        weights = self.opt.get('multitask_weights', [1])
+        sum = 0
+        for i, w in enumerate(self.tasks):
```
emilydinan: you don't seem to use `w` in this loop so you can probably just iterate through `self.task_choices`

jaseweston: fixed

parlai/core/worlds.py
jaseweston: fixed

PR #1454, opened 2019-02-13
Allow to initialize TRA models using --init-model

samhumeau:


emilydinan:
after discussing offline, we don't want to override opts from `init_model`. the rest of the changes can stay but we'll need to remove that. we are going to put a up a PR soon that loudly prints when the opts specified in `init_model` differ from the ones you are loading, and will require you to specify them on the command line.

samhumeau:
Not sure what was is the state of that, but I updated this PR so that it follows what you were saying above. It doesn't override options but now just show loudly the differences, like so:

```bash

!!! WATCHOUT, THE MODEL SPECIFIED BY--init-model WAS TRAINED WITH SOME OPTIONS THAT DIFFER FROM THE CURRENT ONES!!:
key: batchsize value: 1024 (currently 100)
key: pytorch_teacher_task value: None (currently convai2)
key: model value: None (currently transformer/ranker)
key: model_file value: None (currently test_41)
key: dict_class value: None (currently parlai.core.dict:DictionaryAgent)
key: eval_batchsize value: None (currently 30)
key: num_epochs value: -1 (currently 1.0)
key: dimension value: dim_dataset (currently None)
key: dataset value: opensubtitles_64 (currently None)
key: learningrate value: 0.0001 (currently 5e-05)
key: nus value: [0.7] (currently (0.7,))
key: betas value: [0.9, 0.999] (currently (0.9, 0.999))
key: text_truncate value: None (currently 64)
key: label_truncate value: None (currently 64)
key: candidates value: inline (currently batch)
key: eval_candidates value: None (currently inline)
key: train_predict value: False (currently True)
key: init_model value: None (currently /mypath/model_state_200000000)
key: n_layers value: 4 (currently 2)
key: ffn_size value: 1200 (currently 300)
key: learn_positional_embeddings value: True (currently False)
key: share_encoders value: False (currently True)
key: has_memories value: False (currently None)
key: data_parallel value: False (currently True)
key: dict_file value: /mydictionarypath/dictionary (currently /my_other_dictionarypath/dict)
key: pretrained value: True (currently None)
key: num_iterations value: 500000000.0 (currently None)
key: warmup_period value: 10000000.0 (currently None)
key: initial_lr value: 0.0001 (currently None)
key: final_lr value: 1e-05 (currently None)
key: starttime value: Feb12_14-41 (currently Mar04_02-15)

```

stephenroller:
Superseded by #1512.

PR #1455, opened 2019-02-13
Wizard of Wikipedia generative models

stephenroller:
Generative models from Wizard of Wikipedia paper.

Some issues:
- [x] Generalize how model args were being passed in for TGA.
- [x] Generalize loss computations bit
- [x] Generalize 
- [ ] ~~Bring in hierarchical agent somewhere~~ Wait until we get approval for pretrained ranker release
- [x] Pretrained model in model zoo
- [ ] Integration test of end2end model

Needs #1487 to merge first.

### DIFF
--- a/projects/wizard_of_wikipedia/generator/agents.py
+++ b/projects/wizard_of_wikipedia/generator/agents.py
@@ -1,0 +1,319 @@
+#!/usr/bin/env python
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Agents for handling the generation aspects of Wizard.
+"""
+from itertools import chain
+from functools import lru_cache
+
+import torch as th
+import numpy as np
+
+from parlai.core.utils import padded_tensor, round_sigfigs
+
+from parlai.agents.transformer.transformer import TransformerGeneratorAgent
+
+from .modules import EndToEndModel
+from parlai.tasks.wizard_of_wikipedia.agents import TOKEN_KNOWLEDGE
+
+TOKEN_DIALOG = '__dialog__'
+
+
+DEFAULT_OPTS = {
+    "learningrate": 5e-4,
+    "optimizer": "adam",
+    "lr_scheduler": "invsqrt",
+    "warmup_updates": 5000,
+    "clip_norm": 0.1,
+    "ffn_size": 512,
+    "embedding_size": 256,
+    "n_heads": 2,
+    "dropout": 0.2,
+    "n_layers": 5,
+    "betas": "0.9,0.98",
+    "truncate": 128,
+    "dict_textfields": "text,labels,chosen_topic,checked_sentence,knowledge,title",
+}
+
+
+class _GenericWizardAgent(TransformerGeneratorAgent):
+    @classmethod
+    def add_cmdline_args(cls, argparser):
+        argparser.set_defaults(**DEFAULT_OPTS)
+        super(_GenericWizardAgent, cls).add_cmdline_args(argparser)
+
+    def batchify(self, obs_batch):
+        batch = super().batchify(obs_batch)
+        reordered_observations = [obs_batch[i] for i in batch.valid_indices]
+
+        checked_sentences = []
+        for obs in reordered_observations:
+            checked_sentence = '{} {} {}'.format(
+                obs.get('title', ''), TOKEN_KNOWLEDGE, obs.get('checked_sentence', '')
+            )
+            checked_sentences.append(checked_sentence)
+
+        batch['checked_sentence'] = checked_sentences
+        return batch
+
+
+class TwoStageAgent(_GenericWizardAgent):
+    def __init__(self, opt, shared):
+        super().__init__(opt, shared)
+        if shared is not None:
+            # make sure the dialogue token appears
+            self.dict[TOKEN_DIALOG] = 9999999
+
+    def observe(self, obs):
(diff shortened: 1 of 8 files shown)


projects/wizard_of_wikipedia/generator/tasks.py
```diff
@@ -0,0 +1,225 @@
...
+        if self.only_checked_knowledge:
+            # useful for test time evaluation, where it's only ever trained on true
+            # knowledge
+            a['knowledge'] = (
+                a['title'] + ' ' + TOKEN_KNOWLEDGE + ' ' + a['checked_sentence']
+            )
+        return a
+
+
+class GoldWizardTeacher(WizardTeacher):
```
huihuifan: is it preferred in parlai to have a more very specific teachers, or one general teacher that can take in different parameters like 'only_checked_knowledge' or 'ignorant_dropout'?

projects/wizard_of_wikipedia/generator/agents.py
```diff
@@ -0,0 +1,299 @@
...
+            return list(obs['knowledge_parsed'])
+
+        checked_sentence = '{} {} {}'.format(
+            obs['title'], TOKEN_KNOWLEDGE, obs['checked_sentence']
+        )
+        # grab all the nonempty knowledge
+        obs_know = [k.strip() for k in obs.get('knowledge', '').split('\n')]
+        obs_know = [k for k in obs_know if k]
+
+        # we want the correct knowledge to always be in index 0
```
huihuifan: I'm a little lost on this part. Why do you force it to be first?

stephenroller: IIRC later on I use that fact to make loss computation easier or something.

projects/wizard_of_wikipedia/generator/tasks.py
stephenroller: I’m not sure; CLI parameters are more common. This was just a convenient way for me to do it at the time.

parlai/core/torch_generator_agent.py
```diff
@@ -451,7 +461,9 @@ def train_step(self, batch):
         self.zero_grad()
 
         try:
-            scores, preds, _ = self.model(batch.text_vec, batch.label_vec)
+            scores, preds, *_ = self.model(
+                *self._model_input(batch), ys=batch.label_vec
```
stephenroller: cc @alexholdenmiller

stephenroller:
Seeking reviews

parlai/core/torch_generator_agent.py
```diff
@@ -211,13 +214,12 @@ def forward(self, xs, ys=None, cand_params=None, prev_enc=None, maxlen=None):
...
+        encoder_states = prev_enc if prev_enc is not None else self.encoder(*xs)
 
         if ys is not None:
             # use teacher forcing
             scores, preds = self.decode_forced(encoder_states, ys)
         else:
-            bsz = xs.size(0)
             scores, preds = self.decode_greedy(
                 encoder_states,
                 bsz,
```
klshuster: does this protect against bsz being None?

projects/wizard_of_wikipedia/generator/agents.py
```diff
@@ -0,0 +1,358 @@
...
+            fields += [obs['title']]
+        if 'checked_sentence' in obs:
+            fields += [TOKEN_KNOWLEDGE, obs['checked_sentence']]
+        if obs['text'] != '':
+            fields += [TOKEN_DIALOG, obs['text']]
+        obs['text'] = ' '.join(fields)
+
+        # now vectorize with the extra knowledge. It'll all get stored in the
+        # text_vec operation, etc
+        self.vectorize(
```
klshuster: i guess it may not be *necessary* but is it worth setting `obs = self.vectorize(...)` rather than just calling it here? it's not 100% intuitive that vectorize modifies `obs`

projects/wizard_of_wikipedia/generator/agents.py
```diff
@@ -0,0 +1,358 @@
...
+        )
+        # grab all the nonempty knowledge
+        obs_know = [k.strip() for k in obs.get('knowledge', '').split('\n')]
+        obs_know = [k for k in obs_know if k]
+
+        # we want the correct knowledge to always be in index 0
+        try:
+            i = obs_know.index(checked_sentence)
+        except ValueError:
+            # uh oh, couldn't find the sentence in the knowledge. This happens for
```
klshuster: does this really happen still??

stephenroller: Yes, I think it happens once. That particular sentence didn't have a title associated or something.

parlai/core/torch_generator_agent.py
stephenroller: When is bsz None? greedy_decode needs to be explicitly given the batch size. It uses `batch.text_vec.size(0)`.

klshuster: I more meant that `bsz` is an optional arg here, and may be `None` - is this ensured anywhere that `bsz` is in fact not `None` when `ys` is? E.g. I found a snippet [here](https://github.com/facebookresearch/ParlAI/blob/a2535496ec454822b83dea7525ceaa4e856dfb22/parlai/core/torch_generator_agent.py#L744) that calls the model with potentially `ys` as `None` and `bsz` as `None`

stephenroller: Honestly I think that class is nearly entirely broken... But it's a good point. Let's move this into #1469.

PR #1456, opened 2019-02-14
[MTurk] [Easy] fix accidentally dropped imports

JackUrb:
I missed the dropped module import from #1382 and broke the task flow for react tasks with custom components and context (which for now is just me). This fixes it


pringshia:
It would be nice to add linting to the js codebase eventually to quickly catch cases like these

PR #1457, opened 2019-02-14
Offensive username

bharathc346:
Adding a String segmentation algorithm that segments a username string so that we can check if there are any offensive words in it. Added method str_segment function in OffensiveLanguageDetector of parlai/core/utils.py

Example Usage:
           
dict_agent = DictionaryAgent using Wiki Toxic Comments data
old = OffensiveLanguageDector()

split_str = old.str_segment('fucku2', dict_agent, 20)
split_str is 'fuck u 2'

We can then run old.contains_offensive_language(split_str)
which yields the offensive word 'fuck'


parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+
+        def avoid_long_words(word, N):
+            return 10./(N * 10**len(word))
+
+        # def laplace_smoothing(word, N, V):
+        #     return math.log((dict_agent.freq[word] + 1)/(N + V))
+
+        def prob(word):
+            if word in freqs:
+                return math.log(freqs[word]/N)
```
stephenroller: already you should consider precomputing logN and returning math.log(freqs[word]) - logN.

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+            # Return the product of a sequence of numbers
+            return reduce(operator.add, nums, 1)
+
+        def avoid_long_words(word, N):
+            return 10./(N * 10**len(word))
+
+        # def laplace_smoothing(word, N, V):
+        #     return math.log((dict_agent.freq[word] + 1)/(N + V))
+
+        def prob(word):
```
stephenroller: call it logprob

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+
+        @lru_cache(maxsize=None)
+        def segment(text):
+            # Return a list of words that is the best segmentation of text.
+            if not text:
+                return []
+            candidates = ([first]+segment(rem) for first, rem in splits(text))
+            return max(candidates, key=prob_words)
+
+        def splits(text, L=20):
```
stephenroller: What's the rationale behind 20? Seems pretty high but maybe it's good in practice.

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+            # Returns a list of all possible first and remainder tuples where
+            candidates = []
```
stephenroller: If we turn this into a generator, we can save a bunch of memory I believe.

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
         """Determine if text contains any offensive words in the filter."""
         return self.contains_offensive_language(key)
 
+    def str_segment(self, text, dict_agent):
```
stephenroller: Let's add a docstring.

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+
+        def add(nums):
+            # Return the product of a sequence of numbers
+            return reduce(operator.add, nums, 1)
+
+        def avoid_long_words(word, N):
+            return 10./(N * 10**len(word))
+
+        # def laplace_smoothing(word, N, V):
+        #     return math.log((dict_agent.freq[word] + 1)/(N + V))
```
stephenroller: I suggest we go ahead and turn this on and drop avoid_long_words

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+
+        def prob(word):
+            if word in freqs:
+                return math.log(freqs[word]/N)
+            else:
+                return math.log(avoid_long_words(word, N))
+
+        return " ".join(segment(text))
```
stephenroller: Does it make sense to return a string or a list here? What will be more useful down the line?

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+        def splits(text, L=20):
+            # Returns a list of all possible first and remainder tuples where
+            candidates = []
+            for i in range(min(len(text), L)):
+                candidates.append((text[:i+1], text[i+1:]))
+            return candidates
+
+        def prob_words(words):
+            # Returns probability for a sequence of words
+            return add(prob(w) for w in words)
```
stephenroller: you can just use sum!, and then get rid of your definition of add.

On that note, it looks like your starting value for add is off anyway (should be 0 instead of 1)

emilydinan:
@bharathc346 can you update this PR with a description of what it contains?

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+
+        def add(nums):
+            # Return the product of a sequence of numbers
+            return reduce(operator.add, nums, 1)
+
+        def avoid_long_words(word, N):
```
emilydinan: nit: whitespace around operators (applies elsewhere in this PR as well)

parlai/core/utils.py
emilydinan: why is this commented out?

parlai/core/utils.py
emilydinan: Yes, why 20? this should be an argument of the larger function

parlai/core/utils.py
emilydinan: nit: single quotation marks

parlai/core/utils.py
emilydinan: add a comment explaining what this function does. also maybe change the name from `avoid_long_words` to something more descriptive, as this really is for giving the "probability" of out of vocabulary words (in which we assign smaller probabilities to longer words).

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+                return []
+            candidates = ([first] + segment(rem) for first, rem in splits(text, max_length))
+            return max(candidates, key = prob_words)
+
+        def splits(text, max_length):
+            # Returns a list of all possible first and remainder tuples where
+            return [(text[:i+1], text[i+1:])
+                for i in range(min(len(text), max_length))]
+
+        def prob_words(words):
```
emilydinan: this isn't quite the probability, it's really a transformation of the log probability. maybe name it `score` or something.

parlai/core/utils.py
```diff
@@ -713,6 +715,52 @@ def __contains__(self, key):
...
+    def str_segment(self, text, dict_agent, max_length):
+        '''
+        Function that segments a word without spaces into the most
+        probable phrase with spaces
+
+        :param string text: string to segment
+        :param DictionaryAgent dict_agent: Dictionary we use
+            to look at word frequencies
+        :param int max_length: max_length of string to segment
+
```
stephenroller: A little more:

```
:returns: the segmented string
:rtype: str
```

stephenroller:
Make sure travis passes before we merge though.

parlai/core/utils.py
```diff
@@ -713,6 +713,69 @@ def __contains__(self, key):
...
 
+    def str_segment(self, text, dict_agent, max_length):
+        """
+        Function that segments a word without spaces into the most
+        probable phrase with spaces
+
+        :param string text: string to segment
+        :param DictionaryAgent dict_agent: Dictionary we use
+            to look at word frequencies
+        :param int max_length: max_length of string to segment
```
alexholdenmiller: maybe better:
`max_length: max length of a substring (word) in the string. default (None) uses the length of the string.

parlai/core/utils.py
```diff
@@ -713,6 +713,69 @@ def __contains__(self, key):
         """Determine if text contains any offensive words in the filter."""
         return self.contains_offensive_language(key)
 
+    def str_segment(self, text, dict_agent, max_length):
```
alexholdenmiller: should we add `k=1`... and change return type to list? (in case you want to try a few segmentations)

PR #1460, opened 2019-02-14
[MTurk] Fixing the blocking bug that caused SocketManager to get stuck

JackUrb:
Turns out the demons mentioned in #1459 were my own, caused not by #1445, but by #1429. It seems that logging had synchronized packet states in a way that reduced the rate of occurrence for a disconnect bug that would lead all queues to stop sending messages other than un-acked commands to clients that were already dead. It was *I* who caused SocketManager to stop doing it's main duty and spend all of its existence on trying to wake the undead. 

The bug hit on the run that I mentioned in #1459, and thanks to logs being already on I saw what started happening.

Now rather than blocking on what is actually the main socket thread for ack (as I had overlooked my previous architecture change in #1213), I simply mark connections as having a potentially blocking packet, and then skip that connection in the queue if I'm still waiting for an ack.

### DIFF
--- a/parlai/mturk/core/mturk_manager.py
+++ b/parlai/mturk/core/mturk_manager.py
@@ -205,7 +205,8 @@ class MTurkManager():
 
     def _init_logging_config(self):
         """Initialize logging settings from the opt"""
-        if self.use_db and not self.opt['is_debug']:
+        # TODO remove the false flag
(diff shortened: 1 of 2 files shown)


pringshia:
Nice work @JackUrb!

PR #1462, opened 2019-02-15
[MTurk] Adding scaling to HIT_MULT

JackUrb:
# Motivation
Back in the day, we launched all of the HITs at once and created extras to ensure workers didn't take too many all at once. Nowadays, hits are released gradually based on `max_connections`, so the initial reasons for holding extra HITs no longer stands. 1.5x is thus an unrealistic amount to be suggesting that a job will cost

# Implementation
This expands the `HIT_MULT` into 4 buckets that better reflect the upper bound of costs based on experienced disconnect rates as the number of HITs scale up.

# Testing
Launched sandbox jobs in each bucket, asserted the requested $ amount was correct.

### DIFF
--- a/parlai/mturk/core/mturk_manager.py
+++ b/parlai/mturk/core/mturk_manager.py
@@ -26,8 +26,21 @@ import parlai.mturk.core.shared_utils as shared_utils
 # Timeout before cancelling a world start
 WORLD_START_TIMEOUT = 11
 
-# Multiplier to apply when creating hits to ensure worker availibility
-HIT_MULT = 1.5
+# Multiplier to apply when creating hits to ensure worker availibility. As the
+# number of HITs increases, this decreases
+HIT_MULT_SCALE = [
+    # At more than 1000 HITS, most workers will become 'regulars', and we can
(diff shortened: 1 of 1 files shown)


parlai/mturk/core/mturk_manager.py
```diff
@@ -26,8 +26,21 @@
 # Timeout before cancelling a world start
 WORLD_START_TIMEOUT = 11
 
-# Multiplier to apply when creating hits to ensure worker availibility
-HIT_MULT = 1.5
+# Multiplier to apply when creating hits to ensure worker availibility. As the
+# number of HITs increases, this decreases
+HIT_MULT_SCALE = [
```
klshuster: Are these arbitrary multipliers/cutoffs? They look good to me, just curious

JackUrb: Yeah they're mostly derived from a rule of thumb based on claim rates and initial disconnect rates. More than entirely arbitrary, but way less than scientifically concrete.

klshuster: ok good to know it's grounded on some empirical evidence

PR #1464, opened 2019-02-15
[MTurk] Populating the review flow with current worker stats

JackUrb:
# Motivation
When reviewing workers, knowing the previous acceptance and rejection statistics for a worker is helpful for being able to more quickly assess their work. 

# Implementation
From the webapp server perspective, when querying runs we now keep a list of all of the workers who have had a task in that run. We then query worker stats for each of those worker stats and add it to what we return. On the frontend we pull that data and use it to populate the accepted and rejected fields of the `worker_stats` object.


pringshia:
lgtm

PR #1466, opened 2019-02-19
with no_grad in evaluation step.

stephenroller:
Agents generally call `self.model.eval()` in their `eval_step`, so we're accurately computing dropout etc. However, without explicitly disabling the autograd infrastructure, we're wasting memory and compute. This is particularly noticeable when running beam search.

It makes sense to just have a torch nograd around the call to eval step. This will disable it, by default for everyone and probably prevent some headaches later. Users can manually reenable grads using

```python
with torch.enable_grad():
    pass
```

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1205,7 +1205,10 @@ class TorchAgent(Agent):
         if is_training:
             output = self.train_step(batch)
         else:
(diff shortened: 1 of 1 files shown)


stephenroller:
Actually seems to shave ~30s off of travis, but that could be noise.

PR #1471, opened 2019-02-20
Support --reset-optimizer in fairseq.

stephenroller:
Minor fairseq bugfix, where you would crash if you changed optimizers. Just need to pass along a CLI parameter.


stephenroller:
Travis is just a glitch.

PR #1472, opened 2019-02-20
FP16 support

stephenroller:
Just a bit of messing around to see what kind of improvements we would get.

Test cmd:
```
python examples/train_model.py -t convai2 -m transformer/generator \
    -df /tmp/fp16test.dict --dict-lower true \
    -esz 256 -hid 512 -bs 256 -emb fasttext -nl 4 -tr 128 -lr 0.01 -eps 2 -ltim 1 \
    --fp16 true
```

Generative Transformer speedup:
```
fp32 1 gpu: 194.4s
fp32 2 gpu: 102.7s
fp16 1 gpu: 147.1s
fp16 2 gpu: 74.7s
```

Ranker models saw similar speedups:
```
fp32 1 gpu: 160.9s
fp16 1 gpu: 115.7s
```


stephenroller:
(Not quite pushed):

`-m bert_ranker/bi_encoder_ranker -t convai2 -eps 0.5 -cands batch -bs 16`:

FP32 1 GPU: 1660s
FP16 1 GPU: 1216s
FP16 1 GPU:  (with -bs 32 because it fits now): 791s

cc @samhumeau @jaseweston @klshuster

stephenroller:
`-m bert_ranker/cross_encoder_ranker -t convai2 -bs 2 -ltim 10 -eps 0.01 -cands inline` (on a volta32):

FP32 1 GPU: 351s
FP16 1 GPU: 183s (same -bs)
FP16 1 GPU: 141s (now that -bs 4 doesn't OOM)

parlai/core/torch_agent.py
```diff
@@ -1077,8 +1111,12 @@ def vectorize(self, obs, history, add_start=True, add_end=True,
             'cands_vec' fields added.
         """
         self._set_text_vec(obs, history, text_truncate)
-        self._set_label_vec(obs, add_start, add_end, label_truncate)
-        self._set_label_cands_vec(obs, add_start, add_end, label_truncate)
+        self._set_label_vec(
+            obs, add_start=add_start, add_end=add_end, truncate=label_truncate
```
stephenroller: Actually let's just separate this

parlai/core/torch_generator_agent.py
```diff
@@ -433,7 +440,10 @@ def _init_cuda_buffer(self, batchsize, maxlen, force=False):
         if self.use_cuda and (force or not hasattr(self, 'buffer_initialized')):
             try:
                 loss = self.compute_loss(self._dummy_batch(batchsize, maxlen))
-                loss.backward()
+                if self.fp16:
```
stephenroller: @emilydinan what's your opinion? Is it worth building a `def backward(self, loss)` into TorchAgent so that this is abstracted?

parlai/core/torch_generator_agent.py
```diff
@@ -528,7 +538,8 @@ def compute_loss(self, batch, return_output=False):
         model_output = self.model(*self._model_input(batch), ys=batch.label_vec)
         scores, preds, *_ = model_output
         score_view = scores.view(-1, scores.size(-1))
-        loss = self.criterion(score_view, batch.label_vec.view(-1))
+        label_flat = batch.label_vec.view(-1)
+        loss = self.criterion(score_view, label_flat)
```
stephenroller: can undo

parlai/core/torch_ranker_agent.py
```diff
@@ -91,7 +93,7 @@ def __init__(self, opt, shared=None):
             self.set_fixed_candidates(shared)
             self.set_vocab_candidates(shared)
 
-        self.rank_loss = nn.CrossEntropyLoss(reduce=True, size_average=False)
+        self.rank_loss = nn.CrossEntropyLoss(reduce=True, size_average=True)
```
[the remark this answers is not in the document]
stephenroller: Undo this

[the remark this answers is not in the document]
stephenroller: This is the most concerning, because without the size average, the (1) learning rate depends on the batch size and (2) we're more likely to get overflows.

This would explain why the learning rates I see in transformer rankers is always an order of magnitude lower than ones for transformer generators.

stephenroller:
Ready. Remember that it requires you use a volta GPU, so your devfair will actually have a speed DECREASE with --fp16 true.

parlai/core/torch_generator_agent.py
klshuster: given that this shows up a few times, it might be worth it

parlai/core/utils.py
```diff
@@ -24,6 +24,7 @@
 
 """Near infinity, useful as a large penalty for scoring when inf is bad."""
 NEAR_INF = 1e20
+NEAR_INF_FP16 = 65504
```
klshuster: for some reason I find it funny that `NEAR_INF` for FP16 is 65k

stephenroller: it's terrifying, no?

parlai/agents/bert_ranker/helpers.py
```diff
@@ -94,12 +94,12 @@ def forward(self, token_ids, segment_ids, attention_mask):
...
         layer_of_interest = output_bert[self.layer_pulled]
+        dtype = next(self.parameters()).dtype
         if self.add_transformer_layer:
             # Follow up by yet another transformer layer
             extended_attention_mask = attention_mask.unsqueeze(1).unsqueeze(2)
-            extended_attention_mask = extended_attention_mask.to(
-                dtype=next(self.parameters()).dtype)  # fp16 compatibility
-            extended_attention_mask = (1.0 - extended_attention_mask) * -10000.0
+            extended_attention_mask = extended_attention_mask.to(dtype)
+            extended_attention_mask = (1.0 - extended_attention_mask) * neginf(dtype)
```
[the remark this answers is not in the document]
samhumeau: lgtm

parlai/agents/bert_ranker/helpers.py
```diff
@@ -116,7 +116,7 @@ def forward(self, token_ids, segment_ids, attention_mask):
         elif self.aggregation == "max":
             #  consider the max of all the output except CLS
             outputs_of_interest = embedding_layer[:, 1:, :]
-            mask = (attention_mask[:, 1:].float().unsqueeze(2) - 1) * 10000
+            mask = (attention_mask[:, 1:].float().unsqueeze(2) - 1) * neginf(dtype)
```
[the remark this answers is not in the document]
samhumeau: I do not confirm. That line might use a comment that is true: here the attention mask is 1 if present and 0 if not. Your modification make it +infinity when the token is not present

parlai/core/torch_ranker_agent.py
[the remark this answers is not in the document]
samhumeau: I completely approve moving to average here (I did it for bert rankers for instance). But it means lr has to change?

parlai/core/utils.py
samhumeau: Why do we use 65504 instead of 65535?

stephenroller: Because it's not an uint16, it's a float16. That the limits are similar is a coincidence:

```
$ python -c 'import torch; print(torch.HalfTensor([65535]))'
tensor([inf], dtype=torch.float16)
$ python -c 'import torch; print(torch.HalfTensor([65504]))'
tensor([65504.], dtype=torch.float16)
$ python -c 'import torch; print(torch.HalfTensor([65505]))'
tensor([65504.], dtype=torch.float16)
```

parlai/agents/bert_ranker/helpers.py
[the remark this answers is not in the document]
stephenroller: ahh so `attention_mask - 1` is either 0 or -1, which is then multiplied by 10000, to produce a value that's either -10000 or 0.

So this could be fixed with `~attention_mask * neginf`, right?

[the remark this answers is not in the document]
stephenroller: Is this better now @samhumeau ?

parlai/agents/bert_ranker/cross_encoder_ranker.py
```diff
@@ -71,8 +72,8 @@ def score_candidates(self, batch, cand_vecs, cand_encs=None):
         segments_cands = tokens_cands * 0 + 1
         all_tokens = torch.cat([tokens_context, tokens_cands], 1)
         all_segments = torch.cat([segments_context, segments_cands], 1)
-        all_mask = (all_tokens != self.NULL_IDX).long()
-        all_tokens *= all_mask
+        all_mask = (all_tokens != self.NULL_IDX)
+        all_tokens *= all_mask.long()
```
samhumeau: beautiful

parlai/core/torch_generator_agent.py
emilydinan: Sorry I did not see this comment until now -- yes seems fine to me.

parlai/core/torch_agent.py
```diff
@@ -630,6 +647,18 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
 
         optim_class = self.optim_opts()[opt['optimizer']]
         self.optimizer = optim_class(params, **kwargs)
+        if self.fp16:
+            try:
```
emilydinan: since this code is repeated should we factor it into a helper method?

parlai/core/torch_agent.py
```diff
@@ -1426,6 +1465,17 @@ def eval_step(self, batch):
             'Abstract class: user must implement eval_step'
         )
 
+    def backward(self, loss):
```
emilydinan: 👍

parlai/core/torch_agent.py
stephenroller: very good idea. I'm much happier with the new verison

parlai/core/torch_ranker_agent.py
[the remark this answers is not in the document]
stephenroller: Yes, it does. @emilydinan  and @klshuster might have concerns on this, and I think it's a larger discussion.

stephenroller:
Alright, I'm pretty pleased with this. Looking to see some approves now :)

samhumeau:
Tested on crossencoder on convai, I observe speed that is consistent with what's reported above, with up to 2.5x faster.

Screen Shot 2019-03-27 at 6 50 18 PM

In terms of results, it seems to be pretty much on par with float32 too. (evaluated also on convai2 using crossencoder.)

Thanks a lot Stephen, I think this will have a big impact in the next few months.

PR #1473, opened 2019-02-20
starspace: small fix

jaseweston:


parlai/agents/starspace/starspace.py
```diff
@@ -213,7 +213,11 @@ def override_opt(self, new_opt):
 
     def parse(self, text):
         """Convert string to token indices."""
-        return self.dict.txt2vec(text)
+        vec = self.dict.txt2vec(text)
+        if vec == []:
+            vec = [self.dict["_NULL"]]
```
emilydinan: did you mean `self.dict[self.dict.null_token]`? or is this a different special token?

jaseweston: i guess that's the same thing?

stephenroller: that's not the null token we typically use (i.e. by default), and it's user configurable so Emily's suggestion is better.

PR #1477, opened 2019-02-21
Minor changes, now returning top k segmentations of word

bharathc346:
Now returning a list of lists. Allows us to get the top k segmentations of the word

### DIFF
--- a/parlai/core/utils.py
+++ b/parlai/core/utils.py
@@ -793,7 +793,7 @@ class OffensiveLanguageDetector(object):
             count_w = freqs.get(word, 0)
             return math.log(count_w + 1) - logNV
 
-        segmented = segment(text)
+        segment(text)
         res = []
         for i in range(0, k):
             res.append(heapq._heappop_max(max_heap)[1])
--- a/parlai/core/utils.py
+++ b/parlai/core/utils.py
@@ -12,6 +12,8 @@ import os
 import random
 import time
 import warnings
+import heapq
+
 # some of the utility methods are helpful for Torch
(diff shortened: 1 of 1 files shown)


parlai/core/utils.py
```diff
@@ -745,17 +750,31 @@ def str_segment(self, text, dict_agent, max_length):
         V = len(freqs)
 
         logNV = math.log(N + V)
+        max_heap = []
+
+        if not max_length:
+            max_length = len(text)
 
-        @lru_cache(maxsize=None)
+        @lru_cache(maxsize=16)
```
alexholdenmiller: ha you can make this like `2**20` that still would only be O(MB)

parlai/core/utils.py
```diff
@@ -745,17 +750,31 @@ def str_segment(self, text, dict_agent, max_length):
...
-            )
-            return max(candidates, key=score)
+            ]
+
+            nonlocal max_heap
+            max_heap = []
+
+            for c in candidates:
+                cand_score = (score(c), c)  # tuple of (score, candidate)
+                max_heap.append(cand_score)
```
alexholdenmiller: if we have any reason to believe the memory usage here might get heavy (large number of candidates), then it would be better to use the priority queue from the ir baseline class as it never stores more than k elements at once (instead of strong them all and then popping the top k) but I bet that won't come up here, ha

bharathc346: Yeah it's weird, I learned today that there is no fixed sized heap in python

PR #1480, opened 2019-02-25
[Messenger] Get rid of resource warnings

klshuster:
Currently, when uploading attachments, there are a ton of warnings printed about unclosed resources; this makes sure the files are opened and closed safely to remove these warnings.

### DIFF
--- a/parlai/messenger/core/message_sender.py
+++ b/parlai/messenger/core/message_sender.py
@@ -315,21 +315,20 @@ class MessageSender():
                     }
                 }
             }
-            f = open(payload['filename'], 'rb')
-            filedata = {
-                "filedata": (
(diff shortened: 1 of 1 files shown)


parlai/messenger/core/message_sender.py
```diff
@@ -315,10 +315,11 @@ def upload_fb_attachment(self, payload):
                     }
                 }
             }
+            f = open(payload['filename'], 'rb')
```
JackUrb: Prefer `with open(payload['filename'], 'rb') as f:` syntax, rather than manually opening and closing

klshuster: Yeah totally right, changed to that

PR #1481, opened 2019-02-26
[BERT] History bug fix

emilydinan:
Introduction of the History class caused some bugs in BERT, this is a fix.


samhumeau:
I approve it (after the fact) because it seems cleaner. 
However, it shouldn't change anything because _set_text_vec() modifies `obs` in place. Or is there something I do not see?

PR #1485, opened 2019-02-27
Refactor is_valid

emilydinan:
Refactor is valid to be a method belonging to the class so we can more easily override this behavior.

In Torch Ranker Agent, add an option to skip examples for which the set of label candidates do not contain the labels. This defaults to `False`, and prints a better warning when this happens.

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -47,6 +47,11 @@ class TorchRankerAgent(TorchAgent):
             help='Get predictions and calculate mean rank during the train '
                  'step. Turning this on may slow down training.'
         )
+        agent.add_argument(
+            '--ignore-bad-candidates', type='bool', default=False,
+            help='Ignore examples for which the label is not present in the '
+                 'label candidates. Default behavior results in RuntimeError. '
+        )
 
     def __init__(self, opt, shared=None):
         # Must call _get_model_file() first so that paths are updated if necessary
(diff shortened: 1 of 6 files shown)


parlai/core/torch_ranker_agent.py
```diff
@@ -137,6 +142,28 @@ def get_train_preds(self, scores, label_inds, cands, cand_vecs):
...
+        if 'labels_vec' in obs and 'label_candidates_vecs' in obs:
+            cand_vecs = obs['label_candidates_vecs']
+            label_vec = obs['labels_vec']
+            matches = [x for x in cand_vecs if torch.equal(x, label_vec)]
+            if len(matches) == 0:
+                warn_once(
+                    'At least one example has a set of label candidates that '
+                    'does not contain the label.'
+                )
+                good_ex = False
```
stephenroller: I vaguely prefer `return False` here and then `return True` at the bottom.

parlai/core/torch_ranker_agent.py
```diff
@@ -137,6 +142,28 @@ def get_train_preds(self, scores, label_inds, cands, cand_vecs):
             preds = [cands[i][ordering[0]] for i, ordering in enumerate(ranks)]
         return Output(preds)
 
+    def is_valid(self, obs):
+        """Override from TorchAgent."""
+        if not self.opt.get('ignore_bad_candidates', False):
+            return super().is_valid(obs)
+
+        good_ex = 'text_vec' in obs or 'image' in obs
```
jaseweston: i think we should also error if not good_ex, if not on robust mode ?

jaseweston: and i think  this robustness should be for all torch agents, not just rankers

emilydinan: i put it in TRA because if you're not ranking label candidates then this doesn't really matter

PR #1487, opened 2019-02-27
Add regular dropout (like fairseq).

stephenroller:
Equivalent to `--dropout` in fairseq. This dropout apparently works a bit better for very large datasets.


stephenroller:
@samhumeau I believe you requested embedding dropout at some point

samhumeau:
Do we need to be that precise? Can we just have relu_dropout == dropout here?

stephenroller:
I'd prefer to keep the consistency with fairseq and tensor2tensor. @myleott also says he may sometimes tunes all 3 separately when learning on small datasets.

myleott:
Also, in the original AIAYN paper relu_dropout and attn_dropout were zero, but dropout was non-zero :)

PR #1488, opened 2019-02-27
Flag for adding newline to dict.

stephenroller:
This is something that bit @samhumeau and Marie-Anne before -- the dict agent is built using blind versions of the data, not the version processed by the agent. This means for tasks that don't naturally have newlines in them, the history delimiter will be an UNK token.

I'm not convinced this is the right solution, but just using this PR to raise the issue. I used this as a temporary hack to build one dictionary.

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -375,10 +375,26 @@ def load_agent_module(opt):
     model_file = opt['model_file']
     optfile = model_file + '.opt'
     if os.path.isfile(optfile):
+        # Note: new_opt is a bit of a misnomer: opt is what comes from the CLI args,
+        # and new_opt is the options on disk
         new_opt = _load_opt_file(optfile)
         if 'batchindex' in new_opt:
             # This saved variable can cause trouble if we switch to BS=1 at test time
             del new_opt['batchindex']
+
+        # we need to potentially upgrade any old options we might have
+        assert 'model' in new_opt, "Old .opt file doesn't have a model!?"
+        if opt.get('model'):
+            # if the model was specified in the CLI options, use that
+            model_class = get_agent_module(opt['model'])
+            if hasattr(model_class, 'upgrade_saved_opt'):
+                new_opt = model_class.upgrade_saved_opt(new_opt)
+        else:
+            # else, use the model that was specified by the model file on disk
+            model_class = get_agent_module(new_opt['model'])
+            if hasattr(model_class, 'upgrade_saved_opt'):
+                new_opt = model_class.upgrade_saved_opt(new_opt)
+
         # only override opts specified in 'override' dict
         if opt.get('override'):
             for k, v in opt['override'].items():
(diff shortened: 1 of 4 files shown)


samhumeau:
Hum, isn't the default concatenation made with "\n" in torch agent though? (instead of __newline__)?

stephenroller:
The dictionary converts newlines to this word before processing

samhumeau:
Why do we default this to false, can we add it as default field ?

stephenroller:
I think the right way is to put this logic in TA around the —delimeter argument

stephenroller:
There ended up being some messy logic that I tried to abstract in a good way:

We have old models with bad behavior, but we don't want to break them (don't insert this extra token into the dictionary). However, we have some option that *really* should be used on all future models (please do insert this token). As such, we need a "migration" tool.

parlai/core/agents.py
```diff
@@ -375,21 +375,39 @@ def load_agent_module(opt):
     model_file = opt['model_file']
     optfile = model_file + '.opt'
     if os.path.isfile(optfile):
+        # Note: new_opt is a bit of a misnomer: opt is what comes from the CLI args,
```
klshuster: might there be a better name for it then? e.g., `saved_opt`?

parlai/core/agents.py
```diff
@@ -375,21 +375,39 @@ def load_agent_module(opt):
...
+        # and new_opt is the options on disk
```
emilydinan: "options on disk" is a bit confusing, maybe "options loaded from the opt file associated with the model file that is being loaded" ... that might be more confusing

parlai/core/agents.py
```diff
@@ -375,21 +375,39 @@ def load_agent_module(opt):
...
         if 'batchindex' in new_opt:
             # This saved variable can cause trouble if we switch to BS=1 at test time
             del new_opt['batchindex']
+
+        # we need to potentially upgrade any old options we might have
+        assert 'model' in new_opt, "Old .opt file doesn't have a model!?"
+        if opt.get('model'):
+            # if the model was specified in the CLI options, use that
+            model_class = get_agent_module(opt['model'])
+            if hasattr(model_class, 'upgrade_saved_opt'):
```
emilydinan: you can move these two lines outside of the if/else statement since they are repeated in each case

parlai/core/torch_agent.py
```diff
@@ -500,6 +531,8 @@ def __init__(self, opt, shared=None):
             if opt.get('person_tokens'):
                 self.dict[self.P1_TOKEN] = 999999999
                 self.dict[self.P2_TOKEN] = 999999998
+            if self.opt.get('add_delimiter_token') and self.opt.get('delimiter'):
+                self.dict[self.opt['delimiter']] = 999999997
```
emilydinan: what if the delimiter is just a space?

parlai/core/torch_agent.py
```diff
@@ -489,6 +493,33 @@ def add_cmdline_args(cls, argparser):
...
+        Subclasses may *extend* this behavior, but should always call the
+        super-class.
+
+        :param old_opt:
+            The stored options, as loaded directly from disk.
+        :return:
+            The upgraded options. (May be inplace)
+        """
+
+        if 'add_delimiter_token' not in old_opt:
```
emilydinan: this is a bit confusing since it only changes the flag 'add_delimiter_token' with no options to change anything else. is the intended behavior that models would override this?

parlai/core/agents.py
JackUrb: Agreeing with both @emilydinan and @klshuster here, might as well name this something more clearly referring to it being the saved opts associated with the model file. (`saved_opt` being a good choice)

stephenroller:
This goes well to be hand in hand with #1650, but there's still some nuance.

stephenroller:
This isn't happening.

PR #1491, opened 2019-02-28
[BERT] dictionary bug fix

emilydinan:
BERT was not setting the start/end/null token indices properly, causing some things to be tokenized incorrectly (namely, fixed candidates).

### DIFF
--- a/parlai/agents/bert_ranker/bert_dictionary.py
+++ b/parlai/agents/bert_ranker/bert_dictionary.py
@@ -34,9 +34,14 @@ class BertDictionaryAgent(DictionaryAgent):
         self.end_idx = self.tokenizer.convert_tokens_to_ids(['[SEP]'])[
             0]  # should be 102
         self.pad_idx = self.tokenizer.convert_tokens_to_ids(['[PAD]'])[0]  # should be 0
+        # set tok2ind for special tokens
         self.tok2ind[self.start_token] = self.start_idx
         self.tok2ind[self.end_token] = self.end_idx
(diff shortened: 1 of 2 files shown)


parlai/agents/bert_ranker/bert_dictionary.py
```diff
@@ -34,9 +34,9 @@ def __init__(self, opt):
         self.end_idx = self.tokenizer.convert_tokens_to_ids(['[SEP]'])[
             0]  # should be 102
         self.pad_idx = self.tokenizer.convert_tokens_to_ids(['[PAD]'])[0]  # should be 0
-        self[self.start_token] = self.start_idx
-        self[self.end_token] = self.end_idx
-        self[self.null_token] = self.pad_idx
+        self.tok2ind[self.start_token] = self.start_idx
+        self.tok2ind[self.end_token] = self.end_idx
+        self.tok2ind[self.null_token] = self.pad_idx
```
stephenroller: I think you might also need to change self.ind2tok, no?

emilydinan: technically no, since we only ever make calls to "vec2txt" which invokes the bert tokenizer and not the actual dictionary

emilydinan: this fix just allows TorchAgent to set `self.START_IDX` etc properly

stephenroller: yeah but... for completeness at least?

stephenroller:
Travis is having outages. This looks good to merge

PR #1494, opened 2019-03-01
[TA] add 'none' option to use_replies

emilydinan:
- Add option to only track one side of the conversation in the history object

(Also fixed some parentheses to make them consistent 😅 )


stephenroller:
The Parentheses hero we need, but not the one we deserve.

PR #1497, opened 2019-03-04
New candidate mode in TRA: batch-all-cands

samhumeau:
This introduces a new candidate mode in TRA ("batch-all-cands") in addition to "fixed", "batch", "inline" and "vocab".

This I think solves a few limitations: 
 - this way a teacher can introduce hard negatives (elements of history for instance, but maybe other things) in the candidates and we don't ignore them in "batch" mode.
 - It deduplicates candidates, this way we can have duplicated labels across the batch.


parlai/core/torch_ranker_agent.py
```diff
@@ -242,6 +241,13 @@ def _build_candidates(self, batch, source, mode):
                 examples in a batch. This option may be undesirable if it is possible
                 for duplicate labels to occur in a batch, since the second instance of
                 the correct label will be treated as a negative.
+            * batch-all-cands: the set of all candidates in this batch
+                Use all candidates in the batch as candidate set.
+                Note 1: This can result in a very large number of
+                        of candidates.
+                Note 2: In this case we will deduplicate candidates.
+                Note 3: just like with 'batch' the candidate set is identicall
```

parlai/core/torch_ranker_agent.py
```diff
@@ -72,15 +72,14 @@ def __init__(self, opt, shared=None):
                 states = {}
 
         self.rank_loss = nn.CrossEntropyLoss(reduce=True, size_average=False)
+        if self.use_cuda:
```
samhumeau: I might want to put that in another PR. I think model.cuda() should be called before  self.set_fixed_candidates(shared) since this function is using self.model

parlai/core/torch_ranker_agent.py
```diff
@@ -280,6 +286,34 @@ def _build_candidates(self, batch, source, mode):
             cand_vecs = label_vecs
             label_inds = label_vecs.new_tensor(range(batchsize))
 
+        elif source == 'batch-all-cands':
```
stephenroller: I wonder if we can start to refactor this method soon.

parlai/core/torch_ranker_agent.py
stephenroller: I was gonna ask about this... seems okay.

parlai/core/torch_ranker_agent.py
```diff
@@ -280,6 +287,34 @@ def _build_candidates(self, batch, source, mode):
...
+            warn_once(
+                '[ Executing {} mode with all candidates provided in the batch ]'
+                ''.format(mode)
+            )
+            if batch.candidate_vecs is None:
+                raise ValueError(
+                    "If using candidate source 'batch-all-cands', then batch."
+                    "candidate_vecs cannot be None. If your task does not have "
+                    "inline candidates, consider using one of "
+                    "--{m}={{'batch','fixed','vocab'}}."
```
emilydinan: nit: can't use "batch" if candidate vecs is None

samhumeau: With 'batch' only labels are considered.

samhumeau:
@emilydinan
- using "batch" for training is cool, since it uses the other elements of the batch as negative, the whole thing goes fast and that allows to have hundreds of negatives. But you don't use the negatives from the teacher.
- using inline is cool because you use the candidates of the teacher which could be hard negatives, but on the other hand it's slower.
- batch-all-cands allows both.

jaseweston:
i think ti would be good to also be able to sample from one or the other
some % of the time maybe?

samhumeau:
@emilydinan I thought so too that's why this line was changed
https://github.com/facebookresearch/ParlAI/pull/1497/files#diff-40f9d4952e2b8b2c7bf08f6596c2e8bcR214

Is there something I am missing?

samhumeau:
@jaseweston the percentage idea is in order to avoid facing OOM or this kind of things?

samhumeau:
Apparently the unit tests are timing out and that's why they fail.

PR #1498, opened 2019-03-05
faster ranking at prediction time

samhumeau:
In the case of doing predictions with 100k candidates, this part was a bit slow. This modification makes a single prediction with 100k candidates going from 2.6 sec to < 0.1 sec.

- the main part is making sure "ranks" is on CPU. If on GPU, then what follows causes a lot of transfers between GPU mem and RAM.
- I also thought we could cap the number of prediction by default, we probably don't need a ranked list of 100k candidates as output of each observation.


samhumeau:
Not sure why the tests aren't starting (and fail)

```
# No configuration was found in your project. Please refer to https://circleci.com/docs/2.0/ to get started with your configuration.
```

parlai/core/torch_ranker_agent.py
```diff
@@ -190,6 +194,8 @@ def eval_step(self, batch):
                 rank = (ranks[b] == label_inds[b]).nonzero().item()
                 self.metrics['rank'] += 1 + rank
 
+        ranks = ranks.cpu()
+        max_preds = self.opt["cap_num_predictions"]
```
emilydinan: nit: single quotes

PR #1499, opened 2019-03-05
change default parameters for Bert

samhumeau:
So last night, if I got it right:

1) @jaseweston was very sad when he realized  that by default, bert_ranker only fine-tunes the top layer which leads to sub-optimal results, meaning he might have to re-run a ton of experiments.

2) @jaseweston was very happy when he realized  that he actually used `--type-optimization all_encoder_layers` without remembering it.

To avoid more emotional rollercoaster, I suggest we switch this parameter to the one that led to the best results on convai2.

Also I added a download() part in the bert_dictionary, I think that is missing.

### DIFF
--- a/parlai/agents/bert_ranker/bert_dictionary.py
+++ b/parlai/agents/bert_ranker/bert_dictionary.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 from parlai.core.dict import DictionaryAgent
+from parlai.zoo.bert.build import download
 try:
     from pytorch_pretrained_bert import BertTokenizer
 except ImportError:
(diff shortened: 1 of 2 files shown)


samhumeau:
Not sure why the tests aren't starting (and fail)

```
# No configuration was found in your project. Please refer to https://circleci.com/docs/2.0/ to get started with your configuration.
```

parlai/agents/bert_ranker/bert_dictionary.py
```diff
@@ -22,6 +23,7 @@ class BertDictionaryAgent(DictionaryAgent):
     def __init__(self, opt):
         super().__init__(opt)
         # initialize from voab path
+        download(opt['datapath'])
```
emilydinan: why do you need to download here?

samhumeau: So I think building the dictionary is the very first thing done in a training, which means that if you do not have the data downloaded it fails to load the tokenizer.

PR #1507, opened 2019-03-06
Align both encoder with latest changes

samhumeau:
made sure both encoder still worked as expected

In particular
- got rid of surround() now that vectorize_fixed_candidates() exists. 
- got rid of set_text_vec() 
- encode candidates with torch.no_grad()

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -395,6 +395,9 @@ class TorchRankerAgent(TorchAgent):
             cands = self.fixed_candidates
             cand_vecs = self.fixed_candidate_vecs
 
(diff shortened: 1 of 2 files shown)


parlai/core/torch_ranker_agent.py
```diff
@@ -395,6 +395,9 @@ def _build_candidates(self, batch, source, mode):
             cands = self.fixed_candidates
             cand_vecs = self.fixed_candidate_vecs
 
+            if self.opt.get('encode_candidate_vecs', False):
```
emilydinan: this is not quite right and introduces a bug. cand_vecs are not the model encoded candidates, they are simply the tokenized candidates

PR #1510, opened 2019-03-06
Dict Fixes

klshuster:
Fixes for common dictionary problems...

1. If `model_file` specified in eval model, but dict file does not exist, raise exception and inform user
2. If neither `model_file` nor `dict_file` specified in train_model, raise exception and inform user

Includes 2 new unit tests to test these


tests/test_dict.py
```diff
@@ -59,6 +63,48 @@ def test_basic_parse(self):
         assert vec[0] == num_builtin
         assert vec[1] == num_builtin + 1
 
+    def test_mf_without_df(self):
```
stephenroller: expand this name

PR #1514, opened 2019-03-06
[MTurk] Opening up tests that failed on travis to try on circle

JackUrb:
Certain tests were failing on travis because the socket couldn't be properly opened. Maybe having circle means we won't run into this issue?


JackUrb:
Abandoning - Circle seems to fall prey to the same socket port issues. Will need to reinvestigate in the future, but for now local tests are the only way to ensure MTurk validity 😢

stephenroller:
:'(

PR #1517, opened 2019-03-06
TorchAgent doc reformat

stephenroller:
Reformat TorchAgent docs.

I propose this as our standard format.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -133,11 +133,11 @@ class History(object):
     :param maxlen:
         if `vec_type` is 'deque', this sets the maximum length of that object
 
-    :param p1_token: 
+    :param p1_token:
         token indicating 'person 1'; opt must have 'person_tokens' set to True
         for this to be added
 
-    :param p1_token: 
+    :param p1_token:
         token indicating 'person 2'; opt must have 'person_tokens' set to True
         for this to be added
 
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -175,15 +175,11 @@ class History(object):
         self.reset_on_next_update = False
 
     def parse(self, text):
-        """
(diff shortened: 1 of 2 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -6,11 +6,8 @@
 
 
 """
-**BETA**: This module is in beta. Feedback is most welcome, and the API may change
-underneath you.
-
-Generic Pytorch-based Generator agent. Implements quite a bit of boilerplate,
-including Beam search.
+Generic PyTorch-based Generator agent. Implements quite a bit of boilerplate,
```
klshuster: no longer in beta?

parlai/core/torch_agent.py
```diff
@@ -44,51 +44,41 @@ class Batch(AttrDict):
     with additional fields if they would like, though we recommend calling the
     parent function to set up these fields as a base.
 
-    .. py:attribute:: text_vec
-
+    :param text_vec:
```
klshuster: any reason we still use `:attribute:` below in torch generator but not here?

parlai/core/torch_agent.py
```diff
@@ -130,20 +118,31 @@ def __init__(self, text=None, text_candidates=None, **kwargs):
...
-                         'person_tokens' set to True for this to be added
-    :param p1_token:     token indicating 'person 2'; opt must have
-                         'person_tokens' set to True for this to be added
-    :param dict_agent    DictionaryAgent object for tokenizing the history
+    History handles tracking the dialogue history/state over the course of an
+    episode.
+
+    History may also be used to track the history of any field.
+
+    :param field:
```
klshuster: same comment here with `attribute`

parlai/core/torch_agent.py
stephenroller: I will want to update that, good catch

PR #1524, opened 2019-03-06
Warning with ordered data in PytorchDataTeacher

klshuster:
When using the PytorchDataTeacher, and the combination of args results in ordered data, we print a warning.

### DIFF
--- a/parlai/core/pytorch_data_teacher.py
+++ b/parlai/core/pytorch_data_teacher.py
@@ -9,6 +9,7 @@
 
 """
 from .teachers import FixedDialogTeacher
+from parlai.core.utils import warn_once
 from parlai.scripts.build_pytorch_data import build_data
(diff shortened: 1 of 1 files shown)


parlai/core/pytorch_data_teacher.py
```diff
@@ -586,6 +587,11 @@ def __init__(self, opt, shared=None):
                         ('stream' in self.datatype and not opt.get('shuffle')))
         if self.ordered:
             # force index for ordered, so that we see every example
+            warn_once(
+                '\nNote: You are using PytorchDataTeacher with ordered '
```
stephenroller: I think the extra newline is causing the problem?

klshuster: sadly, it does not

PR #1535, opened 2019-03-07
GPU Tests

stephenroller:
Runs GPU tests on CircleCI.


stephenroller:
It works!!!!

The gpu test that fails is an actual test failure, we need to fix the task (Discussed offline).

PR #1538, opened 2019-03-08
Fix dict files once and for all

klshuster:
Fixes DrQA nightly gpu test. Also, fixes a long-standing issue of moving a dictionary.

Tested via running DrQA test locally.

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -391,9 +391,17 @@ def load_agent_module(opt):
             if k not in new_opt:
                 new_opt[k] = v
         new_opt['model_file'] = model_file
-        if new_opt.get('dict_file') and not os.path.isfile(new_opt['dict_file']):
-            raise RuntimeError('WARNING: Dict file does not exist, check to make '
-                               'sure it is correct: {}'.format(new_opt['dict_file']))
+        if not new_opt.get('dict_file'):
+            new_opt['dict_file'] = model_file + '.dict'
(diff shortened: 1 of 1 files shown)


parlai/core/agents.py
```diff
@@ -391,9 +391,17 @@ def load_agent_module(opt):
             if k not in new_opt:
                 new_opt[k] = v
         new_opt['model_file'] = model_file
-        if new_opt.get('dict_file') and not os.path.isfile(new_opt['dict_file']):
-            raise RuntimeError('WARNING: Dict file does not exist, check to make '
-                               'sure it is correct: {}'.format(new_opt['dict_file']))
+        if not new_opt.get('dict_file'):
+            new_opt['dict_file'] = model_file + '.dict'
+        elif new_opt.get('dict_file') and not os.path.isfile(new_opt['dict_file']):
```
jaseweston: i was doing this the other way round.. if you have a   model.dict then it just uses it, overriding whatever is in the dict_file .. is your way better?

klshuster: Your way probably makes more sense in practice, but this way at least respects the case where someone explicitly specifies a `--dict_file` on the command line (which, in your way, we would ignore)

jaseweston:
yea should be fine

stephenroller:
Land this but let's talk about forceful dicts at the next meeting.

PR #1541, opened 2019-03-11
[MTurk] Add ESlint and prettier

pringshia:
Adds linting/auto-formatting to the the mturk web app and task folders.


stephenroller:
😍 is there anyway this could be integrated into lint_changed.sh?
https://github.com/facebookresearch/ParlAI/blob/1bfe2b1fa1f00b9c17b51d2223096118a0d36c89/tests/lint_changed.sh#L14-L21

You would need to also add dependency installations in the circleci step: https://github.com/facebookresearch/ParlAI/blob/1bfe2b1fa1f00b9c17b51d2223096118a0d36c89/.circleci/config.yml#L103-L111

pringshia:
Cool! Yeah I'll add it to the CircleCI process after nailing down the configuration. Thanks for the code pointers

pringshia:
I'll wait for sign-off from @JackUrb as well when he returns from PTO this week before merging this in

PR #1542, opened 2019-03-11
Adding dynamic forms to MTurk React data collection

samhumeau:
I noticed that most frontend overrides on the MTurk data collection was to add form-like input. This PR adds a way to dynamically create Forms while collecting data on MTurk. 

The form is created within python in `world.py`, and stored in `action["task_data"]["respond_with_form"]` (This has to be None whenever we don't need a form)

```python
action["task_data"] = {}
action["task_data"]["respond_with_form"] = None
if self.turns == 3:
    action['id'] = "System"
    action["text"] = ("Thanks for chatting 😻. Now please answer the "
                 "following questions to finish the hit 😃 !")
    action["task_data"]["respond_with_form"] = [
        {
            "type": "choices",
            "question": "How much did you enjoy talking to this user?",
            "choices": ["Not at all", "A little", "Somewhat", "A lot"]
        },
        {
            "type": "choices",
            "question": "Do you think this user is a bot or a human?",
            "choices": ["Definitely a bot", "Probably a bot", "Probably a human", "Definitely a human"]
        },
        {
            "type": "text",
            "question": "Enter any comment here",
        },
    ]
```

It will appear like so: 
Screen Shot 2019-03-11 at 9 51 59 PM

Notice that the "type" ("choices" or "text") determine whether it's a multiple choices question or free form text.

Whenever ["respond_with_form"] is None, we show the classic text input like so:
Screen Shot 2019-03-11 at 9 51 20 PM

The structured result can be read in "task_data" on server side once the form is filled:
```bash
{
    "text": "How much did you enjoy talking to this user?: A lot\nDo you think this user is a bot or a human?: Definitely a bot\nEnter any comment here: Did you notice this is a form?\n",
    "task_data": {
        "How much did you enjoy talking to this user?": "A lot",
        "Do you think this user is a bot or a human?": "Definitely a bot",
        "Enter any comment here": "Did you notice this is a form?"
    },
    "id": "Worker",
    "message_id": "aec68468-f10f-46fd-93e1-5b7fd4263b5d",
    "episode_done": false,
    "duration": 27045
}
```

### DIFF
--- a/parlai/mturk/core/react_server/dev/components/core_components.jsx
+++ b/parlai/mturk/core/react_server/dev/components/core_components.jsx
@@ -839,19 +839,26 @@ class FormResponse extends React.Component {
 
   tryMessageSend() {
     let form_elements = this.props.task_data["respond_with_form"];
-    let response_data = {};
+    let response_data = [];
     let response_text = "";
+    let all_response_filled = true;
     for (let ind in form_elements){
       let question = form_elements[ind]["question"];
       let response = this.state.responses[ind];
-      response_data[question] = response;
+      if (response == ''){
+        all_response_filled = false;
+      }
+      response_data.push({
+        "question": question,
+        "response": response
+      });
       response_text += question + ": " + response + "\n";
     }
 
-    if (this.state.textval != '' && this.props.active && !this.state.sending) {
+    if (all_response_filled && this.props.active && !this.state.sending) {
       this.setState({sending: true});
       this.props.onMessageSend(
-        response_text, response_data,
+        response_text, {"form_responses": response_data},
         () => this.setState({'sending': false}));
     }
   }
@@ -861,9 +868,11 @@ class FormResponse extends React.Component {
     const listFormElements= form_elements.map((form_elem, index) => {
         let question = form_elem["question"];
         if (form_elem["type"] == "choices"){
-          let choices = [<option></option>].concat(
-            form_elem["choices"].map((option_label) => {
-              return <option>{option_label}</option>
+          let choices = [<option key="empty_option"></option>].concat(
+            form_elem["choices"].map((option_label, index) => {
+              return <option key={"option_" + index.toString()}>
+                        {option_label}
+                      </option>
             }));
           return (<FormGroup>
                     <Col
--- a/parlai/mturk/core/react_server/dev/components/core_components.jsx
+++ b/parlai/mturk/core/react_server/dev/components/core_components.jsx
@@ -11,7 +11,7 @@ import ReactDOM from 'react-dom';
 import {
   FormControl, Button, ButtonGroup, InputGroup, FormGroup,
   MenuItem, DropdownButton, Badge, Popover, Overlay,
-  Nav, NavItem,
+  Nav, NavItem, Col,ControlLabel, Form,
 } from 'react-bootstrap';
 import Slider from 'rc-slider';
 import $ from 'jquery';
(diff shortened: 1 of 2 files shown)


parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -822,11 +822,140 @@ class TextResponse extends React.Component {
...
+    let response_data = {};
+    let response_text = "";
+    for (let ind in form_elements){
+      let question = form_elements[ind]["question"];
+      let response = this.state.responses[ind];
+      response_data[question] = response;
+      response_text += question + ": " + response + "\n";
+    }
+
+    if (this.state.textval != '' && this.props.active && !this.state.sending) {
```
JackUrb: This should probably ensure that the form values in responses (`this.state.responses`) are not empty rather than checking `textval`.

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -822,11 +822,140 @@ class TextResponse extends React.Component {
...
+  render() {
+    let form_elements = this.props.task_data["respond_with_form"];
+    const listFormElements= form_elements.map((form_elem, index) => {
+        let question = form_elem["question"];
+        if (form_elem["type"] == "choices"){
+          let choices = [<option></option>].concat(
+            form_elem["choices"].map((option_label) => {
+              return <option>{option_label}</option>
+            }));
+          return (<FormGroup>
```
JackUrb: React might be complaining about elements in an array not having unique keys. You should add keys to these components based on some identifier. I think this should be something that the user can tag rather than `question`

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -822,11 +822,140 @@ class TextResponse extends React.Component {
...
+
+
+  tryMessageSend() {
+    let form_elements = this.props.task_data["respond_with_form"];
+    let response_data = {};
+    let response_text = "";
+    for (let ind in form_elements){
+      let question = form_elements[ind]["question"];
+      let response = this.state.responses[ind];
+      response_data[question] = response;
```
JackUrb: Rather than keying on the displayed `question` text, it might make more sense for users to be able to provide their own keys when they specify the question in their config. This would also give you a key that you can use for the form element's key.

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -822,11 +822,140 @@ class TextResponse extends React.Component {
...
+               </FormGroup>;
+      }
+    );
+    let submit_button = (
+      <Button
+        className="btn btn-primary"
+        style={{'height': '40px', 'width': '100px', 'fontSize': '16px'}}
+        id="id_send_msg_button"
+        disabled={
+          this.state.textval == '' || !this.props.active || this.state.sending}
```
JackUrb: Same as other `textval` comment

samhumeau:
Per conversation with Jack offline, the new format returned is:
```
{                             
    "text": "How much did you enjoy talking to this user?: Not at all\nDo you think this user is a bot or a human?: Definitely a bot\nEnter any comment here: jiijjlj;\n",
    "task_data": {                                                         
        "form_responses": [                                                                               
            {                                                   
                "question": "How much did you enjoy talking to this user?",                                       
                "response": "Not at all"
            },                                                  
            {                                                                                      
                "question": "Do you think this user is a bot or a human?",                         
                "response": "Definitely a bot"                                                                 
            },                                                   
            {               
                "question": "Enter any comment here", 
                "response": "jiijjlj;"
            }
        ]
    },
    "id": "Worker",
    "message_id": "34456401-6331-4844-86d2-8ee291e834a3",
    "episode_done": false,
    "duration": 8263
}
```

samhumeau:
Other remarks of Jack should be taken care of

ethanjperez:
Thanks for this functionality - just what I needed for my task.

Is it possible to clear the form after a response has been submitted? If I ask for consecutive responses via the form, it auto-fills the previous response as the current answer. So I'm getting feedback from Turkers that they are accidentally clicking "Submit" again and sending an incorrect response here or there.

Another related issue is that it's possible for me to receive responses out of the set of responses I'm expecting; if the previous form used a different set of responses, one of those responses could be submitted for the following question.

I'd appreciate any help on this, and thanks for this feature :)

samhumeau:
ethanjperez, thanks so much for your feedback and very glad that you enjoyed this functionality. 
Issue should be addressed in #1599

jazzminewang:
@samhumeau Thanks for this functionality! I'm trying to use this functionality in my worlds.py, with the example task form data:

Right before sending the, my action object looks like this:
```
{'episode_done': False, 'id': 'System', 'text': 'Thanks for reading 😻. Now please answer the following questions to finish the hit 😃 !', 'task_data': {'respond_with_form': [{'type': 'choices', 'question': 'How much did you enjoy talking to this user?', 'choices': ['Not at all', 'A little', 'Somewhat', 'A lot']}, {'type': 'choices', 'question': 'Do you think this user is a bot or a human?', 'choices': ['Definitely a bot', 'Probably a bot', 'Probably a human', 'Definitely a human']}]}}
```

Then, I send it with this line
```self.mturk_agent.observe(validate(action))```

For some reason, I'm only seeing the system text, and not the form response. 
https://cl.ly/fe48d02d7280

Is this the proper way to invoke this component? Can't find code examples in forks as this is quite new. Thanks again for implementing this!

jazzminewang:
Poked around in Chrome developer tools and it looks like the action object is being received correctly but only the text is being displayed:
https://cl.ly/0fc9877e1289

Found some code usage examples in @ethanjperez's and my syntax is the same. Any suggestions for how best to debug would be appreciated! [https://github.com/facebookresearch/ParlAI/compare/master...ethanjperez:master#diff-686464571006f64b581534a8785fc545R512](url)

JackUrb:
Is it possible you're overriding the `TextResponse` or `ResponsePane` components somewhere in the tree of your `custom.jsx` file? Could be related to #1626 as well if so.

kushalchawla:
I have a question.

I tried using 'task_data' to pass data from back-end to front-end React components. However, the moment I add any dummy variable inside task_data, the 'has_context' attribute also turns on, which shows a context tab in the left pane. I am not able to turn this off. Digging up, I found this code in the socket handler:

-------
if (message.task_data !== undefined) {
      let has_context = false;
      if (!this.props.run_static) {
        // We expect static tasks to recieve task_data and display in
        // the main pane, but dynamic tasks have other options.
        // TODO extract above if block out after separating static and dynamic
        // Socket managers
        for (let key of Object.keys(message.task_data)) {
          if (key !== 'respond_with_form') {
            has_context = true;
          }
        }
      }
-------

I am wondering why having a context window has relation with respond_with_form key. What is the best way to turn off the context navigation tab in the Left Pane, when task_data is being used?

JackUrb:
What happened here is that we had overloaded the content of what we wanted task_data to do at the time. Initially it was only to show context that was relevant to a specific point in time, hence showing a context pane when there was anything in it, but later we also used it to transmit the fields of the form used in `FormResponse`. To maintain the context-related idea, this special casing was added. Ultimately the role of `task_data` needed to be better defined, but such an improvement never made it into PMT.

kushalchawla:
Thanks for your replies.

Got it. For now, I have overwritten the left pane to remove the context tab, regardless of whether has_context is true or not.

PR #1544, opened 2019-03-12
Static Website & deployment

stephenroller:
All we need to replace the current website with a static s3-backed version.


website/templates/error.html
```diff
@@ -0,0 +1,8 @@
+<section class="about wrap">
+    <hr style="margin-top: 0px">
+    <div class="container">
+        <div style="margin-top: 30px">
+            Page not found.
+        </div>
+    </div>!
```
pringshia: Abandoned exclamation mark here

stephenroller: omg what a catch

docs/source/_static/css/parlai_theme.css
```diff
@@ -16,7 +16,7 @@ h1, h2, .rst-content .toctree-wrapper p.caption, h3, h4, h5, h6, legend, p.capti
 }
 
 nav p.caption {
-    color: #fcfcfc;
+    color: #fcfcfc !important;
```
pringshia: I would recommend getting rid of the `!important`. You can do what you're trying to achieve by making the selector here [more specific](https://specificity.keegan.st/), e.g. something like **`.wy-menu.wy-menu-vertical p.caption`** should beat out the other selector adding the blue font color (`.wy-menu-vertical p.caption`)

.circleci/config.yml
```diff
@@ -135,6 +136,27 @@ jobs:
...
           command: python setup.py test -s tests.suites.nightly_cpu -q
 
+  deploy_website:
+    executor: standard_cpu
+    working_directory: ~/ParlAI
+    steps:
+      - checkout
+      - setup
+      - installdeps
+      - installtorchcpu
```
pringshia: Curious - is the `installtorchcpu` job really a dependency for this?

stephenroller: Yeah I thought that was funny too, but the docs need it because they torch_agent imports it.

website/generate.py
```diff
@@ -0,0 +1,129 @@
...
+    content = _read_file(os.path.join(TEMPLATES, 'error.html'))
+    html = wrap_base(content, "Error")
+    _write_file('error.html', html)
+
+
+def make_aboutpage():
+    template = _read_file(os.path.join(TEMPLATES, 'about.html'))
+    readme = _read_file(os.path.join(GIT_ROOT_LEVEL, 'README.md'))
+    readme_html = ghmarkdown(readme)
+    readme_html = readme_html.replace("docs/source/\\", "/docs/")
```
pringshia: I also see some links in the page for `/static/docs/` e.g. I see links for these tutorials `http://parl.ai/static/docs/tutorial_mturk.html` and `http://parl.ai/static/docs/tutorial_messenger.html` currently on the page.

Will these break with this new site? Is it also necessary to replace `/static/docs` with `/docs`?

stephenroller: Oh man they totally will. I'll keep the old URL then.

pringshia: Resolving.

For posterity: we set up an S3 redirect.

stephenroller:
Good call @JackUrb

PR #1404, opened 2019-01-30
Self-feeding chatbot code+data release

pemazare:
Integrating all changes related to the self-feeding chatbot project: https://arxiv.org/abs/1901.05415

`export MODELROOT=/tmp/model222;export PARLAIHOME=$(pwd); python -u examples/train_model.py --max-train-time 0 -tblog true -tbmetrics lr,dia_acc,dia_loss -vmt dia_loss -vmm min -ltim 5 -vtim 10 -vp 10 -m projects.metadialog.metadialog_agent:MetadialogAgent --has-memories false --use-memories false -cands batch -ecands batch -histsz 2 --embedding-type fasttext_cc --embedding-size 300 --embeddings-scale false --scores-norm dot --dict-maxtokens 250000 --num-epochs 100 --optimizer adamax --eval-batchsize 20 --relu-dropout 0 --attention-dropout 0 --n-heads 2 --lr-scheduler invsqrt -lr 0.00001 -t metadialog:diaexp:train -tbmetrics lr,dia_acc,dia_loss,exp_acc,exp_loss --dia-train train --exp-train train --exp-weight 0.75 -vmt dia_acc -vmm max --warmup-updates 625 --n-layers 1 -lr 0.001 -bs 128 --ffn-size 32 --model-file $MODELROOT/modelmtl`

PR #1545, opened 2019-03-12
[Torch Ranker Agent] Caching candidate encoding

emilydinan:
Change the `score_candidates` API to possibly have `cand_encs` passed in. This occurs in the case when we want to cache the candidate encodings (possibly when using fixed or vocab candidates). The eval_step in TRA will pass these into score_candidates. It is then up to the model to implement `score_candidates`. See the BiEncoderRanker agent or TransformerRanker agent to see how this is implemented.

Also included, bug fix caused by conflicts in #1404.

cc @samhumeau

### DIFF
--- a/parlai/agents/bert_ranker/bi_encoder_ranker.py
+++ b/parlai/agents/bert_ranker/bi_encoder_ranker.py
@@ -69,9 +69,11 @@ class BiEncoderRankerAgent(TorchRankerAgent):
         self.vocab_candidates will contain a [num_cands] list of strings
         self.vocab_candidate_vecs will contain a [num_cands, 1] LongTensor
         """
+        self.opt['encode_candidate_vecs'] = True
         if shared:
             self.vocab_candidates = shared['vocab_candidates']
             self.vocab_candidate_vecs = shared['vocab_candidate_vecs']
+            self.vocab_candidate_encs = shared['vocab_candidate_encs']
         else:
             if 'vocab' in (self.opt['candidates'], self.opt['eval_candidates']):
                 cands = []
@@ -135,7 +137,7 @@ class BiEncoderRankerAgent(TorchRankerAgent):
                                        self.END_IDX)
         return obs
 
-    def score_candidates(self, batch, cand_vecs):
+    def score_candidates(self, batch, cand_vecs, cand_encs=None):
         # Encode contexts first
         token_idx_ctxt, segment_idx_ctxt, mask_ctxt = to_bert_input(
             batch.text_vec, self.NULL_IDX)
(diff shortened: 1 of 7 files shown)


parlai/agents/bert_ranker/bi_encoder_ranker.py
```diff
@@ -69,9 +69,11 @@ def set_vocab_candidates(self, shared):
         self.vocab_candidates will contain a [num_cands] list of strings
         self.vocab_candidate_vecs will contain a [num_cands, 1] LongTensor
         """
+        self.opt['encode_candidate_vecs'] = True
         if shared:
             self.vocab_candidates = shared['vocab_candidates']
             self.vocab_candidate_vecs = shared['vocab_candidate_vecs']
+            self.vocab_candidate_encs = shared['vocab_candidate_encs']
```
stephenroller: dupe

parlai/agents/bert_ranker/bi_encoder_ranker.py
```diff
@@ -135,23 +137,16 @@ def _set_text_vec(self, *args, **kwargs):
...
-        if (hasattr(self, 'fixed_candidate_encs') and
-                self.fixed_candidate_encs is not None):
-            return embedding_ctxt.mm(self.fixed_candidate_encs.t())
-
-        # evaluating vocab candidates:
-        if (hasattr(self, 'vocab_candidate_encs') and
-                self.vocab_candidate_encs is not None):
-            return embedding_ctxt.mm(self.vocab_candidate_encs.t())
+        if cand_encs is not None:
+            return embedding_ctxt.mm(cand_encs.t())
```
stephenroller: this reads much better

parlai/agents/transformer/modules.py
```diff
@@ -144,6 +144,9 @@ def encode_cand(self, words):
 
     def encode_context_memory(self, context_w, memories_w):
         # [batch, d]
+        if context_w is None:
```
stephenroller: Why did this change here?

stephenroller: oh nvm, I see below. So Transformer wasn't even using the caching before?

stephenroller: Actually, I'm still confused. Nothing about context memory should ever be cached right?

parlai/core/torch_ranker_agent.py
```diff
@@ -119,8 +112,16 @@ def __init__(self, opt, shared=None):
                 broadcast_buffers=False,
             )
 
-    def score_candidates(self, batch, cand_vecs):
-        """Given a batch and candidate set, return scores (for ranking)"""
+    def score_candidates(self, batch, cand_vecs, cand_encs=None):
+        """Given a batch and candidate set, return scores (for ranking).
+
+        :param batch:       a Batch object (defined in torch_agent.py)
```
stephenroller: Mind if we use the new docstring format?

https://github.com/facebookresearch/ParlAI/blob/baa839db2120c837562620386cbe1167bd0e9109/parlai/core/torch_agent.py#L650-L659

parlai/agents/bert_ranker/bi_encoder_ranker.py
emilydinan: sorry i don't understand this comment?

parlai/agents/transformer/modules.py
emilydinan: correct - transformer was not caching before. 

this is just saying if context is None, we want to return None. (useful for when we want to encode just the RHS)

parlai/agents/bert_ranker/bi_encoder_ranker.py
stephenroller: Oh I thought it was a duplicate of the line above but I misread. Please disregard

parlai/agents/transformer/modules.py
stephenroller: So I’m still a tad bit confused why then this guard ends up in the encode_context_memory method

stephenroller: Anyway maybe you can just add a short comment here saying why context_w might be None and then it lgtm.

samhumeau: "this is just saying if context is None, we want to return None. "
I am confused. If context is None, shouldn't we return the embedding of the candidates only?

samhumeau: Ah, nevermind, wrong function

PR #1547, opened 2019-03-13
adding several options for aggregation

samhumeau:
Right now in the bi-encoder we were taking the first output of Bert to reduce the sentence to a single vector. Added possibility to use the average over all the outputs, or the max (In all cases it is followed by a linear layer).

Results on convai2 valid set:
- first output 78.8
- mean: 79.9
- max: 78.1

Reproduce results with:
```
PYTHONPATH=. python -u parlai/scripts/train_model.py -pyt convai2  \
    -m parlai.agents.bert_ranker.bi_encoder_ranker:BiEncoderRankerAgent \
    --batchsize 128 --dict-file ./mydic --model-file mymodel   --eval-batchsize 128 \
    --learningrate 5e-5  --log_every_n_secs 20  --shuffle true --type-optimization all_encoder_layers \
    --lr-scheduler fixed  --lr-scheduler-patience 1 --lr-scheduler-decay 0.45  --data-parallel true \
    --truncate 300 --num-epochs 4.0  -veps 0.999 -vme 8000 --warmup_updates 200  \
    --bert-aggregation (mean|max|first)
```

### DIFF
--- a/parlai/agents/bert_ranker/bi_encoder_ranker.py
+++ b/parlai/agents/bert_ranker/bi_encoder_ranker.py
@@ -195,13 +195,15 @@ class BiEncoderModule(torch.nn.Module):
             BertModel.from_pretrained(opt['pretrained_path']),
             opt['out_dim'],
             add_transformer_layer=opt['add_transformer_layer'],
-            layer_pulled=opt['pull_from_layer']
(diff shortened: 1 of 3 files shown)


stephenroller:
Can we just drop max? It’s worse than both the others so why add it?

PR #1548, opened 2019-03-13
SNLI task fix

emilydinan:
This task breaks because often the label is not included in the label candidates. Fix by fixing label candidates.


jaseweston:
what is '-' ?

emilydinan:
> what is '-' ?

it was exmaples where there is no agreement. these examples were supposed to be skipped, but were not do to a bug in the multinli task which i now fixed.

PR #1549, opened 2019-03-13
[Torch Classifier] Abstract class implementation + BERT Classifier

emilydinan:
- Abstract TorchClassifierAgent (only slightly modified from @samhumeau 's original internal version) meant to take away some of the overhead for training and evaluating classification models.
- Example implementation of a BERT-based classifier.

### DIFF
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -257,12 +257,10 @@ class TorchClassifierAgent(TorchAgent):
 
             if len(examples_per_class) > 1:
                 # get weighted f1
-                percent_of_total = [
-                    x / sum(examples_per_class) for x in examples_per_class
-                ]
                 f1 = 0
+                total_exs = sum(examples_per_class)
                 for i in range(len(self.class_list)):
-                    f1 += (percent_of_total[i] *
+                    f1 += ((self.class_list[i] / total_exs) *
                            m['class_{}_f1'.format(self.class_list[i])])
                 m['weighted_f1'] = f1
 
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -3,6 +3,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+from parlai.core.distributed_utils import is_distributed
 from parlai.core.torch_agent import TorchAgent, Output
 from parlai.core.utils import round_sigfigs, warn_once
 from collections import defaultdict
@@ -103,6 +104,10 @@ class TorchClassifierAgent(TorchAgent):
             self.model.cuda()
             self.classifier_loss.cuda()
             if self.opt['data_parallel']:
+                if is_distributed():
+                    raise ValueError(
+                        'Cannot combine --data-parallel and distributed mode'
+                    )
                 self.model = torch.nn.DataParallel(self.model)
         if shared:
             # We don't use get here because hasattr is used on optimizer later.
@@ -142,9 +147,11 @@ class TorchClassifierAgent(TorchAgent):
     def _update_confusion_matrix(self, batch, predictions):
         """Update the confusion matrix given the batch and predictions.
 
-        :param Batch batch: a Batch object (defined in torch_agent.py)
-        :param list predictions: (list of string of length batchsize) label
-            predicted by the classifier
+        :param batch:
+            a Batch object (defined in torch_agent.py)
+        :param predictions:
+            (list of string of length batchsize) label predicted by the
+            classifier
         """
         for i, pred in enumerate(predictions):
             label = batch.labels[i]
@@ -276,8 +283,10 @@ class TorchClassifierAgent(TorchAgent):
     def score(self, batch):
         """Given a batch and labels, returns the scores.
 
-        :param Batch batch: a Batch object (defined in torch_agent.py)
-        :return: a [bsz, num_classes] FloatTensor containing the score of each
+        :param batch:
+            a Batch object (defined in torch_agent.py)
+        :return:
+            a [bsz, num_classes] FloatTensor containing the score of each
             class.
         """
         raise NotImplementedError(
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -158,7 +158,7 @@ class TorchClassifierAgent(TorchAgent):
         for i, pred_id in enumerate(prediction_id.tolist()):
             prob = round_sigfigs(probs[i][pred_id], 4)
             preds.append('Predicted class: {}\nwith probability: {}'.format(
-                         self.class_list[pred_id],prob))
+                         self.class_list[pred_id], prob))
         return preds
 
     def train_step(self, batch):
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -230,6 +230,29 @@ class TorchClassifierAgent(TorchAgent):
         self.metrics['examples'] = 0
         self.metrics['loss'] = 0.0
 
+    def _report_prec_recall_metrics(self, confmat, class_name, metrics):
+        """Uses the confusion matrix to predict the recall and precision for
+        class `class_name`. Returns the number of examples of each class.
+        """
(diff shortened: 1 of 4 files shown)


parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+        # class arguments
+        parser.add_argument('--classes', type=str, nargs='*', default=None,
+                            help='the name of the classes.')
+        parser.add_argument('--class-weights', type=float, nargs='*', default=None,
+                            help='weight of each of the classes for the softmax')
+        parser.add_argument('--ref-class', type=str, default=None, hidden=True,
+                            help='the class that will be used to compute '
+                                 'precision and recall. By default the first '
+                                 'class.')
+        parser.add_argument('--threshold', type=float, default=0.5,
```
stephenroller: I'm tempted to just say we should use argmax() but I understand why you need to tune this.

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+                            help='during evaluation, threshold for choosing '
+                                 'ref class; only applies to binary '
+                                 'classification')
+        # interactive mode
+        parser.add_argument('--interactive-mode', type='bool', default=False)
+        parser.add_argument('--print-scores', type='bool', default=False,
+                            help='print probability of chosen class during '
+                                 'interactive mode')
+        # miscellaneous arguments
+        parser.add_argument('--data-parallel', type="bool", default=False,
```
stephenroller: I don't think we want data parallel at all. The reason for data parallel in ranker agent is the ability for negative samples. In a classifier agent, where the classes are fixed, this advantage is moot and distributed will be better suited in nearly every way.

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+        parser.add_argument('--ref-class', type=str, default=None, hidden=True,
+                            help='the class that will be used to compute '
+                                 'precision and recall. By default the first '
+                                 'class.')
+        parser.add_argument('--threshold', type=float, default=0.5,
+                            help='during evaluation, threshold for choosing '
+                                 'ref class; only applies to binary '
+                                 'classification')
+        # interactive mode
+        parser.add_argument('--interactive-mode', type='bool', default=False)
```
stephenroller: hmm, we can't infer this from whether a label is provided?

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+        parser.add_argument('--print-scores', type='bool', default=False,
```
stephenroller: do we have a debug or verbose option elsewhere we can use instead of this specialized one?

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+        parser.add_argument('--get-all-metrics', type='bool', default=True,
+                            help='give prec/recall metrics for all classes')
+
+    def __init__(self, opt, shared=None):
+        init_model, _ = self._get_init_model(opt, shared)
+        super().__init__(opt, shared)
+
+        # set up classes
+        if opt.get('classes') is None:
+            raise RuntimeError(
```
stephenroller: good

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+            self.reset_metrics()
+        else:
+            self.class_list = shared['class_list']
+            self.class_dict = shared['class_dict']
+            self.class_weights = shared['class_weights']
+
+        # get reference class; if opt['get_all_metrics'] is False, this is
+        # used to compute metrics
+        # in binary classfication, opt['threshold'] applies to ref class
+        if (opt['ref_class'] is None or opt['ref_class'] not in
```
stephenroller: no wrap needed

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+            raise e
+        labels_tensor = torch.LongTensor(labels_indices_list)
+        if self.use_cuda:
+            labels_tensor = labels_tensor.cuda()
+        return labels_tensor
+
+    def _update_confusion_matrix(self, batch, predictions):
+        """Update the confusion matrix given the batch and predictions.
+
+        :param batch:         a Batch object (defined in torch_agent.py)
```
stephenroller: docstring format

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+
+        # set up threshold, only used in binary classification
+        if len(self.class_list) == 2 and opt.get('threshold', 0.5) != 0.5:
+            self.threshold = opt['threshold']
+        else:
+            self.threshold = None
+
+        # set up model and optimizers
+        weight_tensor = torch.FloatTensor(self.class_weights)
+        self.classifier_loss = torch.nn.CrossEntropyLoss(weight_tensor)
```
stephenroller: add `reduce='mean'`?

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+
+        if self.opt.get('interactive_mode', False):
+            if self.opt.get('print_scores', False):
+                probabilities = []
+                for score in scores:
+                    probabilities.append(probs[0])
+                preds = []
+                for i, pred_id in enumerate(prediction_id):
+                    prob = round_sigfigs(probabilities[i][pred_id], 4)
+                    preds.append('{}: {}'.format(self.class_list[pred_id],
```
stephenroller: I think this is strange behavior. Do you not want to give it back in a structured form?

parlai/core/torch_classifier_agent.py
```diff
@@ -0,0 +1,285 @@
...
+
+        for k, v in m.items():
+            m[k] = round_sigfigs(v, 4)
+
+        return m
+
+    def score(self, batch):
+        """Given a batch and labels, returns the scores.
+
+        :param batch:   a Batch object (defined in torch_agent.py)
```
stephenroller: docstring format

parlai/core/torch_classifier_agent.py
emilydinan: we typically just use the max, but we found this threshold with one of the projects we are using this for

parlai/core/torch_classifier_agent.py
emilydinan: i suppose, the reason i made it explicit is that we've talked about instances where we might not want to provide a label (but it's not interactive mode) but AFAIK this doesn't exist yet

parlai/core/torch_classifier_agent.py
emilydinan: what do you mean by wrap

parlai/core/torch_classifier_agent.py
emilydinan: isn't that the default?

parlai/core/torch_classifier_agent.py
emilydinan: i'll clean this up

parlai/core/torch_classifier_agent.py
emilydinan: what's the harm in keeping it?

parlai/core/torch_classifier_agent.py
emilydinan: this is only in interactive mode

parlai/core/torch_classifier_agent.py
emilydinan: i got rid of this flag in the latest updates

parlai/core/torch_classifier_agent.py
```diff
@@ -272,9 +276,9 @@ def report(self):
     def score(self, batch):
         """Given a batch and labels, returns the scores.
 
-        :param batch:   a Batch object (defined in torch_agent.py)
-        :return:        a [bsz, num_classes] FloatTensor containing the score
-                        of each class.
+        :param Batch batch: a Batch object (defined in torch_agent.py)
+        :return: a [bsz, num_classes] FloatTensor containing the score of each
```
stephenroller: nit on the formatting but *shrug*
```
:return:
    a [bsz, num_classes] ...`
```

emilydinan: lol i'll fix my apologies

PR #1550, opened 2019-03-13
[MTurk] Wrapping requests to parl.ai to prevent site-based outages

JackUrb:
Right now if the parlai webserver goes down, it's impossible to launch an MTurk job without manually overriding some exceptions. This is obviously bad behavior, so I've added try/excepts to handle when the request fails.

### DIFF
--- a/parlai/mturk/core/mturk_manager.py
+++ b/parlai/mturk/core/mturk_manager.py
@@ -54,8 +54,8 @@ SNS_ASSIGN_ABANDONDED = 'AssignmentAbandoned'
 SNS_ASSIGN_SUBMITTED = 'AssignmentSubmitted'
 SNS_ASSIGN_RETURNED = 'AssignmentReturned'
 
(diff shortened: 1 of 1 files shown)


parlai/mturk/core/mturk_manager.py
```diff
@@ -277,7 +277,14 @@ def _upload_worker_data(self):
         worker_data = self.worker_manager.get_worker_data_package()
         data = {'worker_data': worker_data}
         headers = {'Content-type': 'application/json', 'Accept': 'text/plain'}
-        requests.post(PARLAI_MTURK_UPLOAD_URL, json=data, headers=headers)
+        try:
+            requests.post(PARLAI_MTURK_UPLOAD_URL, json=data, headers=headers)
+        except Exception:
+            shared_utils.print_and_log(
+                logging.WARNING,
+                "Unable to log worker statistics to parl.ai",
```
klshuster: nit: double quotes

PR #1552, opened 2019-03-13
Drop CLS token from label in bert.

stephenroller:
Pulled this out of #1472 because it made more sense on its own.

The comment in the code was misleading, and so I updated the code to be a bit more like what it *should* be. Breaks backwards compatibility with any other cross encoder models trained so far though

Also adds a value error around `-cands`. Maybe that is overly aggressive though.


stephenroller:
Jason lamented this will break old models. I'm hoping maybe I can fix this using something similar to #1488.

stephenroller:
Suggesting Kurt or Sam take this over (or officially abandon it), since it affects them more. Doing this with something like upgrade_opt might be the way to go.

PR #1553, opened 2019-03-13
Correct datatests location for error message

JackUrb:
The error message for this test didn't point to the correct location, now it does


stephenroller:
`#docday`

PR #1554, opened 2019-03-13
Interactive web

bharathc346:
Creating a script that can be run on a local web server to interact with models. Same functionality as examples/interactive.py except with a chat UI. Can run the script by passing the same CLI arguments one would to examples/interactive.py and going to http://localhost:8080/

### DIFF
--- a/parlai/scripts/interactive_web.py
+++ b/parlai/scripts/interactive_web.py
@@ -3,10 +3,8 @@ from parlai.scripts.interactive import setup_args
 from parlai.core.agents import create_agent
 from parlai.core.worlds import create_task
 from parlai.core.params import ParlaiParser
-from parlai.core.utils import display_messages
 
 import json
-import socketserver
 import time
 
 HOST_NAME = 'localhost'
@@ -165,6 +163,7 @@ WEB_HTML = """
 </html>
 """
 
+
 class MyHandler(BaseHTTPRequestHandler):
 
     def interactive_running(self, opt, reply_text):
@@ -183,7 +182,6 @@ class MyHandler(BaseHTTPRequestHandler):
         if self.path == '/interact':
             content_length = int(self.headers['Content-Length'])
             body = self.rfile.read(content_length)
-
             model_response = self.interactive_running(SHARED.get('opt'), body.decode('utf-8'))
 
             self.send_response(200)
@@ -226,7 +224,7 @@ def setup_interactive(shared):
     print_parser = parser
 
     if print_parser is not None:
-        if print_parser is True and isinstance(self.server.SHARED['opt'], ParlaiParser):
+        if print_parser is True and isinstance(SHARED['opt'], ParlaiParser):
             print_parser = SHARED['opt']
         elif print_parser is False:
             print_parser = None
--- a/parlai/scripts/interactive_web.py
+++ b/parlai/scripts/interactive_web.py
@@ -4,23 +4,175 @@ from parlai.core.agents import create_agent
 from parlai.core.worlds import create_task
 from parlai.core.params import ParlaiParser
 from parlai.core.utils import display_messages
-from io import BytesIO
 
-# from multiprocessing import Pool, Manager
+import json
 import socketserver
 import time
 
 HOST_NAME = 'localhost'
 PORT_NUMBER = 8080
 SHARED = {}
+STYLE_SHEET = "https://cdnjs.cloudflare.com/ajax/libs/bulma/0.7.4/css/bulma.css"
+FONT_AWESOME = "https://use.fontawesome.com/releases/v5.3.1/js/all.js"
+WEB_HTML = """
+<html>
+    <link rel="stylesheet" href={} />
+    <script defer src={}></script>
+    <head><title> Interactive Run </title></head>
+    <body>
+        <div class="columns">
+            <div class="column is-three-fifths is-offset-one-fifth">
+              <section class="hero is-info is-large has-background-light has-text-grey-dark	">
+                <div id="parent" class="hero-body">
+                    <article class="media">
+                      <figure class="media-left">
+                        <span class="icon is-large">
+                          <i class="fas fa-robot fas fa-2x"></i>
+                        </span>
+                      </figure>
+                      <div class="media-content">
+                        <div class="content">
+                          <p>
+                            <strong>Model</strong>
+                            <br>
+                            Enter a message, and the model will respond interactively.
+                          </p>
+                        </div>
+                      </div>
+                    </article>
+                </div>
+                <div class="hero-foot column is-three-fifths is-offset-one-fifth">
+                  <form id = "interact">
+                      <div class="field is-grouped">
+                        <p class="control is-expanded">
+                          <input class="input" type="text" id="userIn" placeholder="Type in a message">
+                        </p>
+                        <p class="control">
+                          <button id="respond" type="submit" class="button has-text-white-ter has-background-grey-dark">
+                            Submit
+                          </button>
+                        </p>
+                      </div>
+                  </form>
+                </div>
+              </section>
+            </div>
+        </div>
 
+        <script>
+            document.getElementById("interact").addEventListener("submit", function(event){{
+                event.preventDefault()
+                var text = document.getElementById("userIn").value;
+                document.getElementById('userIn').value = "";
+
+                fetch('/interact', {{
+                    headers: {{
+                        'Content-Type': 'application/json'
+                    }},
+                    method: 'POST',
+                    body: text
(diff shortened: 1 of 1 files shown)


parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,260 @@
...
+
+                    span.appendChild(icon);
+                    figure.appendChild(span);
+
+                    article.appendChild(figure);
+                    article.appendChild(media);
+
+                    parDiv.append(article);
+
+                    // Change info for Model response
```
bharathc346: Is there a way to do this by reusing the elements created for the Person response as opposed to creating new elements for the model response? I tried reusing, but couldn't get it to work. @pringshia

bharathc346:
image

jaseweston:
nicee!

stephenroller:
Niiiiice

emilydinan:
awesome!!!

parlai/scripts/interactive_web.py
pringshia: Pretty sure what you're describing should be feasible. Let's debug offline

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+            json_str = json.dumps(model_response)
+            self.wfile.write(bytes(json_str, 'utf-8'))
+
+        else:
+            self.respond({'status': 500})
+
+    def do_GET(self):
+        paths = {
+            '/': {'status': 200},
+            '/favicon.ico': {'status': 202},    # Need for chrome
```
stephenroller: lol

good comment tho

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+            )
+
+            self.send_response(200)
+            self.send_header('Content-type', 'application/json')
+            self.end_headers()
+            json_str = json.dumps(model_response)
+            self.wfile.write(bytes(json_str, 'utf-8'))
+
+        else:
+            self.respond({'status': 500})
```
stephenroller: One way you can have cleaner code in such a situation is invert the condition and fail fast.

```
if self.path != '/interact':
   return self.respond({'status': 500})

content_length = int(self.headers['Content-Length'])
...
```

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def respond(self, opts):
+        response = self.handle_http(opts['status'], self.path)
+        self.wfile.write(response)
+
+
+def setup_interactive(shared):
+    parser = setup_args()
+    SHARED['opt'] = parser.parse_args(print_args=False)
+    print_parser = parser
```
stephenroller: Delete from here...

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+
+    if print_parser is not None:
+        if print_parser is True and isinstance(SHARED['opt'], ParlaiParser):
+            print_parser = SHARED['opt']
+        elif print_parser is False:
+            print_parser = None
+    if isinstance(SHARED['opt'], ParlaiParser):
+        print('[ Deprecated Warning: interactive should be passed opt not Parser ]')
+        SHARED['opt'] = SHARED['opt'].parse_args()
+
```
stephenroller: ...to here.

You copied from old code.

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+    SHARED['opt']['task'] = 'parlai.agents.local_human.local_human:LocalHumanAgent'
+
+    # Create model and assign it to the specified task
+    SHARED['agent'] = create_agent(SHARED.get('opt'), requireModelExists=True)
+    SHARED['world'] = create_task(SHARED.get('opt'), SHARED['agent'])
+
+    if print_parser:
```
stephenroller: delete this block and change the `print_args=True` in the call to `parse_args`

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+if __name__ == '__main__':
+    setup_interactive(SHARED)
+    server_class = HTTPServer
+    Handler = MyHandler
+    Handler.protocol_version = 'HTTP/1.0'
+    httpd = server_class((HOST_NAME, PORT_NUMBER), Handler)
+
+    print(
```
stephenroller: Let's just
`print('http://{}:{}/'.format(HOST_NAME, PORT))`

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+            HOST_NAME,
+            PORT_NUMBER
+        )
+    )
+    try:
+        httpd.serve_forever()
+    except KeyboardInterrupt:
+        pass
+    httpd.server_close()
+    print(time.asctime(), 'Server Stops - %s:%s' % (HOST_NAME, PORT_NUMBER))
```
stephenroller: Don't need this print

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+from parlai.scripts.interactive import setup_args
+from parlai.core.agents import create_agent
+from parlai.core.worlds import create_task
+from parlai.core.params import ParlaiParser
+
+import json
+import time
+
+HOST_NAME = 'localhost'
+PORT_NUMBER = 8080
```
stephenroller: rename to PORT

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from parlai.scripts.interactive import setup_args
+from parlai.core.agents import create_agent
+from parlai.core.worlds import create_task
+from parlai.core.params import ParlaiParser
+
+import json
+import time
```
stephenroller: after addressing my other comments you probably don't need this.

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Basic script which allows human keyboard input to talk to a trained
```
stephenroller: Talk with a model using a web UI.

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,280 @@
...
+FONT_AWESOME = "https://use.fontawesome.com/releases/v5.3.1/js/all.js"
+WEB_HTML = """
+<html>
+    <link rel="stylesheet" href={} />
+    <script defer src={}></script>
+    <head><title> Interactive Run </title></head>
+    <body>
+        <div class="columns">
+            <div class="column is-three-fifths is-offset-one-fifth">
+              <section class="hero is-info is-large has-background-light has-text-grey-dark	">
```
stephenroller: a little extra whitespace here

parlai/scripts/interactive_web.py
```diff
@@ -0,0 +1,244 @@
...
+                  </form>
+                </div>
+              </section>
+            </div>
+        </div>
+
+        <script>
+            function createChatRow(agent, text) {{
+
+            var article = document.createElement("article");
```
pringshia: indentation is off.

Note: Talked to Jai offline and we discussed cleaning up the code a bit by introducing this helper method to generate chat rows. Additionally, instead of introducing another dependency, we are trying to use a lightweight approach to templating here. There are some downsides though, e.g. curly braces have to be escaped, e.g. ``{{`` and ``}}`` and also we have to be more mindful of HTML injection. In this case, `createTextNode` protects us in the case of the text. The agent name could technically be a vector for injection, however it is not modifiable by the user.

pringshia:
This looks really good Jai! Nice work incorporating Bulma, the interface looks clean.

Summarizing an offline discussion:
- Jai to add code to jump the screen to the bottom of the page after a new message is posted. Currently the input box gets pushed off screen. (@bharathc346 I added this functionality for the mTurk chat feature - you can take a look there if helpful: https://github.com/facebookresearch/ParlAI/pull/1382/files#diff-940188883789c3ba879c2dd6252a699eR195)

parlai/scripts/interactive_web.py
```diff
@@ -69,6 +64,46 @@
         </div>
 
         <script>
+            function createChatRow(agent, text) {{
```
stephenroller: I appreciate this refactor

PR #1558, opened 2019-03-18
Adding Mean Reciprocal Rank to the list of metrics

samhumeau:
MRR (Mean Reverse Ranking) is a metric used in some places of the literature to evaluate the quality of a ranking model (e.g. http://workshop.colips.org/dstc7/papers/17.pdf).

It's the average of the inverse of the rank, and designed to be softer than the accuracy. If the groudtruth is ranked 2nd, the model still "gets" 1/2 point for this sample. 1/3 f it is ranked 3rd and so on.

This PR should implement it.

Moreover I noticed mean_rank was wrong at train time, since it was computed 2 times, modified so that it's computed just once.


parlai/core/torch_ranker_agent.py
```diff
@@ -137,10 +137,6 @@ def get_batch_train_metrics(self, scores):
         targets = torch.arange(batchsize, out=targets)
         nb_ok = (scores.max(dim=1)[1] == targets).float().sum().item()
         self.metrics['train_accuracy'] += nb_ok
-        # calculate mean rank
```
samhumeau: self.metrics['rank'] was incremented here.

parlai/core/torch_ranker_agent.py
```diff
@@ -203,6 +200,7 @@ def train_step(self, batch):
         for b in range(batchsize):
             rank = (ranks[b] == label_inds[b]).nonzero().item()
             self.metrics['rank'] += 1 + rank
```
samhumeau: But also here

parlai/core/torch_ranker_agent.py
```diff
@@ -137,10 +137,6 @@ def get_batch_train_metrics(self, scores):
         targets = torch.arange(batchsize, out=targets)
         nb_ok = (scores.max(dim=1)[1] == targets).float().sum().item()
         self.metrics['train_accuracy'] += nb_ok
-        # calculate mean rank
-        above_dot_prods = scores - scores.diag().view(-1, 1)
-        rank = (above_dot_prods > 0).float().sum().item()
-        self.metrics['rank'] += rank
```
emilydinan: i'm confused -- why is this deleted?

emilydinan: nevermind i see -- you can add this code back. the bug is fixed by #1559

stephenroller:
Good catch Sam

parlai/core/torch_ranker_agent.py
samhumeau: ok

PR #1560, opened 2019-03-18
[BERT] skip building dictionary

emilydinan:
BERT dictionary is already built so skip this step

Notes:
- How this works: if BERT agents add the "common args" from helpers, the dictionary should automatically skip being built since the maximum examples to pass through will be set to zero (otherwise, we override "act" to essentially do nothing)
- I considered setting this default in the BertDictionaryAgent (cleaner), but this would require many substantial changes to the way params are parsed
- On the above note, I noticed that in the typical setting of building a dictionary (called from train_model script), the following lines are never exectued because `dict_class` will not be set in params yet: https://github.com/facebookresearch/ParlAI/blob/10e3e271adfdc457dfe3895d17f93a870a76b603/parlai/scripts/build_dict.py#L48-L49

### DIFF
--- a/parlai/agents/bert_ranker/bert_dictionary.py
+++ b/parlai/agents/bert_ranker/bert_dictionary.py
@@ -17,12 +17,11 @@ import os
 
 
 class BertDictionaryAgent(DictionaryAgent):
-    """ Allow to use the Torch Agent with the wordpiece dictionary of Hugging Face.
+    """Allow to use the Torch Agent with the wordpiece dictionary of Hugging Face.
     """
-
     def __init__(self, opt):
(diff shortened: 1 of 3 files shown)


stephenroller:
Looks like this changes the semantics from "maxexs = 0 means use all the examples" to "maxexs = -1 means use all the examples"? I suspect we need to update the help string for that, and maybe the default

stephenroller:
NVM:

https://github.com/facebookresearch/ParlAI/blob/10e3e271adfdc457dfe3895d17f93a870a76b603/parlai/scripts/build_dict.py#L35-L37

PR #1562, opened 2019-03-18
Forceful saving of dictionaries

klshuster:
We now explicitly save a `/path/to/model.dict` file. This makes copying model files a more straightforward process, as you do not have to worry about any number of issues that may cause you to lose access to (or simply forget the location of) a specified dict file. 

This PR also moves the aggregation of the various model components to save to a new function, `get_save_dict`, which is what subclasses of `torch_agent` should override for more specific saving.

### DIFF
--- a/parlai/agents/fairseq/fairseq.py
+++ b/parlai/agents/fairseq/fairseq.py
@@ -447,6 +447,9 @@ class FairseqAgent(TorchAgent):
                 del self.opt['override']
             json.dump(self.opt, handle)
 
+        # force save the dict
+        self.dict.save(path + '.dict', sort=False)
+
     def load(self, path):
         """Load using fairseq's checkpointing."""
         if self.trainer:
--- a/parlai/agents/fairseq/fairseq.py
+++ b/parlai/agents/fairseq/fairseq.py
@@ -447,6 +447,9 @@ class FairseqAgent(TorchAgent):
                 del self.opt['override']
             json.dump(self.opt, handle)
 
+        # force save the dict
+        self.dict.save(path + '.dict', sort=True)
+
     def load(self, path):
         """Load using fairseq's checkpointing."""
         if self.trainer:
--- a/parlai/agents/fairseq/fairseq.py
+++ b/parlai/agents/fairseq/fairseq.py
@@ -447,9 +447,6 @@ class FairseqAgent(TorchAgent):
                 del self.opt['override']
             json.dump(self.opt, handle)
 
-        # force save the dict
-        self.dict.save(path + '.dict')
-
     def load(self, path):
         """Load using fairseq's checkpointing."""
         if self.trainer:
--- a/parlai/agents/fairseq/fairseq.py
+++ b/parlai/agents/fairseq/fairseq.py
@@ -448,7 +448,7 @@ class FairseqAgent(TorchAgent):
             json.dump(self.opt, handle)
 
         # force save the dict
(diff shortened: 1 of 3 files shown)


parlai/agents/fairseq/fairseq.py
```diff
@@ -447,6 +447,9 @@ def save(self, path):
                 del self.opt['override']
             json.dump(self.opt, handle)
 
+        # force save the dict
+        self.dict.save(path + '.dict', sort=True)
```
stephenroller: I don't think we sort=True here, because that is destructive on the ordering. At this point, we absolutely want the ordering to stay fixed.

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -169,29 +167,18 @@ def batchify(self, *args, **kwargs):
...
-
-            with open(path, 'wb') as write:
-                torch.save(model, write)
-
-            # save opt file
-            with open(path + '.opt', 'w') as handle:
-                # save version string
-                self.opt['model_version'] = self.model_version()
-                json.dump(self.opt, handle)
+    def get_save_dict(self):
```
stephenroller: call this `def state_dict(self)`, to mirror pytorch.

parlai/core/torch_agent.py
```diff
@@ -1295,41 +1295,52 @@ def observe(self, observation):
                               text_truncate=self.text_truncate,
                               label_truncate=self.label_truncate)
 
+    def get_save_dict(self):
```
emilydinan: can we rename this something that makes it clear this is referring to the state dict and not hte dictionary? maybe like `get_state_dict` or `get_save_state_dict`?

klshuster:
I've moved the dict saving from `self.save` to `__init__`. The dict for some models (e.g. memnn) is updated after building with special tokens, and we should not save this modified dict as the default dict (since the model **always** adds those tokens)

klshuster:
update: i've reverted to what I had before. I had not realized that `sort` was `True` by default

stephenroller:
Does this resolve the issue you discussed in the meeting?

klshuster:
yes, i had misunderstood the error - dict items weren't being readded, they were being shuffled

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -169,29 +167,18 @@ def batchify(self, *args, **kwargs):
...
-            with open(path, 'wb') as write:
-                torch.save(model, write)
-
-            # save opt file
-            with open(path + '.opt', 'w') as handle:
-                # save version string
-                self.opt['model_version'] = self.model_version()
-                json.dump(self.opt, handle)
+    def state_dict(self):
+        """Get the model states for saving
```
stephenroller: Nit: newline on multiline docstrings.

parlai/agents/seq2seq/seq2seq.py
```diff
@@ -169,29 +167,18 @@ def batchify(self, *args, **kwargs):
...
+
+        Override to include longest_label
```
stephenroller: Nit: Overridden

parlai/core/torch_agent.py
```diff
@@ -1295,41 +1295,52 @@ def observe(self, observation):
                               text_truncate=self.text_truncate,
                               label_truncate=self.label_truncate)
 
+    def state_dict(self):
+        """Get the state dict for saving
```
stephenroller: ditto

PR #1565, opened 2019-03-19
Personality-Captions Website

klshuster:
Website for personality-captions. Note that pretrained models are not included yet.


klshuster:
ok, i added the link

PR #1572, opened 2019-03-22
Transformer GELU activations and XLM/Bert architectures.

malachaux:


parlai/agents/transformer/modules.py
```diff
@@ -28,8 +28,8 @@ def _create_embeddings(dictionary, embedding_size, padding_idx):
     return e
 
 
-def _build_encoder(opt, dictionary, embedding=None, padding_idx=None, reduction=True,
-                   n_positions=1024):
+def _build_encoder(opt, dictionary, embedding=None, padding_idx=None, reduction=True,reduction_type=None,
+                   n_positions=1024,gelu_activation=False,embedding_normalization=False):
```
emilydinan: nit: missing whitespace after commas

parlai/agents/transformer/modules.py
```diff
@@ -67,6 +70,10 @@ def _build_decoder(opt, dictionary, embedding=None, padding_idx=None,
     )
 
 
+def gelu(tensor):
+    return 0.5 * tensor * (1.0 + torch.erf(tensor/math.sqrt(2.0)))
```
emilydinan: nit: white space around binary operators

parlai/agents/transformer/modules.py
```diff
@@ -110,8 +134,11 @@ def __init__(self, opt, dictionary):
             )
         else:
             self.cand_encoder = _build_encoder(
-                opt, dictionary, self.embeddings, self.pad_idx, reduction=True,
+                opt, dictionary, self.embeddings, self.pad_idx,
                 n_positions=n_positions,
+                reduction=True, reduction_type = self.reduction_type,
```
emilydinan: nit: no whitespace around '=' here (same for elsewhere in this PR)

parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -293,6 +303,11 @@ def get(self, episode_idx, entry_idx=0):
         if self.label_type == 'response':
             if 'train' in self.datatype:
                 label_cands = []
+                num_to_gen = self.opt["num_negatives_responses_generated_train"]
```
emilydinan: nit: single quotes

parlai/agents/transformer/modules.py
```diff
@@ -99,9 +108,26 @@ def __init__(self, opt, dictionary):
...
+            self.gelu_activation = opt['gelu_activation']
+        else:
+            self.gelu_activation = False
+
+        if opt.get('embedding_normalization'):
+            self.embedding_normalization = opt['embedding_normalization']
+        else:
+            self.embedding_normalization = False
+
+        if opt.get('reduction_type'):
```
stephenroller: You can simplify these with `self.reduction_type = opt.get('reduction_type', False)`

parlai/agents/transformer/modules.py
```diff
@@ -99,9 +108,26 @@ def __init__(self, opt, dictionary):
         if n_positions < 0:
             raise ValueError('n_positions must be positive')
 
+        if opt.get('gelu_activation'):
```
stephenroller: I would prefer if we had an `--activation` option with choices `{'relu', 'gelu'}`. That would make it easier for us to introduce another activation of choice later.

parlai/agents/transformer/modules.py
```diff
@@ -310,17 +349,23 @@ def forward(self, input):
...
-
         if self.reduction:
-            divisor = mask.float().sum(dim=1).unsqueeze(-1).clamp(min=1e-20)
-            output = tensor.sum(dim=1) / divisor
-            return output
+            if self.reduction_type == 'first':
+                return tensor[:, 0, :]
+            elif self.reduction_type == 'max':
+                return tensor.max(dim=1)[0]
+            else:
```
stephenroller: I would prefer an explicit elif for `mean` here, and then the else would raise a ValueError

parlai/agents/transformer/modules.py
```diff
@@ -335,17 +380,21 @@ def __init__(
...
         self.ffn_dim = ffn_size
+        self.gelu_activation = gelu_activation
         self.attention = MultiHeadAttention(
             n_heads, embedding_size,
             dropout=attention_dropout,  # --attention-dropout
         )
-        self.norm1 = nn.LayerNorm(embedding_size)
-        self.ffn = TransformerFFN(embedding_size, ffn_size, relu_dropout=relu_dropout)
-        self.norm2 = nn.LayerNorm(embedding_size)
+        self.norm1 = nn.LayerNorm(embedding_size, eps=1e-12)
```
stephenroller: pull out the 1e-12 into a constant at the top.

parlai/agents/transformer/modules.py
```diff
@@ -310,17 +349,23 @@ def forward(self, input):
         if self.embeddings_scale:
             tensor = tensor * np.sqrt(self.dim)
         tensor = tensor + self.position_embeddings(positions).expand_as(tensor)
+        if self.embedding_normalization:
+            tensor = self.layer_norm_emb(tensor)
```
stephenroller: So this was a bit of a surprising change to me, because it differs from what fairseq/tensor2tensor do (and what I THOUGHT you were implementing):

https://github.com/pytorch/fairseq/blob/b26d6b58b0bd55c1b59aa89f1b03d79011dbe7aa/fairseq/models/transformer.py#L604-L618

That is, the layer normalizations either happen before the attn/ffn (tensor2tensor mode), or they happen after the residuals (aiayn mode). You're saying there's a third version used in XLM? Where there's just one extra LN around the embeddings, but otherwise it's normal? (cc @glample).

parlai/core/torch_ranker_agent.py
```diff
@@ -65,6 +65,19 @@ def add_cmdline_args(cls, argparser):
...
+        agent.add_argument(
+            '--embedding-normalization', type='bool', default=False,
+            help='normalization after embeddings for transformer. True or False'
+        )
+        agent.add_argument(
+            '--gelu-activation', type='bool', default=False,
+            help='if True use gelu instead of Relu in the FFN. is False use relu'
+        )
+        agent.add_argument(
+            '--reduction-type', type=str, default=None,
```
stephenroller: seems like we should have mean be the default, not None. Additionally, None is not a choice that a user can actually pass in via the command line.

parlai/core/torch_ranker_agent.py
```diff
@@ -65,6 +65,19 @@ def add_cmdline_args(cls, argparser):
             '--ignore-bad-candidates', type='bool', default=False,
             help='Ignore examples for which the label is not present in the '
                  'label candidates. Default behavior results in RuntimeError. ')
+        agent.add_argument(
+            '--embedding-normalization', type='bool', default=False,
+            help='normalization after embeddings for transformer. True or False'
+        )
+        agent.add_argument(
```
stephenroller: gelu and normalization should be part of both the ranker and the generator, please!

parlai/agents/transformer/modules.py
malachaux: I do not use the fairseq implementation but the XLM implemntation ( same as ?BERT) which is here : https://github.com/facebookresearch/XLM

parlai/core/torch_ranker_agent.py
malachaux: I put this because maybe a user doesn't want to have any reduction ie --reduction = False, and --reduction is None ? you see what I mean ?

parlai/agents/transformer/modules.py
samhumeau: We could check here if --reduction is False and set reduction-type to None if that's the case?

parlai/core/torch_ranker_agent.py
samhumeau: If --reduction = False, could we could --reduction-type as None, cf my comment above?

parlai/agents/transformer/modules.py
```diff
@@ -288,6 +322,10 @@ def __init__(
         else:
             nn.init.normal_(self.position_embeddings.weight, 0, embedding_size ** -0.5)
 
+        #embedding normalization
```
samhumeau: lint error here. You can run 
`flake8 path/to/my/dir` in ParlAI to see all of them.

parlai/agents/transformer/modules.py
stephenroller: seems like we can remove the boolean and just have it be a fixed set. `{'none', 'mean', ...}`

parlai/agents/transformer/modules.py
stephenroller: Good, thanks for clarifying that.

I still would like to have some sort of `--variant {'aiayn', 'xlm'}` arg instead of `--embedding-normalization`

parlai/core/torch_ranker_agent.py
stephenroller: Just use 'none' instead of None

parlai/agents/transformer/modules.py
stephenroller: Or `bash tests/lint_changed.sh` to run the exact thing that CI runs

parlai/agents/transformer/modules.py
```diff
@@ -194,7 +214,7 @@ def __init__(self, transformer, hdim):
         self.transformer = transformer
         self.mlp = nn.Sequential(
             nn.Linear(dim, hdim),
-            nn.ReLU(),
+            nn.ReLU(),  # TODO: should this also be gelu?
```
stephenroller: @emilydinan  opinions?

parlai/agents/transformer/modules.py
```diff
@@ -42,13 +46,16 @@ def _build_encoder(opt, dictionary, embedding=None, padding_idx=None, reduction=
         padding_idx=padding_idx,
         learn_positional_embeddings=opt.get('learn_positional_embeddings', False),
```
emilydinan: why is this an opt.get? can we change it to opt['learn_positional_embeddings'] for consistency with the other args?

parlai/agents/transformer/modules.py
```diff
@@ -42,13 +46,16 @@ def _build_encoder(opt, dictionary, embedding=None, padding_idx=None, reduction=
...
         embeddings_scale=opt['embeddings_scale'],
-        reduction=reduction,
+        reduction_type=reduction_type,
         n_positions=n_positions,
+        n_segments=n_segments,
+        activation=opt['activation'],
```
emilydinan: related question -- is this going to break backwards compatiibilty? i wonder if we'll see opt loading issues here ...

parlai/agents/transformer/modules.py
```diff
@@ -98,9 +112,15 @@ def __init__(self, opt, dictionary):
         if n_positions < 0:
             raise ValueError('n_positions must be positive')
 
+        self.activation = opt.get('activation', 'relu')
```
emilydinan: not sure if we need these (activation, variant ... ) since these are passed into `build_encoder` via opt

parlai/agents/transformer/modules.py
```diff
@@ -295,34 +341,55 @@ def __init__(
...
+        if self.reduction_type == 'first':
+            return tensor[:, 0, :]
+        elif self.reduction_type == 'max':
+            return tensor.max(dim=1)[0]
+        elif self.reduction_type == 'mean':
+            divisor = mask.float().sum(dim=1).unsqueeze(-1).clamp(min=1e-20)
             output = tensor.sum(dim=1) / divisor
             return output
-        else:
+        elif self.reduction_type == 'none' or self.reduction_type is None:
```
emilydinan: 'none' is kinda weird, what is this here for? is this legacy stuff?

parlai/agents/transformer/modules.py
```diff
@@ -334,17 +401,23 @@ def __init__(
...
+        self.activation = activation
+        self.variant = variant
         self.attention = MultiHeadAttention(
             n_heads, embedding_size,
             dropout=attention_dropout,  # --attention-dropout
         )
-        self.norm1 = nn.LayerNorm(embedding_size)
-        self.ffn = TransformerFFN(embedding_size, ffn_size, relu_dropout=relu_dropout)
-        self.norm2 = nn.LayerNorm(embedding_size)
+        self.norm1 = nn.LayerNorm(embedding_size, eps=LAYER_NORM_EPS)
```
emilydinan: why did you change the epsilon here?

tests/test_transformers.py
```diff
@@ -120,6 +120,71 @@ def test_backcomp(self):
...
+            "valid hits@1 = {}\nLOG:\n{}".format(valid['hits@1'], stdout)
+        )
+        self.assertGreaterEqual(
+            test['hits@1'],
+            0.90,
+            "test hits@1 = {}\nLOG:\n{}".format(test['hits@1'], stdout)
+        )
+
+    @testing_utils.retry(ntries=3)
+    def test_alt_reduction(self):
```
emilydinan: test name is a bit unclear if we're specifically testing reduction_type == 'first'

parlai/agents/transformer/transformer.py
```diff
@@ -41,6 +41,14 @@ def add_common_cmdline_args(argparser):
     argparser.add_argument('--n-positions', type=int, default=None, hidden=True,
                            help='Number of positional embeddings to learn. Defaults '
                                 'to truncate or 1024 if not provided.')
+    argparser.add_argument('--n-segments', type=int, default=0,
+                           help='The number of segments that support the model. '
+                                'If zero no segment and no langs_embedding.')
+    argparser.add_argument('--variant', choices={'aiayn', 'xlm'}, default='aiayn',
+                           help='Chooses locations of layer norms, etc.')
+    argparser.add_argument('--activation', choices={'relu', 'gelu'}, default='relu',
```
emilydinan: nice these comments are helpful

parlai/agents/transformer/modules.py
emilydinan: hmmm... good question. would have to try some experiments, but maybe we can pass teh activation in as an arg

parlai/agents/transformer/modules.py
stephenroller: So the main None is actually what the generative model needs to ensure there's no reduction. The `'none'` was just imagining that it *could* be passed as something... eventually.

parlai/agents/transformer/modules.py
stephenroller: because it was specified in some places and not in others. I figure it's better to use a consistent value everywhere.

tests/test_transformers.py
stephenroller: It was meant to be a test that just tries something /other/ than `mean`. `max` we know doesn't work that great.

parlai/agents/transformer/modules.py
stephenroller: Indeed we do not.

parlai/agents/transformer/modules.py
stephenroller: so new runs should have the default arg which is `'relu'`.

stephenroller: It's worth noting we have a unit test for transformer backwards compatibility.

emilydinan: cool

PR #1453, opened 2019-02-13
fix interactive for personachat seq2seq

alexholdenmiller:
Fixes #1430. I think got a transient error but I got these errors when I ran that script myself

Issue #1430, opened 2019-02-07
error in running seq2seq_withprofile_interactive.py


mohamad-mehdi-jafari:

mohamad-mehdi-jafari:
I encountered the following error after download and store pre_trained models

python projects/personachat/scripts/seq2seq_withprofile_interactive.py 
[ optional arguments: ] 
[  display_examples: False ]
[ Main ParlAI Arguments: ] 
[  datapath: /home/mehdi/git/parlai/data ]
[  datatype: train ]
[  download_path: /home/mehdi/git/parlai/downloads ]
[  hide_labels: False ]
[  image_mode: raw ]
[  numthreads: 1 ]
[  show_advanced_args: False ]
[  task: parlai.agents.local_human.local_human:LocalHumanAgent ]
[ Batching Arguments: ] 
[  batchsize: 1 ]
[  include_labels: True ]
[ PytorchData Arguments: ] 
[  batch_length_range: 5 ]
[  batch_sort_cache_type: pop ]
[  batch_sort_field: text ]
[  numworkers: 4 ]
[  pytorch_context_length: -1 ]
[  pytorch_datapath: None ]
[  pytorch_include_labels: True ]
[  pytorch_preprocess: False ]
[  pytorch_teacher_batch_sort: False ]
[  pytorch_teacher_dataset: None ]
[  pytorch_teacher_task: None ]
[  shuffle: False ]
[ ParlAI Model Arguments: ] 
[  dict_class: None ]
[  model: projects.personachat.persona_seq2seq:PersonachatSeqseqAgentBasic ]
[  model_file: /home/mehdi/git/parlai/data/models/personachat/seq2seq_personachat/seq2seq_no_dropout0.2_lstm_1024_1e-3 ]
[ ParlAI Image Preprocessing Arguments: ] 
[  image_cropsize: 224 ]
[  image_size: 256 ]
[ Local Human Arguments: ] 
[  local_human_candidates_file: None ]
[  single_turn: False ]
[ Dictionary Arguments: ] 
[  bpe_debug: False ]
[  dict_endtoken: __end__ ]
[  dict_file: /home/mehdi/git/parlai/data/models/personachat/profile_memory/fulldict.dict ]
[  dict_initpath: None ]
[  dict_language: english ]
[  dict_lower: False ]
[  dict_max_ngram_size: -1 ]
[  dict_maxtokens: -1 ]
[  dict_minfreq: 0 ]
[...]

PR #1581, opened 2019-03-25
[Personachat] Deprecate profile memory models

emilydinan:
The profile memory model contained many bugs (see for example #1453) so we are removing it from the codebase. See also #1430 and comments on #662.


emilydinan:
Unit test failure is unrelated to this diff

PR #1582, opened 2019-03-25
[Torch Ranker Agent] print only relevant metrics

emilydinan:
Print train_accuracy and mean_rank only when these metrics are calculated.

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -87,7 +87,6 @@ class TorchRankerAgent(TorchAgent):
             else:
                 states = {}
 
-        self.is_training = False  # track whether model is training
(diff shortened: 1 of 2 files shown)


parlai/core/torch_ranker_agent.py
```diff
@@ -226,6 +228,7 @@ def train_step(self, batch):
 
     def eval_step(self, batch):
         """Evaluate a single batch of examples."""
+        self.is_training = False
```
stephenroller: you could do all this in batch_act of TA and then we'd have it in TGA too

emilydinan: yes good idea i'll move it up a level in case we need this info elsewhere

PR #1584, opened 2019-03-26
nightly check for Bert

samhumeau:
Adding a nightly test for Bert, train for about 100 samples on convai2 for bi and cross ranker.

I wasn't sure how to keep the final evaluation of the train loop under a certain number of samples, 
so I created this debug LimitedSelfOriginalTeacher, that can be used for such occasions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,6 +174,7 @@ workflows:
       - lint
       - datatests
       - unittests
+      - nightly_gpu_tests
       - deploy_website:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,11 +80,11 @@ commands:
           working_directory: ~/ParlAI/parlai/mturk
           command: npm install
 
-  installhfbert:
-    description: "Install Hugging Face version of Bert"
+  installgpudeps:
+    description: "Install Dependencies specific to GPU testing"
     steps:
       - run:
-          name: Install Bert
+          name: GPU dependencies
           command: |
             pip3 install --progress-bar off pytorch-pretrained-bert
 
@@ -137,7 +137,7 @@ jobs:
       - setup
       - installtorchgpu
       - installdeps
-      - installhfbert
+      - installgpudeps
       - run:
           name: Nightly GPU tests
           command: python setup.py test -s tests.suites.nightly_gpu -q
@@ -183,7 +183,6 @@ workflows:
       - lint
       - datatests
       - unittests
-      - nightly_gpu_tests
       - deploy_website:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,6 +80,14 @@ commands:
           working_directory: ~/ParlAI/parlai/mturk
           command: npm install
 
+  installhfbert:
+    description: "Install Hugging Face version of Bert"
+    steps:
+      - run:
+          name: Install Bert
+          command: |
+            pip3 install --progress-bar off pytorch-pretrained-bert
+
 jobs:
   datatests:
     executor: standard_cpu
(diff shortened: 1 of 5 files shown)


samhumeau:
Following instructions from dataset test: 
```
.
Ran 1 test in 194.066s

OK

```

samhumeau:
So it remains to be seen if it can run on the GPU that take care of the test.
On my machine where I have access to a 16GB GPU, nvidia-smi tells me that this test uses 13GB, but I am not sure how much of that is actually used.

tests/nightly/gpu/test_bert.py
```diff
@@ -0,0 +1,60 @@
...
+            warmup_updates=100,
+        ))
+        # The cross encoder reaches an interesting state MUCH faster
+        # accuracy should be present and somewhere between 0.2 and 0.8
+        # (large interval so that it doesn't flake.)
+        self.assertGreaterEqual(
+            test['accuracy'], 0.2,
+            'test accuracy = {}\nLOG:\n{}'.format(test['accuracy'], stdout)
+        )
+        self.assertLessEqual(
```
stephenroller: I think we can get rid of the lessEqual constraints. Mostly I think we can just ensure this doesn't crash.

parlai/tasks/convai2/agents.py
```diff
@@ -69,6 +69,18 @@ def __init__(self, opt, shared=None):
         super().__init__(opt, shared)
 
 
+class LimitedSelfOriginalTeacher(SelfOriginalTeacher):
```
stephenroller: can we just do --num-epochs .001 instead of making a new teacher?

stephenroller:
How long does the test take?

If you add "- nightly_gpu_tests" here, the test will run on this branch:

'https://github.com/facebookresearch/ParlAI/blob/b412c4d88298a7c531eaab9b5e105ab5588f3a87/.circleci/config.yml#L176

tests/nightly/gpu/test_bert.py
```diff
@@ -0,0 +1,60 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+import parlai.core.testing_utils as testing_utils
+
+
+@testing_utils.skipUnlessGPU
+class TestBertModel(unittest.TestCase):
+    """Checks that Both Biencoder and CrossEncoder of Bert can be trained
```
stephenroller: Do you mind editing the docstring whitespace to be more like:

https://github.com/facebookresearch/ParlAI/blob/b412c4d88298a7c531eaab9b5e105ab5588f3a87/parlai/core/torch_agent.py#L1468-L1474

parlai/tasks/convai2/agents.py
samhumeau: This does not seem to apply for the final eval of the train loop (on valid and test set) does it?

tests/nightly/gpu/test_bert.py
samhumeau: I'd leave it just to make sure no bug produces an accuracy of 1.0, but maybe that's an overstretch

samhumeau:
The tests took less than a minute on my machine

parlai/tasks/convai2/agents.py
stephenroller: Oh jeez the damn validation; you're right.

stephenroller: I’m going to make a flag to help us with that. I ran into this exact issue two weeks ago

tests/nightly/gpu/test_bert.py
stephenroller: oh good idea

.circleci/config.yml
```diff
@@ -129,6 +137,7 @@ jobs:
       - setup
       - installtorchgpu
       - installdeps
+      - installhfbert
```
stephenroller: maybe consider renaming this to "installgpudeps" or something

parlai/tasks/convai2/agents.py
stephenroller: cf #1606

PR #1586, opened 2019-03-26
[WoW] retrieval zoo fixes/updates

emilydinan:
- bug in retrieval model download path (passing in opt, not datapath)
- update readme
- update zoo list


emilydinan:
> Any chance you could make an integration test :)

done, going to land now

stephenroller:
Does your test really need all those extra options?

emilydinan:
> Does your test really need all those extra options?

yes 😢

PR #1589, opened 2019-03-28
[Wizard] zoo fix

emilydinan:
Needed to rename zoo file for wizard model.

Tests should pass now.


.circleci/config.yml
```diff
@@ -176,6 +176,7 @@ workflows:
       - lint
       - datatests
       - unittests
+      - nightly_gpu_tests
```
stephenroller: just remember to undo these at the end

stephenroller:
wow somehow that didn't do the trick?

PR #1596, opened 2019-04-01
[wip] Hybrid DataParallel/DDP

stephenroller:
Work in progress, you can ignore for now.

DistributedDataParallel has a mode where it acts like a DataParallel locally, but still broadcast gradients across all workers. This allows for a sort of "hybrid" training, where a local machine can use all GPUs for `-cands batch` negative samples, but still get the speedup from using multiple workers. Also can be run locally on one machine where it just behaves like DataParallel, getting rid of the need for having both code paths.

This is a "simpler" solution to the idea proposed by #1416. I suspect it will work well since individuals have reported there can be a sweet spot with number of negative candidates.

Issues:
- [x] DDP in this mode is broken internal fairseq env 20190211 (hard crashes). [cc @myleott] Fixed in torch stable (1.0.1post2).
- [ ] Cannot recover from a CUDA OOM. Need to file a bug with pytorch. For some reason, catching the RuntimeError doesn't clear any CUDA memory. This results in the dummy pass also OOMing, making the error unrecoverable. Frustratingly, this happens even when using only a world size of 1.
- [x] Make a minimum test case. File a pytorch bug.

### DIFF
--- a/collect_env.py
+++ b/collect_env.py
@@ -1,0 +1,386 @@
+# This script outputs relevant system environment info
+# Run it with `python collect_env.py`.
+from __future__ import absolute_import, division, print_function, unicode_literals
+import re
+import subprocess
+import sys
+import os
+from collections import namedtuple
+
+try:
+    import torch
+    TORCH_AVAILABLE = True
+except (ImportError, NameError, AttributeError):
+    TORCH_AVAILABLE = False
+
+PY3 = sys.version_info >= (3, 0)
+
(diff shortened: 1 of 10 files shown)


stephenroller:
Ugh, looks like you can't mark PRs is draft once you accidentally make them non-draft.

stephenroller:
Tracking a larger bug in pytorch/pytorch#18853

PR #1599, opened 2019-04-02
clean form field after sending

samhumeau:
Following up a remark from @ethanjperez in #1542 , 
the fields of the form were not clearing themselves after sending an answer.
They should do now.
Screen Shot 2019-04-02 at 11 57 13 AM


ethanjperez:
Thanks for this fix, this is great!

PR #1600, opened 2019-04-02
Fix Dict Saving

klshuster:
Dict saving used to iterate over the length of the `ind2tok` dictionary; however, this fails if certain tokens are added at different indices (e.g. `CLS` in bert is `101`). We now iterate over the dict keys


klshuster:
note: tested locally on bert tests to ensure they pass

samhumeau:
Thanks a lot Kurt!

PR #1602, opened 2019-04-02
[MTurk] Re-enabling socket tests on circle

JackUrb:
The long running mystery of why socket tests were failing is now solved - sockets were crashing because the wrong requirement was being downloaded from `requirements.txt`. This specifies the correct requirements, so we should be back to good.


stephenroller:
lgtm, let's land!

PR #1605, opened 2019-04-02
Fix nightly CPU tests

stephenroller:
- Ensures nightly CPU tests are given plenty of timeout
- Enables Teeing the captured stdout, which is very helpful when debugging a particular test. Sam and I had discussed this at one point
- Fixup the pytorch test so it runs much faster and has a looser requirement for "task solved".
- Realized seq2seq tests weren't using the LR parameter properly. NBD because the default was the same, but should fix this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -154,6 +154,7 @@ jobs:
       - installtorchcpu
       - run:
           name: All nightly CPU tests
+          no_output_timeout: 30m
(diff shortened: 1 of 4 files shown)


stephenroller:
If you set DEBUG to True, the output will NOT be suppressed. We also store it in the StringIO so that tests can still access that output.

https://en.wikipedia.org/wiki/Tee_(command)

PR #1609, opened 2019-04-03
[circleci] Build website step (&& squash with master.)

stephenroller:
- Add a step that ensures the website builds okay in CircleCI.
- Fix a bug that should've been done in #1607 (shield showing on About page).
- Remove a redundant numpy install. It's already in requirements.
- Get CircleCI to merge with master before running tests if it can.
- Get rid of some warnings in the website generation

PR #1610, opened 2019-04-03
[MTurk] Adding more support for 1 person static tasks.

JackUrb:
## Motivation
Long term, I think that ParlAI-MTurk can support pretty much any crowdsourcing use case while leveraging the same UI and tooling for worker and task management. Initially it was built for the more complicated use cases of connecting multiple workers together or having a worker in direct communication with a local model. From here I'll define these as tasks that have _dynamic_ action flows, in that the task needs to communicate back with the server at multiple time steps to make the task completable. The other task type would be _static_ tasks, or tasks where everything in order to complete the task can be supplied up-front and all the task logic can exist on the frontend. This PR takes the first step towards making this type of task easy to create in ParlAI.

## Implementation
The basic idea of this PR is that we can strip away a lot of the complicated worker state tracking out of MTurkManager and related classes, and then fork the frontend code to create a version that isn't in constant communication with the server. This new version will have two important points of communication. The first is the initial connect, which proceeds as would in the dynamic case, except when a worker alives, if they are determined eligible we put them directly into the task pool without checking state. The regular flow then starts up a world for them, which selects and sends the related `task_data` to the frontend as a regular act message. At this point communication across the socket is suspended (no alives or other monitoring). The second point is at task completion, when a worker clicks the "Done With This HIT" button, the task data is packaged up and sent as a post to the router. The router then mimics the behavior of what would be expected for sending an act as a packet from the frontend, thus allowing it to be delivered as an act from the `MTurkAgent` in the world.

As an example, this PR also includes a *very bad* demo that I don't intend to commit in the current form. It generally shows the flow of how a static task only needs to define a flow for creating the relevant `task_data` for an assignment and the frontend for the task, but otherwise doesn't need to deal with a `worlds` file at all. In a refactor this can be simplified to an API that is simply `get_task_data` and `return_task_data`, removing almost all of the boilerplate for static tasks.
**EDIT** The demo has been removed to replaced with a better one when I get v1 of this functionality implemented.

## Detailed changes
`mturk_utils`
- Adds a parameter that allows for selection of the template that will be rendered in the frontend, either static or dynamic.

`react_server/app.jsx`
- Creates a new `StaticApp` class, which essentially is a stripped down version of the regular app.
- Key method difference is `onValidData()` which is responsible for updating whether or not the task is done. The validation responsibility is now on the frontend as we have no backend connections.

`core_components.jsx`
- Creates a series of static versions of components intent on replacing the right pane of the existing template with arbitrary content. These mostly exist as piping to be able to deliver arbitrary content to that pane. With refactoring this will clean up to remove unnecessary features (such as volume) but I wanted a starting point.

`mturk_submit_form.jsx`
- Adds the required logic to be able to have a submit that first POSTs the required data to the router. Also includes it as HIT data to submit <- ideally we replace this with *only* submitting the response data to Amazon if the POST to our server fails, as we don't want to prevent a worker from being able to submit work when we run into server issues, but we also don't want to lose data.

`socket_handler.jsx`
- Fixes a small ACK bug that would cause certain messages to be sent forever in a loop. We shouldn't be overriding something we've marked as ACK to SENT ever.
- creates a run_static flag that, when enabled, makes it so the socket handler doesn't send heartbeats, and doesn't treat task_data as special (as all tasks will communicate through task_data)

`server.js`
- Creates the endpoint for packaging post data and simulating an act message from it
- Forwards the templating variable defined in `mturk_utils`

`index.html`
- Adds the templating variable for determining which frontend to render

`StaticSocketManager`
- new class overrides channel management to ensure we're not sending heartbeats or checking liveliness in any way really. In a refactor I could even do away with separate sending queues entirely, as we only expect two outgoing messages and one incoming message.
- no liveliness instantiation in open_channel

`StaticMTurkTaskWorld`
- Exists as a new base world that abstracts the process of observing the task_data to begin a task, and then waiting for response data into a proper ParlAI world. 

`StaticMTurkManager`
- New class overrides certain opts to ensure things work for the static case
- `setup_socket` overriden to use a `StaticSocketManager`
- `on_alive` is a much simplified form of the original version that puts almost no effort into state tracking because we don't actually need to manage the state of a worker. They're either working on a task or done. (This can probably be cleaned up, but makes more sense to do in a refactor).

`image_captions_demo`
- creates some demo files that show the path of creating task data on the python side and displaying that data as well as handling validation on the frontend. Arbitrarily complex react components can have arbitrarily complex task flows.
**EDIT** The demo has been removed to be re-added in a future PR when this is more final

## Demo Video
demo_vid

## Important notes and future work
- I did test a normal task, dynamic tasks appear to be working just fine 
- Right now response_data is being submitted to amazon's HIT data servers as a backup in all cases. I'd like to make it so this *only* ever happens if the submission to ParlAI fails. There's also a lot of behavior that will need to be addressed with respect to syncing state in the case of a dropped message
- This doesn't actually track worker completion time in any way other than listening for explicit abandon/return events, so I expect it to be possible for a HIT to be completed but the world never cleans up. Will need to add logic that reaps a task after the HIT duration
- This *in no way* has been tested on the webapp. I expect that the infrastructure will be able to be extended pretty reasonably to support viewing static tasks inline in the same way I have it working for dynamic tasks, but I haven't investigated this yet at all.


parlai/mturk/core/mturk_manager.py
```diff
@@ -1859,3 +1862,186 @@ def email_worker(self, worker_id, subject, message_text):
...
+    workflow, and more, but only for tasks that require just 2 connections
+    to the server: an initial task request and the submission of results
+    """
+
+    def __init__(self, opt, is_test=False):
+        """No interaction means only ever one agent, so that's what we get"""
+        opt['max_connections'] = 0  # Max connections doesn't make sense here
+        opt['count_complete'] = True  # No other way to count static HITs
+        opt['frontend_template_type'] = 'static'
+        super().__init__(opt, ['worker'], is_test, True)
```
klshuster: is `True` for a kwarg? Not really sure what this does here

parlai/mturk/core/mturk_manager.py
```diff
@@ -1859,3 +1862,186 @@ def email_worker(self, worker_id, subject, message_text):
...
+        self.hit_mult = 1  # No need to pad HITs if they're static
+        base_required_hits = self.num_conversations * len(self.mturk_agent_ids)
```
klshuster: assuming this will always be equal to `self.num_conversations` as `len(self.mturk_agent_ids) == 1`

stephenroller:
Pull from master if you get a chance.

parlai/mturk/core/mturk_manager.py
```diff
@@ -1071,14 +1073,15 @@ def start_new_run(self):
...
-        except Exception:
+        except Exception as e:
             self.topic_arn = None
             shared_utils.print_and_log(
                 logging.WARN,
                 'Botocore couldn\'t subscribe to HIT events, '
                 'perhaps you tried to register to localhost?',
                 should_print=True
             )
+            print(repr(e))
```
stephenroller: is this redundant

parlai/mturk/core/mturk_manager.py
```diff
@@ -1859,3 +1862,186 @@ def email_worker(self, worker_id, subject, message_text):
...
+                # no information is held, and add to the pool if not already in
+                # the pool
+                if not conversation_id:
+                    self._restore_agent_state(worker_id, assign_id)
+            elif agent.get_status() == AssignState.STATUS_IN_TASK:
+                # Reconnecting to the onboarding world or to a task world
+                # should resend the messages already in the conversation
+                if not conversation_id:
+                    self._restore_agent_state(worker_id, assign_id)
+            elif (agent.get_status() == AssignState.STATUS_DISCONNECT or
```
stephenroller: this or statement might be good as a helper

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -235,6 +238,170 @@ class MainApp extends React.Component {
...
+          onRequestMessage={() => {}}
+          onTaskDone={() => {}}
+          onInactiveDone={inactive_text =>
+            this.setState({
+              task_done: true,
+              chat_state: 'done',
+              done_text: inactive_text,
+            })
+          }
+          onForceDone={allDoneCallback}
```
klshuster: why isnt this `staticAllDoneCallback`?

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -235,6 +238,170 @@ class MainApp extends React.Component {
...
+// prevents this class from setting a whole lot of dummy methods
+class StaticApp extends React.Component {
+  constructor(props) {
+    super(props);
+    let initialization_status = 'initializing';
+    if (!doesSupportWebsockets()) {
+      initialization_status = 'websockets_failure';
+    }
+
+    // TODO move constants to props rather than state
```
stephenroller: this looks like you haven't finalized it

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -235,6 +238,170 @@ class MainApp extends React.Component {
...
+    if (isMobile() && BLOCK_MOBILE) {
+      task_description =
+        '<h1>Sorry, this task cannot be completed on mobile devices. ' +
+        'Please use a computer.</h1><br>Task Description follows:<br>' +
+        data['task_description'];
+    }
+
+    this.setState({
+      task_description: task_description,
+      frame_height: data['frame_height'] || 650,
```
stephenroller: magic constant?

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -235,6 +238,170 @@ class MainApp extends React.Component {
...
+  }
+
+  onValidData(valid, response_data) {
+    this.setState({task_done: valid, response_data: response_data});
+  }
+
+  render() {
+    let socket_handler = null;
+    if (!this.state.is_cover_page) {
+      socket_handler = (
```
stephenroller: this seems like it might be good to refactor into a helper method, this is a really long if

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -1326,6 +1364,21 @@ class ContentLayout extends React.Component {
   }
 }
 
+class StaticContentLayout extends React.Component {
+  render() {
+    let layout_style = '2-PANEL'; // Currently the only layout style is 2 panel
```
stephenroller: should this be an enum?

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -1368,27 +1421,75 @@ class BaseFrontend extends React.Component {
...
+          <XLeftPane {...this.props} />
+        </div>
+      );
+    } else if (this.props.initialization_status == 'initializing') {
+      content = <div id="ui-placeholder">Initializing...</div>;
+    } else if (this.props.initialization_status == 'websockets_failure') {
+      content = (
+        <div id="ui-placeholder">
+          Sorry, but we found that your browser does not support WebSockets.
+          Please consider updating your browser to a newer version and check
```
klshuster: should we also recommend switching browsers?

parlai/mturk/core/socket_manager.py
```diff
@@ -718,3 +717,73 @@ def shutdown(self):
...
+                            if not packet:
+                                # This packet was deleted out from under us
+                                continue
+                            if packet.status is not Packet.STATUS_ACK:
+                                # either need to send initial packet
+                                # or resend not-acked packet
+                                self._send_packet(packet, connection_id, t)
+                    except Empty:
+                        pass
+                except BaseException as e:
```
stephenroller: why baseexception, why capture e?

parlai/mturk/tasks/image_captions_demo/run.py
```diff
@@ -0,0 +1,90 @@
...
+
+            return world.prep_save_data(workers)
+
+        mturk_manager.start_task(
+            eligibility_function=check_worker_eligibility,
+            assign_role_function=assign_worker_roles,
+            task_function=run_conversation
+        )
+
+    except BaseException:
```
stephenroller: I still don't like BaseException instead of Exception

random suggestion: 

have you considered using withs? https://effbot.org/zone/python-with-statement.htm

```python
with mturk_manager:
    pass
```
which would have:
```python
def __exit__(self, type, value, tradeback):
   self.expire_all_unassigned_hits()
   self.shutdown()
```

Just a random thought for future stuff.

stephenroller: (at the very least, it seems like you should do a try-finally instead of a try-catch-finally)

parlai/mturk/tasks/image_captions_demo/run.py
```diff
@@ -0,0 +1,90 @@
...
+
+        def check_worker_eligibility(worker):
+            return True
+
+        def assign_worker_roles(workers):
+            for w in workers:
+                w.id = 'worker'
+
+        def run_conversation(mturk_manager, opt, workers):
+            image = '/Users/jju/ParlAI/data/yfcc_images/d06b45856ea94a4a26aa112f79a292e2.jpg'
```
klshuster: perhaps put another image here, this could be copyright protected

stephenroller:
Looks like #1609 broke travis, I'll put a fix.

parlai/mturk/core/mturk_manager.py
JackUrb: If referring to the `repr` call, `repr` actually gives more context on the error object than its tostring does. If referring to a print after the print and log, it's actually just in here for debug and I should remove before merging.

parlai/mturk/core/mturk_manager.py
JackUrb: True - this was copypasta

parlai/mturk/core/mturk_manager.py
JackUrb: I actually already have this as a helper (`agent.is_final()`), but I didn't want to obfuscate the switch-case like flow that this if block is demonstrating. This logic should be reduced in the v1 implementation down the road anyways though.

parlai/mturk/core/react_server/dev/app.jsx
JackUrb: I'm not sure I follow. This and many TODOs are due to the fact that this feature is a v0 baseline implementation and I want pointers that let me remember what I have to fix in future work

parlai/mturk/core/react_server/dev/app.jsx
JackUrb: Moving

parlai/mturk/core/react_server/dev/app.jsx
JackUrb: This whole thing is going to be replaced in v1,  and really it's a single assign statement, but otherwise agreed.

parlai/mturk/core/react_server/dev/app.jsx
JackUrb: good catch! Really it should be for completeness in v0, but ultimately forceDone should never be called by a static world as we gate the actions that are sent from a static run.

parlai/mturk/core/react_server/dev/components/core_components.jsx
JackUrb: It should be, and eventually it will be once we support additional layouts. This is copied directly from the existing component.

parlai/mturk/core/react_server/dev/components/core_components.jsx
JackUrb: Sure I'll put that in here

parlai/mturk/core/socket_manager.py
JackUrb: No particular reason other than inheritance from old code. Should be replaced with Exception. We want to capture though because sockets can have a varied series of errors and if this thread dies the whole HIT dies too. I really should contain the scope of this at some point, but haven't had the bandwidth.

parlai/mturk/tasks/image_captions_demo/run.py
JackUrb: Oh definitely - as noted in the PR description I'm going to be modifying this task heavily. Probably will make it pull the ParlAI logo.

stephenroller:
Fix in #1611 landed.

parlai/mturk/tasks/image_captions_demo/run.py
JackUrb: In this case, `BaseException` is required to catch `KeyboardInterrupt` and others. If the `with` syntax is able to manage that I can switch to it, but I just hadn't tried that kind of syntax before. 

To some degree this discussion belongs elsewhere though as this run file is going to be cut down quite a lot in v1 when it's finally committed.

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -1368,27 +1421,75 @@ class BaseFrontend extends React.Component {
...
+  XDoneResponse: ['XDoneResponse', DoneResponse],
+  XIdleResponse: ['XIdleResponse', IdleResponse],
+  XDoneButton: ['XDoneButton', DoneButton],
+  XChatPane: ['XChatPane', ChatPane],
+  XWaitingMessage: ['XWaitingMessage', WaitingMessage],
+  XMessageList: ['XMessageList', MessageList],
+  XChatMessage: ['XChatMessage', ChatMessage],
   XTaskDescription: ['XTaskDescription', TaskDescription],
   XReviewButtons: ['XReviewButtons', ReviewButtons],
-  XContextView: ['XContextView', ContextView],
```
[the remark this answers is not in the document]
JackUrb: Good catch, this shouldn't've been removed. No tasks use it yet, but it's still a mistake

Issue #1594, opened 2019-04-01
Wikipedia Retriever missing dictionary


JohannesTK:
Both [Wikipedia Retriever (Used For Open Squad)
](http://parl.ai.s3-website.us-east-2.amazonaws.com/docs/zoo.html#wikipedia-retriever-used-for-open-squad) and [Wikipedia Retriever (Used For Wizard Of Wikipedia)
](http://parl.ai.s3-website.us-east-2.amazonaws.com/docs/zoo.html#wizard-of-wikipedia-end-to-end-generator) seem to be missing their dictionaries. 

```
python -m parlai.scripts.interactive --model tfidf_retriever -mf models:wikipedia_2016-12-21/tfidf_retriever/drqa_docs
[building data: /home/ubuntu/ParlAI/data/models/wikipedia_2016-12-21/tfidf_retriever/drqa_docs.tgz]
[ downloading: http://parl.ai/downloads/_models/wikipedia_2016-12-21/tfidf_retriever/drqa_docs.tgz to /home/ubuntu/ParlAI/data/models/wikipedia_2016-12-21/tfidf_retriever/drqa_docs.tgz ]
Downloading drqa_docs.tgz: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 7.72G/7.72G [05:06<00:00, 25.2MB/s]
unpacking drqa_docs.tgz
[ warning: overriding opt['model_file'] to /home/ubuntu/ParlAI/data/models/wikipedia_2016-12-21/tfidf_retriever/drqa_docs (previously: wiki_full_notitle )]
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/envs/pytorch_p36/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/ubuntu/anaconda3/envs/pytorch_p36/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/ubuntu/ParlAI/parlai/scripts/interactive.py", line 76, in <module>
    interactive(parser.parse_args(print_args=False), print_parser=parser)
  File "/home/ubuntu/ParlAI/parlai/scripts/interactive.py", line 54, in interactive
    agent = create_agent(opt, requireModelExists=True)
  File "/home/ubuntu/ParlAI/parlai/core/agents.py", line 552, in create_agent
    model = load_agent_module(opt)
[...]

PR #1613, opened 2019-04-05
Missing dict file

emilydinan:
Related to #1594.

Let's (loudly) warn instead of raising a RuntimeError, some models do not require us building a dictionary (like the TFIDF retriever and BERT models).

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -38,6 +38,7 @@ This module also provides a utility method:
 """
 
 from parlai.core.build_data import modelzoo_path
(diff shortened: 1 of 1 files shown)


stephenroller:
I think it'd be nice for such agents to have a way to indicate they don't need a dict. (`dictionary_class` returns null?)

I think this is a reasonable solution though. People will still get the embedding mismatch though.

emilydinan:
@stephenroller i had the same thought-- i am happy to change the PR now

PR #1631, opened 2019-04-11
Creating duplicates of mturk core for legacy and dev branching

JackUrb:
I'm about to make some breaking changes to MTurk. To handle this and allow future breaking changes if necessary, but also to offer stability to people who want to work on a stable branch, I'll be moving to a 3 branch structure. `legacy` will have the core code as it exists in this moment. `stable` will eventually have the next iteration of the platform which I'll be working on in `dev` for the upcoming few weeks. Then `dev` is where I can work on things that are partially complete. This PR just introduces the file hierarchy so that future PRs will be cleaner.

Upcoming PRs for structure before I start working in `dev` on the real changes
- Moving run_data to `parlai/mturk/run_data`
- Moving scripts to `parlai/mturk/scripts` (while persisting the old location for now)
- Creating test runner scripts in `parlai/mturk/test` to run tests properly in the subdirectories of `legacy` and `dev` depending on changed file paths.


parlai/mturk/core/mturk_manager.py
```diff
@@ -14,6 +14,15 @@
 import errno
 import requests
 
+# TODO uncomment once stable is created.
+# logging.warn(
+#     'Directly importing parlai.mturk.core.<module> is pending deprecation, '
+#     'please update your callsites to use either '
+#     'parlai.mturk.core.stable.<module>  '
+#     'or parlai.mturk.core.legacy.<module>. \n'
```
stephenroller: maaaaaaaybe "legacy" is too generic?

JackUrb: It's trying to refer to the legacy version of core vs the stable version of core. I'm not sure what would be better though I get what you're saying. Perhaps `original_impl`?

PR #1634, opened 2019-04-11
[MTurk] Moving scripts directory

JackUrb:
Directory updates to bring scripts to a new location, as they aren't actually part of core


parlai/mturk/core/scripts/auto_complete_hit.py
```diff
@@ -24,6 +24,11 @@
 import time
 import random
 
+print(
+    'This script will be removed from this location May 1st 2019. '
```
stephenroller: go ahead and make an issue?

parlai/mturk/core/scripts/bonus_workers.py
```diff
@@ -7,6 +7,11 @@
 import argparse
 import parlai.mturk.core.mturk_utils as mturk_utils
 
+print(
+    'This script will be removed from this location May 1st 2019. '
```
stephenroller: lol maybe refactor out your deprecation warning?

JackUrb: I'm not sure there are enough cases of this to be worth refactoring out.

PR #1636, opened 2019-04-11
[MTurk] Fixing onboarding test possibly

JackUrb:
Ran this a lot locally, going to try to have it hit circle a few times.

Found the issue - the test only passed on fast hardware due to a race condition where the worker would be removed from the onboarding pool before the onboarding pool expiration message was sent.


JackUrb:
Tests passed 18/18 rebuilds, I think that's probably safe enough

PR #1638, opened 2019-04-12
Newface

kstats:
Merging in new, refactored FACE for both Seq2Seq and transformer


kstats:
Apologies, I meant to merge this into my local fork!

stephenroller:
JFYI ParlAI supports a `parlai_internal` module, which we use for private development while doing our own research. This is a separate git repo that we keep private, and has a similar structure to the external parlai

```
$ ls parlai_internal
agents  __init__.py  messenger  mturk  projects  __pycache__  README.md  tasks
```

You can do things like:

```
python examples/train_model.py -t internal:my_internal_task -m internal:my_internal_agent
```
this will map automatically to parlai_internal/agents/my_internal_agent, etc.

PR #1642, opened 2019-04-12
[MTurk] Using a tmp dir for scratch work. 

JackUrb:
Rather than arbitrarily dumping things in core, like in #1634 and #1637 I'm moving things out to the `mturk` directory. This provides an `mturk/tmp` directory to put temporary heroku files related to launching a server, along with other build artifacts. I'll update the webapp to use this on the next webapp pass I take.

This also adds a command line argument that lets us specify the temporary scratch directory, `--tmp-dir`. Should save us time building on distributed filesystems if we have a local scratch directory and use that instead.

Note you only need to check the 4 files once between dev, legacy_2018, and core. I probably should've made this change before the split!

### DIFF
--- a/parlai/mturk/core/dev/server_utils.py
+++ b/parlai/mturk/core/dev/server_utils.py
@@ -15,11 +15,12 @@ import shlex
 import shutil
 import subprocess
 import time
+import parlai.mturk.core.shared_utils as shared_utils
 
 region_name = 'us-east-1'
 user_name = getpass.getuser()
 
-parent_dir = os.path.dirname(os.path.abspath(__file__))
+parent_dir = shared_utils.get_core_dir()
 legacy_server_source_directory_name = 'server_legacy'
 server_source_directory_name = 'react_server'
 heroku_server_directory_name = 'heroku_server'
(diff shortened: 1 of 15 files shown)


parlai/mturk/core/server_utils.py
```diff
@@ -447,18 +450,18 @@ def setup_heroku_server(task_name, task_files_to_copy=None,
     os.chdir(parent_dir)
```
emilydinan: is this supposed to be tmp_dir?

parlai/mturk/core/dev/server_utils.py
```diff
@@ -447,18 +450,18 @@ def setup_heroku_server(task_name, task_files_to_copy=None,
     os.chdir(parent_dir)
```
emilydinan: same question here

JackUrb: I can actually drop this line and any other `os.chdir`'s as I'm using absolute paths for everything now, but I'll do that in the dev branch cleanup coming soon.

PR #1650, opened 2019-04-19
Pull out the dictionary construction

stephenroller:
Makes a good spot for adding special tokens.


stephenroller:
Done, what do you think?

PR #1651, opened 2019-04-19
Lengthen timeout for nightly tests. [long]

stephenroller:
Just lengthen the timeout before a test is marked as nonresponsive. Hoping this cleans up the cpu nightly test.


JackUrb:
Hm it doesn't seem this branch passed those tests, so I guess it's not enough?

stephenroller:
#1688 was the true fix.

PR #1655, opened 2019-04-19
Add padding mask to Memnn

### LINKED ISSUE #1649: MemNN padding issue
I'm seeing different results with MemNN while varying batch size.  The problem seems to relate to attention and padding.

If I have a memory that gets padded (due to another training sample in its batch being longer), we add 0's to the end of the memory so that their dimensions match.  When we [do attention](https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/memnn/modules.py#L203-L204) we don't treat these paddings any differently which will lead to different results in `probs` after applying the softmax.  For example:

```Python
>>> F.softmax(torch.Tensor([1, 4, 5, 3]))
tensor([0.0120, 0.2418, 0.6572, 0.0889])
>>> F.softmax(torch.Tensor([1, 4, 5, 3, 0, 0, 0]))
tensor([0.0119, 0.2386, 0.6486, 0.0878, 0.0044, 0.0044, 0.0044])
>>>
```

This causes the results to diverge at each hop.  [Fairseq gets around this](https://github.com/pytorch/fairseq/blob/master/fairseq/modules/multihead_attention.py#L186-L189) by setting padded values to `-inf`

```Python
>>> F.softmax(torch.Tensor([1, 4, 5, 3, float('-inf'), float('-inf'), float('-inf')]))
tensor([0.0120, 0.2418, 0.6572, 0.0889, 0.0000, 0.0000, 0.0000])
```

lematt1991:
This allows us to avoid attending to padded values and gives
deterministic results regardless of padding/batchsize

Fixes #1649


parlai/agents/memnn/modules.py
```diff
@@ -188,19 +191,22 @@ def __init__(self, embedding_size, rotate=True):
...
                              for outputs
-
+        :param pad_mask      (bsz x num_mems) optional mask indicating which tokens
+                             correspond to padding
         :returns: (bsz x esz) output state
         """
         # rotate query embeddings
         attn = torch.bmm(query_embs.unsqueeze(1), in_mem_embs).squeeze(1)
+        if pad_mask is not None:
+            attn[pad_mask] = float('-inf')
```
stephenroller: use `parlai.core.utils.neginf`:

https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/utils.py#L50

stephenroller:
Looks like you've got a hogwild deadlock. Last time I had one of those, it was because one of the threads threw an exception.

Edit parlai/core/testing_utils.py to make DEBUG to `True`, then run `CUDA_VISIBLE_DEVICES=0 python tests/test_memnn.py`.

stephenroller:
Edit: I gave the wrong `CUDA_VISIBLE_DEVICES` flag before (but it looks like you figured it out)

parlai/agents/memnn/memnn.py
```diff
@@ -83,7 +83,12 @@ def build_model(self):
 
     def score_candidates(self, batch, cand_vecs, cand_encs=None):
         mems = self._build_mems(batch.memory_vecs)
-        scores = self.model(batch.text_vec, mems, cand_vecs)
+        # Check for rows that have no non-null tokens
+        pad_mask = None
+        if mems is not None:
+            non_null = mems != self.dict.txt2vec(self.dict.null_token)[0]
```
stephenroller: you can use `self.NULL_IDX`

parlai/agents/memnn/memnn.py
```diff
@@ -182,20 +187,20 @@ def _build_mems(self, mems):
...
 
         # NOTE: currently below we are adding tf's to every memory,
         # including emtpy ones. above commented-out code adds only to filled
         # ones but is significantly slower to run.
-        if self.use_time_features:
-            nm = num_mems - 1
-            for i in range(num_mems):
-                # put lowest time feature in most recent memory
-                padded[:, nm - i, -1] = self.dict[self._time_feature(i)]
+        # if self.use_time_features:
```
stephenroller: Delete

parlai/agents/memnn/modules.py
```diff
@@ -61,13 +61,15 @@ def _score(self, output, cands):
...
 
-    def forward(self, xs, mems, cands=None):
+    def forward(self, xs, mems, cands=None, pad_mask=None):
         """One forward step.
 
-        :param xs:    (bsz x seqlen) LongTensor queries to the model
-        :param mems:  (bsz x num_mems x seqlen) LongTensor memories
-        :param cands: (num_cands x seqlen) or (bsz x num_cands x seqlen)
-                      LongTensor with candidates to rank
+        :param xs:       (bsz x seqlen) LongTensor queries to the model
```
stephenroller: Go ahead and reformat to the newer format since you're editing this, please

Example:
https://github.com/facebookresearch/ParlAI/blob/1b4f655dbb277a241d332e8fff70308088faaa4b/parlai/core/torch_agent.py#L875-L890

parlai/agents/memnn/modules.py
```diff
@@ -120,7 +123,7 @@ def _reduce(self, embs, input):
         elif self.reduction == 'mean':
             # this is more fair than mean(-2) since mean includes null tokens
             sum = embs.sum(-2)
-            lens = input.ne(0).sum(-1).unsqueeze(-1).float()
+            lens = input.ne(0).sum(-1).unsqueeze(-1).float().clamp_(min=1)
```
stephenroller: clever

stephenroller: Does the 0 need to be NULL_IDX though?

parlai/agents/memnn/modules.py
```diff
@@ -188,19 +191,22 @@ def __init__(self, embedding_size, rotate=True):
...
+    def forward(self, query_embs, in_mem_embs, out_mem_embs, pad_mask):
         """Compute MemNN Hop step.
 
         :param query_embs:   (bsz x esz) embedding of queries
         :param in_mem_embs:  bsz list of (num_mems x esz) embedding of memories
                              for activation
         :param out_mem_embs: bsz list of (num_mems x esz) embedding of memories
                              for outputs
-
+        :param pad_mask      (bsz x num_mems) optional mask indicating which tokens
```
stephenroller: ditto please

parlai/agents/memnn/memnn.py
```diff
@@ -182,20 +187,20 @@ def _build_mems(self, mems):
         padded = torch.LongTensor(bsz, num_mems, seqlen).fill_(0)
 
         for i, mem in enumerate(mems):
-            # tf_offset = len(mem) - 1
+            tf_offset = len(mem) - 1
```
stephenroller: weird that this was commented before?

stephenroller:
When you get a chance, also rebase or merge master for my sanity with the unit tests, please.

parlai/agents/memnn/memnn.py
emilydinan: yeah, why were these commented out?

stephenroller: Okay I'm still confused why these lines were commented out before.

Issue #1659, opened 2019-04-21
DailyDialog examples containing __SILENCE__ tokens


ricsinaruto:
There's a lot of examples in the dataset where the first utterance is just __SILENCE__. Is this intended, and if so what's the purpose of it?

I feel like this __SILENCE__ token is making the learning task harder, because it's observed with a lot of responses in the data (see image below). Is there any way to use the dataset where examples containing this token are left out?

Command run:
```
python examples/train_model.py -t dailydialog -m transformer/generator --shuffle true -tr 32 --dict-maxtokens 16384 -histsz 1 -bs 256
```

![image](https://user-images.githubusercontent.com/18282017/56474810-7484df80-647f-11e9-9fd4-1b6cbed74b40.png)

stephenroller:
This is done to be consistent with Convai2, where you need to simulate being the first person of the conversation. However, this is done entirely in the agent:

https://github.com/facebookresearch/ParlAI/blob/733b627ae456d6b11a2fc4624088a781bc6c1d03/parlai/tasks/dailydialog/agents.py#L66-L85

If you wanted to add another teacher that skipped the beginning entries, I would accept the PR.

ricsinaruto:
Okay, thanks! I will look into it.

ricsinaruto:
Solved by #1661

PR #1661, opened 2019-04-22
New DailyDialog Teacher to exclude __SILENCE__ tokens 

ricsinaruto:
Made the modifications according to #1659 
Basically added a modified dailydialog version where there is no START_ENTRY.

I still tried to keep the two speaker setup however, to have as many examples as possible. Basically, when the agent should start the dialog the first utterance is skipped, and the second utterance is the input. 

However, when the dialog consists of only 2 turns we can't shift the dialog, so I just use the same dialog regardless of who should start the dialog. Because all episodes are duplicated and there has to be at least 1 entry for each episode, I don't see another way to solve this, but I don't think it should be a big problem that the few 2-turn dialogs that are in the dataset are duplicated.

The naming of the datasets might not be the best, I'm open for suggestions :)

Tested with the display_data script, seems to work as expected (22236 episodes with a total of 87170 examples)
image

I tried to run this command from the contributing guidelines, but it gave me an error:
```
python setup.py test -s tests.suites.datatests
```
Error:
```
running test
Searching for markdown<3.0
Best match: Markdown 2.6.11
Processing Markdown-2.6.11-py3.6.egg

Using /mnt/permanent/home/csaky/miniconda3/envs/python-3.6-gpu/lib/python3.6/site-packages/Markdown-2.6.11-py3.6.egg
running egg_info
writing parlai.egg-info/PKG-INFO
writing dependency_links to parlai.egg-info/dependency_links.txt
writing requirements to parlai.egg-info/requires.txt
writing top-level names to parlai.egg-info/top_level.txt
reading manifest file 'parlai.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'parlai.egg-info/SOURCES.txt'
running build_ext
datasets (unittest.loader._FailedTest) ... ERROR

ERROR: datasets (unittest.loader._FailedTest)
AttributeError: module 'tests.suites' has no attribute 'datasets'

Ran 1 test in 0.000s

FAILED (errors=1)
Test failed: <unittest.runner.TextTestResult run=1 errors=1 failures=0>
error: Test failed: <unittest.runner.TextTestResult run=1 errors=1 failures=0>
```

### DIFF
--- a/parlai/tasks/dailydialog/agents.py
+++ b/parlai/tasks/dailydialog/agents.py
@@ -91,8 +91,9 @@ class Convai2Teacher(FixedDialogTeacher):
 
 
 class NoStartTeacher(Convai2Teacher):
-    """ Same as default teacher, but it doesn't contain __SILENCE__ entries.
-        If we are the first speaker, then the first utterance is skipped.
+    """
+    Same as default teacher, but it doesn't contain __SILENCE__ entries.
+    If we are the first speaker, then the first utterance is skipped.
     """
     def __init__(self, opt, shared=None):
         super().__init__(opt, shared)
@@ -100,7 +101,7 @@ class NoStartTeacher(Convai2Teacher):
         # Calculate the correct number of examples.
         self.num_exs = sum(len(d['dialogue']) - 1 for d in self.data)
 
-        # Store all episodes in self.data
+        # Store all episodes separately, so we can deal with 2-turn dialogs.
         self.all_eps = self.data + [d for d in self.data if len(d['dialogue']) > 2]
         self.num_eps = len(self.all_eps)
 
@@ -110,7 +111,7 @@ class NoStartTeacher(Convai2Teacher):
 
         # Sometimes we're speaker 1 and sometimes we're speaker 2.
         # We can't be speaker 1 if dialog has only 2 turns.
-        speaker_id = episode_idx % 2 if len(entries) > 2 else 0
+        speaker_id = int(episode_idx >= len(self.data))
 
         their_turn = entries[speaker_id + 2 * entry_idx]
         my_turn = entries[1 + speaker_id + 2 * entry_idx]
--- a/parlai/tasks/dailydialog/agents.py
+++ b/parlai/tasks/dailydialog/agents.py
@@ -96,8 +96,7 @@ class NoSilenceTeacher(Convai2Teacher):
         super().__init__(opt, shared)
 
         # Calculate the correct number of examples.
-        dialogs = [d['dialogue'] for d in self.data]
-        self.num_exs -= len([0 for d in dialogs if len(d) >= 2])
+        self.num_exs = sum(len(d['dialogue']) - 1 for d in self.data)
 
     def get(self, episode_idx, entry_idx=0):
         # Sometimes we're speaker 1 and sometimes we're speaker 2
--- a/parlai/tasks/dailydialog/agents.py
+++ b/parlai/tasks/dailydialog/agents.py
@@ -90,30 +90,31 @@ class Convai2Teacher(FixedDialogTeacher):
         return shared
 
 
-class NoSilenceTeacher(Convai2Teacher):
-    # Same as the default teacher, but it doesn't contain __SILENCE__ entries.
+class NoStartTeacher(Convai2Teacher):
+    """ Same as default teacher, but it doesn't contain __SILENCE__ entries.
+        If we are the first speaker, then the first utterance is skipped.
+    """
     def __init__(self, opt, shared=None):
         super().__init__(opt, shared)
 
         # Calculate the correct number of examples.
         self.num_exs = sum(len(d['dialogue']) - 1 for d in self.data)
 
+        # Store all episodes in self.data
+        self.all_eps = self.data + [d for d in self.data if len(d['dialogue']) > 2]
+        self.num_eps = len(self.all_eps)
+
     def get(self, episode_idx, entry_idx=0):
-        # Sometimes we're speaker 1 and sometimes we're speaker 2
-        speaker_id = episode_idx % 2
-        full_eps = self.data[episode_idx // 2]
+        full_eps = self.all_eps[episode_idx]
         entries = full_eps['dialogue']
 
-        # If we are first speaker, then the first utterance is skipped.
-        # If dialog is only 2 turns, then we can't shift it.
-        if len(entries) > 2:
-            their_turn = entries[speaker_id + 2 * entry_idx]
-            my_turn = entries[1 + speaker_id + 2 * entry_idx]
-            episode_done = 2 * entry_idx + speaker_id + 1 >= len(entries) - 2
-        else:
-            their_turn = entries[0]
-            my_turn = entries[1]
-            episode_done = True
+        # Sometimes we're speaker 1 and sometimes we're speaker 2.
+        # We can't be speaker 1 if dialog has only 2 turns.
+        speaker_id = episode_idx % 2 if len(entries) > 2 else 0
+
+        their_turn = entries[speaker_id + 2 * entry_idx]
+        my_turn = entries[1 + speaker_id + 2 * entry_idx]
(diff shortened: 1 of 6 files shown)


ricsinaruto:
That one unit test that failed seems totally unrelated to my modifications

stephenroller:
Are you sure you ran `python setup.py test -s tests.suites.datatests`? It looks like you ran `python setup.py test -s tests.suites.datasets`, but I'm not discounting the possibility of outdated docs or a bug.

ricsinaruto:
> Are you sure you ran `python setup.py test -s tests.suites.datatests`? It looks like you ran `python setup.py test -s tests.suites.datasets`, but I'm not discounting the possibility of outdated docs or a bug.

My bad!
Here is the output:
```
running egg_info
writing parlai.egg-info/PKG-INFO
writing dependency_links to parlai.egg-info/dependency_links.txt
writing requirements to parlai.egg-info/requires.txt
writing top-level names to parlai.egg-info/top_level.txt
reading manifest file 'parlai.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'parlai.egg-info/SOURCES.txt'
running build_ext
test_verify_data (test_new_tasks.TestNewTasks) ... ok

Ran 1 test in 46.728s

OK
```

ricsinaruto:
> I'm not sure the number of examples is correct. It's set to the total number of utterances before, but now one utterance per conversation needs to be excluded, no?

You are right, the number of examples should be less, I have no idea why the display_data script returns this number.

stephenroller:
https://github.com/facebookresearch/ParlAI/blob/da41e279af0ba3c63842f12318ec686981acce38/parlai/tasks/dailydialog/agents.py#L47-L57

ricsinaruto:
Okay the reason is, that it is pre-computed by a function (num_examples) in the original dataset, so I have to change this.

ricsinaruto:
Added a commit to fix it, the correct number of examples is 76052.

parlai/tasks/dailydialog/agents.py
```diff
@@ -92,6 +92,13 @@ def share(self):
 
 class NoSilenceTeacher(Convai2Teacher):
     # Same as the default teacher, but it doesn't contain __SILENCE__ entries.
+    def __init__(self, opt, shared=None):
+        super().__init__(opt, shared)
+
+        # Calculate the correct number of examples.
+        dialogs = [d['dialogue'] for d in self.data]
+        self.num_exs -= len([0 for d in dialogs if len(d) >= 2])
```
stephenroller: Can we use an explicit sum here?

I think it's `self.num_exs = sum(len(d['dialogue']) - 1 for d in self.data))`?

ricsinaruto: Done!

parlai/tasks/dailydialog/agents.py
```diff
@@ -88,3 +88,43 @@ def share(self):
         shared = super().share()
         shared['data'] = self.data
         return shared
+
+
+class NoSilenceTeacher(Convai2Teacher):
```
stephenroller: What do you think about `NoStart`? `NoSilence` is fine but I agree we could have a slightly better name

parlai/tasks/dailydialog/agents.py
```diff
@@ -88,3 +88,43 @@ def share(self):
...
+    # Same as the default teacher, but it doesn't contain __SILENCE__ entries.
```
stephenroller: Make this into a docstring instead please.

parlai/tasks/task_list.py
```diff
@@ -889,6 +889,20 @@
             "for more information."
         )
     },
+    {
```
stephenroller: Let's not add this to task list.

parlai/tasks/dailydialog/agents.py
```diff
@@ -88,3 +88,43 @@ def share(self):
...
+
+    def get(self, episode_idx, entry_idx=0):
+        # Sometimes we're speaker 1 and sometimes we're speaker 2
+        speaker_id = episode_idx % 2
+        full_eps = self.data[episode_idx // 2]
+        entries = full_eps['dialogue']
+
+        # If we are first speaker, then the first utterance is skipped.
+        # If dialog is only 2 turns, then we can't shift it.
+        if len(entries) > 2:
```
stephenroller: Honestly it feels like we should be able to do this without special casing using correct math. If a conversation only has 2 turns, then we shouldn't have it included in num_episodes. This makes all of this much more nuanced :(

ricsinaruto: I will think about it, it might be easier than I originally thought

ricsinaruto:
I think it should be okay now, I just store all episodes separately so the 2-turn dialogs are only stored once.
21753 episodes with a total of 76052 examples.

parlai/tasks/dailydialog/agents.py
```diff
@@ -88,3 +88,44 @@ def share(self):
...
+        self.all_eps = self.data + [d for d in self.data if len(d['dialogue']) > 2]
+        self.num_eps = len(self.all_eps)
+
+    def get(self, episode_idx, entry_idx=0):
+        full_eps = self.all_eps[episode_idx]
+        entries = full_eps['dialogue']
+
+        # Sometimes we're speaker 1 and sometimes we're speaker 2.
+        # We can't be speaker 1 if dialog has only 2 turns.
+        speaker_id = episode_idx % 2 if len(entries) > 2 else 0
```
stephenroller: I think this should now be `speaker_id = int(episode_idx >= len(self.data))`, right?

parlai/tasks/dailydialog/agents.py
```diff
@@ -88,3 +88,44 @@ def share(self):
...
+        If we are the first speaker, then the first utterance is skipped.
+    """
+    def __init__(self, opt, shared=None):
+        super().__init__(opt, shared)
+
+        # Calculate the correct number of examples.
+        self.num_exs = sum(len(d['dialogue']) - 1 for d in self.data)
+
+        # Store all episodes in self.data
+        self.all_eps = self.data + [d for d in self.data if len(d['dialogue']) > 2]
```
stephenroller: good

parlai/tasks/dailydialog/agents.py
```diff
@@ -88,3 +88,44 @@ def share(self):
         shared = super().share()
         shared['data'] = self.data
         return shared
+
+
+class NoStartTeacher(Convai2Teacher):
+    """ Same as default teacher, but it doesn't contain __SILENCE__ entries.
```
stephenroller: indenting
```python
"""
Same as default teacher, but it doesn't contain __SILENCE__ entries.
If we are the first speaker, then the first utterance is skipped.
"""
```

parlai/tasks/dailydialog/agents.py
ricsinaruto: Yes, that's actually a lot better.

stephenroller:
Cool, one more run of `python setup.py test -s tests.suites.datatests` and lgtm.

ricsinaruto:
> Cool, one more run of `python setup.py test -s tests.suites.datatests` and lgtm.

```
running test
Searching for markdown<3.0
Best match: Markdown 2.6.11
Processing Markdown-2.6.11-py3.6.egg

Using /mnt/permanent/home/csaky/miniconda3/envs/python-3.6-gpu/lib/python3.6/site-packages/Markdown-2.6.11-py3.6.egg
running egg_info
writing parlai.egg-info/PKG-INFO
writing dependency_links to parlai.egg-info/dependency_links.txt
writing requirements to parlai.egg-info/requires.txt
writing top-level names to parlai.egg-info/top_level.txt
reading manifest file 'parlai.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'parlai.egg-info/SOURCES.txt'
running build_ext
test_verify_data (test_new_tasks.TestNewTasks) ... ok

Ran 1 test in 80.768s

OK
```

PR #1664, opened 2019-04-22
[MTurk] Treat abandons as a special case of returns

JackUrb:
This only particularly affects static tasks. As of now on certain SNS messages we would update the state of an agent to reflect the amazon event. Abandon events wouldn't really occur with any frequency on dialogue/connected tasks, and even if a worker didn't finish in the allotted time, by leaving the page they'd be considered a disconnect. As the `MTurkAgent.act()` method uses a task being returned as an exit condition, this PR makes abandons equivalent to returns, but retains the abandoned flag as it can be an important signal for a task that has too short of a duration set.


JackUrb:
@stephenroller It's only temporary - once `dev` is done and becomes `stable` I'm not touching legacy again 😅

Issue #1622, opened 2019-04-08
Wizard Generator having issues loading in public


stephenroller:
Pulled in from bottom of #1587. Likely related to #1620.

Appears to be an issue that sometimes pretrained models in the zoo are having trouble with overriding options of some sort.

cc @jazzminewang

Steps to reproduce:
```
python examples/eval_model.py -bs 64 -t wizard_of_wikipedia:generator:topic_split -mf models:wizard_of_wikipedia/end2end_generator/model
```

Traceback:
```python-traceback
/Users/jasminewang/Development/RLLab/mturk_dialog_eval/parlai/agents/transformer/transformer.py:18: UserWarning: Public release transformer models are currently in beta. The name of command line options may change or disappear before a stable release. We welcome your feedback. Please file feedback as issues at https://github.com/facebookresearch/ParlAI/issues/new
  "Public release transformer models are currently in beta. The name of "
[ warning: overriding opt['task'] to wizard_of_wikipedia:generator:topic_split (previously: wizard_of_wikipedia:generator:random_split )]
[ warning: overriding opt['model_file'] to /Users/jasminewang/Development/RLLab/mturk_dialog_eval/data/models/wizard_of_wikipedia/end2end_generator/model (previously: /tmp/wizard_endtoend_model )]
Dictionary: loading dictionary from /Users/jasminewang/Development/RLLab/mturk_dialog_eval/data/models/wizard_of_wikipedia/end2end_generator/model.dict
[ num words =  34883 ]
/private/home/roller/working/parlai/data/models/fasttext_vectors/wiki.en.vec: 0.00B [00:00, ?B/s]
Traceback (most recent call last):
  File "examples/eval_model.py", line 17, in <module>
    eval_model(opt, print_parser=parser)
  File "/Users/jasminewang/Development/RLLab/mturk_dialog_eval/parlai/scripts/eval_model.py", line 68, in eval_model
    agent = create_agent(opt, requireModelExists=True)
  File "/Users/jasminewang/Development/RLLab/mturk_dialog_eval/parlai/core/agents.py", line 552, in create_agent
    model = load_agent_module(opt)
  File "/Users/jasminewang/Development/RLLab/mturk_dialog_eval/parlai/core/agents.py", line 429, in load_agent_module
    return model_class(new_opt)
  File "/Users/jasminewang/Development/RLLab/mturk_dialog_eval/projects/wizard_of_wikipedia/generator/agents.py", line 99, in __init__
[...]

PR #1666, opened 2019-04-23
Don't copy datapath from disk opt files.

stephenroller:
Seems to be manifesting in #1622, and replicated by @koustuvsinha.

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -388,6 +388,10 @@ def load_agent_module(opt):
     optfile = model_file + '.opt'
     if os.path.isfile(optfile):
         new_opt = _load_opt_file(optfile)
+        # TODO we need a better way to say these options are never copied...
+        if 'datapath' in new_opt:
+            # never use the datapath from an opt dump
+            del new_opt['datapath']
         if 'batchindex' in new_opt:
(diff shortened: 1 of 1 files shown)


parlai/core/agents.py
```diff
@@ -388,6 +388,10 @@ def load_agent_module(opt):
     optfile = model_file + '.opt'
     if os.path.isfile(optfile):
         new_opt = _load_opt_file(optfile)
+        # TODO we need a better way to say these options are never copied...
+        if 'datapath' in new_opt:
+            # never use the datapath from an opt dump
+            del new_opt['datapath']
```
emilydinan: i'm not sure this is what we want to do. we should probably do something like `new_opt['datapath'] = opt['datapath']`, right? if you check line 446, you'll see that the model is loaded with `new_opt`

stephenroller: The case you're concerned about is handled in lines: 406-408

https://github.com/facebookresearch/ParlAI/blob/cc8dfa4e4becc528f16e1f44e23fb97ed726c929/parlai/core/agents.py#L406-L408

emilydinan: ah thanks

PR #1668, opened 2019-04-23
Fix setting of PARLAI_DATAPATH and PARLAI_DOWNPATH env variables

klshuster:
**Patch description**

We originally were overriding the `PARLAI_DOWNPATH` and `PARLAI_DATAPATH` env variables within params, as both `--datapath` and `--download-path` had default values. Now, we only use the default values if neither the arg is set nor the environment variable is set.

**Testing steps**
Run any standard ParlAI command and see if the datapath is set correctly;
Set the `PARLAI_DATAPATH` env variable to a value different from the default data path.

**Expected behavior**
ParlAI will use the default datapath when nothing is specified, and will use the `PARLAI_DATAPATH` env variable when specified

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -595,6 +595,8 @@ class ParlaiParser(argparse.ArgumentParser):
         self.opt['parlai_home'] = self.parlai_home
 
         # set environment variables
+        # Priority for setting the datapath (same applies for download_path):
+        # --datapath -> os.environ['PARLAI_DATAPATH'] -> <self.parlai_home>/data
         if self.opt.get('download_path'):
             os.environ['PARLAI_DOWNPATH'] = self.opt['download_path']
         elif os.environ.get('PARLAI_DOWNPATH') is None:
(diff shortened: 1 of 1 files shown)


parlai/core/params.py
```diff
@@ -599,8 +597,15 @@ def parse_args(self, args=None, namespace=None, print_args=True):
...
         if self.opt.get('download_path'):
             os.environ['PARLAI_DOWNPATH'] = self.opt['download_path']
+        elif os.environ.get('PARLAI_DOWNPATH') is None:
+            os.environ['PARLAI_DOWNPATH'] = os.path.join(self.parlai_home, 'downloads')
         if self.opt.get('datapath'):
             os.environ['PARLAI_DATAPATH'] = self.opt['datapath']
+        elif os.environ.get('PARLAI_DATAPATH') is None:
+            os.environ['PARLAI_DATAPATH'] = os.path.join(self.parlai_home, 'data')
+
+        self.opt['download_path'] = os.environ['PARLAI_DOWNPATH']
```
stephenroller: the self.opt = os.environ bit is surprising but I think your logic is correct. Maybe you can leave a comment saying the priorities. `# --datapath > environ[PARLAI_DATAPATH] > parlai_dir/downloads (default)`.

klshuster: added

PR #1671, opened 2019-04-26
Update train_end2end.py

### LINKED ISSUE #1670: embeddingsize or embedding_size
When I search 'embeddingsize' in this repository, I see many files referencing `opt['embeddingsize']` and similarly for 'embedding_size'. Unless there is a real reason for having both, could you please merge the two options 'embeddingsize' and 'embedding_size'? This threw me off. Here is one example set of files:

'embeddingsize'
https://github.com/facebookresearch/ParlAI/blob/a43f2880719c5a048fdf3d0aa5d5b25eeb9a1a41/projects/wizard_of_wikipedia/generator/train_end2end.py#L21

'embedding_size'
https://github.com/facebookresearch/ParlAI/blob/8ab911a29dbbe5cfb7d3e615cccf8f4c76066ff1/projects/wizard_of_wikipedia/generator/agents.py#L33

stephenroller:
**Patch description**
There was a typo in the supplied arguments for a pretraining script. Fixes #1670.

**Testing steps**
Checked the arguments carefully again.

### DIFF
--- a/projects/wizard_of_wikipedia/generator/train_end2end.py
+++ b/projects/wizard_of_wikipedia/generator/train_end2end.py
@@ -18,7 +18,7 @@ if __name__ == '__main__':
         n_heads=2,
         dropout=0.20,
         ffn_size=512,
-        embeddingsize=256,
+        embedding_size=256,
         log_every_n_secs=10,
         validation_patience=12,
         validation_metric='ppl',


g-karthik:
This pull request does not fix everything mentioned in #1670.

Like I mentioned in that issue, you can search this repo with the query "embeddingsize" and you will find several results, not just the Wizard of Wikipedia files I listed and specifically mentioned as one example.

Please re-open #1670 (I would if I could) until the option "embeddingsize" is completely merged with "embedding_size".

stephenroller:
That's not going to happen. Unfortunately, there's just inconsistencies throughout the repo, and it's model dependent as to which is correct:

```
devfair0237 parlai fixdatapathpretrained » ag -- '--embeddingsize'
docs/source/tutorial_seq2seq.rst
303:        agent.add_argument('-esz', '--embeddingsize', type=int, default=128,

parlai/agents/example_seq2seq/example_seq2seq.py
92:        agent.add_argument('-esz', '--embeddingsize', type=int, default=128,

parlai/agents/language_model/language_model.py
40:        agent.add_argument('-esz', '--embeddingsize', type=int, default=200,

parlai/agents/legacy_agents/seq2seq/seq2seq_v1.py
51:        agent.add_argument('-esz', '--embeddingsize', type=int, default=128,

parlai/agents/legacy_agents/seq2seq/seq2seq_v0.py
127:        agent.add_argument('-esz', '--embeddingsize', type=int, default=128,

parlai/agents/seq2seq/seq2seq.py
40:        agent.add_argument('-esz', '--embeddingsize', type=int, default=128,

parlai/agents/starspace/starspace.py
73:            '--embeddingsize',

parlai/agents/ibm_seq2seq/ibm_seq2seq.py
60:        agent.add_argument('-esz', '--embeddingsize', type=int, default=128,

projects/personachat/kvmemnn/kvmemnn.py
154:        agent.add_argument('-esz', '--embeddingsize', type=int, default=128,
devfair0237 parlai fixdatapathpretrained » ag -- '--embedding-size'
parlai/agents/legacy_agents/memnn/memnn_v0.py
31:        arg_group.add_argument('--embedding-size', type=int, default=128,

parlai/agents/memnn/memnn.py
30:            '-esz', '--embedding-size', type=int, default=128,

parlai/agents/transformer/transformer.py
26:    argparser.add_argument('-esz', '--embedding-size', type=int, default=300,

projects/babi/memnn/memnn_bAbI_10k_task_parameter_sweep.sh
70:            -bs ${bs} -lr ${lr} --embedding-size ${esz} --hops ${hops} \

projects/babi/memnn/memnn_bAbI_1k_task_parameter_sweep.sh
67:            -bs ${bs} -lr ${lr} --embedding-size ${esz} --hops ${hops} \

projects/memnn_feedback/agent/memnn_feedback.py
43:        arg_group.add_argument('--embedding-size', type=int, default=128,

**Comment by g-karthik** [comment]
So are you saying the inconsistency is a feature, not a bug? If no, I think everyone would agree that this needs to be fixed.

**Comment by stephenroller** [comment]
No, it's not a feature. It's an unfortunate legacy issue that cannot be fixed as trivially as a global substitution, due to our priority of maintaining backwards compatibility of old model files.

**Comment by g-karthik** [comment]
Just saw your edits to your previous comment, thanks for making it more clear.

Well, then this is basically going to lead to an explosion of the .opt file dict. As it is, the opt dict is pretty ginormous - if more options are added carelessly without making sure that new options aren't already present in some form, it will become hard for consumers of ParlAI to really debug what's going on if they're facing issues.

My suggestion is that you version your .opt files to handle backward compatibility. One possible approach is to put in a version number key in the .opt file - v1 could have `embeddingsize` but v2 would have `embedding_size`, both won't co-exist in a .opt file that way.

**Comment by stephenroller** [comment]
Yes, I accidentally hit the enter key too early, apologies.

Yeah the proposal I'm thinking about is something like this upgrade_opt in #1488, but it's something that has to be considered carefully.

A possible temporary mediation would be to have most models take both:
```python
arg_group.add_argument('-esz', '--embeddingsize', '--embedding-size', dest='embedding_size')
```
And have it be such that dest is always set to what the model expects. This would work fine for the command line (where I usually just use `-esz` anyway) but it wouldn't work too well for config file type settings, like the unit tests and train scripts we usually distribute.

PR #1675, opened 2019-04-27
Next task button for static apps

hadasah:
**Patch description**
Created a next button so that a single hit can have >1 subtask presented individually. 
Static App now expects task_data to be passed to it as an array of subtask data objects, and onValidData will now trigger the Next button to be shown if there are more subtasks, otherwise the Done with Hit button will be shown.

**Testing steps**
Modifying any demo static task that uses the done with hit button to have task_data arrays of size 2 will work, or the task I'm working on.

**Expected behavior**
Next button should show up for the first n-1 subtasks.

@JackUrb @pringshia the UI for the next task button looks *exactly* like the done with hit button right now, but it should probably look different. Do either of you have thoughts regarding 1) removing the glyph, 2) changing the colour? (image here)

Screen Shot 2019-04-26 at 19 44 58


pringshia:
For the button, we could change the className to `btn-default` maybe or one of these other ones: https://getbootstrap.com/docs/3.4/css/#buttons-options

And yeah, maybe we can just use `"glyphicon glyphicon-chevron-right"` instead for the icon?

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -308,7 +312,35 @@ class StaticApp extends React.Component {
...
+      task_done = false;
+    }
+    this.setState(
+      {
+        show_next_task_button: show_next_task_button,
+        subtask_done: valid,
+        task_done: task_done,
+        response_data: all_response_data,
+      },
+      this.render
```
JackUrb: Is this line necessary? setState should cause an implied re-render when the state changes

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -308,7 +312,35 @@ class StaticApp extends React.Component {
...
+
+  nextButtonCallback() {
+    let next_subtask_index = this.state.current_subtask_index + 1;
+    this.setState(
+      {
+        current_subtask_index: next_subtask_index,
+        task_data: Object.assign(this.state.task_data, this.state.all_tasks_data[next_subtask_index]),
+        subtask_done: false,
+      },
+      this.reset
```
JackUrb: Similarly I'm not sure what reset is doing here, the state change in theory should be sufficient to handle this

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -707,6 +707,68 @@ class ReviewButtons extends React.Component {
   }
 }
 
+class NextButton extends React.Component {
+  // This component is responsible for initiating the click
```
JackUrb: nit: Update this comment to reflect the real functionality

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -707,6 +707,68 @@ class ReviewButtons extends React.Component {
...
+        Next
+      </button>
+    );
+    if (this.props.display_feedback) {
+      if (this.state.feedback_shown) {
+        let XReviewButtons = getCorrectComponent(
+          'XReviewButtons',
+          this.props.v_id
+        );
+        review_flow = (
```
JackUrb: You should be able to cut out the review flow entirely.

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -794,9 +857,11 @@ class DoneResponse extends React.Component {
       paddingRight: '25px',
       float: 'left',
     };
-    let done_button = <XDoneButton {...this.props} />;
-    if (!this.props.task_done) {
-      done_button = null;
+    let button = null;
+    if (this.props.task_done) {
+      button = <XDoneButton {...this.props} />;
+    } else if (this.props.subtask_done && this.props.show_next_task_button !== null && this.props.show_next_task_button) {
```
JackUrb: Nit: There's no need to do the explicit !== null check considering you're already checking for truthiness of `show_next_task_button`.

parlai/mturk/core/react_server/dev/components/core_components.jsx
```diff
@@ -707,6 +707,37 @@ class ReviewButtons extends React.Component {
...
+    let next_button = (
+      <button
+        id="next-button"
+        type="button"
+        className="btn btn-default btn-lg"
+        onClick={() => this.props.nextButtonCallback()}
+      >
+        <span
+          className="glyphicon glyphicon-chevron-right" aria-hidden="true" />{' '}
+        Next
```
pringshia: Small nit - maybe having the icon after the wording looks better here?

i.e. "Next >" instead of "> Next"

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -308,7 +312,34 @@ class StaticApp extends React.Component {
...
+    );
+  }
+
+  nextButtonCallback() {
+    let next_subtask_index = this.state.current_subtask_index + 1;
+    this.setState(
+      {
+        current_subtask_index: next_subtask_index,
+        task_data: Object.assign(
+          this.state.task_data, this.state.all_tasks_data[next_subtask_index]),
```
pringshia: [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) isn't the most intuitive but you'd actually want to do `Object.assign({}, this.state.task_data, this.state.all_tasks_data[next_subtask_index])` as `Object.assign` modifies the first argument.

Without having an empty object literal `{}` as the first argument, you are unintentionally modifying the state directly (i.e. `this.state.task_data`)

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -322,7 +353,11 @@ class StaticApp extends React.Component {
           }}
           onNewTaskData={new_task_data =>
             this.setState({
-              task_data: Object.assign(this.state.task_data, new_task_data),
+              all_tasks_data: Object.assign(
```
pringshia: Provide `{}` as first argument. See explanation in other comment

parlai/mturk/core/react_server/dev/app.jsx
```diff
@@ -322,7 +353,11 @@ class StaticApp extends React.Component {
...
+                this.state.all_tasks_data, new_task_data),
+              task_data: Object.assign(this.state.task_data, new_task_data[0]),
```
pringshia: `{}`

pringshia:
I think having this multi-step component is great! My only concern is how does this change the API surface of `StaticApp` for tasks that are simple and do not require multiple steps? I think having multiple steps is the special case here, in that most other uses of `StaticApp` for an mTurk task will most probably NOT use a multi-step approach. Given that, does it make sense to change the API surface (as we are here) to treat multiple steps as a first-class citizen of the API? Are we cramming too much into what should otherwise try to be straightforward abstraction?

Maybe the solution here is to keep `StaticApp` as it was previously, and to introduce a separate component called `MultiStepStaticApp` that includes your work in this PR. What do you think? Does this reasoning make sense?

hadasah:
@pringshia That's a good point. My thought was: Having multiple subtasks within a hit is actually quite common, and there isn't necessarily a huge distinction between cases where you want to present multiple subtasks within one hit vs. just one, so it might not make sense to have two separate static apps. Also, most people are likely to not touch the static app or core component code at all, so the only change necessary is to put their task_data inside an array container in run.py before passing it over to the app. @JackUrb: thoughts?

JackUrb:
@pringshia I think the multi-hit use case is pretty generic just based on the number of times I've seen it utilized in the type of tasks that static tasks is intending to capture. Putting as many tasks as possible into one chunk is generally used to improve worker experience by reducing churn between tasks.

JackUrb:
My only remaining comment at this point is to check lint

pringshia:
Sounds good then - I haven't had too much experience with actual flows yet (aside from just the single one I am working on) so was waxing philosophical. LGTM.

JackUrb:
The external perspective is always appreciated! The hope here is that we'd be able to support batching any arbitrary task into a multitask using this framework. Whether or not we'll actually be able to extend to that is an open question 😅

PR #1677, opened 2019-04-28
[MTurk] removing legacy from dev code

JackUrb:
**Patch description**
I've removed some code from the dev branch of the mturk core that was used to support the legacy (django) version of the frontend. This refactor also adds the capability for local support again (running ParlAI MTurk using just one server, such as on an EC2 instance rather than needing an additional heroku pass-through) to the new tasks. 

**Testing**
Launched a task after importing the `dev` mturk manager rather than the default. The flow works properly.


pringshia:
So happy for local support! <3

PR #1694, opened 2019-05-03
[wip] Add OS X tests.

stephenroller:
**Patch description**
Enables running unit tests on OS X for CircleCI.

**Testing steps**
CircleCI should pass with a new section.


stephenroller:
Looks like this will require some hassle of the environment more. Just gonna leave it as temp work.

PR #1700, opened 2019-05-08
[MTurk] Adding test debug message for a test that is flaky on circle

JackUrb:
**Patch description**
This test occasionally fails on circle and I'm not sure why, so I've added some logging to it to make things a little bit more clear. This test is going to be deprecated soon but in the meantime it would be nice to figure it out.


JackUrb:
Right now it only prints the packet id on failure, which is pretty useless as they are randomly generated. By printing the whole packet, I can actually know what came in out of order as there are a limited number of packets generated for this test.

PR #1706, opened 2019-05-11
Brighter announcements with user suppression.

stephenroller:
**Patch description**
1. Refactor the variety of announcements we've placed everywhere into a single call to `print_announcements()` so we can do this in the future a bit better.
2. Add terminal color formatting, and a daysleft counter (but only if outputting to a terminal).

**Testing steps**
Ran some commands. and saw the output

**Expected behavior**
Screen Shot 2019-05-11 at 1 52 53 PM

### DIFF
--- a/parlai/core/testing_utils.py
+++ b/parlai/core/testing_utils.py
@@ -34,7 +34,7 @@ except ImportError:
     GIT_AVAILABLE = False
 
 
-DEBUG = False  # change this to true to print to stdout anyway
+DEBUG = True  # change this to true to print to stdout anyway
 
 
 def is_this_circleci():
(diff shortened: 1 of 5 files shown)


stephenroller:
Changed it to say "Please fill it out at" so it would be a better call to action.

parlai/core/testing_utils.py
```diff
@@ -34,7 +34,7 @@
     GIT_AVAILABLE = False
 
 
-DEBUG = False  # change this to true to print to stdout anyway
+DEBUG = True  # change this to true to print to stdout anyway
```
hadasah: @stephenroller Is this supposed to stay True?

PR #1714, opened 2019-05-15
Transformer encoder fp16 compatibility

malachaux:
**Patch description**
In the transformer encoder reduction ,there is a conversion in float32, which is not compatible with float16.

**Testing steps**
finetune locally a memnet that uses transformer encoder in fp16 and fp32.


parlai/agents/transformer/modules.py
```diff
@@ -378,7 +378,7 @@ def forward(self, input, positions=None, segments=None):
         elif self.reduction_type == 'max':
             return tensor.max(dim=1)[0]
         elif self.reduction_type == 'mean':
-            divisor = mask.float().sum(dim=1).unsqueeze(-1).clamp(min=1e-20)
+            divisor = mask.type(tensor.dtype).sum(dim=1).unsqueeze(-1).clamp(min=1e-20)
```
stephenroller: `divisor = mask.float().sum(dim=1).unsqueeze(-1).clamp(min=1).type_as(tensor)`

PR #1720, opened 2019-05-23
Add a new test_website step to CircleCI.

stephenroller:
**Patch description**
Adds a small suite of tests to CircleCI which attempts to load of pages on the website. This is to remedy an issue with the CSS on the website being down.

These tests are run nightly, and every time we commit to master.

**Testing steps**
CircleCI passes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -251,7 +251,7 @@ workflows:
       - deploy_website:
           filters:
             branches:
-              only: testwebsite # change back to master before landing
+              only: master
       - test_website:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -231,7 +251,10 @@ workflows:
       - deploy_website:
           filters:
             branches:
-              only: master
+              only: testwebsite # change back to master before landing
```
stephenroller: note this will be fixed after review.

PR #1723, opened 2019-05-24
Tighten the lint a bit

stephenroller:
**Patch description**
Adds bugbear to our linter. This is used in fbcode and marks most of the "fatal" lints we get on a sync.

This diff also fixes a bunch of those lints.

**Testing steps**
Run `flake8 parlai tests`. Still not brave enough to go into projects

**Expected behavior**
Lint passes.

**Logs**
```
$ flake8 parlai tests
flake8 parlai tests  50.69s user 1.76s system 1420% cpu 3.692 total
```


stephenroller:
Since it was just a lint run, I don't really want to run datasets tests...

PR #1731, opened 2019-05-29
Quick start and Basic Documentation modifications

hadasah:
**Patch description**
Making some modifications to the docs, starting with Quick Start and the first half of Basic. I'm not happy with it yet, but I wanted to get some comment threads started.


docs/source/tutorial_quick.rst
```diff
@@ -57,26 +59,89 @@ Let's try asking the model a question ourselves.
...
   ...
   Enter your message: John went to the hallway.\n Where is John?
 
 Hopefully the model gets this right!
 
 
-Simple model
-------------
+
+Train a Transformer on Twitter
```
stephenroller: 👑

parlai/mturk/tasks/react_task_demo/react_custom_with_deps/frontend/package-lock.json
```diff
@@ -0,0 +1,5771 @@
+{
```
emilydinan: just a heads up, i think these files got in here by accident!

docs/source/tutorial_quick.rst
```diff
@@ -57,26 +59,89 @@ Let's try asking the model a question ourselves.
...
+
+  # display predictions for model saved at specific file on twitter
+  python examples/display_model.py -t twitter -mf /tmp/tr_twitter -ecands batch
+
+And finally, we can talk to the model interactively as before:
+
+.. code-block:: bash
+
+  # interact with saved model
+  python examples/interactive.py -mf /tmp/tr_twitter
```
emilydinan: This looks good! Note here, this command won't work. We'll probably have to specify a set of fixed candidates for the file to rank at interactive time (i.e. `-ecands fixed -fcp <some txt file with candidates to rank`)... although some have suggested that we should make the generation of this file automatic if we specify the task, so may change in the future

hadasah: Yes! I had meant to ask about this and completely forgot - since we use -ecands vocab for babi but we can't do that here. Hmm it would be nice if we made a file for this part of the task or provided a script to generate this file for a particular task.

hadasah: ok I'm going to remove this code block for now since it's not very important and I don't want to explain -ecands or -cands in the quick start

PR #1735, opened 2019-05-31
Internal Zoo shortcuts

emilydinan:
**Patch description**
Shortcut for models contained in an internal zoo. Useful for non-public repositories. Specify the path to your internal model zoo in your parlai_internal repository at `parlai_internal/zoo/.internal_zoo_path`. Then, when you specify your model with `-mf internal:<rest of model file path>` it automatically checks the specified repository. 

Also changed shortcut for public zoo from `models:` to `zoo:`. Note that `models:` will still work to preserve backwards compatibility.

**Testing steps**
Test by adding models to internal repository and running.

### DIFF
--- a/parlai/core/build_data.py
+++ b/parlai/core/build_data.py
@@ -16,6 +16,8 @@ import requests
 import shutil
 import tqdm
 
+from parlai.core.utils import warn_once
+
 
 def built(path, version_string=None):
     """
@@ -244,9 +246,9 @@ def modelzoo_path(datapath, path):
     """
     if path is None:
         return None
-    if not path.startswith('models:'):
+    if not path.startswith('models:') and not path.startswith('internal:'):
         return path
-    else:
+    elif path.startswith('models:'):
         # Check if we need to download the model
         animal = path[7:path.rfind('/')].replace('/', '.')
         if '.' not in animal:
@@ -259,3 +261,17 @@ def modelzoo_path(datapath, path):
             pass
 
         return os.path.join(datapath, 'models', path[7:])
+    else:
+        # Internal path --  useful for non-public projects.
+        # Save the path to your internal model zoo in
+        # parlai_internal/.internal_zoo_path
+        zoo_path = 'parlai_internal/zoo/.internal_zoo_path'
+        if not os.path.isfile('parlai_internal/zoo/.internal_zoo_path'):
+            warn_once('Please specify the path to your internal zoo in the '
+                      'file parlai_internal/zoo/.internal_zoo_path in your '
+                      'internal repository.')
+            return None
+        else:
+            with open(zoo_path, 'r') as f:
+                zoo = f.read().split('\n')[0]
+            return os.path.join(zoo, path[9:])
--- a/parlai/core/build_data.py
+++ b/parlai/core/build_data.py
@@ -16,8 +16,6 @@ import requests
 import shutil
 import tqdm
 
-from parlai.core.utils import warn_once
-
 
 def built(path, version_string=None):
     """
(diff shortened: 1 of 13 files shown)


stephenroller:
One thing I think we should do for doc day is build up a skeleton `parlai_internal` that we can share with external people. (I'm thinking `example_internal` in this repo). It would contain a README on how to set up your own `parlai_internal` and the features this give you.

parlai/core/build_data.py
```diff
@@ -259,3 +261,17 @@ def modelzoo_path(datapath, path):
             pass
 
         return os.path.join(datapath, 'models', path[7:])
+    else:
+        # Internal path --  useful for non-public projects.
+        # Save the path to your internal model zoo in
+        # parlai_internal/.internal_zoo_path
+        zoo_path = 'parlai_internal/zoo/.internal_zoo_path'
+        if not os.path.isfile('parlai_internal/zoo/.internal_zoo_path'):
+            warn_once('Please specify the path to your internal zoo in the '
```
stephenroller: Seems like this should be a `RuntimeError` or a `ValueError`.

parlai/core/build_data.py
```diff
@@ -244,9 +246,9 @@ def modelzoo_path(datapath, path):
     """
     if path is None:
         return None
-    if not path.startswith('models:'):
+    if not path.startswith('models:') and not path.startswith('internal:'):
```
stephenroller: What do you think about `internalzoo:`?

I personally think `models:` is obtuse and it might make sense for us to "switch" to `zoo:`.

parlai/core/build_data.py
```diff
@@ -246,11 +244,14 @@ def modelzoo_path(datapath, path):
     """
     if path is None:
         return None
-    if not path.startswith('models:') and not path.startswith('internal:'):
+    if (not path.startswith('models:') and not path.startswith('zoo:') and not
+            path.startswith('izoo:')):
         return path
-    elif path.startswith('models:'):
+    elif path.startswith('models:') or path.startswith('zoo:'):
+        zoo = path.split(':')[0]
```
stephenroller: s/zoo/prefix/

parlai/core/build_data.py
```diff
@@ -246,11 +244,14 @@ def modelzoo_path(datapath, path):
     """
     if path is None:
         return None
-    if not path.startswith('models:') and not path.startswith('internal:'):
+    if (not path.startswith('models:') and not path.startswith('zoo:') and not
```
stephenroller: alternative:

```python
if not any(path.startswith(prefix) for prefix in ['models:', 'zoo:', 'izoo:']):
```

parlai/core/build_data.py
```diff
@@ -246,11 +244,14 @@ def modelzoo_path(datapath, path):
...
+            path.startswith('izoo:')):
         return path
-    elif path.startswith('models:'):
+    elif path.startswith('models:') or path.startswith('zoo:'):
+        zoo = path.split(':')[0]
+        zoo_len = len(zoo) + 1
         # Check if we need to download the model
-        animal = path[7:path.rfind('/')].replace('/', '.')
+        animal = path[zoo_len:path.rfind('/')].replace('/', '.')
```
stephenroller: prefix_len

parlai/core/build_data.py
```diff
@@ -246,11 +244,14 @@ def modelzoo_path(datapath, path):
     """
```
stephenroller: Update the docstring please

emilydinan:
> Can we add a `# TODO: test the internal zoo.`

where would i add this todo? doesn't seem relevant to put in the public facing files

stephenroller:
Just in the logic that looks for izoo?

PR #1736, opened 2019-05-31
Add docstring linting to ParlAI

stephenroller:
**Patch description**
This patch enables `flake8-docstrings` for the entire repo, with the same change-it-fix-it pattern we have now. The main thing this does is require we add docstrings to all new and changed code. It also mandates some minimal style on docstrings. Additionally makes sure that the docstrings will compile fine for the actual docs.

Additionally, I fix a bunch of these lints for a large portion of parlai.core.

**Testing steps**
Run a bunch of lints on parlai.core. This has not exhaustively fixed all of them.

Still need to test the website.


parlai/core/agents.py
```diff
@@ -113,8 +136,9 @@ def shutdown(self):
 
 class Teacher(Agent):
     """
-    Basic Teacher agent which keeps track of how many times it's received
-    messages. Teachers provide the ``report()`` method to get back metrics.
+    Basic Teacher agent that keeps track of how many times it's received messages.
+
```
emilydinan: why does flake8 put these giant spaces between sentences?

parlai/core/utils.py
```diff
@@ -893,6 +894,7 @@ def add_field(name, data):
     return txt.rstrip('\t')
 
 
+# DEPREATION DAY: DELETE
```
emilydinan: spelling

parlai/core/utils.py
```diff
@@ -631,10 +632,10 @@ def __contains__(self, key):
         """Determine if text contains any offensive words in the filter."""
         return self.contains_offensive_language(key)
 
+    # TODO: can we delete this?
```
emilydinan: not yet, still in use in internal stuff

parlai/core/agents.py
stephenroller: https://www.python.org/dev/peps/pep-0257/

Docstrings should be a one-line summary, possibly followed by a longer description or additional details.

parlai/core/agents.py
```diff
@@ -131,30 +155,44 @@ def __init__(self, opt, shared=None):
...
         return t
 
     def epoch_done(self):
+        """Return whether the epoch is done."""
         return self.epochDone
 
     # Default unknown length
     def num_examples(self):
+        """
+        Return the number of examples (individual utterances) in the dataset.
```
klshuster: Should we maybe say `... number of examples (e.g. individual utterances)...` to specify that "individual utterances" is more of an example/archetype than the absolute norm for each dataset? E.g. for captioning datasets this will be an image, not an utterance.

parlai/core/agents.py
```diff
@@ -131,30 +155,44 @@ def __init__(self, opt, shared=None):
...
+
+        Default implementation returns `None`, indicating an unknown number.
+        """
         return None
 
     def num_episodes(self):
+        """
+        Return the number of episodes (conversations) in the dataset.
```
klshuster: ditto `e.g.`

parlai/core/build_data.py
```diff
@@ -155,17 +175,6 @@ def untar(path, fname, deleteTar=True):
         os.remove(fullpath)
 
 
-def cat(file1, file2, outfile, deleteFiles=True):
```
klshuster: I notice this function is used here: https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/twitter/build.py#L95

parlai/core/params.py
```diff
@@ -788,8 +798,3 @@ def ag_add_argument(*args, **kwargs):
 
         arg_group.add_argument = ag_add_argument  # override _ => -
         return arg_group
-
-    def error(self, message):
```
klshuster: is there a reason we are removing this?

parlai/core/pytorch_data_teacher.py
```diff
@@ -745,9 +769,13 @@ def act(self):
...
     def shutdown(self):
+        """Shut down."""
         super().shutdown()
         BatchSortCache.destroy()
 
 
 class DefaultTeacher(PytorchDataTeacher):
+    """Default teacher."""
+
+    # TODO: isn't having a DefaultTeacher here a bad idea? What is this used for?
```
klshuster: This is to mimic behavior of having a `DefaultTeacher` in other classes, i.e. so you do not have to specify `-t task:TeacherToUse`. It makes the logic easier to handle in the `get_task_module` function in `agents.py` (here: https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/agents.py#L625)

parlai/core/worlds.py
```diff
@@ -89,9 +90,11 @@ def __init__(self, opt, agents=None, shared=None):
 
     def parley(self):
         """
-        The main method, that does one step of actions for the agents
-        in the world. This is empty in the base class.
+        Perform one step of actions for the agents in the world.
+
+        This is empty in the base class.
         """
+        # TODO: mark as abstract?
```
klshuster: this is probably a good idea

parlai/core/worlds.py
```diff
@@ -380,8 +402,9 @@ def shutdown(self):
 
 class ExecutableWorld(MultiAgentDialogWorld):
     """
-    A world where messages from agents can be interpreted as _actions_ in the
-    world which result in changes in the environment (are executed). Hence a grounded
+    World where messages from agents can be interpreted as _actions_.
+
+    Actions  result in changes in the environment (are executed). Hence a grounded
```
klshuster: nit: 2 spaces after `Actions`

parlai/core/params.py
stephenroller: it looked unused to me. DEPRECATE ALL THE THINGS

klshuster: Actually, this specific function is related to the "Arg failures errors should happen AFTER the help message, not before" doc day task. It is what is called when there is an argparse error with our parser (you can un-delete it and i will take care of it tomorrow)

PR #1738, opened 2019-06-03
[Torch Ranker Agent] Hard reset LR if finetune

### LINKED ISSUE #1737: lr scheduler does not reset during fine turning in local model
**Bug description**
When train_model.py is loading model with init-model argument with , lr scheduler seems to continue from the last learning rate recorded, based on model status stored in init-model location, rather than have a hard reset.

Notice that:
* When bert bi ranker model is first downloaded and loaded via --pretrained-path argument, the issue does NOT persist. Only when model is loaded locally.
* I am not sure this should be the expected behavior or not, I assume not. Main basis of my assumptions are based on @emilydinan advice at #823 for fine-turn. In addition, @stephenroller 's comment on #1734 indicates that LR scheduler are expected to be reset during fine-turn.

**Reproduction steps**
Model 1 is first trained locally on convai2 task (with bert bi ranker) and complete training. Later fine turning with --init-model argument point to model 1 path, combine with --model-file argument point to folder does not exist (and expect to create). The LR scheduler does not reset, but rather continue decay based on last learning rate recorded.

**Expected behavior**
LR scheduler expected to reset to default state based on arguments given at command line inputs.

**My silly opinions**
In my opinion, it should be expected to perform one of the followings:
* At command line interface for train_model.py, a new argument should be allow to pass in to control "hard_reset" on lr scheduler, or scheduler do automatically hard reset when --init-model paramter is used. 
* At [torch ranker](https://github.com/facebookresearch/ParlAI/blob/46f656abf28c4fba2818c62a2199b25d484e0b41/parlai/core/torch_ranker_agent.py#L118), the "[hard_reset](https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/torch_agent.py#L686)" parameter should be controlled or accessed from outside of ranker.

emilydinan:
**Patch description**
Fixes #1737. Hard reset learning rate if we are finetuning.

**Testing steps**
Train and then finetune a model. Check LR.

### DIFF
--- a/tests/test_transformers.py
+++ b/tests/test_transformers.py
@@ -9,369 +9,436 @@ import unittest
 import parlai.core.testing_utils as testing_utils
 
 
-class TestTransformerRanker(unittest.TestCase):
-    """Checks that transformer_ranker can learn some very basic tasks."""
+# class TestTransformerRanker(unittest.TestCase):
+#     """Checks that transformer_ranker can learn some very basic tasks."""
+#
+#     @testing_utils.retry(ntries=3)
+#     def test_repeater(self):
+#         stdout, valid, test = testing_utils.train_model(dict(
+#             task='integration_tests:candidate',
+#             model='transformer/ranker',
+#             optimizer='adamax',
+#             learningrate=7e-3,
+#             batchsize=16,
+#             validation_every_n_epochs=5,
+#             validation_patience=2,
+#             n_layers=1,
+#             n_heads=4,
+#             ffn_size=64,
+#             embedding_size=32,
+#             candidates='batch',
+#             eval_candidates='inline',
+#             gradient_clip=0.5,
+#         ))
+#
+#         self.assertGreaterEqual(
+#             valid['hits@1'],
+#             0.90,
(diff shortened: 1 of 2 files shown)


tests/test_transformers.py
```diff
@@ -385,65 +387,82 @@ def test_resuming(self):
...
-                valid4['num_updates'],
-                valid1['num_updates'],
-                'LR scheduler change reset failed (num_updates).\n' + stdout4
-            )
-            self.assertEqual(
-                valid4['lr'],
-                1e-3,
-                'LR is not correct in final resume.\n' + stdout4
-            )
+        for BASE_ARGS in [GENERATOR_ARGS, RANKER_ARGS]:
```
stephenroller: As is, when this fails, it will be slightly annoying to tell whether it's the ranker model or the generator model that is failing.

It would be better to refactor out this code into a helper method, and then call it from two differently named tests with the proper args.

emilydinan: if you look at my latest commit i print in the error message which model it is

stephenroller: Alright I still think it be cleaner, but that’s what’s important :)

PR #1740, opened 2019-06-04
Fix fp16 candidate encodings.

stephenroller:
**Patch description**
If a model was trained in fp16, and then candidates were encoded using the GPU, they will be stored to disk in fp16 mode. This is fine, but we also need to potentially convert them back to fp32 in case we're using CPU model

**Testing steps**
Ran Jason's command and tested interactive mode on CPU.

**Logs**
```
$ python projects/convai2/interactive.py -mf removed --fixed-candidates-path removed  --encode-candidate-vecs True -ecands fixed --fp16 False --no-cuda
Enter [DONE] if you want a new partner at any time.
Enter Your Message: hi
/private/home/roller/working/parlai/parlai/core/torch_ranker_agent.py:440: UserWarning: [ Executing eval mode with a common set of fixed candidates (n = 126027). ]
  "(n = {}). ]".format(mode, len(self.fixed_candidates))
[removed]: hello . i am joan and i love to cook . how about you ?
```

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -538,7 +538,7 @@ class TorchAgent(Agent):
             if not shared and opt['gpu'] != -1:
                 torch.cuda.set_device(opt['gpu'])
         # indicate whether using fp16
-        self.fp16 = self.opt.get('fp16', False)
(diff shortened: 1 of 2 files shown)


jaseweston:
fankooo <3

PR #1742, opened 2019-06-04
[BERT Classifier] CLS + SEP tokens, plus segments

emilydinan:
**Patch description**
- Add option to add CLS token to the front of the text_vec. This defaults to True. We keep it as an option for backwards compatibility as some models were not doing this before. (Note: old models will still break unless we set the flag correctly, which is a tradeoff.)
- Add option to separate the last utterance in the history: adds a separator token and uses a different segment index for these tokens.
- Also add option to pass in a list to BERT's dictionary `vec2txt` function.

**Testing steps**
Train models.

### DIFF
--- a/parlai/agents/bert_classifier/bert_classifier.py
+++ b/parlai/agents/bert_classifier/bert_classifier.py
@@ -53,6 +53,8 @@ class BertClassifierAgent(TorchClassifierAgent):
         self.pretrained_path = os.path.join(opt['datapath'], 'models',
                                             'bert_models', MODEL_PATH)
         opt['pretrained_path'] = self.pretrained_path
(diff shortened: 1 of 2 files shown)


stephenroller:
This is exactly the kind of thing we need upgrade_opt for

samhumeau:
It looks good to me, (but the non backward compatibility is a bit annoying. Maybe a 'version' attribute?)

emilydinan:
> It looks good to me, (but the non backward compatibility is a bit annoying. Maybe a 'version' attribute?)

good idea -- i'll fix

emilydinan:
thoughts about this backwards compatibility fix @stephenroller @samhumeau ?

PR #1753, opened 2019-06-05
ABCs for TA, TRA, & TGA

emilydinan:
**Patch description**
Abstract base classes for Torch Agent, Torch Ranker Agent, and Torch Generator Agent.

Plus a change so that memnn works.

**Testing steps**
Run our tests 😸

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -696,7 +696,7 @@ class TorchRankerAgent(TorchAgent):
         :param padded_cands:
             The padded candidates.
         """
-        raise RuntimeError(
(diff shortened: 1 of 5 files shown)


parlai/core/torch_ranker_agent.py
```diff
@@ -697,8 +696,13 @@ def encode_candidates(self, padded_cands):
         :param padded_cands:
             The padded candidates.
         """
-        # TODO: describe input type and return type.
-        pass
+        raise RuntimeError(
```
stephenroller: isn't `NotImplementedError` better?

stephenroller:
gif

PR #1761, opened 2019-06-05
Arg provenance: trace your opt history

emilydinan:
**Patch description**
Ever find your self chasing through ParlAI to find out where that arg got set? Have we got a solution for you!

Now Opt is a subclass of dict. This new object traces the history of args being set. Simply call `opt.display_history(arg)` for some arg to see the file/line numbers in which this arg was changed. 

For instance, if we run `python examples/train_model.py -t convai2 -m transformer/ranker -mf /tmp/test_opt --learningrate 0.5` and then inside the TransformerRankerAgent call `self.opt.display_history('learningrate')` the output is as follows:

```
1. learningrate was set to 0.5 at:
  File "/private/home/edinan/ParlAI/parlai/core/agents.py", line 451, in load_agent_module
    new_opt[k] = v
```

Diff also includes a unittest to test this functionality.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -9,6 +9,8 @@ from parlai.core.utils import round_sigfigs
 from parlai.core.utils import set_namedtuple_defaults
 from parlai.core.utils import padded_tensor
 from parlai.core.utils import argsort
+from parlai.core.utils import Opt
+from copy import deepcopy
 import time
 import unittest
 import torch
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -118,6 +120,20 @@ def test_argsort(self):
 
         assert np.all(argsort(torch_keys, torch_keys)[0].numpy() == np.arange(1, 6))
 
+    def test_opt(self):
```
stephenroller: tessssts!

PR #1766, opened 2019-06-06
Writing a TGA tutorial

stephenroller:
**Patch description**
This patch contains a tutorial for writing a TorchGeneratorAgent from scratch, and the bare minimum pieces you need. The model is a simple LSTM encoder-decoder without attention.

Early feedback welcome, of course.

What's left:
- [x] This contains the example agent, which is well commented, but the tutorial isn't actually there.
- [x] ~This makes me feel like we could simplify the constructor for TGA~
- [x] Unit test that runs this model/playbook

**Testing steps**
Ran the code on the example. It gets some reasonable ppl; playing with beam size changes results.


docs/source/tutorial_torch_generator_agent.rst
```diff
@@ -0,0 +1,31 @@
...
+
+   python examples/train_model -m transformer/generator -t convai2 -mf /tmp/testtransformer --beam-size 5 -bs 16
+
+
+Creating a Model
+----------------
+
+In order to write a generative model, your agent should extend
+:py:class:`~parlai.core.torch_generator_agent.TorchGeneratorAgent`. This parent
+class implements ``train_step`` and ``eval_step``, so you only need to implement
```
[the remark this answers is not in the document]
stephenroller: Yeah i should find a better wording.

parlai/agents/example_tga/agents.py
```diff
@@ -0,0 +1,205 @@
...
+
+        Input *must* be input_tokens, which are the context tokens given
+        as a matrix of lookup IDs.
+
+        :param input_tokens:
+            Input tokens as a bsz x seqlen LongTensor.
+            Likely will contain padding.
+
+        :return:
+            You can return anything you like; it is will be passed verbatim
```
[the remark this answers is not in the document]
stephenroller: i'll add examples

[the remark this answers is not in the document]
EricMichaelSmith: @stephenroller What examples had you been thinking of adding? I think the last sentence of this docstring is a good example: I'm rewording it to "This particular implementation returns the hidden and cell states from the LSTM."

docs/source/tutorial_torch_generator_agent.rst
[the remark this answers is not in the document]
EricMichaelSmith: Fixed this

docs/source/tutorial_torch_generator_agent.rst
```diff
@@ -0,0 +1,31 @@
...
+parent class that provides functionality for building autoregressive generative
+models. Extending TorchGeneratorAgent requires your model conform to a strict
+interface, but then provides you rich functionality like beam search and sampling.
+
+
+Example Models
+--------------
+
+Two major models in ParlAI inheret from TorchGeneratorAgent: seq2seq and transformer.
+You can try one of these with the example below:
```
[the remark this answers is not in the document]
EricMichaelSmith: Fixed this

docs/source/tutorial_torch_generator_agent.rst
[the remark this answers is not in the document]
EricMichaelSmith: I've added a tiny detail to this paragraph ("and instantiate it through `build_model`"), but I do discuss reorder_encoder_states later on in the tutorial. Lemme know if this is what you had in mind

parlai/agents/examples/seq2seq.py
```diff
@@ -0,0 +1,202 @@
...
+            '-hid', '--hidden-size', type=int, default=1024, help='Hidden size.'
+        )
+
+    def build_model(self):
+        """
+        Construct the model.
+        """
+
+        model = ExampleModel(self.dict, self.opt['hidden_size'])
+        # we're responsible for setting the embeddings ourselves, but TorchAgent
```
stephenroller: Expand on this a bit more, say that we're initializing pretrained embeddings

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -0,0 +1,154 @@
+# Using Torch Generator Agent
+
+**Authors**: Stephen Roller, Eric Smith
```
stephenroller: I'm not an author, this is all you.

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -0,0 +1,154 @@
...
+
+`parlai.core.torch_generator_agent.TorchGeneratorAgent` is an abstract parent class that provides functionality for building autoregressive generative models. Extending `TorchGeneratorAgent` requires your model conform to a strict interface, but then provides you rich functionality like beam search and sampling.
```
stephenroller: Maybe let's just use the short name, `TorchGeneratorAgent`, not the full absolute path? You make the call.

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -0,0 +1,154 @@
...
+## Creating a Model
+
+In order to write a generative model, your agent should extend `parlai.core.torch_generator_agent.TorchGeneratorAgent`. This parent class implements `train_step` and `eval_step`, so you only need to implement your model and instantiate it through `build_model`. However, your model should implement the `parlai.core.torch_generator_agent.TorchGeneratorModel` interface.
+
+
+## Tutorial
+
+This tutorial will walk you through creating a simple generative model, found at `parlai.agents.examples.seq2seq`, that consists of an LSTM-based encoder and decoder.
+
+### Extending `TorchGeneratorAgent`
```
stephenroller: Worth adding a sentence saying the overall steps: we make an Agent, and a Model.

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -0,0 +1,154 @@
...
+
+The full code for the agent can be seen [here](https://github.com/facebookresearch/ParlAI/tree/master/parlai/agents/examples/seq2seq.py). To call training:
+
+```
+python examples/train_model.py -m examples/seq2seq \
+    -mf /tmp/example_model \
+    -t convai2 -bs 32 -eps 2 --truncate 128
+```
+
+You should get a perplexity of around 135 and a token accuracy of around 28% on the ConvAI2 validation/test set.
```
stephenroller: Also add some bits about generation, using beam search, nucleus, etc.

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -0,0 +1,154 @@
...
+Two major models in ParlAI inherit from `TorchGeneratorAgent`: seq2seq and transformer. You can try the transformer with the example below:
+
+```
+python examples/train_model -m transformer/generator -t convai2 -mf /tmp/testtransformer \
+  --beam-size 5 -bs 16
+```
+
+## Creating a Model
+
+In order to write a generative model, your agent should extend `parlai.core.torch_generator_agent.TorchGeneratorAgent`. This parent class implements `train_step` and `eval_step`, so you only need to implement your model and instantiate it through `build_model`. However, your model should implement the `parlai.core.torch_generator_agent.TorchGeneratorModel` interface.
```
emilydinan: i think we should sell this a little more -- what else doess TGA automatically take care of for you? (i.e. various sampling methods, beams earch...)

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -0,0 +1,154 @@
...
+
+
+## Tutorial
+
+This tutorial will walk you through creating a simple generative model, found at `parlai.agents.examples.seq2seq`, that consists of an LSTM-based encoder and decoder.
+
+### Extending `TorchGeneratorAgent`
+
+A minimal `TorchGeneratorAgent` only needs to implement `build_model()`, but if you want to specify any command-line arguments, you'll need to add `add_cmdline_args()` as well. This method first adds flags for `TorchGeneratorAgent` and then adds a `--hidden-size` flag for the hidden dimension of the LSTMs.
```
emilydinan: "This method" --> "the implementation below" ?

emilydinan: also, of waht LSTMs? can you describe teh example model you're implementing first?

tests/test_examples_seq2seq.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
```
emilydinan: nit: shouldn't this file be named `test_example_seq2seq`?

parlai/agents/examples/seq2seq.py
```diff
@@ -0,0 +1,202 @@
...
+        # gives us a nice helper
+        if self.opt['embedding_type'] != 'random':
```
emilydinan: why do we need these lines? doesn't TGA handle this for you? CC @stephenroller

parlai/agents/examples/seq2seq.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # get the new output and decoder incremental state
+        output, incr_state = self.lstm(embedded, state)
+
+        return output, incr_state
+
+
+class ExampleModel(tga.TorchGeneratorModel):
+    """
+    ExampleModel implements the final components of TorchGeneratorModel.
```
emilydinan: a bit confusing, doesn't this implement the entire model?

parlai/agents/examples/seq2seq.py
```diff
@@ -0,0 +1,202 @@
...
+"""  # noqa: E501
+
+import torch.nn as nn
+import torch.nn.functional as F
+import parlai.core.torch_generator_agent as tga
+
+
+class Encoder(nn.Module):
+    """
+    Example encoder.
```
emilydinan: can you elaborate on this a bit more? (and decoder too).. i.e. it instantiates a 1 layer LSTM with hidden size...

parlai/agents/examples/seq2seq.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+r"""
+Example TorchGeneratorAgent model.
+
+This demonstrates the minimal structure of a building a generative model.
```
emilydinan: can you add more detail ehre too?

parlai/agents/examples/seq2seq.py
EricMichaelSmith: yes - reworded

parlai/agents/examples/seq2seq.py
EricMichaelSmith: Yes - added more detail

tests/test_examples_seq2seq.py
stephenroller: I wouldn't mind if we consolidate with the example that dexter is making and just had `test_examples.py`

parlai/agents/examples/seq2seq.py
stephenroller: No, particularly because you could have 3 sets of embeddings: encoder input, decoder input, and decoder output. we typically tie them but the seq2seq model supports not tying them.

I think the `!= random` line can be removed though, or at least moved into `_copy_embeddings`

tests/test_examples_seq2seq.py
EricMichaelSmith: Yeah, I like this - I'm not sure we need multiple test_example scripts at the moment

> I wouldn't mind if we consolidate with the example that dexter is making and just had `test_examples.py`

Yeah, I like this - I'm not sure we need multiple test_example scripts at the moment

parlai/agents/examples/seq2seq.py
EricMichaelSmith: > No, particularly because you could have 3 sets of embeddings: encoder input, decoder input, and decoder output. we typically tie them but the seq2seq model supports not tying them.
> 
> I think the `!= random` line can be removed though, or at least moved into `_copy_embeddings`

Yes, that's cleaner - changing this

docs/source/tutorial_torch_generator_agent.md
EricMichaelSmith: Makes sense - added this to the "Creating a Model" section

docs/source/tutorial_torch_generator_agent.md
EricMichaelSmith: Right, good point - added this

docs/source/tutorial_torch_generator_agent.md
EricMichaelSmith: Right, was a bit vague - added a little more detail on this to this section and the one above

parlai/agents/examples/seq2seq.py
EricMichaelSmith: Yes - just added

EricMichaelSmith:
Thanks for the feedback @stephenroller @emilydinan . I made changes - lemme know what you think! (Stephen, I'm not sure if I can add you as a reviewer, even if I unassign you to this task, because you created the PR)

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -0,0 +1,156 @@
...
+We next define a function to project the output of the decoder back into the token space:
+
+```
+    def output(self, decoder_output):
+        return F.linear(decoder_output, self.embeddings.weight)
+```
+
+Lastly, we define two functions to reindex the latent states of the encoder and decoder along the sample dimension. We reindex the encoder at the very beginning of beam search and when ranking candidates during eval, and we reindex the decoder after each step of beam search. Since our encoder and decoder both are based on LSTMs, these encoder/decoder states are the hidden and cell states:
+```
+    def reorder_encoder_states(self, encoder_states, indices):
```
emilydinan: can you add a comment explaining what the indices passed in here are?

parlai/agents/examples/seq2seq.py
```diff
@@ -0,0 +1,210 @@
...
+        )
+        self.embeddings = nn.Embedding(len(dictionary), esz)
+        self.encoder = Encoder(self.embeddings, hidden_size)
+        self.decoder = Decoder(self.embeddings, hidden_size)
+
+    def output(self, decoder_output):
+        """
+        Perform the final output -> logits transformation.
+        """
+        return F.linear(decoder_output, self.embeddings.weight)
```
emilydinan: dumb q-- how is this returning logits if it's just a linear transformation

EricMichaelSmith: Oy. Very good question. I just checked, and `TransformerGeneratorModel` uses the same terminology: https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/transformer/modules.py#L939 @stephenroller Can we still consider these "logits" if they aren't mathematically the log of the odds of the probability, but they're still mathematically used in the same manner as logits for computing losses?

docs/source/tutorial_torch_generator_agent.md
EricMichaelSmith: Yeah, that was vague - I added this

parlai/agents/examples/seq2seq.py
EricMichaelSmith: Oh, makes sense - just talked to Stephen, and they *are* mathematically logits, because when the output of this function gets passed to `CrossEntropyLoss`, that loss criterion *performs* the softmax on the output, making them mathematically the log-odds of probabilities

PR #1769, opened 2019-06-07
Support fixed random seeds.

stephenroller:
**Patch description**
Support using a fixed random training seed.

**Testing steps**
Ran twice with two different seeds for 1 epoch, got exact some result on output.


stephenroller:
Ehhh i'm not confident this going to be a good idea when resuming training so I'm going to abandon this.

PR #1770, opened 2019-06-07
Don't output the LR warning when finetuning.

stephenroller:
**Patch description**
Previously, when you started fine tuning a model, the LR scheduler would reset. But it would also show a weird cryptic message when it reset.

Now, this cryptic message only shows when the LR scheduler actually changed; not when the user explicitly knows they're fine tuning.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -769,7 +769,14 @@ class TorchAgent(ABC, Agent):
             # first make sure there are no null pointers
             states = {}
 
-        if states and states.get('lr_scheduler_type') != self.opt['lr_scheduler']:
+        if (
+            # there is already an old LR scheduler saved on disk
+            states and
+            # and the old LR scheduler is different
+            states.get('lr_scheduler_type') != self.opt['lr_scheduler'] and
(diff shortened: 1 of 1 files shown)


parlai/core/torch_agent.py
```diff
@@ -769,7 +769,14 @@ def _invsqrt_lr(step):
             # first make sure there are no null pointers
             states = {}
 
-        if states and states.get('lr_scheduler_type') != self.opt['lr_scheduler']:
+        if (
+            # there is already an old LR scheduler saved on disk
+            states and
+            # and the old LR scheduler is different
+            states.get('lr_scheduler_type') != self.opt['lr_scheduler'] and
```
emilydinan: only question... do we need an opt.get('lr_scheduler') here?

stephenroller: A super old model might have been trained before we introduce lr_schedulers.

emilydinan: then why didn't you put an opt.get there?

stephenroller: Hmm, right above it will `raise ValueError` if it's not in the dictionary so i guess it's handled

PR #1771, opened 2019-06-07
Fix loading older fp16 optimizer checkpoints.

stephenroller:
**Patch description**
Sometime in the past two months, NVIDIA's APEX library changed the way it's storing/loading optimizer state from disk. As such, some models I trained in mid-february were unable to load their optimizer state successfully. This patch fixes by only loading the non-fp16 wrapped optimizer state from disk, and waiting to do the wrapping last.

This also happens to fix a bug where switching from fp16 training to fp32 training would fail to load the state from disk

**Testing steps**
- Tried loading an old model that contains the bad fp16 params
- Tried training and then resuming a checkpoint with the new fp16 params
- Tried taking the fp16 trained state and resuming in fp32 mode.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -665,15 +665,23 @@ class TorchAgent(ABC, Agent):
 
         optim_class = self.optim_opts()[opt['optimizer']]
         self.optimizer = optim_class(params, **kwargs)
-        if self.fp16:
-            self.optimizer = fp16_optimizer_wrapper(self.optimizer)
 
         if optim_states:
             if saved_optim_type != opt['optimizer']:
-                print('WARNING: not loading optim state since optim class '
-                      'changed.')
+                print('WARNING: not loading optim state since optim class changed.')
             else:
+                if 'loss_scaler' in optim_states:
+                    # previously, this code used to load up the fp16 optimizer
+                    # states directly, including the current fp16 scalars and all.
+                    # due to some changes in the APEX state_dict, this
+                    # auxillary information is no longer valid.  it makes sense
+                    # to simply to recover the underlying fp32 optimizer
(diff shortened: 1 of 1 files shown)


parlai/core/torch_agent.py
```diff
@@ -665,15 +665,23 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
...
-            self.optimizer = fp16_optimizer_wrapper(self.optimizer)
 
         if optim_states:
             if saved_optim_type != opt['optimizer']:
-                print('WARNING: not loading optim state since optim class '
-                      'changed.')
+                print('WARNING: not loading optim state since optim class changed.')
             else:
+                if 'loss_scaler' in optim_states:
+                    # previously, this code used to load up the fp16 optimizer
```
emilydinan: wowza that's one big comment

emilydinan: but very helpful 😄

parlai/core/torch_agent.py
```diff
@@ -1370,8 +1381,15 @@ def state_dict(self):
...
-        if hasattr(self, 'optimizer'):  # save optimizer params
-            states['optimizer'] = self.optimizer.state_dict()
+        # save optimizer params
+        if hasattr(self, 'optimizer'):
+            # don't store the fp16 wrapper
+            if self.fp16:
+                # store the "true" optimizer, not the fp16 wrapped version
+                states['optimizer'] = self.optimizer.optimizer.state_dict()
+            else:
+                # nothing abnormal.
```
emilydinan: kind of a weird comment lol

stephenroller:
This didn't work like i expected, needs further work.

Abandoning for to mark that.

PR #1772, opened 2019-06-08
Address subtleties in fp16/fp32 checkpoint loading

stephenroller:
**Patch description**
This is a reattempt of #1771.

This patch solves two issues: The initial one was that APEX broke backwards compatibility, forcing me to find a work around.

In doing so, I noticed that fine-tuning an optimizer produced different behaviors if one switched from fp16 to fp32 (or vice versa) compared to when one *stayed* in the original training mode.

Previously, staying in the original training mode would produce resume the optimizer exactly where it left off. However, switching between *modes* modes would cause the optimizer to be reset to a fresh state and forgetting all of the moments previously stored.

After some discussions with @aszlam and @jaseweston, I ran some experiments and found that fine tuning is better when we keep the old moments around, compared to when we reset them (keeping everything else constant). So this patch implements this choice: now no matter what setting you start and resume, the weights stay around. Due to changes in precision, it's most recommended you stay in the same regime.

Finally, at the end of the optimizer loading was some old dead code. After testing, this seems to be no longer necessary, so I'm also deleting it.

**Testing steps**
1. Trained a model in fp32 for 1 epoch. Compared resuming in fp16 and fp32 for another epoch, and confirmed results were very close
2. Trained a model in fp16 for 1 epoch, resumed in fp16 and fp32 for another epoch, and confirmed results were very close.
3. Pray that apex doesn't break backwards compatibility again.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -668,21 +668,43 @@ class TorchAgent(ABC, Agent):
         if self.fp16:
             self.optimizer = fp16_optimizer_wrapper(self.optimizer)
 
-        if optim_states:
-            if saved_optim_type != opt['optimizer']:
-                print('WARNING: not loading optim state since optim class '
-                      'changed.')
+        # TODO: we might want to hard reset optimizers here in the
+        # case of fine tuning. Some rudimentary experiments seemed to
+        # indicate that keeping adam weights around was desirable, so this
+        # will remain the behavior for the time being.
+        if optim_states and saved_optim_type != opt['optimizer']:
+            # we changed from adam to adamax, or sgd to adam, or similar
+            print('WARNING: not loading optim state since optim class changed.')
+        elif optim_states:
+            # check for any fp16/fp32 conversions we need to do
+            optimstate_fp16 = ('loss_scaler' in optim_states)
+            if self.fp16 and optimstate_fp16:
+                # previously trained in fp16, now we're training in fp16.
+                # ideally no action needed, but APEX broke backwards
+                # compatibility and this is the hack around it.
+                optim_states['loss_scaler'] = (
+                    self.optimizer.state_dict()['loss_scaler']
+                )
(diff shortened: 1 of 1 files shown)


parlai/core/torch_agent.py
```diff
@@ -668,21 +668,43 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
...
+            # we changed from adam to adamax, or sgd to adam, or similar
+            print('WARNING: not loading optim state since optim class changed.')
+        elif optim_states:
+            # check for any fp16/fp32 conversions we need to do
+            optimstate_fp16 = ('loss_scaler' in optim_states)
+            if self.fp16 and optimstate_fp16:
+                # previously trained in fp16, now we're training in fp16.
+                # ideally no action needed, but APEX broke backwards
+                # compatibility and this is the hack around it.
+                optim_states['loss_scaler'] = (
```
emilydinan: 😮

parlai/core/torch_agent.py
```diff
@@ -668,21 +668,43 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
...
+        # TODO: we might want to hard reset optimizers here in the
+        # case of fine tuning. Some rudimentary experiments seemed to
+        # indicate that keeping adam weights around was desirable, so this
+        # will remain the behavior for the time being.
+        if optim_states and saved_optim_type != opt['optimizer']:
+            # we changed from adam to adamax, or sgd to adam, or similar
+            print('WARNING: not loading optim state since optim class changed.')
+        elif optim_states:
+            # check for any fp16/fp32 conversions we need to do
+            optimstate_fp16 = ('loss_scaler' in optim_states)
```
emilydinan: is this the best / only way to check if the previous model was using fp16?

stephenroller:
probably. i'll leave it for another PR though.

PR #1775, opened 2019-06-11
Report correct agent-reported metrics during evaluation when multi-tasking

emilydinan:
**Patch description**
Currently during multitasking, agent reported metrics aren't reported correctly. This changes makes sure that the metrics are reported correctly during evaluation by creating multiple worlds and iterating through them sequentially, collecting separate metrics for each world and then finally aggregating them. 

Note that in this case, if we specify `validation_max_exes`, it splits the number of examples equally across tasks. In the future, this could change to be different proportions.

**Testing steps**
Train and evaluate models, check that metrics are reported correctly. [Edit: should probably write unit tests but will wait until feedback from initial review...]

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -59,7 +59,7 @@ def aggregate_task_reports(reports, tasks, micro=True):
 
     :param reports: list of report dicts from separate tasks
     :param tasks: list of tasks
-    :param mico: average per example if True, else average over t
+    :param micro: average per example if True, else average over t
 
     :return: aggregated report dicts
     """
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -53,6 +53,46 @@ def normalize_answer(s):
     return white_space_fix(remove_articles(remove_punc(lower(s))))
 
 
+def aggregate_task_reports(reports, tasks, micro=True):
+    """
+    Aggregate separate task reports into a single report.
+
+    :param reports: list of report dicts from separate tasks
+    :param tasks: list of tasks
+    :param mico: average per example if True, else average over t
+
+    :return: aggregated report dicts
+    """
+    if len(reports) == 1:
+        # singular task
+        return reports[0]
+    # multiple tasks, aggregate metrics
+    metrics = {}
+    exs = {}
+    total_report = {'tasks': {}}
+    # collect metrics from all reports
+    for i, report in enumerate(reports):
+        total_report['tasks'][tasks[i]] = report
+        for metric, val in report.items():
+            if metric == 'exs':
+                exs[tasks[i]] = val
+            else:
+                metrics.setdefault(metric, {})[tasks[i]] = val
+    # now aggregate
+    total_exs = sum(exs.values())
+    total_report['exs'] = total_exs
+    for metric, task_vals in metrics.items():
+        if micro:
+            # average over the number of examples
+            vals = [task_vals[task] * exs[task] for task in tasks]
+            total_report[metric] = round_sigfigs(sum(vals) / total_exs, 4)
+        else:  # macro
+            # average over tasks
+            vals = task_vals.values()
+            total_report[metric] = round_sigfigs(sum(vals) / len(vals), 4)
+    return total_report
+
+
 def _exact_match(guess, answers):
     """Check if guess is a (normalized) exact match with any answer."""
     if guess is None or answers is None:
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -169,11 +169,6 @@ def aggregate_metrics(reporters):
     if num_tasks > 0:
         for k in sums.keys():
             m[k] = round_sigfigs(sums[k] / num_tasks, 4)
-    m['exs'] = total
-    m['accuracy'] = 0
-    if num_tasks > 0:
-        for k in sums.keys():
-            m[k] = round_sigfigs(sums[k] / num_tasks, 4)
     return m
 
 
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -90,6 +90,11 @@ def aggregate_task_reports(reports, tasks, micro=True):
             # average over tasks
             vals = task_vals.values()
             total_report[metric] = round_sigfigs(sum(vals) / len(vals), 4)
+    # add a warning describing how metrics were averaged across tasks.
+    total_report['warning'] = 'metrics are averaged across tasks'
+    if micro:
+        total_report['warning'] += (' and weighted by the number of examples '
+                                    'per task')
     return total_report
 
 
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -149,21 +149,29 @@ def aggregate_metrics(reporters):
         sums['rouge-L'] = 0.0
     num_tasks = 0
     total = 0
+    agent_metrics = {}
     for i in range(len(reporters)):
-        tid = reporters[i].getID()
-        mt = reporters[i].report()
-        while tid in m['tasks']:
+        task_id = reporters[i].getID()
+        task_report = reporters[i].report()
+        for metric, val in task_report.items():
+            agent_metrics.setdefault(metric, []).append(val)
+        while task_id in m['tasks']:
             # prevent name cloberring if using multiple tasks with same ID
-            tid += '_'
-        m['tasks'][tid] = mt
-        total += mt['exs']
+            task_id += '_'
+        m['tasks'][task_id] = task_report
+        total += task_report['exs']
         found_any = False
         for k in sums.keys():
-            if k in mt:
-                sums[k] += mt[k]
+            if k in task_report:
+                sums[k] += task_report[k]
                 found_any = True
         if found_any:
             num_tasks += 1
+    # aggregate agent reported metrics across tasks (average)
+    if len(reporters) > 0:
+        m['tasks']['all'] = {}
+        for metric, vals in agent_metrics.items():
+            m['tasks']['all'][metric] = round_sigfigs(sum(vals) / len(vals), 4)
     m['exs'] = total
     m['accuracy'] = 0
     if num_tasks > 0:
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -149,14 +149,11 @@ def aggregate_metrics(reporters):
         sums['rouge-L'] = 0.0
     num_tasks = 0
     total = 0
-    agent_metrics = {}
     for i in range(len(reporters)):
         task_id = reporters[i].getID()
         task_report = reporters[i].report()
-        for metric, val in task_report.items():
-            agent_metrics.setdefault(metric, []).append(val)
         while task_id in m['tasks']:
-            # prevent name cloberring if using multiple tasks with same ID
+            # prevent name clobbering if using multiple tasks with same ID
             task_id += '_'
         m['tasks'][task_id] = task_report
         total += task_report['exs']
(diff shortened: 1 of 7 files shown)


parlai/scripts/train_model.py
```diff
@@ -251,25 +257,43 @@ def run_eval(valid_world, opt, datatype, max_exs=-1, write_log=False):
...
-            print(valid_world.report())
-        cnt += valid_world.opt['batchsize']
-        if max_exs > 0 and cnt >= max_exs:
-            # note this max_exs is approximate--some batches won't always be
-            # full depending on the structure of the data
-            break
-    valid_report = valid_world.report()
-    valid_world.reset()  # this makes sure agent doesn't remember valid data
+    reports = []
+    for v_world in valid_worlds:
```
stephenroller: If we kept as a MultiWorld and split this into a recursive function we could address Jason’s concerns from the meeting while also not changing the type of the original variable (making it simpler diff)

emilydinan: sorry i don't understand how that would fix it if we kept as a multiworld?

parlai/scripts/eval_model.py
```diff
@@ -64,42 +64,65 @@ def eval_model(opt, printargs=None, print_parser=None):
...
+        for i, report in enumerate(reports):
+            total_report['tasks'][tasks[i]] = report
+            for metric, val in report.items():
+                agent_metrics.setdefault(metric, []).append(val)
+        total_report['tasks']['all'] = {}
+        for metric, vals in agent_metrics.items():
+            total_report['tasks']['all'][metric] = \
+                round_sigfigs(sum(vals) / len(vals))
+
+        print_announcements(opt)
```
stephenroller: ?

parlai/scripts/eval_model.py
```diff
@@ -64,42 +64,65 @@ def eval_model(opt, printargs=None, print_parser=None):
...
+        # multiple tasks, aggregate metrics
+        agent_metrics = {}
+        total_report = {'tasks': {}}
+        for i, report in enumerate(reports):
+            total_report['tasks'][tasks[i]] = report
+            for metric, val in report.items():
+                agent_metrics.setdefault(metric, []).append(val)
+        total_report['tasks']['all'] = {}
+        for metric, vals in agent_metrics.items():
+            total_report['tasks']['all'][metric] = \
```
stephenroller: does this really need to be wrapped.

parlai/scripts/eval_model.py
```diff
@@ -64,42 +64,65 @@ def eval_model(opt, printargs=None, print_parser=None):
...
+    tasks = opt['task'].split(',')
+    for task in tasks:
+        task_opt = opt.copy()  # copy opt since we're editing the task
+        task_opt['task'] = task
+        world = create_task(task_opt, agent)  # create worlds for tasks
+
+        if print_parser:
+            # Show arguments after loading model
+            print_parser.opt = agent.opt
+            print_parser.print_args()
```
stephenroller: ?

parlai/scripts/eval_model.py
```diff
@@ -34,98 +34,99 @@ def setup_args(parser=None):
     parser.add_argument('-d', '--display-examples', type='bool', default=False)
     parser.add_argument('-ltim', '--log-every-n-secs', type=float, default=2)
     parser.add_argument('--metrics', type=str, default="all",
-                        help="list of metrics to show/compute, e.g. "
-                             "ppl,f1,accuracy,hits@1."
-                             "If 'all' is specified [default] all are shown.")
+                        help='list of metrics to show/compute, e.g. '
```
stephenroller: lol

stephenroller: you're ruthless

parlai/scripts/train_model.py
```diff
@@ -242,6 +241,45 @@ def load_eval_worlds(agent, opt, datatype):
...
+            print(valid_world.report())
+        cnt += valid_world.opt['batchsize']
+
+    valid_report = valid_world.report()
+    valid_world.reset()  # make sure world doesn't remember valid data
+
+    return valid_report
+
+
+def _aggregate_reports(reports, valid_worlds):
```
stephenroller: are these substantially different? Why not just import it from the train_model. (or vice versa)? Should aggregate (where there's already an aggregator) go into metrics?

parlai/scripts/train_model.py
```diff
@@ -242,6 +241,45 @@ def load_eval_worlds(agent, opt, datatype):
...
+    agent_metrics = {}
+    total_report = {'tasks': {}}
+    for i, report in enumerate(reports):
+        total_report['tasks'][tasks[i]] = report
+        for metric, val in report.items():
+            agent_metrics.setdefault(metric, []).append(val)
+    total_report['tasks']['all'] = {}
+    for metric, vals in agent_metrics.items():
+        total_report['tasks']['all'][metric] = round_sigfigs(
+            sum(vals) / len(vals), 4)
```
stephenroller: so much lost precision from round_sigfigs at every level (no change needed in this patch but SAD!)

parlai/scripts/train_model.py
```diff
@@ -242,6 +241,45 @@ def load_eval_worlds(agent, opt, datatype):
...
+
+    return valid_report
+
+
+def _aggregate_reports(reports, valid_worlds):
+    if len(reports) == 1:
+        return reports[0]
+
+    # multiple tasks, aggregate metrics
+    tasks = [world.opt['task'] for world in valid_worlds]
```
stephenroller: i have a weird feeling this is gonna break with -pytd task1,task2

stephenroller: hopefully it shouldn't because we previously disabled -pytd for valid/test

parlai/scripts/train_model.py
```diff
@@ -264,46 +302,23 @@ def run_eval(valid_worlds, opt, datatype, max_exs=-1, write_log=False):
...
-                agent_metrics.setdefault(metric, []).append(val)
-        total_report['tasks']['all'] = {}
-        for metric, vals in agent_metrics.items():
-            total_report['tasks']['all'][metric] = \
-                round_sigfigs(sum(vals) / len(vals), 4)
-        valid_report = total_report
-
-    metrics = datatype + ':' + str(valid_report)
+        task_report = _run_single_eval(opt, v_world,
+                                       max_exs / len(valid_worlds))
```
stephenroller: does this really need to wrap?

Jason already blacked this file; you should black it too. `black -l 88 -S filename.py`

parlai/scripts/train_model.py
```diff
@@ -242,6 +241,45 @@ def load_eval_worlds(agent, opt, datatype):
...
+    valid_world.reset()
+
+    cnt = 0
+    max_cnt = max_exs if max_exs > 0 else float('inf')
+    while not valid_world.epoch_done() and cnt < max_cnt:
+        valid_world.parley()
+        if cnt == 0 and opt['display_examples']:
+            print(valid_world.display() + '\n~~')
+            print(valid_world.report())
+        cnt += valid_world.opt['batchsize']
```
stephenroller: previously this was a call to `.get('batchsize', 1)`, right? do we need that?

parlai/scripts/train_model.py
jaseweston: don't we have eval tasks flag as well?

parlai/scripts/train_model.py
```diff
@@ -475,12 +514,16 @@ def validate(self):
...
-            validation_metric = opt['validation_metric'].split('/')[1]
+        if 'tasks' in valid_report:
+            if '/' in opt['validation_metric']:
+                # if you are multitasking and want your validation metric to be
+                # a metric specific to a subtask, specify your validation metric
+                # as -vmt subtask/metric
+                subtask = opt['validation_metric'].split('/')[0]
+                validation_metric = opt['validation_metric'].split('/')[1]
+            else:
+                subtask = 'all'
```
stephenroller: Previously the meta-averages ended up in the root report, not in a sub dictionary. where do they end up now? is the report gonna end up being a dictionary with exactly one key ('tasks')?

One could just as easily use `--validation-metric all/metric` i guess?

parlai/scripts/train_model.py
```diff
@@ -227,21 +226,66 @@ def load_eval_world(agent, opt, datatype):
...
 
-    valid_world = create_task(opt, valid_agent)
-    return valid_world
+    tasks = opt['task'].split(',')
+    worlds = []
+    for task in tasks:
+        task_opt = opt.copy()  # copy opt since we edit the task
+        task_opt['task'] = task
+        if opt.get('validation_share_agent', False):
+            valid_agent = create_agent_from_shared(agent.share())
```
stephenroller: this is actually a pretttty subtle change; i'm not sure it matters.

Previously (with multiworld running things). We're strictly creating more agents here than previously, and makes me wonder about our discussion in yesterday's meeting.

parlai/scripts/train_model.py
```diff
@@ -188,15 +188,14 @@ def setup_args(parser=None) -> ParlaiParser:
     return parser
 
 
-def _maybe_load_eval_world(agent, opt, datatype):
+def _maybe_load_eval_worlds(agent, opt, datatype):
```
stephenroller: we really need a distributed test :(

stephenroller:
Looks like you broke kvmemnn somehow :-X

parlai/scripts/train_model.py
emilydinan: i didn't change anything here

parlai/scripts/train_model.py
emilydinan: can't put it in metrics because these are being aggregated post-report. can put a utility function in metrics and import it here though. might be cleaner.

parlai/scripts/train_model.py
emilydinan: oof maybe, i'll test

parlai/scripts/train_model.py
emilydinan: i didn't change this code, but i can do that call.

parlai/scripts/train_model.py
emilydinan: i can put it in the root. i guess i wanted to make the distinction clear that this is really an average across tasks, and not a true per-example measure. my solution (putting them in `all`) isn't great either. maybe we can discuss in person

parlai/scripts/train_model.py
stephenroller: it's within a loop; before it wasn't.

parlai/scripts/train_model.py
stephenroller: saul goodman

stephenroller:
cross encoder failing is probably bad luck.

parlai/core/metrics.py
```diff
@@ -53,6 +53,51 @@ def lower(text):
     return white_space_fix(remove_articles(remove_punc(lower(s))))
 
 
+def aggregate_task_reports(reports, tasks, micro=True):
+    """
+    Aggregate separate task reports into a single report.
+
+    :param reports: list of report dicts from separate tasks
+    :param tasks: list of tasks
+    :param mico: average per example if True, else average over t
```
klshuster: nit: micro

parlai/scripts/eval_model.py
```diff
@@ -33,76 +34,89 @@ def setup_args(parser=None):
...
-            print_parser = opt
-        elif print_parser is False:
-            print_parser = None
-    if isinstance(opt, ParlaiParser):
-        print('[ Deprecated Warning: eval_model should be passed opt not Parser ]')
-        opt = opt.parse_args()
-
-    random.seed(42)
+def _eval_single_world(opt, agent, task):
+    print('[ Evaluating task {} on using datatype {}. ] '.format(
```
klshuster: nit: `[ Evaluating task {} using datatype {} ...]` (extra `on`)

PR #1776, opened 2019-06-11
Parallelize tests on CircleCI.

stephenroller:
**Patch description**
Adds parallelism to CircleCI, so unittests run on multiple machines (hopefully much faster)

**Testing steps**
CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,6 @@ jobs:
 
   unittests_36:
     <<: *standard_cpu36
-    parallelism: 4
     working_directory: ~/ParlAI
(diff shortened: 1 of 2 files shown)


stephenroller:
https://circleci.com/gh/facebookresearch/ParlAI/5373 Looks like gpu tests only take 32 minutes now. That's quite an improvement.

dexterju:
#impact 
LGTM! 
Thanks!

stephenroller:
Holding the land for now because there's SOME sort of issue with cloudflare or similar. Could be the public outage, could be something else.

PR #1778, opened 2019-06-12
allow to add a start and end token to the text vec if needed

samhumeau:
**Patch description**
Since the history class has been put in place, I think we can no longer add a start and end token and that all classes that need it (the bert_ranker classes for instance) are doing so a bit artisanally. This allows to override _set_text_vec to add start and end token.

**Testing steps**
Create an agent that overrides _set_text_vec, for instance:
```
    def _set_text_vec(self, *args, **kwargs):
        kwargs['add_start'] = True
        kwargs['add_end'] = True
        return super()._set_text_vec(*args, **kwargs)
```
Observe that start and end tokens are added.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1040,7 +1040,8 @@ class TorchAgent(ABC, Agent):
         else:
             return vec[:truncate]
 
-    def _set_text_vec(self, obs, history, truncate):
+    def _set_text_vec(self, obs, history, truncate, add_start=False,
+                      add_end=False):
         """
         Set the 'text_vec' field in the observation.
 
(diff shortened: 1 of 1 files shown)


stephenroller:
Yeah i'm not opposed to this, but Emily's point is true. You could just override set_text_vec, call super(), and then call the add_start_end.

samhumeau:
Humm, so something like this:

```
def _set_text_vec(self, *args, **kwargs):
        kwargs['add_start'] = True
        kwargs['add_end'] = True
        obs = super()._set_text_vec(*args, **kwargs) 
        if 'text_vec' in obs:
            obs['text_vec'] = self._add_start_end_tokens(obs['text_vec'], True, True)
        return obs
```

It's true... the only argument I have against is that it's weird that this capability (adding start and end token) is there for labels but not for text.

stephenroller:
Ya i agree with you there. Part of it those is these semantics are set by `vectorize`.

https://github.com/facebookresearch/ParlAI/blob/b91e6ade544c060cbb4cd4c662228676fed7dffd/parlai/core/torch_agent.py#L1143-L1163

samhumeau:
Hum, I see. Ok I close this then

PR #1780, opened 2019-06-12
TRA heuristic repeats from conversation blocking

jaseweston:


parlai/core/torch_ranker_agent.py
```diff
@@ -52,6 +52,11 @@ def add_cmdline_args(cls, argparser):
             choices=['batch', 'inline', 'fixed', 'vocab', 'batch-all-cands'],
             help='The source of candidates during evaluation (defaults to the same'
                  'value as --candidates if no flag is given)')
+        agent.add_argument(
+            '--repeat_blocking_heuristic', type='bool', default=True,
```
emilydinan: i think the convention we're using here is `-` instead of `_`

parlai/core/torch_ranker_agent.py
```diff
@@ -313,9 +318,29 @@ def eval_step(self, batch):
             cand_preds_generator = (cand_list[rank] for rank in ordering)
             cand_preds.append(list(islice(cand_preds_generator, max_preds)))
 
+        if (self.opt.get('repeat_blocking_heuristic', False) and
```
emilydinan: why do we do false here if it defaults to true above?

parlai/core/torch_ranker_agent.py
```diff
@@ -313,9 +318,29 @@ def eval_step(self, batch):
...
+            self.opt.get('eval_candidates') == 'fixed'):
+            cand_preds = self.block_repeats(cand_preds)
+
         preds = [cand_preds[i][0] for i in range(batchsize)]
         return Output(preds, cand_preds)
 
+    def block_repeats(self, cand_preds):
+        """Heuristic to block a model repeating a line from the history."""
+        history_strings = []
```
emilydinan: i don't quite understand what this is doing. why do we have to split on new line here?

parlai/core/torch_ranker_agent.py
```diff
@@ -313,9 +318,29 @@ def eval_step(self, batch):
             cand_preds_generator = (cand_list[rank] for rank in ordering)
             cand_preds.append(list(islice(cand_preds_generator, max_preds)))
 
+        if (self.opt.get('repeat_blocking_heuristic', False) and
+            self.opt.get('eval_candidates') == 'fixed'):
+            cand_preds = self.block_repeats(cand_preds)
```
emilydinan: hmm. i think that we should alter "preds" instead of "cand_preds". i.e., maintain the order of cands predicted by the model, but select the prediction that was not repeated. thoughts?

parlai/core/torch_ranker_agent.py
emilydinan: is this for like personas or something?

jaseweston: yes for personas and stuff

parlai/core/torch_ranker_agent.py
jaseweston: if the option isn't there this is backwards compatible to behavior before i guess.. but ok i can change it

parlai/core/torch_ranker_agent.py
jaseweston: well if the convention was '_' everywhere because then it's easier to search in a file to find the same thing...  alex made it so that it converts to '-' in the actual flags..

parlai/core/torch_ranker_agent.py
jaseweston: hmm.. yes.. could do it that way.. not really a big difference i guess

jaseweston: actually thinking about it.. in say comment battle kurt picks from the top N to show possibilities.. so it is best if the repeat is just removed from the cands list

parlai/core/torch_ranker_agent.py
jaseweston: so you're ok with this not being backwards compatible, right?

parlai/core/torch_ranker_agent.py
emilydinan: ah good point

parlai/core/torch_ranker_agent.py
emilydinan: i don't think this will effect anything because if the argument didn't exist beforehand it will default to True anyway

parlai/core/torch_ranker_agent.py
jaseweston: so my suggestion is to change the convention..

parlai/core/torch_ranker_agent.py
jaseweston: oh... what code does that?

parlai/core/torch_ranker_agent.py
stephenroller: you can just grep for repeat.blocking.heuristic to find them both if that's really what you want. 

```bash
$ ag 'add_argument.*--[a-z]+-' | wc -l
383
```

I think the convention is a bit stronger than one preference here...

parlai/core/torch_ranker_agent.py
stephenroller: The parser will insert the default I think.

emilydinan: yep i believe that is correct

jaseweston: so in general this is bad to set a new default which isn't backwards compatible, but done anyway. i wonder if we could have a     default_if_missing=False     in the  agent.add_argument function?

stephenroller: Yeah this is kinda all the domain of `upgrade_opt` that we've been debating.

jaseweston: yes, but this suggestion wouldn't require to write a new function for this standard use-case

PR #1784, opened 2019-06-13
More TRA fixes

jaseweston:
Two TRA fixes:

(1) when no label was present, we still want it to be fast (e.g. in self-play) but it wasn't doing that.

(2) it didn't respect 'ignore_bad_candidates' despite the message, so if there was a label, it would just bug out with a fixed candidate set

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -552,7 +552,6 @@ class TorchRankerAgent(TorchAgent):
                             'set `--ignore-bad-candidates True`.'
                         )
 
-                    
         elif source == 'fixed':
             warn_once(
                 "[ Executing {} mode with a common set of fixed candidates "
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -550,7 +550,7 @@ class TorchRankerAgent(TorchAgent):
                             'At least one of your examples has a set of label candidates '
                             'that does not contain the label. To ignore this error '
                             'set `--ignore-bad-candidates True`.'
-        )
+                        )
 
                     
         elif source == 'fixed':
@@ -588,6 +588,7 @@ class TorchRankerAgent(TorchAgent):
                             'At least one of your examples has a set of label candidates '
                             'that does not contain the label. To ignore this error '
                             'set `--ignore-bad-candidates True`.'
+                        )
 
         elif source == 'vocab':
             warn_once(
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -531,6 +531,7 @@ class TorchRankerAgent(TorchAgent):
             )
             if label_vecs is not None:
                 label_inds = label_vecs.new_empty((batchsize))
+                bad_batch = False
                 for i, label_vec in enumerate(label_vecs):
                     label_vec_pad = label_vec.new_zeros(cand_vecs[i].size(1)).fill_(
                         self.NULL_IDX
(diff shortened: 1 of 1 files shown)


parlai/core/torch_ranker_agent.py
```diff
@@ -387,14 +387,11 @@ def _set_label_cands_vec(self, *args, **kwargs):
...
-            if 'labels' in obs
-            else 'eval_candidates'
-            if 'eval_labels' in obs
-            else None
-        )
-        if cands_key is not None and self.opt[cands_key] not in [
+        if 'labels' in obs:
+            cands_key = 'candidates'
+        else:
+            cands_key = 'eval_candidates'
```
stephenroller: this changes the semantics of above ever so slightly.

parlai/core/torch_ranker_agent.py
```diff
@@ -542,7 +540,19 @@ def _build_candidates(self, batch, source, mode):
                         label_vec = label_vec[0: cand_vecs[i].size(1)]
                     label_vec_pad[0: label_vec.size(0)] = label_vec
                     label_inds[i] = self._find_match(cand_vecs[i], label_vec_pad)
+                    if label_inds[i] == -1:
```
stephenroller: You could do this without changing the semantics of `_find_match`:

```python
try:
    label_inds[i] = self._find_match(cand_vecs[i], label_vec_pad)
except RuntimeError:
    # got a bad batch.
    if opt.get('ignore_bad_candidates'):
        label_inds = None:
    else:
        raise
```

parlai/core/torch_ranker_agent.py
jaseweston: it is meant to

parlai/core/torch_ranker_agent.py
jaseweston: i don't consider that better.
i actually think the way we are computing the loss isn't right, but it was too big a change for now. if we can track which ones we don't have labels for (set to -1) we can compute the loss correctly when labels are missing (they shouldn't be bad, they should just have 0 hits@1)

parlai/core/torch_ranker_agent.py
stephenroller: just checking

parlai/core/torch_ranker_agent.py
```diff
@@ -387,14 +387,11 @@ def _set_label_cands_vec(self, *args, **kwargs):
...
         obs = args[0]
-        cands_key = (
-            'candidates'
-            if 'labels' in obs
-            else 'eval_candidates'
-            if 'eval_labels' in obs
-            else None
-        )
-        if cands_key is not None and self.opt[cands_key] not in [
+        if 'labels' in obs:
```
emilydinan: can you explain this change? is there an instance where 'labels' and 'eval_labels' don't exist but 'label_candidates' do?

parlai/core/torch_ranker_agent.py
```diff
@@ -567,6 +578,17 @@ def _build_candidates(self, batch, source, mode):
                         label_vec = label_vec[0: cand_vecs[i].size(1)]
                     label_vec_pad[0: label_vec.size(0)] = label_vec
                     label_inds[i] = self._find_match(cand_vecs, label_vec_pad)
+                    if label_inds[i] == -1:
+                        bad_batch = True
+                if bad_batch:
+                    if self.opt.get('ignore_bad_candidates'):
+                        label_inds = None
```
emilydinan: this *may* break some things. hopefully this only happens when people are doing this intentionally. theoretically, it should only break training, so maybe we should have a check for `self.is_training` here too (i.e. `if self.opt.get('ignore_bad_candidates') and not self.is_training`)

parlai/core/torch_ranker_agent.py
jaseweston: discussed offline: during interactive chat you dont have a label but do have fixed cands

parlai/core/torch_ranker_agent.py
jaseweston: added, thanks

PR #1787, opened 2019-06-14
BERT for QA

fabiopetroni:
**Patch description**
Added an agent that uses BERT for Question Answering (QA) based on the huggingface implementation.

**Testing steps**
`python examples/train_model.py -m bert_qa -t squad -mf /tmp/bertqa -bs 10 -tr 320 -lr 3e-5 --bert_model "bert-large-uncased" --do_lower_case -eps 2`

[ running eval: test ]
test:{'exs': 10570, 'accuracy': 0.6844, 'f1': 0.7835, 'bleu': 0.1502, 'lr': 3e-05, 'num_updates': 17520, 'examples': 175200, 'loss': 16070.0, 'mean_loss': 0.09174}


parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,421 @@
...
+    @staticmethod
+    def dictionary_class():
+        return BertSpanDictionaryAgent
+
+    def build_model(self):
+        self.model = BertForQuestionAnswering.from_pretrained(
+            self.pretrained_path)
+
+    def init_optim(self, params, optim_states=None, saved_optim_type=None):
+        # self.optimizer = get_bert_optimizer([self.model],
```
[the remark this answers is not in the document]
alexholdenmiller: can you put exactly the training command you ran that didn't decrease the loss here?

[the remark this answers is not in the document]
fabiopetroni: The command is the same (e.g., `python examples/train_model.py -m bert_qa -t squad -mf /tmp/bertqa -bs 24 -tr 200`). The loss does not decreases if I use `get_bert_optimizer` (i.e., uncomment line 362 and comment the rest of the method, lines 363-421 )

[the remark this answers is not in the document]
stephenroller: What's your learning rate? The default in torchagent is set to 1.0, but maybe it's overridden by the BERT add_common_args

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,421 @@
...
+        shared = super().share()
+        shared['model'] = self.model
+        shared['metrics'] = self.metrics
+        return shared
+
+    @staticmethod
+    def add_cmdline_args(parser):
+        TorchAgent.add_cmdline_args(parser)
+        parser = parser.add_argument_group('BERT Classifier Arguments')
+        parser.add_argument('--type-optimization',
```
stephenroller: I wouldn't mind retiring this argument. We haven't seen an instance where you only want to fine tune a subset and it actually work?

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,421 @@
...
+                            type=str,
+                            default='all_encoder_layers',
+                            choices=[
+                                'additional_layers', 'top_layer',
+                                'top4_layers', 'all_encoder_layers', 'all'
+                            ],
+                            help='which part of the encoders do we optimize '
+                            '(defaults to all layers)')
+        parser.add_argument(
+            "--max_text_length",
```
stephenroller: how is this different than --truncate/--text-truncate/--label-truncate?

Our convventions use `--hyphens-instead-of-underscores`.

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,421 @@
...
+            "--bert_model",
+            default='bert-base-cased',
+            type=str,
+            help=
+            "Bert pre-trained model selected in the list: bert-base-uncased, "
+            "bert-large-uncased, bert-base-cased, bert-large-cased, bert-base-multilingual-uncased, "
+            "bert-base-multilingual-cased, bert-base-chinese.",
+        )
+        parser.add_argument(
+            "--warmup_proportion",
```
stephenroller: --warmup-updates is supposed to be this

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,421 @@
...
+            default=0.1,
+            type=float,
+            help=
+            "Proportion of training to perform linear learning rate warmup for. E.g., 0.1 = 10%% "
+            "of training.",
+        )
+        parser.add_argument(
+            "--loss_scale",
```
stephenroller: Would you be willing to just let the optimizer be the default implementation of torchagent? It doesn't have the careful weight decay but you're reimplementing a *lot* of behavior from torchagent.

parlai/agents/bert_qa/bert_qa.py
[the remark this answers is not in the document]
fabiopetroni: I tried different learning rates, I think that's not the issue - the loss does not decrease.

parlai/agents/bert_qa/bert_qa.py
fabiopetroni: I tried torchagent optimizer, but didn't learn.

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,443 @@
...
+    def dictionary_class():
+        return BertSpanDictionaryAgent
+
+    def build_model(self):
+        self.model = BertForQuestionAnswering.from_pretrained(self.pretrained_path)
+        self.n_gpu = 0
+        if self.use_cuda:
+            self.n_gpu = torch.cuda.device_count()
+            if self.n_gpu > 0:
+                self.model = torch.nn.DataParallel(self.model)
```
stephenroller: is there any reason to avoid using DistributedDataParallel?

https://github.com/facebookresearch/ParlAI/blob/27c78c4ebeb599306eccd50f9f08e4d3c20d37bd/parlai/core/torch_generator_agent.py#L410-L413

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,443 @@
...
+            m["mean_loss"] = self.metrics["loss"] / examples
+
+        for k, v in m.items():
+            # clean up: rounds to sigfigs and converts tensors to floats
+            base[k] = round_sigfigs(v, 4)
+        return base
+
+    def train_step(self, batch):
+
+        tensors = self._vectorize(batch["observations"])
```
stephenroller: these should already be available already I think

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,443 @@
...
+        loss = self.model(*tensors)
+
+        if self.n_gpu > 1:
+            loss = loss.mean()  # mean() to average on multi-gpu.
```
stephenroller: distributed mode would avoid this

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,443 @@
...
+
+        self.metrics["examples"] += len(batch["observations"])
+        self.metrics["loss"] += loss
+
+        self.backward(loss)
+        self.update_params()
+        self.zero_grad()
```
stephenroller: zero grad should go at the beginning of train step. this ensures we don't accidentally leak from validation in, and correctly handles gradient accumulation.

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,443 @@
...
+
+        # predictions
+        with torch.no_grad():
```
stephenroller: this doesn't make much sense to me; why wouldn't we want to backpropagate through the model here?

parlai/agents/bert_qa/bert_qa.py
fabiopetroni: I need custom vectors, those in the batch are not usable

parlai/agents/bert_qa/bert_qa.py
alexholdenmiller: rather, should probably collect the b_start_logits and b_end_logits during the first forward pass above rather than doing two forward passes during the train step, one for learning and one for predictions

parlai/agents/bert_qa/bert_qa.py
alexholdenmiller: the other tensors are still being generated, which is wasting computation then and slowing down the training. should find a way to replace the other tokenization then (or do post-processing on it) instead of this approach.

(we can discuss in-person when you are back from PTO)

alexholdenmiller: we may need to make changes to torch agent to make this easier.

parlai/agents/bert_qa/bert_qa.py
```diff
@@ -0,0 +1,443 @@
...
+        _NbestPrediction = collections.namedtuple(  # pylint: disable=invalid-name
+            "NbestPrediction", ["text", "start_logit", "end_logit"]
+        )
+
+        nbest = []
+        for pred in prelim_predictions:
+            if len(nbest) >= self.opt["n_best_size"]:
+                break
+
+            if pred.start_index > 0:  # this is a non-null prediction
```
samhumeau: Can it be <=0 anyway, given l.216?

parlai/agents/bert_qa/bert_qa.py
fabiopetroni: --warmup-updates is an integer. 
--warmup_proportion is a double in [0,1]

parlai/agents/bert_qa/bert_qa.py
fabiopetroni: mmmm, I don't think so. The first token is always [CLS], so a valid start_index should be >= 1

parlai/agents/bert_qa/bert_qa.py
fabiopetroni: unfortunately the forward function returns the loss during training and the logits during evaluation. This is the reason for the double call

PR #1791, opened 2019-06-17
Bi, Poly and Cross encoder not dependent from Hugging Face code 

samhumeau:
**Patch description**
This code intends to reproduce the results of https://arxiv.org/abs/1905.01969, without depending from the code of Hugging Face in order to limit dependencies.

** Steps **

- [x] Reproduce results of the Bi-encoder on convai2
- [x] Reproduce results of the cross-encoder on convai2
- [x] Reproduce results of the poly-encoder on convai2


parlai/agents/transformer/polyencoder.py
```diff
@@ -13,6 +13,7 @@
 from .transformer import TransformerRankerAgent
 from .modules import BasicAttention, MultiHeadAttention
 import torch
+import pdb
```
stephenroller: heads up

parlai/agents/transformer/polyencoder.py
```diff
@@ -0,0 +1,267 @@
...
+            self.rank_loss.cuda()
+        self.data_parallel = opt.get('data_parallel') and self.use_cuda
+        if self.data_parallel:
+            from parlai.core.distributed_utils import is_distributed
+            if is_distributed():
+                raise ValueError(
+                    'Cannot combine --data-parallel and distributed mode'
+                )
+            self.model = torch.nn.DataParallel(self.model)
+        self.START_IDX = self.END_IDX
```
emilydinan: i'd be careful about this call. not sure it will mess anything up, but you might want to fix the dictionary

parlai/agents/transformer/polyencoder.py
```diff
@@ -0,0 +1,267 @@
...
+        kwargs['add_start'] = True
+        kwargs['add_end'] = True
+        obs = super().vectorize(*args, **kwargs)
+        return obs
+
+    def _set_text_vec(self, *args, **kwargs):
+        """ Add start and end tokens for set_text_vec()
+        """
+        obs = super()._set_text_vec(*args, **kwargs)
+        if 'text_vec' in obs:
```
emilydinan: you can also call `_add_start_end_tokens` from torch_agent

parlai/agents/transformer/polyencoder.py
```diff
@@ -0,0 +1,267 @@
...
+            obs['text_vec'] = surround(obs['text_vec'],
+                                       self.START_IDX,
+                                       self.END_IDX)
+
+    def score_candidates(self, batch, cand_vecs, cand_encs=None):
+        bsz = batch.text_vec.size(0)
+        ctxt_rep, ctxt_rep_mask, _ = self.model(ctxt_tokens=batch.text_vec)
+
+        if cand_encs is not None:
+            raise Exception('cand_encs is not yet taken into account')
```
emilydinan: this should be pretty straight forward to handle, maybe just call `cand_rep = cand_encs`? will have to check that it's the appropriate shape. i added coded to do this in parlai_internal, can send if helpful

parlai/agents/transformer/transformer.py
```diff
@@ -49,6 +52,11 @@ def add_common_cmdline_args(argparser):
     argparser.add_argument('--activation', choices={'relu', 'gelu'}, default='relu',
                            help='Nonlinear activation to use. AIAYN uses relu, but '
                                 'more recent papers prefer gelu.')
+    argparser.add_argument('--output-scaling', type=float, default=1.0,
+                           help='scale the output of every transformer by this quantity.')
+    argparser.add_argument('--share_word_embeddings', type='bool', default=True,
```
emilydinan: nit -- we've been using `-` instead of `_`

parlai/agents/transformer/polyencoder.py
samhumeau: This method seems to expect a python list though. Maybe i should rewrite it so that it also handle tensors?

emilydinan: ooo good call... yes i think that might be helfpul. thoughts @stephenroller ?

parlai/agents/transformer/transformer.py
samhumeau: thx

parlai/agents/transformer/polyencoder.py
samhumeau: The same effect can be obtained with `--dict-endtoken __start__`

parlai/agents/transformer/polyencoder.py
```diff
@@ -0,0 +1,267 @@
...
+        ctxt_final_rep = self.attention(cand_embed,
+                                        ctxt_rep,
+                                        ctxt_rep_mask)
+        scores = torch.sum(ctxt_final_rep * cand_embed, 2)
+        return scores
+
+    def forward(self, ctxt_tokens=None, cand_tokens=None,
+                      ctxt_rep=None, ctxt_rep_mask=None, cand_rep=None):
+        """ Due to a limitation of parlai, we have to have one single model
+            in the agent. And because we want to be able to use data-parallel,
```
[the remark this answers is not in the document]
samhumeau: hum. different models? 
I wanted to keep one model so that all the save/load functions of Parlai would work straightforward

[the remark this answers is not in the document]
emilydinan: ah sorry meant to say different calls to the model forward. you could factor out the score function from the forward so that you just call score instead of model(...). see how it is done here for instance: https://github.com/facebookresearch/ParlAI/blob/ba3d6204e6d740ef711d82f6e0375ee5fea4db57/parlai/agents/transformer/transformer.py#L192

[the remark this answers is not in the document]
samhumeau: Your exemple does not need any model parameter in the _score function. _score could basically be static.

Here the score function makes use of attention (multi head attention) which adds a few parameters (query, key and value matrices) which have to be stored in the model.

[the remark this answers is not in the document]
emilydinan: ahhh i see. ok, in this case this seems like the cleanest solution 😐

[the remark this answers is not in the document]
samhumeau: Actually something a bit cleaner that I had tried before would be to add an argument `operation_type`
```
self.model('encode', ...)
self.model('score', ...)
```
Unfortunately this seems to bug with data-parallel when batchsize < num_gpus so I didn't find better than what's there right now

parlai/agents/transformer/polyencoder.py
samhumeau: I updated _add_start_end_tokens to handle tensors

samhumeau:
Ok I feel that we're good for a review. 

Most of the things are explained in that README:
https://github.com/facebookresearch/ParlAI/blob/crossencoder/parlai/zoo/pretrained_transformers/README.md

Things to try:
Discuss with the polyencoder (89+).
```
PYTHONPATH=. python examples/interactive.py -m transformer/polyencoder -mf zoo:pretrained_transformers/model_poly --encode-candidate-vecs true --eval-candidates fixed  --fixed-candidates-path data/models/pretrained_transformers/convai_trainset_cands.txt
```

 Watch out, it expects a syntax very close from the train set so for example try to initiate the conversation with:
```
your persona: i love to drink fancy tea.\nyour persona: i have a big library at home.\nyour persona: i'm a museum tour guide.\nhi how are you doing ?
```

parlai/agents/transformer/polyencoder.py
```diff
@@ -0,0 +1,289 @@
...
+            codes = torch.nn.init.uniform_(codes)
+            self.codes = torch.nn.Parameter(codes)
+
+            # The attention for the codes.
+            if self.codes_attention_type == 'multihead':
+                self.code_attention = MultiHeadAttention(
+                                        self.codes_attention_num_heads,
+                                        embed_dim,
+                                        opt['dropout'])
+            elif self.codes_attention_type == 'basic_sqrt':
```
emilydinan: can we rename the attn param `basic_sqrt` instead of `sqrt` so we just call `self.code_attention = BasicAttention(dim=2, attn=self.codes_attention_type...`

parlai/agents/transformer/polyencoder.py
```diff
@@ -0,0 +1,289 @@
...
+            elif self.codes_attention_type == 'basic':
+                self.code_attention = BasicAttention(dim=2, attn='basic', get_weights=False)
+
+        # The final attention (the one that takes the candidate as key)
+        if self.attention_type == 'multihead':
+            self.attention = MultiHeadAttention(
+                                self.attention_num_heads,
+                                opt['embedding_size'],
+                                opt['dropout'])
+        elif self.attention_type == 'basic_sqrt':
```
emilydinan: same here

parlai/agents/transformer/polyencoder.py
```diff
@@ -0,0 +1,289 @@
...
+        if cand_tokens is not None:
+            assert len(cand_tokens.shape) == 3
+            bsz = cand_tokens.size(0)
+            num_cands = cand_tokens.size(1)
+
+        cand_embed = None
+        ctxt_rep = None
+        ctxt_rep_mask = None
+
+        if cand_tokens is not None:
```
emilydinan: can't we put these under the if statements above (lines 217, 220)

parlai/agents/transformer/polyencoder.py
```diff
@@ -0,0 +1,289 @@
...
+            cand_embed = self.encoder_cand(cand_tokens.view(bsz * num_cands, -1))
+            cand_embed = cand_embed.view(bsz, num_cands, -1)
+
+        if ctxt_tokens is not None:
+            # get context_representation. Now that depends on the cases.
+            ctxt_out, ctxt_mask = self.encoder_ctxt(ctxt_tokens)
+            dim = ctxt_out.size(2)
+
+            if self.type == 'codes':
+                # Basic Attention and MultiHeadAttention share the same API.
```
emilydinan: 😸

samhumeau: Actually got a message from Marie Anne saying that this is not true, I'll modify that

parlai/agents/transformer/modules.py
```diff
@@ -305,6 +320,8 @@ class TransformerEncoder(nn.Module):
     :param variant:
         Which transformer architecture to use. Could be AIAYN or XLM.
         Future versions may support things like GPT-2, ...
+    :param output_scaling:
```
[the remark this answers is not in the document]
samhumeau: Yeah for the polyencoder best results were with this parameteR == 0.06. That's probably because we've used basic instead of basic sqrt

parlai/agents/transformer/modules.py
```diff
@@ -119,6 +139,10 @@ def __init__(self, opt, dictionary):
         self.embeddings = _create_embeddings(
             dictionary, opt['embedding_size'], self.pad_idx
         )
+        if not opt.get('share_word_embeddings'):
```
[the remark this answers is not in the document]
samhumeau: oops, bad merge of a previous PR, sorry

parlai/agents/transformer/modules.py
```diff
@@ -107,6 +108,25 @@ def gelu(tensor):
     return 0.5 * tensor * (1.0 + torch.erf(tensor / math.sqrt(2.0)))
 
 
+def get_n_positions_from_options(opt):
+    if opt is None:
+        return 1024
+    n_positions = 1024
```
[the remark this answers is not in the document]
samhumeau: Well apparently it's useless indeed

parlai/agents/transformer/modules.py
```diff
@@ -127,23 +151,13 @@ def __init__(self, opt, dictionary):
             )
 
         if not opt.get('learn_embeddings'):
+            if not opt.get('share_word_embeddings'):
```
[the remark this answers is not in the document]
samhumeau: Also due to bad merge.

parlai/agents/transformer/polyencoder.py
samhumeau: hum, isn't that not backward compatible?

parlai/agents/transformer/polyencoder.py
samhumeau: Done

parlai/agents/transformer/polyencoder.py
samhumeau: Same question as above, I'd rather not change a parameter of BasicAttention (so far I have only added some)

parlai/agents/transformer/modules.py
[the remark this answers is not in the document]
emilydinan: still need to remove 😄

parlai/zoo/pretrained_transformers/README.md
```diff
@@ -0,0 +1,108 @@
+This directory contains:
```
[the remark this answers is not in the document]
samhumeau: Done

parlai/agents/transformer/crossencoder.py
```diff
@@ -0,0 +1,148 @@
...
+class CrossencoderAgent(TorchRankerAgent):
+    """ Equivalent of bert_ranker/crossencoder but does not rely on an external
+        library (hugging face).
+    """
+
+    @classmethod
+    def add_cmdline_args(cls, argparser):
+        """Add command-line arguments specifically for this agent."""
+        TransformerRankerAgent.add_cmdline_args(argparser)
+        agent = argparser.add_argument_group('Cross Arguments')
```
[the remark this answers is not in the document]
samhumeau: removed

parlai/agents/transformer/polyencoder.py
samhumeau: Dine

parlai/agents/transformer/crossencoder.py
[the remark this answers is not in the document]
klshuster: ahh sorry, i just meant to remove the line `agent = argparser.add_argument_group('Cross Arguments')` - we might still need `TransformerRankerAgent.add_cmdline_args(argparser)`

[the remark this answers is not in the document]
samhumeau: Right, I have been slightly too optimistic there :-)

PR #1793, opened 2019-06-17
[Opt] add ability to pickle this object

emilydinan:
**Patch description**
Opt was not pickleable. Old models pickled opt files, so this caused problems (cc @dexterju). Added `__getstate__`, `__setstate__`, and `__reduce__` methods to allow these objects to be pickled. 

**Testing steps**
Pickle and unpickle opt object.


stephenroller:
What is this black magic?

Should we instead just convert whatever opt dexter is having trouble with?

emilydinan:
> What is this black magic?
> 
> Should we instead just convert whatever opt dexter is having trouble with?
here are the pickle docs which explain what the setstate/getstate/reduce methods do: https://docs.python.org/3/library/pickle.html

this stack overflow post had a pretty good explanation of why we need to implement reduce as well in the case of subclassing a dict: https://stackoverflow.com/questions/21144845/how-can-i-unpickle-a-subclass-of-dict-that-validates-with-setitem-in-pytho

stephenroller:
(ignore py35 ofc)

Issue #1794, opened 2019-06-17
`cross_encoder_ranker` + ConvAI2 parameters


bkj:
The [polyencoder paper](https://arxiv.org/pdf/1905.01969.pdf) discuesses a cross-encoder system that gets very high hits@1 on ConvAI2.  I'm trying to reproduce these experiments using the cross-encoder implementation [here](https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/bert_ranker/cross_encoder_ranker.py).

In issue #1713, I was able to (more or less) successfully reproduce the reported performance of the bi-encoder.  Making a couple of modifications described in the paper (different batch size, longer warmup period, using `inline` candidates), this is my best guess at the correct parameterization:

```bash
python -u parlai/scripts/train_model.py                          \
    --task                      convai2                          \
    --model                     bert_ranker/cross_encoder_ranker \
    --dict-file                 $OUTPATH/dictionary              \
    --model-file                $OUTPATH/model                   \
    --shuffle                   true                             \
    --log_every_n_secs          10                               \
    --eval-batchsize            16                               \
    --data-parallel             true                             \
    --type-optimization         all_encoder_layers               \
    --batchsize                 16                               \
    --learningrate              5e-5                             \
    --history-size              20                               \
    --label-truncate            72                               \
    --text-truncate             360                              \
    --lr-scheduler              reduceonplateau                  \
    --lr-scheduler-patience     0                                \
    --lr-scheduler-decay        0.4                              \
    --validation-every-n-epochs 0.5                              \
    --warmup_updates            1000                             \
    --fp16                      true
```

I'm planning to run this on a p3.16xlarge EC2 instance w/ 8xV100 -- but before I do, does anyone see anything wrong w/ the parameters above?  Or (even better) is it possible to share the scripts that were used to run the experiments in the paper?

Thanks!
~ Ben

stephenroller:
when you launch on 8 gpus with dataparallel, you should be able to 8x that batchsize, maybe even more with the fp16 on?

bkj:
The paper says that they use a batch size of 16 but -- there are 19 distractors for every record in ConvAI1, so the number of records being processed per minibatch is 16 * 20 = 320.  Based on my experiments w/ the bi-encoder, I think that's going to get pretty close to the GPU memory limits.

stephenroller:
Oh yes you're right. I get confused with all the models. :)

stephenroller:
[...]

PR #1795, opened 2019-06-18
allowing to fix the size of the context

samhumeau:
Related to issue #1794, this is a fix that we used for this paper https://arxiv.org/pdf/1905.01969.pdf but did not make his way to public ParlAI yet. 

For the crossencoder in bert_ranker, the context is concatenated with the candidates. However, in case of batching, some padding are inserted between the context and the candidate. This does not change the performance, but leads to results that slightly vary during eval depending on the batch size.

This provides a fix by setting the context width to be constant.

### DIFF
--- a/parlai/agents/bert_ranker/cross_encoder_ranker.py
+++ b/parlai/agents/bert_ranker/cross_encoder_ranker.py
@@ -13,7 +13,6 @@ from .helpers import (BertWrapper, BertModel, get_bert_optimizer,
 
 import os
 import torch
-import pdb
 
 
 class CrossEncoderRankerAgent(TorchRankerAgent):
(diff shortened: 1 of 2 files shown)


parlai/agents/bert_ranker/cross_encoder_ranker.py
```diff
@@ -67,6 +67,15 @@ def score_candidates(self, batch, cand_vecs, cand_encs=None):
         nb_cands = cand_vecs.size()[1]
         size_batch = cand_vecs.size()[0]
         text_vec = batch.text_vec
+
+        if self.opt['fix_size'] != -1:
```
stephenroller: I'm not sure I quite get this (the patch is clear, the need is not). Shouldn't it be bounded by (text-truncate + label_truncate) already?

klshuster: hmm yes wouldn't `text-truncate` take care of limiting the context length?

PR #1798, opened 2019-06-18
Bug in the qualification_flow_example.py

easonnie:
It seems that we need to compare the answer['text'] and the actual answer rather than the answer dictionary.


parlai/mturk/tasks/qualification_flow_example/worlds.py
```diff
@@ -94,7 +94,7 @@ def parley(self):
             }
             self.mturk_agent.observe(validate(ad))
             answer = self.mturk_agent.act()
-            if answer == self.questions[self.curr_question][1]:
+            if answer['text'] == self.questions[self.curr_question][1]:
```
emilydinan: good catch

PR #1799, opened 2019-06-19
loading opts from a json file

jaseweston:
Due to linting, the PR is a bit obscured.
Look for "_load_known_opts"  and "_load_opts", it's only a few lines:

The behavior is you can now load opts from a file, with:

- Any command line arguments have higher priority and will take precedence over  the new file opts. This allows you quickly to load a set of defaults and change one thing in the command line.

- Both opts from file and command line args have priority over model opts that would be loaded from '-mf' flag. That is, the file opts are kind of treated like command line opts in this regard.

To do those things, I needed to get the calls in the right places, so all defaults and overrides worked.. think I've done it right...

Running it:

(base) jase@devfair0328:~/src/ParlAI$ cat /tmp/test.opt
{"task": "convai2", "model": "ir_baseline" }
(base) jase@devfair0328:~/src/ParlAI$ pyd examples/eval_model.py -o /tmp/test.opt
/private/home/jase/miniconda3/lib/python3.6/site-packages/sklearn/externals/joblib/externals/cloudpickle/cloudpi$
kle.py:47: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentati$
n for alternative uses
  import imp
/private/home/jase/src/ParlAI/parlai/core/metrics.py:32: UserWarning: Rouge metrics require py-rouge. Please run
`pip install py-rouge`.
  warn_once('Rouge metrics require py-rouge. Please run `pip install py-rouge`.')
WARNING: model_file unset but model has a `load` function.
[ optional arguments: ]
[  aggregate_micro: True ]
[  display_examples: False ]
[  log_every_n_secs: 2 ]
[  metrics: all ]
[  num_examples: -1 ]
[ Main ParlAI Arguments: ]
[  batchsize: 1 ]
[  datapath: /private/home/jase/src/ParlAI/data ]
[  datatype: valid ]
[  download_path: /private/home/jase/src/ParlAI/downloads ]
[  hide_labels: False ]
[  image_mode: raw ]
[  init_opt: /tmp/test.opt ]
[  multitask_weights: [1] ]
[  numthreads: 1 ]
[  show_advanced_args: False ]
[  task: convai2 ]
[ ParlAI Model Arguments: ]
[  dict_class: None ]
[  init_model: None ]
[  model: ir_baseline ]
[  model_file: None ]
[ PytorchData Arguments: ]
[  batch_length_range: 5 ]
[  batch_sort_cache_type: pop ]
[  batch_sort_field: text ]
[  numworkers: 4 ]
[  pytorch_context_length: -1 ]
[  pytorch_datapath: None ]
[  pytorch_include_labels: True ]
[  pytorch_preprocess: False ]
[  pytorch_teacher_batch_sort: False ]
[  pytorch_teacher_dataset: None ]
[  pytorch_teacher_task: None ]
[  shuffle: False ]
[ Tensorboard Arguments: ]
[  tensorboard_comment:  ]
[  tensorboard_log: False ]
[  tensorboard_metrics: None ]
[  tensorboard_tag: None ]
[ ParlAI Image Preprocessing Arguments: ]
[  image_cropsize: 224 ]
[  image_size: 256 ]
[ IrBaseline Arguments: ]
[  history_size: 1 ]
[  label_candidates_file: None ]
[  length_penalty: 0.5 ]
[ Evaluating task convai2 using datatype valid. ]
[creating task(s): convai2]
[loading fbdialog data:/private/home/jase/src/ParlAI/data/ConvAI2/valid_self_original.txt]
2s elapsed: {'exs': 3388, '%done': '43.43%', 'time_left': '2s', 'accuracy': 0.1594, 'f1': 0.2529, 'hits@1': 0.159, 'hits@5': 0.426, 'hits@10': 0.646, 'hits@100': 1.0, 'bleu': 0.16}
4s elapsed: {'exs': 6762, '%done': '86.68%', 'time_left': '0s', 'accuracy': 0.1569, 'f1': 0.2482, 'hits@1': 0.157, 'hits@5': 0.423, 'hits@10': 0.642, 'hits@100': 1.0, 'bleu': 0.1572}
[ Finished evaluating tasks ['convai2'] using datatype valid ]
{'exs': 7801, 'accuracy': 0.1573, 'f1': 0.2483, 'hits@1': 0.157, 'hits@5': 0.424, 'hits@10': 0.644, 'hits@100': 1.0, 'bleu': 0.1576}

OR

pyd examples/eval_model.py -o /tmp/test.opt --length-penalty 100
..<snip>...
{'exs': 7801, 'accuracy': 0.05563, 'f1': 0.1344, 'hits@1': 0.0556, 'hits@5': 0.263, 'hits@10': 0.531, 'hits@100': 1.0, 'bleu': 0.05571}

or

pyd examples/eval_model.py -o /tmp/test.opt -model repeat_label
..<snip>...
{'exs': 7801, 'accuracy': 0, 'f1': 0.1394, 'bleu': 0.001687}

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -229,8 +229,7 @@ class MultiTaskTeacher(Teacher):
                 if k:
                     opt_singletask = copy.deepcopy(opt)
                     opt_singletask['task'] = k
-                    self.tasks.extend(create_task_agent_from_taskname(
-                        opt_singletask))
+                    self.tasks.extend(create_task_agent_from_taskname(opt_singletask))
         self.task_idx = -1
         self.new_task = True
         self.random = opt.get('datatype') == 'train'
@@ -280,7 +279,8 @@ class MultiTaskTeacher(Teacher):
             if self.random:
                 # select random teacher
                 self.task_idx = random.choices(
-                    self.task_choices, cum_weights=self.cum_task_weights)[0]
+                    self.task_choices, cum_weights=self.cum_task_weights
+                )[0]
             else:
                 # do at most one full loop looking for unfinished task
                 for _ in range(len(self.tasks)):
@@ -352,7 +352,7 @@ def name_to_agent_class(name):
     words = name.split('_')
     class_name = ''
     for w in words:
-        class_name += (w[0].upper() + w[1:])
+        class_name += w[0].upper() + w[1:]
     class_name += 'Agent'
     return class_name
 
@@ -369,13 +369,15 @@ def compare_init_model_opts(opt, curr_opt):
 
     extra_opts = {}
     different_opts = {}
-    exempt_opts = ['model_file', 'dict_file', 'override', 'starttime',
-                   'init_model']
+    exempt_opts = ['model_file', 'dict_file', 'override', 'starttime', 'init_model']
 
     # search through init model opts
     for k, v in init_model_opt.items():
-        if (k not in exempt_opts and k in init_model_opt and
-                init_model_opt[k] != curr_opt.get(k)):
+        if (
+            k not in exempt_opts
+            and k in init_model_opt
+            and init_model_opt[k] != curr_opt.get(k)
+        ):
             if isinstance(v, list):
                 if init_model_opt[k] != list(curr_opt[k]):
                     different_opts[k] = ','.join([str(x) for x in v])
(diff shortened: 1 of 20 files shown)


parlai/core/params.py
```diff
@@ -638,6 +832,24 @@ def parse_known_args(self, args=None, namespace=None, nohelp=False):
             args = [a for a in args if a != '-h' and a != '--help']
         return super().parse_known_args(args, namespace)
 
+    def _load_known_opts(self, optfile, parsed):
+        with open(optfile, 'r', encoding='utf-8') as handle:
```
emilydinan: can you call this existing utility function instead to allow for backwards compatibility with pickled opt files? https://github.com/facebookresearch/ParlAI/blob/9e1a417de7ca7346fd3bc2995bf6c847fb051c58/parlai/core/agents.py#L412

parlai/core/params.py
```diff
@@ -638,6 +832,24 @@ def parse_known_args(self, args=None, namespace=None, nohelp=False):
...
+            new_opt = json.load(handle)
+        for key, value in new_opt.items():
+            # existing commandline parameters take priority.
+            if key not in parsed or parsed[key] is None:
+                parsed[key] = value
+
+    def _load_opts(self, opt):
+        optfile = opt.get('init_opt')
+        with open(optfile, 'r', encoding='utf-8') as handle:
```
emilydinan: same here

parlai/core/params.py
```diff
@@ -638,6 +832,24 @@ def parse_known_args(self, args=None, namespace=None, nohelp=False):
...
+            new_opt = json.load(handle)
+        for key, value in new_opt.items():
+            # existing commandline parameters take priority.
+            if key not in self.opt['override']:
+                self.opt[key] = value
```
emilydinan: quick question: you pass in opt as an arg here, so why are you calling `self.opt` instead of `opt`?

jaseweston: fixed, they're actually the same anyway rn..

samhumeau:
The quantity of lint related refactor here would probably have deserved its own PR :-)

parlai/core/params.py
jaseweston: fixed

jaseweston:
"i don't understand the difference between _load_opts and _load_known_opts`. they look very similar, can you explain?"

The param code (before my PR) does this thing where it first works out which model and task we have in the args, and then pulls in the arguments for those specific to the task and model, etc, so we can parse them correctly.
If you want to set model and task also from the opt file then it needs to load them first there, pull in the args, and then reparse the options file again now it has the options. As this is done in the current code in two places in slightly different ways I just made two functions.  I might be able to sort of merge them, but it will probably be harder to read.

parlai/core/agents.py
```diff
@@ -416,18 +416,6 @@ def compare_init_model_opts(opt, curr_opt):
         print('*' * 75)
 
 
-def _load_opt_file(optfile):
```
emilydinan: this function is called in https://github.com/facebookresearch/ParlAI/blob/cdfa62dfb2eae7bd23bcba02ab42a45f7f5118d6/parlai/agents/bert_classifier/bert_classifier.py#L12

so you'll have to fix the function name & correct import there

jaseweston: done

PR #1801, opened 2019-06-19
Add Black support

stephenroller:
**Patch description**
This PR sets up our environment for black. If you want the optional pre-commit hooks to autorun black on your code, you can run once `pip install pre-commit; pre-commit install`.

Similar to lint, this test will check that you blacked your code before it's allowed to run. It's recommend you just hook it into your editor.

A separate PR will actually *black* the codebase.

We'll get lots and lots and lots of merge conflicts on any partial changes we have now. That will be a little bit of pain, but the easiest way to reduce them is to additionally black any PRs you're working on, which should bring both branches to a consistent format.

**Testing steps**
The CI tests in this repo here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,12 +140,6 @@ jobs:
       - <<: *fixgit
       - <<: *setup
       - run:
-          name: JS Lint
-          working_directory: ~/ParlAI/parlai/mturk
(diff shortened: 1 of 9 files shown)


stephenroller:
Alright, this last one finally did exactly what was needed.

PR #1802, opened 2019-06-19
Black ALL THE THINGS

stephenroller:
**Patch description**
black all the things.jpg)

https://twitter.com/mkbhd/status/778376455373803521

Actually runs black on all the code.

**Testing steps**
Just ran the autoformatter, hopefully nothing dies.

I refuse to run data tests since this should be functionally idempotent.

**Logs**
```
...
All done!
438 files reformatted, 316 files left unchanged.
black .  165.36s user 6.88s system 1272% cpu 13.538 total
```


emilydinan:
omg

stephenroller:
the rest of lint is stuff I won't fix. It's too old and I have no guarantee of its safety.

pringshia:
Maybe we can explicitly exclude the files that are not having lint fixed? Unless IIRC lint_changed only checks against the files touched in the PR... then it should be fine.

pringshia:
Hmm, so I know this PR is onto the `black` branch. I would probably merge in the current `black` branch into master with the black config first before merging these blacked files in. This way, other open PRs can rebase and get just the black config instead of all the blacked files... they can then black everything themselves and then put up a PR. This is to avoid each PR having to deal with resolving merge conflicts when trying to pull in just the black config. Not sure if you already were planning to sequence it this way, but just thought it was worth mentioning.

PR #1803, opened 2019-06-20
[MTurk] Agent state change refactor

JackUrb:
**Patch description**
Following the last cleanup, agents were in an awkward state where they had the ability to manage their own state updates on state changes, but both `mturk_manager` and `worker_manager` were still getting their hands dirty micromanaging the transition. Now `mturk_manager` just tells the agent to update its status, and the agent handles its own state update and request forming. 

Other reasoning comments for the changes appear inline

**Testing steps**
Smoke test by running a sandbox task to completion. More complete (and automated) testing to come upon completion of the migration.


parlai/mturk/core/dev/agents.py
```diff
@@ -14,13 +14,11 @@
 import parlai.mturk.core.dev.shared_utils as shared_utils
 
 # Special act messages for failure states
+# TODO replace with raised errors
 MTURK_DISCONNECT_MESSAGE = '[DISCONNECT]'  # Turker disconnected from conv
 TIMEOUT_MESSAGE = '[TIMEOUT]'  # the Turker did not respond but didn't return
 RETURN_MESSAGE = '[RETURNED]'  # the Turker returned the HIT
 
-# TODO move time management into another class, this way we can handle it
```
JackUrb: This was actually already handled, agents don't manage their own heartbeats anymore

parlai/mturk/core/dev/agents.py
```diff
@@ -46,13 +44,11 @@ def __init__(self, status=None):
             status = self.STATUS_NONE
         self.status = status
         self.messages = []
-        self.last_command = None
```
JackUrb: Last command has been removed, as the state on the heroku server has this detail stored locally already

parlai/mturk/core/dev/agents.py
```diff
@@ -90,76 +80,13 @@ def is_final(self):
             self.status == self.STATUS_DISCONNECT
             or self.status == self.STATUS_DONE
             or self.status == self.STATUS_PARTNER_DISCONNECT
-            or self.status == self.STATUS_PARTNER_DISCONNECT_EARLY
+            or
+            # TODO REMOVE THIS TYPE
```
JackUrb: `PARTNER_DISCONNECT_EARLY` will be phased out in a followup, as it is against MTurk TOS. Instead we will put people back into the queue on failure.

parlai/mturk/core/dev/agents.py
```diff
@@ -90,76 +80,13 @@ def is_final(self):
...
+            self.status == self.STATUS_PARTNER_DISCONNECT_EARLY
             or self.status == self.STATUS_RETURNED
             or self.status == self.STATUS_EXPIRED
         )
 
-    def get_inactive_command_text(self):
```
JackUrb: All of these have been moved to being managed on the server-side in `status_messages.js`. An agent is really only responsible for its own state, not necessarily what we're going to be displaying on the frontend in particular states.

parlai/mturk/core/dev/agents.py
```diff
@@ -177,9 +104,31 @@ class MTurkAgent(Agent):
     def __init__(self, opt, mturk_manager, hit_id, assignment_id, worker_id):
         super().__init__(opt)
 
-        self.conversation_id = None
-        self.mturk_manager = mturk_manager
+        # all MTurkManager functions explicitly used by agents extracted here
```
JackUrb: I wanted `mturk_manager` dependencies to all be explicit rather requiring the full context of a manager in the agent state. (This way it's more obviously going to break if we make a change in that class that breaks here too.)

parlai/mturk/core/dev/agents.py
```diff
@@ -177,9 +104,31 @@ class MTurkAgent(Agent):
...
+        self.m_send_state_change = mturk_manager.send_state_change
+        self.m_send_message = mturk_manager.send_message
+        self.m_handle_turker_timeout = mturk_manager.handle_turker_timeout
+        self.m_send_command = mturk_manager.send_command
+        # TODO update or remove this function
+        self.m_get_agent_work_status = mturk_manager.get_agent_work_status
```
JackUrb: This function should be replaced when I update to a better assignment state tracking mechanism

parlai/mturk/core/dev/agents.py
```diff
@@ -298,52 +243,21 @@ def get_connection_id(self):
         """Returns an appropriate connection_id for this agent"""
         return "{}_{}".format(self.worker_id, self.assignment_id)
 
-    def log_reconnect(self):
```
JackUrb: Reconnects and related state tracking are no longer needed as they are handled on the server-side

parlai/mturk/core/dev/agents.py
```diff
@@ -298,52 +243,21 @@ def get_connection_id(self):
...
-        return False
+        """Simple check for an agent being in a task"""
+        return self.get_status() == AssignState.STATUS_IN_TASK
 
     def observe(self, msg):
         """Send an agent a message through the mturk_manager"""
-        self.mturk_manager.send_message(self.worker_id, self.assignment_id, msg)
+        if 'message_id' not in msg:
+            msg['message_id'] = str(uuid.uuid4())
+        self.state.append_message(msg)
```
JackUrb: I made the agent responsible for its own message tracking, as having this happen in various other classes didn't make sense when all messages were coming through here at some point.

parlai/mturk/core/dev/mturk_manager.py
```diff
@@ -1281,10 +1282,6 @@ def send_message(
             'Manager sending: {}'.format(packet),
             should_print=self.opt['verbose'],
         )
-        # Push outgoing message to the message thread to be able to resend
```
JackUrb: Handled in agents now

parlai/mturk/core/dev/mturk_manager.py
```diff
@@ -1811,93 +1807,8 @@ def _assert_opts(self):
...
-
-    def _on_alive(self, pkt):
-        """Notes a new connection from an agent. In the Static case, this
-        should only be called once per task. If it's called again, the
-        task world will re-send the task details.
+    def _onboard_new_agent(agent):
+        """Override onboarding to go straight to the pool
+        for static stasks
         """
-        shared_utils.print_and_log(logging.DEBUG, 'on_agent_alive: {}'.format(pkt))
```
JackUrb: Turns out this was all duplicated just to be able to map `_onboard_new_agent(agent)` to `add_agent_to_pool(agent)`. Silly silly

parlai/mturk/core/dev/react_server/server/server.js
```diff
@@ -43,23 +46,13 @@ function uuidv4() {
 
 // ===================== <Agent state> ====================
 
-const STATUS_NONE = 'none';
```
JackUrb: Moved to constants file

parlai/mturk/core/dev/worker_manager.py
```diff
@@ -334,20 +335,10 @@ def _get_agent_from_pkt(self, pkt):
             self._log_missing_agent(worker_id, assignment_id)
         return agent
 
-    def change_agent_conversation(self, agent, conversation_id, new_agent_id):
-        """Handle changing a conversation for an agent, takes a callback for
-        when the command is acknowledged
+    def register_to_conv(self, agent, conversation_id):
```
JackUrb: This is the only functionality that `worker_manager` actually needs to be capable or aware of in what used to take place in `change_agent_conversation`.

parlai/mturk/core/dev/react_server/server/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "server.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "node server.js"
+    "start": "node -r esm server.js"
```
stephenroller: this looks like a weird change.

stephenroller:
tho that lint looks like a real bug

PR #1804, opened 2019-06-21
[WIP]: Interactive mode for torch agents 

jaseweston:
Goal, be able to do:

py interactive.py  -mf /pathto/TRA_model  

instead of:

py interactive.py  -mf /pathto/TRA_model  --encode-candidate-vecs True -ecands fixed   -fcp /checkpoint/parlai/zoo/beatthebot/model.candspair2   --ignore_bad_candidates True

..and similarly to avoid having to know about skip-generation for TGA
Basically so you don't need to know about all these flags if you just want to talk to a model, you just point to the model file and go..

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -703,7 +703,7 @@ class TorchAgent(ABC, Agent):
         self.rank_candidates = opt['rank_candidates']
         self.add_person_tokens = opt.get('person_tokens', False)
         # set interactive mode or not according to options.
-        self.interactive_mode(opt['interactive_mode'])
+        self.interactive_mode(opt['interactive_mode'], shared)
 
     def build_dictionary(self):
         """
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -395,13 +395,13 @@ class TorchAgent(ABC, Agent):
             '-i',
             '--interactive-mode',
             type='bool',
-            default=True,
-            help='Whether in full interactive mode or not,  which means generating text or '
-            ' retrieving from a full set of candidates, which is necessary to actually '
-            ' do full dialogue. However, during training or quick validation (e.g. PPL for '
-            ' generation or ranking a few candidates for ranking models) you might want these '
-            ' set to off. '
-            ' Typically, scripts can set their preferred default behavior at the start, '
+            default=False,
+            help='Whether in full interactive mode or not,  which means generating text or'
+            ' retrieving from a full set of candidates, which is necessary to actually'
+            ' do full dialogue. However, during training or quick validation (e.g. PPL for'
+            ' generation or ranking a few candidates for ranking models) you might want these'
+            ' set to off.'
+            ' Typically, scripts can set their preferred default behavior at the start,'
             ' e.g. eval scripts.',
         )
         # pretrained embedding arguments
@@ -1549,7 +1549,12 @@ class TorchAgent(ABC, Agent):
                 with open(path + '.opt', 'w', encoding='utf-8') as handle:
                     if hasattr(self, 'model_version'):
                         self.opt['model_version'] = self.model_version()
-                    json.dump(self.opt, handle)
+                    saved_opts = deepcopy(self.opt)
+                    if 'interactive_mode' in saved_opts:
+                        # We do not save the state of interactive mode, it is only decided
+                        # by scripts or command line.
+                        del saved_opts['interactive_mode']
+                    json.dump(saved_opts, handle)
                     # for convenience of working with jq, make sure there's a newline
                     handle.write('\n')
 
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -391,6 +391,18 @@ class TorchAgent(ABC, Agent):
     def add_cmdline_args(cls, argparser):
         """Add the default commandline args we expect most agents to want."""
         agent = argparser.add_argument_group('TorchAgent Arguments')
+        agent.add_argument(
+            '-i',
+            '--interactive-mode',
+            type='bool', default=True,
+            help='Whether in full interactive mode or not,  which means generating text or '
+            ' retrieving from a full set of candidates, which is necessary to actually '
+            ' do full dialogue. However, during training or quick validation (e.g. PPL for '
+            ' generation or ranking a few candidates for ranking models) you might want these '
+            ' set to off. '
+            ' Typically, scripts can set their preferred default behavior at the start, '
+            ' e.g. eval scripts.'
+        )
         # pretrained embedding arguments
         agent.add_argument(
             '-emb',
@@ -689,7 +701,9 @@ class TorchAgent(ABC, Agent):
         self.is_training = False  # track whether model is training
         self.rank_candidates = opt['rank_candidates']
         self.add_person_tokens = opt.get('person_tokens', False)
-
+        # set interactive mode or not according to options.
+        self.interactive_mode(opt['interactive_mode'])
+        
     def build_dictionary(self):
         """
         Return the constructed dictionary, which will be set to self.dict.
@@ -1616,6 +1630,11 @@ class TorchAgent(ABC, Agent):
     @abstractmethod
     def eval_step(self, batch):
         """[Abstract] Process one batch but do not train on it."""
+        pass
+
+    @abstractmethod
+    def interactive_mode(self, mode):
+        """[Abstract] Set interactive mode on or off."""
         pass
 
     def backward(self, loss):
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1633,9 +1633,9 @@ class TorchAgent(ABC, Agent):
         """[Abstract] Process one batch but do not train on it."""
         pass
 
-    @abstractmethod
-    def interactive_mode(self, mode):
-        """[Abstract] Set interactive mode on or off."""
+    def set_interactive_mode(self, mode, shared):
+        """ Set interactive mode on or off."""
+        # Base class is a no-op.
         pass
 
     def backward(self, loss):
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -703,7 +703,7 @@ class TorchAgent(ABC, Agent):
         self.rank_candidates = opt['rank_candidates']
         self.add_person_tokens = opt.get('person_tokens', False)
         # set interactive mode or not according to options.
-        self.interactive_mode(opt['interactive_mode'], shared)
+        self.set_interactive_mode(opt['interactive_mode'], shared)
 
     def build_dictionary(self):
         """
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -394,14 +394,15 @@ class TorchAgent(ABC, Agent):
         agent.add_argument(
             '-i',
             '--interactive-mode',
-            type='bool', default=True,
+            type='bool',
+            default=True,
             help='Whether in full interactive mode or not,  which means generating text or '
             ' retrieving from a full set of candidates, which is necessary to actually '
             ' do full dialogue. However, during training or quick validation (e.g. PPL for '
             ' generation or ranking a few candidates for ranking models) you might want these '
             ' set to off. '
             ' Typically, scripts can set their preferred default behavior at the start, '
-            ' e.g. eval scripts.'
+            ' e.g. eval scripts.',
         )
         # pretrained embedding arguments
         agent.add_argument(
@@ -703,7 +704,7 @@ class TorchAgent(ABC, Agent):
         self.add_person_tokens = opt.get('person_tokens', False)
         # set interactive mode or not according to options.
         self.interactive_mode(opt['interactive_mode'])
-        
+
     def build_dictionary(self):
         """
         Return the constructed dictionary, which will be set to self.dict.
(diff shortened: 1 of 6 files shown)


parlai/core/torch_agent.py
```diff
@@ -391,6 +391,19 @@ def history_class(cls):
     def add_cmdline_args(cls, argparser):
         """Add the default commandline args we expect most agents to want."""
         agent = argparser.add_argument_group('TorchAgent Arguments')
+        agent.add_argument(
+            '-i',
+            '--interactive-mode',
+            type='bool',
+            default=True,
+            help='Whether in full interactive mode or not,  which means generating text or '
+            ' retrieving from a full set of candidates, which is necessary to actually '
```
emilydinan: nit: you have spaces on each end of the line

parlai/core/torch_agent.py
```diff
@@ -699,8 +714,8 @@ def build_dictionary(self):
         """
         d = self.dictionary_class()(self.opt)
         if self.opt.get('person_tokens'):
-            d[self.P1_TOKEN] = 999999999
-            d[self.P2_TOKEN] = 999999998
+            d[self.P1_TOKEN] = 999_999_999
```
emilydinan: 😮 what is this?

parlai/core/torch_agent.py
```diff
@@ -1618,6 +1633,11 @@ def eval_step(self, batch):
         """[Abstract] Process one batch but do not train on it."""
         pass
 
+    @abstractmethod
+    def interactive_mode(self, mode):
```
emilydinan: shouldn't `shared` be an arg here?

parlai/core/torch_ranker_agent.py
```diff
@@ -176,6 +177,27 @@ def __init__(self, opt, shared=None):
...
+    def interactive_mode(self, mode, shared=False):
+        if mode:
+            if not shared:
+                # Only print in the non-shared version.
+                print("[" + self.id + ': full interactive mode on.' + ']')
+            self.eval_candidates = 'fixed'
+            self.ignore_bad_candidates = True
+            self.encode_candidate_vecs = True
+            if self.opt['fixed_candidates_path'] is None:
+                # Attempt to get a standard candidate set for the given task
```
emilydinan: since we're leaving this as a TODO for now, can you throw an error here that explains that you  need to specify the fixed candidates file?

parlai/core/torch_ranker_agent.py
```diff
@@ -176,6 +177,27 @@ def __init__(self, opt, shared=None):
...
+                # TODO(fill this bit in and built a cand set if it doesn't exist)
+                # path = self.task_candidate_set_path(self.opt['task'])
+                if not shared:
+                    print("[setting fixed_candidates path to: ] " + path)
+                self.fixed_candidates_path = path
+        else:
+            self.eval_candidates = self.opt['eval_candidates']
```
emilydinan: why is `self.eval_candidates` set here and `self.candidates` set in the init? this is a bit confusing

parlai/core/torch_agent.py
stephenroller: python 3.6 feature; the _ is just a comment symbol for more readability.

parlai/core/torch_agent.py
```diff
@@ -391,6 +391,19 @@ def history_class(cls):
     def add_cmdline_args(cls, argparser):
         """Add the default commandline args we expect most agents to want."""
         agent = argparser.add_argument_group('TorchAgent Arguments')
+        agent.add_argument(
+            '-i',
+            '--interactive-mode',
+            type='bool',
+            default=True,
```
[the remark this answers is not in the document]
stephenroller: one very ugly possibility is we could not actually add is as a command line flag in the agent, and instead set it in the scripts? that sounds like some terrible leakage.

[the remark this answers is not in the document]
emilydinan: i think it should probably default to FAlse, right?

parlai/core/torch_ranker_agent.py
```diff
@@ -176,6 +177,27 @@ def __init__(self, opt, shared=None):
...
+            self.eval_candidates = 'fixed'
+            self.ignore_bad_candidates = True
+            self.encode_candidate_vecs = True
+            if self.opt['fixed_candidates_path'] is None:
+                # Attempt to get a standard candidate set for the given task
+                # TODO(fill this bit in and built a cand set if it doesn't exist)
+                # path = self.task_candidate_set_path(self.opt['task'])
+                if not shared:
+                    print("[setting fixed_candidates path to: ] " + path)
+                self.fixed_candidates_path = path
```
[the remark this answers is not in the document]
emilydinan: why would it be different for every turker?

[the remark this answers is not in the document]
stephenroller: i'm sorry; it's different for every instance of the agent (which corresponds to each HIT)

parlai/core/torch_agent.py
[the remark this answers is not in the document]
jaseweston: The scripts should set their preferred default,  train to off, interactive to on

parlai/core/torch_ranker_agent.py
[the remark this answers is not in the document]
jaseweston: why is it different?

parlai/core/torch_ranker_agent.py
jaseweston: this is WIP, i plan to fix it before merge

parlai/core/torch_agent.py
jaseweston: ya, guess so

parlai/core/torch_agent.py
jaseweston: i thought we weren't doing lint. and i ran the black thing.. and you're STILL on my back?? omg omg omg

parlai/core/torch_ranker_agent.py
jaseweston: that one isn't affected by interactive mode, but i can move it if you want

parlai/core/torch_agent.py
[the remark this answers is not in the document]
jaseweston: i thought it might remind you to implement it, no?

parlai/core/torch_agent.py
stephenroller: I think she means that the strings concatenate, so you've written double spaces everywhere. (i.e. a space after "or" and before "retrieving")

parlai/scripts/build_candidates.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Build the candidate responses for a retrieval model.
```
stephenroller: niiice.

parlai/scripts/build_candidates.py
```diff
@@ -0,0 +1,92 @@
...
+def main():
+    random.seed(42)
+    # Get command line arguments
+    parser = ParlaiParser()
+    parser.add_argument(
+        '-n',
+        '--num-examples',
+        default=-1,
+        type=int,
+        help='Total number of exs to convert, -1 to convert \
```
stephenroller: this is going to render really weird.

parlai/scripts/build_candidates.py
```diff
@@ -0,0 +1,92 @@
...
+    else:
+        outfile = opt['outfile']
+
+    if opt['num_examples'] == -1:
+        num_examples = world.num_examples()
+    else:
+        num_examples = opt['num_examples']
+    log_timer = TimeLogger()
+
+    print('[ starting to build candidates from task.. (ex:' +
```
stephenroller: this is good; since we axed py35, you can now write:

```python
print(f'[ starting to build candidates from task.. (ex:{num_examples})]')
```

maybe you like this new syntax; it matters not.

parlai/scripts/build_candidates.py
```diff
@@ -0,0 +1,92 @@
...
+    print('[ saving output to {} ]'.format(outfile))
+    cands = []
+    for _ in range(num_examples):
+        world.parley()
+        acts = world.acts[0]
+        if type(acts) == dict:
+            # turn into a batch of 1 example.
+            acts = [ acts ]
+        for a in acts:
+            candidate = a.get('labels', a.get('eval_labels', None))[0]
```
stephenroller: on the really awful chance that labels and eval_labels are both off, this will error. is that ever possible? probably not.

parlai/core/torch_ranker_agent.py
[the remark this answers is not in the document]
stephenroller: idk, it looks good now to me.

parlai/core/torch_agent.py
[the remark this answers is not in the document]
stephenroller: I think the problem is that whatever gets dumped to disk overrides whatever is the default for the script.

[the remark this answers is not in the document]
jaseweston: ah ok, fixing..

parlai/core/torch_agent.py
```diff
@@ -391,6 +391,19 @@ def history_class(cls):
     def add_cmdline_args(cls, argparser):
         """Add the default commandline args we expect most agents to want."""
         agent = argparser.add_argument_group('TorchAgent Arguments')
+        agent.add_argument(
+            '-i',
+            '--interactive-mode',
```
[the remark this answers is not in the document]
jaseweston: i do not see them with a grep in agents/ or core/ .. .?

parlai/core/torch_agent.py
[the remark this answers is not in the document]
emilydinan: Yeah this is going to break a lot of stuff if we set it default to True

parlai/core/torch_agent.py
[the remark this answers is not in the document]
emilydinan: @jaseweston -- all models that inherit from it don't have it implemented will break now if we make it an abstractmethod. if you think this is good behavior maybe we can set a `#TODO: make required abstract method`

parlai/core/torch_agent.py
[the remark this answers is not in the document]
emilydinan: lots of internal models use it!

parlai/core/torch_ranker_agent.py
emilydinan: yeah i think we should move it to the same place that candidates are set so people can see more clearly how this works. then in the case of interactive_mode we just override that

parlai/core/torch_ranker_agent.py
```diff
@@ -176,6 +177,45 @@ def __init__(self, opt, shared=None):
...
+            self.ignore_bad_candidates = self.get('ignore_bad_candidates', False)
+            self.encode_candidate_vecs = self.opt['encode_candidate_vecs']
+            self.fixed_candidates_path = self.opt['fixed_candidates_path']
+
+
+    def get_task_candidates_path(self):
+        path = self.opt['model_file'] + '.cands-' + self.opt['task'] + '.cands'
+        if os.path.isfile(path) and self.opt['fixed_candidate_vecs'] == 'reuse':
+            return path
+        print("[ building candidates file as they do not exist: " + path + ' ]')
```
emilydinan: sweet this is nice

emilydinan: can we make that print statement a little more loud, in case you provided a filepath and it oculdn't find it?

parlai/core/torch_agent.py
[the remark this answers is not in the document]
jaseweston: ah.. ok .. let me check that

parlai/core/torch_ranker_agent.py
jaseweston: ok

parlai/core/torch_ranker_agent.py
jaseweston: how do i make it loud?  like "**** " ?

parlai/core/torch_ranker_agent.py
```diff
@@ -176,6 +176,44 @@ def __init__(self, opt, shared=None):
...
+            if self.fixed_candidates_path is None or self.fixed_candidates_path == '':
+                # Attempt to get a standard candidate set for the given task
+                path = self.get_task_candidates_path()
+                if path:
+                    if not shared:
+                        print("[setting fixed_candidates path to: " + path + " ]")
+                    self.fixed_candidates_path = path
+        else:
+            self.eval_candidates = self.opt['eval_candidates']
+            self.ignore_bad_candidates = self.get('ignore_bad_candidates', False)
```
emilydinan: self.opt.get

parlai/scripts/build_candidates.py
```diff
@@ -0,0 +1,91 @@
...
+
+    print('[ starting to build candidates from task.. (ex:' + str(num_examples) + ')]')
+    print('[ saving output to {} ]'.format(outfile))
+    cands = []
+    for _ in range(num_examples):
+        world.parley()
+        acts = world.get_acts()[0]
+        if type(acts) == dict:
+            # turn into a batch of 1 example.
+            acts = [acts]
```
emilydinan: a bit confusing because you grab the first act in the list in line 46 (which is presumably a dict) and then turn it back into a list

parlai/core/torch_ranker_agent.py
jaseweston: done

PR #1806, opened 2019-06-25
Add upgrade_opt support.

stephenroller:
**Patch description**
`upgrade_opt` gives you an escape hatch to handle backwards compatible behavior in your options. It's useful for things like deprecating options, renaming them, etc. 

TODO:
- [x] Add some sort of automated test.
- [x] Check the documentation.

**Testing steps**
1. Added support in the BERT classifier
2. Added a unit test.


parlai/core/torch_agent.py
```diff
@@ -690,6 +690,40 @@ def __init__(self, opt, shared=None):
...
+        When loading a previously trained model file, we may wish to ensure
+        backwards compatibility. For example, we may introduce a new option
+        today that wasn't available initially, but we want the default on all
+        new models to be the new behavior.
+
+        ``upgrade_opt`` provides an opportunity for such checks for backwards
+        compatibility. It is called shortly after loading the opt file from
+        disk, and is called before the Agent is initialized.
+
+        Implementations of ``upgrade_opt`` should conform to high standards,
```
emilydinan: 🍸

parlai/agents/bert_classifier/bert_classifier.py
```diff
@@ -106,26 +110,31 @@ def add_cmdline_args(parser):
...
+        """Upgrade opts from older model files."""
+        super(BertClassifierAgent, cls).upgrade_opt(opt_on_disk)
+
+        # 2019-06-25: previous versions of the model did not add a CLS token
+        # to the beginning of text_vec.
+        if 'add_cls_token' not in opt_on_disk:
             warn_once('Old model: overriding `add_cls_token` to False.')
-            opt['add_cls_token'] = False
-        return
+            opt_on_disk['add_cls_token'] = False
```
emilydinan: nice

parlai/core/torch_agent.py
```diff
@@ -690,6 +690,40 @@ def __init__(self, opt, shared=None):
...
+        due to the risk of these methods becoming complicated. We recommend
+        the following behaviors:
+
+            1. ``upgrade_opt`` should only be used to provide backwards
+            compatibility.  Other behavior should find a different behavior.
+            2. Children should always call the parent's ``upgrade_opt`` first.
```
klshuster: any way we can enforce this?

parlai/core/torch_agent.py
```diff
@@ -690,6 +690,40 @@ def __init__(self, opt, shared=None):
...
+            4. Include Comments annotating the date and purpose of each upgrade.
+
+        :param Opt opt_on_disk:
+            The opt file, as loaded from the ``.opt`` file on disk.
+        :return:
+            The modified options
+        :rtype:
+            Opt
+        """
+        # Currently no-op for TorchAgente.
```
klshuster: nit: TorchAgent (unless we're targeting Spanish ParlAI)

stephenroller: 🇪🇸

parlai/core/torch_agent.py
stephenroller: high standards in code review

stephenroller: maybe tho? Pytorch has some way of knowing if you failed to call super().__init__()?

stephenroller: https://github.com/pytorch/pytorch/blob/5b87049c66f0c64b405145ff57869de98729dae9/torch/nn/modules/module.py#L116-L118

here's how pytorch does it. it just looks for an initialized private variable, and then somewhere else that you don't override checks and complains.

parlai/agents/bert_classifier/bert_classifier.py
stephenroller: is opt_from_disk better? opt_on_disk sounds subtly like you're updating disk.

emilydinan: i like opt_from_disk

parlai/core/torch_agent.py
klshuster: I actually kinda like that

parlai/agents/bert_classifier/bert_classifier.py
```diff
@@ -3,6 +3,9 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+
+"""BERT classifier agent uses bert embeddings to make an utterance-level calssification."""
```
emilydinan: spelling of classification

parlai/core/torch_agent.py
```diff
@@ -690,6 +690,40 @@ def __init__(self, opt, shared=None):
...
+
+        Implementations of ``upgrade_opt`` should conform to high standards,
+        due to the risk of these methods becoming complicated. We recommend
+        the following behaviors:
+
+            1. ``upgrade_opt`` should only be used to provide backwards
+            compatibility.  Other behavior should find a different behavior.
+            2. Children should always call the parent's ``upgrade_opt`` first.
+            3. ``upgrade_opt`` should always warn when an option was overwritten.
+            4. Include Comments annotating the date and purpose of each upgrade.
```
emilydinan: why is comments uppercased

parlai/core/torch_agent.py
emilydinan: 🚸

parlai/core/torch_agent.py
stephenroller: i'm not used to this new fancy keyboard lol

parlai/core/torch_agent.py
stephenroller: rofl

stephenroller: This turns out to be not possible here because it's a class agent, so any private variable would be a Singleton. :(

stephenroller:
This is ready for a final review.

1. Ended up concluding that we cannot do the nn.Module trick of checking that the parent was called, because it's a class method. So we would need to modify the return opt to check and that see
2. Putting off legacy param values for another PR. We need to spec it out appropriately.
3. Added a test.

parlai/core/agents.py
```diff
@@ -131,6 +131,40 @@ def shutdown(self):
...
 
+    @classmethod
+    def upgrade_opt(cls, opt_from_disk):
+        """
+        Upgrade legacy options when loading an opt file from disk.
+
+        When loading a previously trained model file, we may wish to ensure
+        backwards compatibility. For example, we may introduce a new option
+        today that wasn't available initially, but we want the default on all
+        new models to be the new behavior.
```
JackUrb: Super nitty doc comment with extended reasoning: This feels a little unclear as it's talking about what the intended effect of `upgrade_opts` implies (all new models can happily use the new default) without explicitly stating the requirement that `upgrade_opts` should satisfy for older models ("To keep the behavior of old models consistent, `upgrade_opts` should disable the new option if it's not present in the loaded opts."). I think this is partially implied by "backwards compatibility," but that term can also accurately be read as "still works without breaking" rather than "works in exactly the way it used to."

tests/test_params.py
```diff
@@ -0,0 +1,56 @@
...
+from parlai.core.params import ParlaiParser
+import parlai.core.agents as agents
+import parlai.core.testing_utils as testing_utils
+
+
+class _ExampleUpgradeOptAgent(agents.Agent):
+    def __init__(self, opt, shared=None):
+        super().__init__(opt)
+        assert 'is_upgraded' in opt
+        assert opt['is_upgraded'] is True
```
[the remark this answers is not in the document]
stephenroller: Probably but god forbid it was like `1` or something. In this case I think I prefer the explicit comparison.

parlai/core/agents.py
```diff
@@ -131,6 +131,40 @@ def shutdown(self):
...
+
+        Implementations of ``upgrade_opt`` should conform to high standards,
+        due to the risk of these methods becoming complicated. We recommend
+        the following behaviors:
+
+            1. ``upgrade_opt`` should only be used to provide backwards
+            compatibility.  Other behavior should find a different behavior.
+            2. Children should always call the parent's ``upgrade_opt`` first.
+            3. ``upgrade_opt`` should always warn when an option was overwritten.
+            4. Include comments annotating the date and purpose of each upgrade.
```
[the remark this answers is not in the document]
stephenroller: lol another check_extra_test. `git diff master | grep upgrade_opt && "CALLING THE UPGRADE_OPT POLICE"`

PR #1808, opened 2019-06-25
[Train Loop] kill it faster

emilydinan:
**Patch description**
Don't save or evaluate if you ctrl-c 🔥 🔥 🔥

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -8,7 +8,7 @@ Worlds are the basic environments which define how agents interact with one anot
 
     ``World(object)`` provides a generic parent class, including ``__enter__``
     and ``__exit__`` statements which allow you to guarantee that the shutdown
-    method is called and KeyboardInterrupts are less noisy (if desired).
+    method is called.
 
     ``DialogPartnerWorld(World)`` provides a two-agent turn-based dialog setting.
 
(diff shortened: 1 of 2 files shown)


stephenroller:
Screen Shot 2019-06-25 at 3 41 49 PM

idk what in the world this is on about tho

parlai/scripts/train_model.py
```diff
@@ -326,11 +325,6 @@ class TrainLoop:
     """TrainLoop contains the core training loop logic."""
 
     def __init__(self, opt):
-        # if python is called from a non-interactive shell, like a bash script,
-        # it will by-default ignore SIGINTs, and KeyboardInterrupt exceptions are
-        # not produced. This line brings them back
-        signal.signal(signal.SIGINT, signal.default_int_handler)
```
stephenroller: hmm so there's one small question around how this works with SLURM.

parlai/core/worlds.py
```diff
@@ -185,10 +185,8 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
-        """After ``with`` statement, call shutdown."""
```
stephenroller: we can leave the docstring

PR #1811, opened 2019-06-25
[TA] sharing model

emilydinan:
**Patch description**
Share model in torch agent.


parlai/core/torch_generator_agent.py
```diff
@@ -29,10 +29,10 @@
 import torch.nn as nn
 import torch.nn.functional as F
 
+from parlai.core.distributed_utils import is_distributed
```
stephenroller: is this you sorting the imports?

PR #1812, opened 2019-06-26
Fix a bug in resuming LR with warmup and reduceonplateau

samhumeau:
**Patch description**
In some conditions, restarting a training from checkpoint would multiply the learning rate by 1e-4 due to a weird behavior of the warmup scheduler.
Those conditions were: 
- ReduceOnPlateau is used
- warmup_updates is used
- TransformerRankerAgent is used


samhumeau:
Sounds good. Added a test that fails on master and succeeds on this branch.

PR #1816, opened 2019-06-28
Change quickstart tutorial to use `-bs 1` instead

stephenroller:
**Patch description**
This is a temporary mitigation for #1805 to update the quickstart docs to suggest users use `--batchsize 1` instead of `--batchsize 8` when training the hogwild memnn.

There's something broken about hogwild with a batchsize on OSX (but not on Linux). I can't begin to fathom why (perhaps @alexholdenmiller has an idea?).

**Testing steps**
Just a doc update.


stephenroller:
The mystery gets deeper...

PR #1817, opened 2019-06-28
[Pretrained transformers] small fix; only save once

emilydinan:
**Patch description**
Every time we used pretrained transformers from the zoo, a call to `torch.save` was being made. This led to a race condition when many processes were trying to load these models at once. Edited to only make this call once.

Eventually will edit the files uploaded to the zoo so that these calls do not have to be made at all.

**Testing steps**
Load up these models using a call to the zoo. Delete folder and try again.

### DIFF
--- a/parlai/agents/transformer/transformer.py
+++ b/parlai/agents/transformer/transformer.py
@@ -166,7 +166,7 @@ class TransformerRankerAgent(TorchRankerAgent):
             '--reduction-type',
             type=str,
             default='mean',
-            choices=['first', 'max', 'mean', 'none'],
+            choices=['first', 'max', 'mean'],
             help='Type of reduction at the end of transformer',
         )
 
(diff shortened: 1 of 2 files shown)


parlai/agents/transformer/transformer.py
```diff
@@ -166,7 +166,7 @@ def add_cmdline_args(cls, argparser):
             '--reduction-type',
             type=str,
             default='mean',
-            choices=['first', 'max', 'mean'],
+            choices=['first', 'max', 'mean', 'none'],
             help='Type of reduction at the end of transformer',
```
stephenroller: Should this really be in there?

emilydinan: we check for it in modules https://github.com/facebookresearch/ParlAI/blob/45b7a9565ce980cad3f29e4747defa66fef96024/parlai/agents/transformer/modules.py#L480

but we can't actual invoke this option since there is a list of choices and a default

PR #1824, opened 2019-07-01
Install `url-loader` into react_server.

easonnie:
`url-loader` need to be install to support `{test: /\.(svg|png)$/, loader: 'url-loader?limit=100000',}` in the `ParlAI/parlai/mturk/core/react_server/webpack.config.js` file.

### DIFF
--- a/parlai/mturk/core/react_server/package.json
+++ b/parlai/mturk/core/react_server/package.json
@@ -28,6 +28,7 @@
     "babel-loader": "^8.0.2",
     "css-loader": "^1.0.0",
     "style-loader": "^0.23.0",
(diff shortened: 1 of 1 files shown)


JackUrb:
Can you add the changes to `webpack.config.js` and any other changes you needed to make your task work to this PR too? (Was the static change in `server_utils` necessary?). Thanks for digging into this and finding out what was going on though!

easonnie:
We can just put everything in the `css` or `components` folder then it will work. `static` change in `server_utils` is not necessary. Sorry about the mess. I made another RP for changes in `webpack.config.js` in #1825.

PR #1826, opened 2019-07-03
Message object

emilydinan:
**Patch description**
Many people experience issues from editing the observation/act dict and not realizing that this may cause problems as these objects are mutable. To avoid having to copy this dict in every observe/act, we print a warning if someone edits one of the special fields.

Left to do:
- [x] Write a unit test
- [x] Ensure all existing tests pass.
- [ ] Test time difference during training
- [ ] Consider adding this to all teachers.
- [ ] Update docs

Tested training a TA model on twitter before and after this change; there is no time difference. [EDIT: still testing..]


parlai/core/utils.py
```diff
@@ -181,6 +181,24 @@ def load_opt_file(optfile):
...
+    """
+    Class for observations and actions.
+
+    Functions like a dict, but warns when writing to fields that already exist.
+    """
+    def __setitem__(self, key, val):
+        if key in self:
+            warn_once(
+                'Message already contains key `{}`. Please make sure '
+                'this change is intentional.'.format(key)
```
JackUrb: I think this warning is a little too weak to inform users of what's going on when they write to these fields. As an end user I might read this and say "yeah, I intended to change the value, that's why I did it" and disregard. Something like a second sentence explaining the risks of mutating a `Message` object before the 'please make sure' might be valuable.

emilydinan: Maybe something along the lines of `Agents may edit fields in unexpected ways`?

JackUrb: I'm thinking more noting possible consequences of making edits to the Message object. Even one concrete example can provide enough context as to why editing this dict can be a bad thing if you don't know what you're doing. `Unexpected ways` doesn't really lead me to understanding when it makes sense to make intentional changes to this and when it could ruin my experiments or cause issues. Of course, its possible that we'll want to go more in-depth into the implementation details somewhere else.

emilydinan:
this is getting to be a bit messy but i don't see a clear/clean way around it

stephenroller:
Current unit test failures appear to be that the test is outdated.

What if we add a bit in `TorchAgent.observe()` that does essentially:

```
if not isinstance(obs, Message):
    raise TypeError(
        "TorchAgent got a Message instead of a dict. This indicates a legacy "
        "teacher or world which hasn't gotten upgraded (see "
        "https://github.com/facebookresearch/ParlAI/pull/1826). Please file "
        "an issue on GitHub with the command you ran."
    )
```

emilydinan:
> ```python
> if not isinstance(obs, Message):
>     raise TypeError(
>         "TorchAgent expects a Message instead of a dict. This indicates a legacy "
>         "teacher or world which hasn't gotten upgraded (see "
>         "https://github.com/facebookresearch/ParlAI/pull/1826). Please file "
>         "an issue on GitHub with the command you ran."
>     )
> ```

i'm not sure we want to do this yet. this would require changing every teacher everywhere.

emilydinan:
ready for review

parlai/core/utils.py
```diff
@@ -181,6 +181,28 @@ def load_opt_file(optfile):
...
+                'Message already contains key `{}`. If this was intentional, '
+                'please use the function `set_key(key, value)`.'.format(key)
+            )
+        super().__setitem__(key, val)
+
+    def force_set(self, key, val):
+        super().__setitem__(key, val)
+
+    def copy(self):
+        return Message(self)
```
stephenroller: I think we're also going to need to support deepcopy.

parlai/core/utils.py
```diff
@@ -1011,7 +1033,7 @@ def convert(key, value):
         if key not in ignore_fields.split(','):
             msg[key] = convert(key, value)
     msg['episode_done'] = msg.get('episode_done', False)
-    return msg
+    return Message(msg)
```
stephenroller: What, does this handle it for FbDialogTeacher or ParlaiDialogTeacher? This is obviously correct, but I don't know who is using this...

tests/test_utils.py
```diff
@@ -122,6 +125,18 @@ def test_argsort(self):
 
         assert np.all(argsort(torch_keys, torch_keys)[0].numpy() == np.arange(1, 6))
 
+    def test_message(self):
```
stephenroller: please rip out into an entire new class (probably test_message.py)

parlai/core/utils.py
emilydinan: yeah parlaidialogteacher uses this

parlai/core/utils.py
emilydinan: it should automatically support deepcopy, there are no additional attributes that need to be copied over except for self.__dict__

emilydinan:
so the controllable dialogue tests are failing because we no longer append the person tokens to the string in the text field in torch agent. rather, we put the edited text with the history inside of "full_text". i'll let @stephenroller weigh in on how i should handle this error instead.

stephenroller:
oh no

parlai/core/message.py
```diff
@@ -17,7 +17,7 @@ def __setitem__(self, key, val):
         if key in self:
             raise RuntimeError(
                 'Message already contains key `{}`. If this was intentional, '
-                'please use the function `set_key(key, value)`.'.format(key)
+                'please use the function `force_set(key, value)`.'.format(key)
```
stephenroller: good, thanks

emilydinan:
i think this is finally ready for review

parlai/core/agents.py
```diff
@@ -495,6 +495,12 @@ def load_agent_module(opt):
                         "previously: {} )]".format(k, v, new_opt.get(k, None))
                     )
                 new_opt[k] = v
+
+        model_class = get_agent_module(new_opt['model'])
+
+        if hasattr(model_class, 'upgrade_opt'):
```
stephenroller: I think we need to separate this into another PR.

emilydinan: ok

PR #1833, opened 2019-07-09
Divide the loss by update_freq when using gradient accumulation.

stephenroller:
**Patch description**
Presently in ParlAI, gradient accumulate just continuously sums gradients into the appropriate `grad` tensors. However, to be roughly more congruent with simply increasing the batch size, fairseq and others actually *average* over the gradient accumulation. This ensures the semantics of a learning rate stay *roughly* fixed over different levels of gradient accumulation.

**Testing steps**
Been using on a private branch for a few weeks now. Doesn't make a huge difference, but I prefer to be congruent with others.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1666,6 +1666,10 @@ class TorchAgent(ABC, Agent):
         loss.backward(), for integration with distributed training and FP16
         training.
         """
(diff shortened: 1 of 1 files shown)


stephenroller:
Seems reasonable to me. Changing the default.

PR #1837, opened 2019-07-09
Fix resume-from-checkpoint bugs.

stephenroller:
**Patch description**
Right now if you're training for a long time and you get preempted on the scheduler and you have `--load-from-checkpoint true`, the train *loop* will load the statistics from `model.checkpoint.trainstats`, but the *model* currently ignores that and loads from just `model`, which is wrong! This is because the model tries to ignore `--init-model` when resuming! It's something of a failure due to the dual use of `--init-model`.

**Testing steps**
~This needs more testing. **Should not be approved yet.** Just raising awareness.~
@klshuster was a hero and took it over. CircleCI is the test now.


stephenroller:
I don't think so :( Too many legacy models expecting --init-model.

klshuster:
@stephenroller I've added a test for this; github won't let me add you as a reviewer, but please review at your convenience

parlai/agents/test_agents/dummy_torch_agent.py
```diff
@@ -0,0 +1,94 @@
...
+        return [self[tok] for tok in txt.split()]
+
+    def save(self, path, sort=False):
+        pass
+
+
+class TorchAgent(TorchAgent):
+    """Use MockDict instead of regular DictionaryAgent."""
+
+    # def __init__(self, opt, shared=None):
```
stephenroller: Strip the dead code if it's unneeded, please.

tests/test_torch_agent.py
```diff
@@ -103,11 +42,13 @@ def get_agent(**kwargs):
 class TestTorchAgent(unittest.TestCase):
```
stephenroller: you could do a the `@unittest.skipIf` around the class instead of doing it on every `def`, and it will skip all of them.

parlai/agents/test_agents/dummy_torch_agent.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Test TorchAgent."""
```
stephenroller: Do you mind adding a line explaining that this is a mock agent for unit testing purposes, and it's not meant to be actually invoked?

parlai/agents/test_agents/dummy_torch_agent.py
klshuster: ahh shoot I meant to delete this, will do

parlai/agents/test_agents/dummy_torch_agent.py
klshuster: yes, and I will probably rename the agent to reflect this

klshuster:
Re-requesting review from @emilydinan given changes

tests/test_torch_agent.py
```diff
@@ -882,6 +815,57 @@ def test_batch_act(self):
...
+                self.assertTrue(is_finetune)
+                valid, test = tl.train()
+                # now, train the model for another epoch
+                opt = get_opt('{}.checkpoint'.format(mf), mf)
+                opt['load_from_checkpoint'] = True
+                popt, tl = get_popt_and_tl(opt)
+                agent = tl.agent
+                init_model_file, is_finetune = agent._get_init_model(popt, None)
+                self.assertEqual(init_model_file, '{}.checkpoint'.format(mf))
+                self.assertFalse(is_finetune)
```
emilydinan: might want to add error messages to these for easy debugging when something breaks

PR #1838, opened 2019-07-09
Report gradient norm and clipping statistics

stephenroller:
**Patch description**
Generally useful metrics to have when training a model (borrowing from fairseq). `gnorm` reports the 2-norm of the parameter weights (treated as a vector). `clip` reports the fraction of updates whose gradient was clipped.

Currently only reported when actually using `--gradient-clip` but we could change that.

**Testing steps**
Been using extensively for a few weeks. Needs another behavior when not using clipping.


parlai/core/torch_agent.py
```diff
@@ -215,6 +215,13 @@ def reset(self):
         self.history_strings = []
         self.history_vecs = []
 
+    def reset_metrics(self):
```
emilydinan: why is this in the History object?????? super().reset_metrics() here isn't going to do anything?

parlai/core/torch_agent.py
```diff
@@ -941,6 +948,11 @@ def report(self):
             current_lr = round_sigfigs(self.optimizer.param_groups[0]['lr'], 4)
             metrics['lr'] = round_sigfigs(current_lr, 4)
         metrics['num_updates'] = self._number_training_updates
+
+        steps = max(1, self.metrics['updates'])  # prevent divide by zero
```
emilydinan: again,  i don't think this is  in the right place. the history objects doesn't have a metrics attribute

stephenroller: O_o what in teh world!

stephenroller:
Alright, ready for another review. no rush.

parlai/core/torch_agent.py
```diff
@@ -1101,6 +1116,14 @@ def share(self):
         Subclasses will likely want to share their model as well.
         """
         shared = super().share()
+
+        if self.opt.get('numthreads', 1) > 1 and isinstance(self.metrics, dict):
+            torch.set_num_threads(1)
```
emilydinan: what does this call do? looks like we weren't calling this in TGA before if `self.opt['numthreads'] > 1`

parlai/core/torch_agent.py
```diff
@@ -941,6 +951,11 @@ def report(self):
             current_lr = round_sigfigs(self.optimizer.param_groups[0]['lr'], 4)
             metrics['lr'] = round_sigfigs(current_lr, 4)
         metrics['num_updates'] = self._number_training_updates
+
+        steps = max(1, self.metrics['updates'])  # prevent divide by zero
+        if steps > 0 and self.opt.get('gradient_clip', -1) > 0:
+            metrics['gnorm'] = round_sigfigs(self.metrics['gnorm'] / steps, 4)
+            metrics['clip'] = round_sigfigs(self.metrics['clip'] / steps, 2)
```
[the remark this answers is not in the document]
stephenroller: 4 is gratuitous

parlai/core/torch_agent.py
stephenroller: It prevents torch from parallelizing the matrix multiplication across cores (resulting in hogwild stomping each other).

We actually already do it in `__init__` so I can drop this line.

stephenroller:
Comments addressed

PR #1839, opened 2019-07-09
Cosine Learning rate

stephenroller:
**Patch description**
Implements Cosine Learning Rate annealing with warm restarts ("SGDR", https://arxiv.org/abs/1608.03983).

Also adds `--lr-period-updates` which can stretch the length of `--lr-scheduler invsqrt` to be different than that of `--warmup-updates`. The default behavior remains the same still.

Needs to be resolved with #1812 still. I don't think it will work when merged into master.

Additionally, I think this code is begging for a refactor now, but not sure if I want to put in the love right now.

**Testing steps**
Desperately needs a unit test. 

**Logs**
```
```

**Other information**

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -504,6 +504,14 @@ class TorchAgent(ABC, Agent):
             'when it is lowered.',
         )
         lr_group.add_argument(
+            '--lr-period-updates',
+            type=int,
+            default=None,
+            hidden=True,
+            help='LR period in updates. Defaults to same as --warmup-updates. '
+            'Used by cosine and invsqrt scheduler.',
+        )
+        lr_group.add_argument(
             '--warmup-updates',
             type=int,
             default=-1,
@@ -855,6 +863,10 @@ class TorchAgent(ABC, Agent):
         else:
             self.warmup_scheduler = None
 
+        lr_period = self.opt.get('lr_period_updates')
+        if lr_period is None or lr_period <= 0:
+            lr_period = self.opt.get('warmup_updates', -1)
+
         patience = self.opt.get('lr_scheduler_patience', 3)
         decay = self.opt.get('lr_scheduler_decay', 0.5)
 
@@ -874,25 +886,24 @@ class TorchAgent(ABC, Agent):
         elif self.opt.get('lr_scheduler') == 'fixed':
             self.scheduler = optim.lr_scheduler.StepLR(optimizer, patience, gamma=decay)
         elif self.opt.get('lr_scheduler') == 'invsqrt':
-            if self.opt.get('warmup_updates', -1) <= 0:
+            if lr_period is None or lr_period <= 0:
                 raise ValueError(
-                    '--lr-scheduler invsqrt requires setting --warmup-updates'
+                    '--lr-scheduler invsqrt requires setting --warmup-updates '
+                    'or --lr-period-updates'
                 )
-            warmup_updates = self.opt['warmup_updates']
-            decay_factor = np.sqrt(max(1, warmup_updates))
+            decay_factor = np.sqrt(max(1, lr_period))
 
             def _invsqrt_lr(step):
                 return decay_factor / np.sqrt(max(1, step))
 
             self.scheduler = optim.lr_scheduler.LambdaLR(optimizer, _invsqrt_lr)
         elif self.opt.get('lr_scheduler') == 'cosine':
-            if self.opt.get('warmup_updates', -1) <= 0:
+            if lr_period is None or lr_period <= 0:
                 raise ValueError(
-                    '--lr-scheduler cosine requires setting --warmup-updates'
+                    '--lr-scheduler cosine requires setting --warmup-updates '
+                    'or --lr-period-updates'
                 )
-            self.scheduler = optim.lr_scheduler.CosineAnnealingLR(
-                optimizer, self.opt['warmup_updates']
-            )
+            self.scheduler = optim.lr_scheduler.CosineAnnealingLR(optimizer, lr_period)
         else:
             raise ValueError(
                 "Don't know what to do with lr_scheduler '{}'".format(
@@ -908,12 +919,10 @@ class TorchAgent(ABC, Agent):
         if (
             # there is already an old LR scheduler saved on disk
             states
-            and
             # and the old LR scheduler is different
-            states.get('lr_scheduler_type') != self.opt['lr_scheduler']
-            and
+            and states.get('lr_scheduler_type') != self.opt['lr_scheduler']
             # and we're not already using a fresh scheduler
-            not hard_reset
+            and not hard_reset
         ):
             # the LR scheduler changed, start things fresh
             warn_once("LR scheduler is different from saved. Starting fresh!")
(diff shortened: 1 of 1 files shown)


parlai/core/torch_agent.py
```diff
@@ -1662,6 +1706,52 @@ def update_params(self):
...
+                passes = np.floor(np.log((lr_mult - 1) * n + 1) / np.log(lr_mult))
+                last_reset_step = (
+                    self._lr_period * (1 - lr_mult ** passes) / (1 - lr_mult)
+                )
+                step = cosine_steps - last_reset_step
+                # cycle_steps contains the total length of this period
+                cycle_steps = self._lr_period * lr_mult ** passes
+                self.scheduler.T_max = cycle_steps
+
+            # the height of the peak of the cosine wave is multiplied by
```
stephenroller: improve this comment

parlai/core/torch_agent.py
```diff
@@ -1662,6 +1706,52 @@ def update_params(self):
...
+            warmup_updates = max(0, self.opt['warmup_updates'])
+            cosine_steps = self._number_training_updates - warmup_updates
+
+            # factor we lengthen the cosine period by
+            lr_mult = self.opt['lr_period_mult']
+
+            # passes contains the number of periods we have iterated
+            # step contains the number of steps relative to the last reset
+            if lr_mult == 1:
+                # special case, do this
```
stephenroller: improve this comment

parlai/core/torch_agent.py
```diff
@@ -420,88 +420,119 @@ def add_cmdline_args(cls, argparser):
...
             'fixed scheduler, LR is decayed every <patience> validations.',
         )
-        agent.add_argument(
+        lr_group.add_argument(
             '--lr-scheduler-decay',
             type=float,
             default=0.5,
-            help='Decay factor for LR scheduler, or how much LR is multiplied by '
-            'when it is lowered.',
+            help=(
```
emilydinan: do you need these parentheses

parlai/core/torch_agent.py
```diff
@@ -420,88 +420,119 @@ def add_cmdline_args(cls, argparser):
...
+                'Decay factor for LR scheduler, or how much LR is multiplied by '
+                'when it is lowered (on plataue with --lr-scheduler reduceonplateau '
+                'and on periods in --lr-scheduler cosine)'
+            ),
         )
-        agent.add_argument(
+        lr_group.add_argument(
+            '--lr-period-mult',
```
emilydinan: would be nice if we were able to have nested args ... maybe some day!

parlai/core/torch_agent.py
```diff
@@ -844,6 +876,10 @@ def _warmup_lr(step):
         else:
             self.warmup_scheduler = None
 
+        self._lr_period = self.opt.get('lr_period_updates')
+        if self._lr_period is None or self._lr_period <= 0:
+            self._lr_period = self.opt.get('warmup_updates', -1)
```
emilydinan: why do we set it to warmup updates here?

parlai/core/torch_agent.py
```diff
@@ -1662,6 +1706,52 @@ def update_params(self):
...
+                step = cosine_steps - last_reset_step
+                # cycle_steps contains the total length of this period
+                cycle_steps = self._lr_period * lr_mult ** passes
+                self.scheduler.T_max = cycle_steps
+
+            # the height of the peak of the cosine wave is multiplied by
+            # decay every period
+            mult_max = self.opt['lr_scheduler_decay'] ** passes
+            lr = self.opt['learningrate']
+            self.scheduler.base_lrs = [lr * mult_max for _ in self.scheduler.base_lrs]
```
emilydinan: what does this do?

parlai/core/torch_agent.py
stephenroller: No, my opinion was that it was just more readable this way. You prefer the other?

parlai/core/torch_agent.py
stephenroller: This is legacy behavior. Currently `--lr-scheduler` uses `warmup_updates` to set the decay rate. This lets the user have separate those values, but still defaults to the old behavior of depending on `--warmup-updates`. (Previously it was actually not possible to use invsqrt without warmup)

parlai/core/torch_agent.py
stephenroller: It's the way you set the peak learning rate. Every time you iterate through a cosine (half period), the coefficient resets to 1, but the max rate decays. I should scare up an illustration.

stephenroller: Okay here's how it looks. This corresponds to

`--lr-scheduler-decay 0.75 --lr-period-mult 2 --lr-period-updates 1000`

Screen Shot 2019-07-10 at 10 51 03 AM

Notice at the end of a period, the peak is multiplied by `0.75` and the length of the new period is doubled.

parlai/core/torch_agent.py
emilydinan: doesn't really matter, other way looks a bit more consistent

parlai/core/torch_agent.py
emilydinan: nice thanks for the picture

emilydinan:
curious, have you tried this and found that it helps?

parlai/core/torch_agent.py
mwillwork: The paper sometimes uses a schedule/not a multiplicative decay factor (i.e. 0.050, 0.025, 0.01 and 0.005) - do we care?

parlai/core/torch_agent.py
mwillwork: Sorry if stupid/intentional - is there a reason we don't incorporate their LR min value from the paper? (Seems like 0 is always used in the arg to the torch CosineAnnealingLR object). Too much trouble to add another config arg?

stephenroller: Laziness tbh. I actually added it, it made a bug, and chose to delete the relevant code rather than figure out my bug. It's real unclear to me what's a right value for eta_min.

@glample you use 0? even though you said you only do a single period...

mwillwork: Well, not a huge deal either way, but 0 seems a bit pointless no? The last few updates (esp the last one) before each restart are a waste in that case

stephenroller: Only one is a waste, the rest are just REALLY tiny ;)

parlai/core/torch_agent.py
stephenroller: I don’t think so. I’m mimicking fairseq for this bit. Both come out to roughly the same. With decay of 0.5 you get (.5, .25, .125. .0625, ...) so it’s the right ballpark

stephenroller:
@glample told me he likes it with —lr-period-updates 1000000. So you never get to the end of the first period... If we set eta min like @mwillwork suggests we could at least have it be smooth. Getting down to that super tiny LR (1e-8) was helpful for a very final part of the learning, but the sudden jump up hurts.

parlai/core/torch_agent.py
mwillwork: Sure - more a question of if there are times when it wouldn't be similar. Say you might want something other than multiplicative (i.e. a defined schedule or log or something like that). I'm happy to leave it though if you are. I need to get calibrated as to what extent we care about such things!

mwillwork:
@stephenroller - What is our philosophy on implementing exactly what's in a cited paper? i.e. I don't mind if you don't see the need for the min value, but yes if we have a min and set a smart decay factor, the restart need not be so jarring.

stephenroller:
I'm abandoning this because:

1. I have higher priority items, but deserves to be properly experimented with;
2. As Mary points out, it's unfaithful in some ways to the original paper, potentially adding confusion and trouble down the line;
3. Learning rate schedulers deserve a refactor so that their behavior can be somewhat isolated from TorchAgent.

Marking on the TODO & Roadmap for future revisiting.

PR #1841, opened 2019-07-10
Option for lenient argparser

JackUrb:
**Patch description**
Sometimes it's appropriate to use a ParlAI argparser in a context where there might be other argparsers, some that may have different possible command line arguments attached to them. Rather than forcing a user to pass their specific argparser along to the downstream class that wants a ParlaiParser, or forcing the user to manipulate arguments after they've parsed out what they wanted, this allows a user to specify that the ParlaiParser shouldn't be so strict as to having to know all of the arguments passed to it. 

**Testing steps**
Ran tests
Used a parser in a script that accepted command line arguments way upstream. The script fails without setting `force_known` to False, and passes when the flag is set false.

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -881,7 +881,7 @@ class ParlaiParser(argparse.ArgumentParser):
 
         return opt
 
-    def parse_args(self, args=None, namespace=None, print_args=True):
+    def parse_args(self, args=None, namespace=None, print_args=True, force_known=True):
         """
         Parse the provided arguments and returns a dictionary of the ``args``.
 
@@ -890,7 +890,10 @@ class ParlaiParser(argparse.ArgumentParser):
         return ``None``.
         """
         self.add_extra_args(args)
-        self.args = super().parse_args(args=args)
+        if force_known:
+            self.args = super().parse_args(args=args)
+        else:
+            self.args, _unknown = super().parse_known_args(args=args)
         self.opt = Opt(vars(self.args))
 
         # custom post-parsing
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -941,6 +941,13 @@ class ParlaiParser(argparse.ArgumentParser):
         self.opt['starttime'] = datetime.datetime.today().strftime('%b%d_%H-%M')
 
     def parse_and_process_known_args(self, args=None):
+        """
+        Parse provided arguments and return parlai opts and unknown arg list.
+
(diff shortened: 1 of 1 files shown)


stephenroller:
This is... fine, but why not use the same add_cmdline_args idiom we use in the rest of ParlAI?

JackUrb:
Definitely open to discussion here. My thought is I'm not sure it makes sense to pollute the downstream opt dict with server configuration options and such, and the idea of throwing a dozen arguments into a parser just to ensure it doesn't yell at me for not understanding some existing arguments seems a bit off.

One such configuration I'm working on involves a run script that uses command line arguments to initialize a server. That server creates game instances, and those game instances initialize agents that want ParlAI opts. The instance class is aware of the underlying architecture running ParlAI, but the server should be ParlAI agnostic.

I can add a warning when using it to note that it can be dangerous (due to the chance that args might change in a breaking way), but there feels like a place for this somewhere. Happy to hear contrasting opinions

emilydinan:
Seems fine to me, but can you explain why this is better than just calling `parse_known_args` on ParlAIParser?

stephenroller:
This is all very abstract. Can you show us examples internally?

JackUrb:
@emilydinan yeah of course - the main difference is in the want to maintain the post processing that `parse_args` does. I'm thinking a better solution would be to abstract that post processing into `_process_parsed_args` or something, then override `parse_known_args` and have both parsing methods point back to the helper.

@stephenroller yeah I'll link some stuff in chat

parlai/core/params.py
```diff
@@ -890,7 +890,10 @@ def parse_args(self, args=None, namespace=None, print_args=True):
         return ``None``.
         """
         self.add_extra_args(args)
-        self.args = super().parse_args(args=args)
+        if force_known:
```
stephenroller: i would feel more comfortable if there was instead a special `parse_known_args_with_postprocessing` that had the behavior you're seeking (force_known=False), and parse_args() keeps it's current behavior. Refactoring out the custom post-parsing would then be the necessary bit.

parlai/core/params.py
```diff
@@ -952,6 +940,24 @@ def parse_args(self, args=None, namespace=None, print_args=True, force_known=Tru
         # add start time of an experiment
         self.opt['starttime'] = datetime.datetime.today().strftime('%b%d_%H-%M')
 
+    def parse_and_process_known_args(self):
```
stephenroller: Please add a docstring :)

PR #1842, opened 2019-07-11
[WoW] Fix sharing for retrieval interactive agent.

emilydinan:
**Patch description**
Fix sharing for wizard interactive retrieval agent.

**Testing steps**
Share  model and create an agent from shared.


stephenroller:
WoW

PR #1846, opened 2019-07-12
Fix distributed initialization of model copies.

stephenroller:
**Patch description**

If you've been using distributed mode without setting a random seed yourself (which we don't do in ParlAI), then all the copies of the model were using different initializations, making learning very difficult: gradients would be synchronized, so models would head in the right direction, but still produce highly varied gradients.

This manifested in a few strange behaviors:
- As the number of distributed copies and model size increased, model training became more difficult.
- Models could appear to overfit extremely quickly, with training loss continuing to go down but validation loss exploding. This is because some hosts would be improving and broadcasting their lower losses, but the validation host (rank 0) would be super far off.
- If training was preempted, then loss would significantly improve very rapidly after resuming, since model weights would be initialized from disk and therefore synchronized.

With synchronized parameters, this issue goes away.

Plan:
- [x] Add fast GPU unittests to CircleCI (same as what runs on our dev machines).
- [x] Add a distributed test, which ensures training happens
- [x] ~Maybe: support CPU distributed?~
- [x] Check synced parameters manually in a test
- [x] Actually fix the distributed sync bug.

**Testing steps**
Fixed manually once, saw issues go away. The finalized version of this PR should contain unit tests that will catch all these issues, and add distributed tests.

**Logs**
```
$ python tests/test_distributed.py
/private/home/roller/working/parlai/parlai/agents/transformer/transformer.py:18: UserWarning: Public release transformer models are currently in beta. The name of command line options may change or
disappear before a stable release. We welcome your feedback. Please file feedback as issues at https://github.com/facebookresearch/ParlAI/issues/new
  "Public release transformer models are currently in beta. The name of "
/private/home/roller/working/parlai/parlai/scripts/train_model.py:696: UserWarning: Distributed training outputs average-per-worker metrics during training, and may be slightly distorted. Validation/test are unadulterated.
  "Distributed training outputs average-per-worker metrics during "
.
Ran 1 test in 13.685s

OK
```

### DIFF
--- a/.circleci/triggers.py
+++ b/.circleci/triggers.py
@@ -13,7 +13,7 @@ words into your commit messages.
 [gpu]: Run the nightly GPU tests
 [mturk]: Run the mturk tests
 [data]: Run the data tests
-[long]: run all above
+[long] or [all]: run all above
 """
 
 import parlai.core.testing_utils as testing_utils
@@ -21,7 +21,7 @@ import parlai.core.testing_utils as testing_utils
 
 def detect_all():
     """Check if we should run all tests."""
-    return '[long]' in testing_utils.git_commit_messages()
+    return any(kw in testing_utils.git_commit_messages() for kw in ['[all]', '[long'])
 
 
 def detect_gpu():
--- a/.circleci/triggers.py
+++ b/.circleci/triggers.py
@@ -21,7 +21,7 @@ import parlai.core.testing_utils as testing_utils
 
 def detect_all():
     """Check if we should run all tests."""
-    return any(kw in testing_utils.git_commit_messages() for kw in ['[all]', '[long'])
+    return any(kw in testing_utils.git_commit_messages() for kw in ['[all]', '[long]'])
 
 
 def detect_gpu():
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -378,6 +378,7 @@ class TorchGeneratorAgent(TorchAgent):
 
             self.build_criterion()
             with torch.random.fork_rng(range(torch.cuda.device_count())):
+                # TODO: turn this into an option
                 torch.manual_seed(42)
                 self.build_model()
             check_synced_parameters(self.model)
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -29,7 +29,7 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
-from parlai.core.distributed_utils import is_distributed
+from parlai.core.distributed_utils import is_distributed, check_synced_parameters
 from parlai.core.thread_utils import SharedTable
 from parlai.core.torch_agent import TorchAgent, Batch, Output
 from parlai.core.utils import padded_tensor, round_sigfigs, warn_once, neginf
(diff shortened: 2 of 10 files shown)


tests/test_distributed.py
```diff
@@ -0,0 +1,92 @@
...
+                parser = build_dict.setup_args()
+                build_dict.build_dict(popt)
+
+                valid, test = mp_train.launch_and_train(popt, 31337)
+
+        return (output.getvalue(), valid, test)
+
+    def tearDown(self):
+        # we need to de-initialize the distributed world, otherwise other
+        # tests will they're we're distributed when we're really not.
```
stephenroller: Sp

.circleci/triggers.py
```diff
@@ -13,15 +13,15 @@
...
+[long] or [all]: run all above
 """
 
 import parlai.core.testing_utils as testing_utils
 
 
 def detect_all():
     """Check if we should run all tests."""
-    return '[long]' in testing_utils.git_commit_messages()
+    return any(kw in testing_utils.git_commit_messages() for kw in ['[all]', '[long'])
```
klshuster: nit: believe you want `'[long]']` rather than `'[long']` (though i suppose it'll work regardless)

parlai/core/torch_generator_agent.py
```diff
@@ -377,7 +377,10 @@ def __init__(self, opt, shared=None):
                 print('[ Saving dot beam logs in {} ]'.format(self.beam_dot_dir))
 
             self.build_criterion()
-            self.build_model()
+            with torch.random.fork_rng(range(torch.cuda.device_count())):
+                torch.manual_seed(42)
```
klshuster: should we make this seed configurable? or is that too dangerous...

parlai/scripts/multiprocessing_train.py
```diff
@@ -62,50 +64,60 @@ def multiprocess_train(rank, opt, port=61337, gpu=None, hostname='localhost'):
...
     spawncontext = torch.multiprocessing.spawn(
         multiprocess_train,
         (opt, port),
-        nprocs=opt['distributed_world_size'],
+        nprocs=opt['distributed_world_size'] - 1,
         join=False,
     )
 
     try:
+        # rank is offset by -1 to make the root process be rank 0
```
klshuster: can you elaborate a little bit why you offset by -1 here and then offset by +1 above?

tests/test_distributed.py
```diff
@@ -0,0 +1,92 @@
...
+                n_layers=1,
+                n_heads=1,
+                ffn_size=32,
+                embedding_size=32,
+                beam_size=1,
+            )
+        )
+
+        self.assertLessEqual(
+            valid['ppl'], 1.20, "valid ppl = {}\nLOG:\n{}".format(valid['ppl'], stdout)
```
klshuster: pretty solid ppl if you ask me

tests/test_pytorch_data_teacher.py
```diff
@@ -56,6 +56,7 @@ def solved_task(str_output, valid, test):
     )
 
 
+@testing_utils.skipIfGPU
```
klshuster: is this broken on gpu?

.circleci/triggers.py
stephenroller: lol good catch

parlai/core/torch_generator_agent.py
stephenroller: probably

parlai/scripts/multiprocessing_train.py
stephenroller: Sure. It should be added as a comment then.

the calls to spawn() always start at 0, and that's nonconfigurable. However, we want rank 0 to be the main thread, so that we can successfully capture the response and return the retval properly.

tests/test_distributed.py
stephenroller: It's a repeat-after-me task. It's trivial to solve.

tests/test_pytorch_data_teacher.py
stephenroller: No, but it doesn't mix well with the multiprocessing_train. (Multiprocessing_train changes the spawn method which creates issues).

parlai/core/torch_generator_agent.py
stephenroller: left a TODO comment

PR #1848, opened 2019-07-16
Fix empathetic dialogue description in task list.

stephenroller:
**Patch description**
Docs were rendering in correctly, mostly due to differences in expectations about restructured text and how it renders things.

This patches that output.

**Testing steps**
Screen Shot 2019-07-16 at 2 56 54 PM


EricMichaelSmith:
Looks good!

PR #1855, opened 2019-07-17
[MTurk] Better HIT management on shutdown

JackUrb:
**Patch description**
Shutting down HITs was a painful process in the past due to how hit status is managed. Unfortunately, after this PR it's still a painful process that can take up to the full duration of the HIT that you launched would take (as if a worker still has the assignment in their queue, we can't expire and shutdown so we have to wait for that HIT to expire). The real difference here is that now the shutdown is guaranteed to happen. 

I did some experiments with HIT state transitions, and have confirmed that we don't need to be expiring HITs repeatedly, but rather just be checking them ad-infinitum (at least until we've elapsed the duration that your hits would be running). 

The content of this PR line-wise is split between cleaning up the expiration and hit checking functions, and better threading management of the HIT status thread.

**Testing steps**
Running qa_data_collection in 3 different forms -
Complete the task: shutdown happens immediately
Return the task: shutdown happens after a few seconds
Accept the task, then exit: shutdown takes 30 mins waiting for the task to expire naturally.
All of these exited cleanly.


parlai/mturk/core/dev/mturk_manager.py
```diff
@@ -1195,23 +1191,32 @@ def _wait_for_task_expirations(self):
         """
         start_time = time.time()
         min_wait = self.opt['assignment_duration_in_seconds']
+        wait_time = 0.1
         while time.time() - start_time < min_wait and len(self.hit_id_list) > 0:
-            self.expire_all_unassigned_hits()
-            time.sleep(max(self.opt['assignment_duration_in_seconds'] / 60, 0.1))
+            for hit_id in self.hit_id_list.copy():
+                self.check_hit_status(hit_id)
+            wait_time *= 1.5
```
emilydinan: interesting. why do you wait longer and longer each time?

JackUrb: I'm using [Exponential Backoff](https://en.wikipedia.org/wiki/Exponential_backoff) to prevent overloading the MTurk API's request system. Being stuck here means the assignment is waiting in someone's task queue, so there's no good way of knowing when it'll expire. The previous version would sometimes exceed the rate limit.

parlai/mturk/core/dev/mturk_manager.py
```diff
@@ -1598,22 +1604,37 @@ def get_assignments_for_hit(self, hit_id):
         assignments_info = client.list_assignments_for_hit(HITId=hit_id)
         return assignments_info.get('Assignments', [])
 
+    # FIXME move this into mturk_utils
+    def check_hit_status(self, hit_id):
```
klshuster: just a note that the side effect of removing the hit from the hit_id list is a little ambiguous given the function name (I figured this was happening based on code above but was not sure until examining the function)

parlai/mturk/core/dev/mturk_manager.py
```diff
@@ -1598,22 +1604,37 @@ def get_assignments_for_hit(self, hit_id):
...
             should_print=not self.is_test,
         )
-        completed_ids = self.worker_manager.get_complete_hits()
         for hit_id in self.hit_id_list:
-            if hit_id not in completed_ids:
-                # TODO get confirmation that the HIT is acutally expired
-                mturk_utils.expire_hit(self.is_sandbox, hit_id)
+            mturk_utils.expire_hit(self.is_sandbox, hit_id)
+
+    def expire_all_hits(self):
```
klshuster: how does this function differ from the one directly above it? they appear to be removing from the same `self.hit_id_list`

JackUrb: Definitely a bug - fixing!

parlai/mturk/core/dev/mturk_manager.py
JackUrb: Yeah this is somewhat unclear from the name. Picking a better one.

stephenroller:
Merge in master to fix tests.

PR #1856, opened 2019-07-17
Explicit metric shortcutting with --metrics

dexterju:
**Patch description**
This allows us to select what metrics we want to compute, not only filtering them, but also preventing them from being computed.
For all :  `['correct', 'rouge-1', 'rouge-2', 'rouge-L', 'bleu' 'accuracy', 'f1', 'loss', 'ppl']`
For default : ` ['correct', 'bleu', 'accuracy', 'f1', 'loss', 'ppl']`
'correct' is always added since it is required for computing accuracy, and we don't want dependency in metrics. 

Usage: You can choose 'all', 'default', or you can add your metrics you like as 'rouge-1', 'bleu'.

**Testing steps**
```

python examples/train_model.py -m transformer/generator  -t quac  -bs 32 -mf /tmp/tg01  --truncate 1024 --metrics all
{'exs': 384, 'lr': 1, 'num_updates': 22, 'loss': 112.6, 'token_acc': 0.05282, 'nll_loss': 9.398, 'ppl': 12060.0}


python examples/train_model.py -m transformer/generator  -t quac  -bs 32 -mf /tmp/tg01  --truncate 1024 
 time:6.0s total_exs:1056 epochs:0.01 ] {'exs': 384, 'lr': 1, 'num_updates': 33, 'loss': 107.9, 'token_acc': 0.05236, 'nll_loss': 8.995, 'ppl': 8067.0}
```

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -10,8 +10,9 @@ Uses locking and shared memory when ``numthreads`` is set to >1 to share metrics
 between processes.
 """
 
-DEFAULT_METRICS = ['correct', 'bleu', 'accuracy', 'f1']
-ALL__METRICS = set(DEFAULT_METRICS + ['rouge-1', 'rouge-2', 'rouge-L'])
+DEFAULT_METRICS = set(['correct', 'bleu', 'accuracy', 'f1'])
+ROUGE__METRICS = set(['rouge-1', 'rouge-2', 'rouge-L'])
+ALL__METRICS = set(DEFAULT_METRICS | ROUGE__METRICS)
 
 from parlai.core.thread_utils import SharedTable
 from parlai.core.utils import round_sigfigs, no_lock
@@ -183,8 +184,8 @@ def _rouge(guess, answers):
 
     scores_rouge1 = [score['rouge-1']['r'] for score in scores]
     scores_rouge2 = [score['rouge-2']['r'] for score in scores]
-    scores_rougel = [score['rouge-l']['r'] for score in scores]
-    return max(scores_rouge1), max(scores_rouge2), max(scores_rougel)
+    scores_rougeL = [score['rouge-l']['r'] for score in scores]
+    return max(scores_rouge1), max(scores_rouge2), max(scores_rougeL)
 
 
 def aggregate_metrics(reporters):
@@ -192,15 +193,20 @@ def aggregate_metrics(reporters):
     # reporters is a list of teachers or worlds
     m = {}
     m['tasks'] = {}
-    sums = {'accuracy': 0, 'f1': 0, 'loss': 0, 'ppl': 0}
-    if nltkbleu is not None:
-        sums['bleu'] = 0
-    if rouge is not None:
-        sums['rouge-1'] = 0.0
-        sums['rouge-2'] = 0.0
-        sums['rouge-L'] = 0.0
+    sums = {}
     num_tasks = 0
     total = 0
+    for i in range(len(reporters)):
+        task_id = reporters[i].getID()
+        task_report = reporters[i].report()
+        for each_metric, value in task_report.items():
+            if isinstance(value, float):
+                sums[each_metric] = 0.0
+                m[each_metric] = 0.0
+            elif isinstance(value, Number):
+                sums[each_metric] = 0
+                m[each_metric] = 0
+
     for i in range(len(reporters)):
         task_id = reporters[i].getID()
         task_report = reporters[i].report()
@@ -232,8 +238,6 @@ class Metrics(object):
         self.metrics['cnt'] = 0
         self.metrics_list = set()
         optional_metrics_list = []
-        self.compute_rouge_n = 0
-        self.compute_rouge_l = False
         metrics_arg = opt.get('metrics', 'default')
         if metrics_arg == 'default':
             optional_metrics_list = DEFAULT_METRICS
@@ -244,14 +248,7 @@ class Metrics(object):
             optional_metrics_list.add('correct')
         for each_m in optional_metrics_list:
             if each_m.startswith('rouge') and rouge is not None:
-                self.metrics_list.add(each_m)
-                if each_m == 'rouge-L':
-                    self.compute_rouge_l = True
-                elif each_m.split('-')[-1].isdigit():
-                    if int(each_m.split('-')[-1]) > self.compute_rouge_n:
-                        self.compute_rouge_n = int(each_m.split('-')[-1])
-                        for each_rouge_idx in range(1, self.compute_rouge_n):
-                            self.metrics_list.add('rouge-{}'.format(each_rouge_idx))
+                self.metrics_list = self.metrics_list | ROUGE__METRICS
             elif each_m == 'bleu' and nltkbleu is None:
                 # only compute bleu if we can
                 pass
@@ -332,10 +329,8 @@ class Metrics(object):
                 f1 = _f1_score(prediction, labels)
             if 'bleu' in self.metrics_list:
                 bleu = _bleu(prediction, labels)
-            if self.compute_rouge_l or self.compute_rouge_n > 0:
-                rougen, rougel = _rouge(
-                    prediction, labels, self.compute_rouge_n, self.compute_rouge_l
-                )
+            if 'rouge-L' in self.metrics_list:
+                rouge1, rouge2, rougeL = _rouge(prediction, labels)
 
             with self._lock():
                 if 'f1' in self.metrics:
@@ -344,13 +339,13 @@ class Metrics(object):
                 if 'bleu' in self.metrics:
                     self.metrics['bleu'] += bleu
                     self.metrics['bleu_cnt'] += 1
-                if self.compute_rouge_l:
-                    self.metrics['rouge-L'] += rougel
+                if 'rouge-L' in self.metrics:
+                    self.metrics['rouge-1'] += rouge1
+                    self.metrics['rouge-1_cnt'] += 1
+                    self.metrics['rouge-2'] += rouge2
+                    self.metrics['rouge-2_cnt'] += 1
+                    self.metrics['rouge-L'] += rougeL
                     self.metrics['rouge-L_cnt'] += 1
-                if self.compute_rouge_n:
-                    for idx in range(1, self.compute_rouge_n):
-                        self.metrics['rouge-{}'.format(idx)] += rougen[idx - 1]
-                        self.metrics['rouge-{}_cnt'.format(idx)] += 1
 
         # Ranking metrics.
         self._update_ranking_metrics(observation, labels)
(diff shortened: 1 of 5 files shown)


parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -31,11 +31,17 @@
 
 
 def _first_val(dictionary):
-    return list(dictionary.values())[0]
+    vals = list(dictionary.values())
```
dexterju: @stephenroller did you add that when you were helping me during the debugging?

stephenroller: @emilydinan  did this in #1849.

parlai/core/metrics.py
```diff
@@ -10,6 +10,9 @@
 between processes.
 """
 
+__ALL__METRICS__ = ['correct', 'rouge-1', 'rouge-2', 'rouge-L', 'bleu' 'accuracy', 'f1', 'loss', 'ppl']
```
stephenroller: `__` prefixes are only for hidden variables that you don't want to expose. It's not unreasonable to have these variables with `ALL_METRICS` and `DEFAULT_METRICS`.

loss/ppl shouldn't be in this list; they are agent-level metrics.

ALL_METRICS should be 'DEFAULT_METRICS + ['rouge-1', ...]`

parlai/core/metrics.py
```diff
@@ -159,12 +162,17 @@ def _bleu(guess, answers):
     )
 
 
-def _rouge(guess, answers):
+def _rouge(guess, answers, rougen, rougel):
```
stephenroller: `rougeN` or `rouge_n` and `rouge_L` would be easier to read imho.

parlai/core/metrics.py
```diff
@@ -177,27 +185,29 @@ def _rouge(guess, answers):
...
-        sums['rouge-L'] = 0.0
+    sums = {}
     num_tasks = 0
     total = 0
+    for i in range(len(reporters)):
+        task_id = reporters[i].getID()
+        task_report = reporters[i].report()
+        for each_metric, value in task_report.items():
+            if isinstance(value, Number):
+                sums[each_metric] = 0.0
```
stephenroller: you should respect ints/floats here. There's some nuance (floats get rounded by round_sigfigs later on, ints don't).

parlai/core/metrics.py
```diff
@@ -227,15 +237,33 @@ class Metrics(object):
...
+            optional_metrics_list = __DEFAULT_METRICS__
+        elif metrcis_arg == 'all':
+            optional_metrics_list = __ALL__METRICS__
+        else:
+            optional_metrics_list = set(metrcis_arg.split(','))
+            optional_metrics_list.add('correct')
+        for each_m in optional_metrics_list:
+            if each_m.startswith('rouge') and rouge is not None:
+                self.metrics_list.add(each_m)
+                if each_m == 'rouge-L':
```
stephenroller: Hmm, it might be cleaner to just have `rouge`, which computes all the rouge metrics, or doesn't.

parlai/core/metrics.py
```diff
@@ -227,15 +237,33 @@ class Metrics(object):
...
-        if rouge is not None:
-            # only compute rouge if we can
-            self.metrics_list.append('rouge-1')
-            self.metrics_list.append('rouge-2')
-            self.metrics_list.append('rouge-L')
+        self.metrics_list = set()
+        optional_metrics_list = []
+        self.compute_rouge_n = 0
+        self.compute_rouge_l = False
+        metrcis_arg = opt.get('metrics', 'default')
```
stephenroller: spelling here and around.

parlai/core/metrics.py
```diff
@@ -307,39 +335,35 @@ def update(self, observation, labels):
...
-                if k not in [
-                    'correct',
-                    'f1',
-                    'hits@k',
-                    'bleu',
-                    'rouge-1',
-                    'rouge-2',
-                    'rouge-L',
-                ]:
+                if k not in __ALL__METRICS__:
```
stephenroller: if we keep this condition we might need to make all metrics a set instead I think.

parlai/core/metrics.py
```diff
@@ -352,7 +376,7 @@ def update(self, observation, labels):
                             # no need to lock because not SharedTable
                             if k not in self.metrics:
                                 self.metrics[k] = v
-                                self.metrics_list.append(k)
+                                self.metrics_list.add(k)
```
stephenroller: i don't understand this one, I don't see why custom metrics should be polluting the Teacher's metric computation list.

parlai/tasks/wizard_of_wikipedia/agents.py
dexterju: Then I should remove that from this commit.

parlai/core/metrics.py
dexterju: Thanks! I didn't know that.

parlai/core/metrics.py
dexterju: But we may want some control over which rouge we want to compute? 
1. Rouge-L is time consuming since it involves a LCS computation.
2. You mentioned before we want to make this n a parameter?

parlai/core/metrics.py
dexterju: I think I just keep it as it is? or you suggest sth else?

stephenroller:
It seems reasonable to add a unit test for default/all/pick one. it can just be done with eval_model probably, and assert that a slow metrics is/isn't in the list and a fast metric is in both.

dexterju:
ping  @stephenroller 

I think that is all we need?

parlai/core/metrics.py
```diff
@@ -18,6 +18,11 @@
 
 import re
 
+DEFAULT_METRICS = set(['correct', 'bleu', 'accuracy', 'f1'])
+ROUGE__METRICS = set(['rouge-1', 'rouge-2', 'rouge-L'])
+ALL__METRICS = set(DEFAULT_METRICS | ROUGE__METRICS)
```
stephenroller: still have the DOUBLE__UNDERSCORES

parlai/core/metrics.py
```diff
@@ -307,22 +326,26 @@ def update(self, observation, labels):
...
 
             # F1 and BLEU metrics.
-            f1 = _f1_score(prediction, labels)
-            bleu = _bleu(prediction, labels)
-            rouge1, rouge2, rougel = _rouge(prediction, labels)
+            if 'f1' in self.metrics_list:
+                f1 = _f1_score(prediction, labels)
+            if 'bleu' in self.metrics_list:
+                bleu = _bleu(prediction, labels)
+            if 'rouge-L' in self.metrics_list:
```
stephenroller: given that you only check `if 'rouge-L' in metrics_list`, i'd prefer if we just had one key for all of the rouges. `if 'rouge' in metrics_list:`. Maybe @emilydinan  or @jaseweston will find that too obtuse though.

parlai/scripts/eval_model.py
```diff
@@ -44,12 +44,14 @@ def setup_args(parser=None):
...
         '--metrics',
         type=str,
-        default='all',
-        help='list of metrics to show/compute, e.g. '
-        'ppl, f1, accuracy, hits@1.'
-        'If `all` is specified [default] all are shown.',
+        default='default',
+        help='list of metrics to show/compute, e.g. all, default,'
+        'or give a list split by , like '
+        'ppl, f1, accuracy, hits@1, rouge, bleu'
```
stephenroller: I think you need no spaces, right?

tests/test_eval_model.py
```diff
@@ -36,11 +36,79 @@ def test_output(self):
...
-                self.assertEqual(score['rouge-L'], 1, 'rouge-L != 1')
+            self.assertEqual(
+                'accuracy' in score, True, "Accuracy is missing from default"
+            )
+            self.assertEqual(score['accuracy'], 1, "Accuracy != 1")
+            self.assertEqual(
+                'rouge-1' in score, False, "Rouge is in the default metrics"
+            )
+
+    def test_metrics_all(self):
```
stephenroller: thaaannnks!

tests/test_eval_model.py
```diff
@@ -36,11 +36,79 @@ def test_output(self):
...
+    def test_metrics_select(self):
+        """Test output of running eval_model"""
+        parser = setup_args()
+        parser.set_defaults(
+            task='integration_tests',
+            model='repeat_label',
+            datatype='valid',
+            num_examples=5,
+            display_examples=False,
+            metrics='accuracy, rouge',
```
stephenroller: O_o I don't understand how your thing handles the spaces O_o

parlai/core/metrics.py
```diff
@@ -18,6 +18,11 @@
 
 import re
 
+DEFAULT_METRICS = set(['correct', 'bleu', 'accuracy', 'f1'])
```
stephenroller: you can also do `DEFAULT_METRICS = {'correct', 'bleu', ...}`. The syntax is a bit prettier.

dexterju: that is cool! Thanks!

parlai/core/metrics.py
stephenroller: (for clarification, you should only have single underscores)

stephenroller: https://hackernoon.com/understanding-the-underscore-of-python-309d1a029edc

parlai/scripts/eval_model.py
dexterju: Sorry I don't understand, what do you mean by space here?

stephenroller: I'm genuinely surprised it's `--metrics 'ppl, f1, accuracy'` instead of  `--metrics 'ppl,f1,accuracy'`. Where do you handle that?

dexterju: aha, sorry I didn't consider that....I was just brainlessly following the coding habits.

parlai/core/metrics.py
emilydinan: yes i'd prefer that

tests/test_eval_model.py
emilydinan: yeah... where are those handled? i'm also missing that

parlai/core/metrics.py
dexterju: I am not sure that is a good idea, you will need to deal with every rouge inside it separately, instead of simplying looping inside self.metrics_list when you init the self.metrics dict....

stephenroller: you enumerate them all anyway... Let's do it the single key way, please.

dexterju: Sure.

PR #1859, opened 2019-07-19
Support setting adam epsilon.

stephenroller:
**Patch description**
Support setting adam epsilon with --adam-eps.

Most users will never care about this flag, but when training very large models I've found it setting it to 1e-6 increases stability, and others in the lab report this. 

**Testing steps**
Been using it. Added it as a flag in one of the transformer tests to make sure it has coverage in CircleCI.


stephenroller:
I'll just change the default to match pytorch; there's a very little chance they change it.

PR #1862, opened 2019-07-22
Bug fixes in true distributed.

stephenroller:
**Patch description**
#1846 introduced a unit test for multiprocessing. It was fine and dandy but it actually broke train_distributed.py. GG.

This ensures both work again, makes the code a LITTLE bit more readable, and makes the manual_seed in a more reasonable place such that it's less likely we'll have this problem with TRA/TQA/...

**Testing steps**
- Went back to training my distributed models on this code.
- Ran unit test locally.
- CircleCI gpu tests.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -373,10 +373,7 @@ class TorchGeneratorAgent(TorchAgent):
                 print('[ Saving dot beam logs in {} ]'.format(self.beam_dot_dir))
 
             self.build_criterion()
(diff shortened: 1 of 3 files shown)


stephenroller:
That is built into TGA now. TorchGeneratorAgent hard crashes if the parameters aren't synced. That was part of #1846.

https://github.com/facebookresearch/ParlAI/blob/e89b26e9800d5b5abc06cc8ce369fc13b3fae5e4/parlai/core/torch_generator_agent.py#L380

PR #1868, opened 2019-07-25
Build history

stephenroller:
**Patch description**
Generalizes the history creation into `build_history` so that one can more easily customize how the history object is constructed.

**Testing steps**
CircleCI.


stephenroller:
cc @shrimai

PR #1870, opened 2019-07-26
OS X unit testing.

stephenroller:
**Patch description**
Based on #1805, it seems like there are some issues which may only crop up on OS X. Although we don't use OS X for development, it seems like many of our OSS users do. Therefore it seems reasonable to just have the unittests tracked in OS X.

This patch adds OS X testing to our CI framework.

**Testing steps**
CircleCI.


stephenroller:
Screen Shot 2019-07-26 at 10 33 56 AM

PR #1885, opened 2019-07-30
[WIP] Optionally evaluate EmpatheticDialogues on listener responses only

EricMichaelSmith:
**Patch description**
If the `--reactions-only` flag is used, only evaluate the EmpatheticDialogues dataset on Listener responses and not Speaker utterances, as is done when running evaluation in https://arxiv.org/abs/1811.00207 .

(Thanks to @hadasah for her help with this a few months back.)

**Testing steps**
- Displaying both Listener responses and Speaker utterances as labels: `python examples/display_data.py -t empathetic_dialogues -ne 100 -dt valid`
- Displaying only Listener responses as labels: `python examples/display_data.py -t empathetic_dialogues -ne 100 -dt valid --reactions-only`

### DIFF
--- a/parlai/tasks/empathetic_dialogues/agents.py
+++ b/parlai/tasks/empathetic_dialogues/agents.py
@@ -119,16 +119,16 @@ class EmpatheticDialogueTeacher(FixedDialogTeacher):
                     continue
 
                 dialogue_parts = [
-                        contextt,
-                        label,
-                        prompt,
-                        sit,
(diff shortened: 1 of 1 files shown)


EricMichaelSmith:
Hmm, closing this for the time being - I think this isn't needed for what I'm doing currently (evaluation that forces responses only with no option), so I'll re-open this later on if I need to

PR #1888, opened 2019-08-01
[Bug Fix]Fix Metrics type issue in convai2 circleci test

dexterju:
**Patch description**
The new metrics clear function ignores the original type and set all these to float, this cause issues in https://circleci.com/gh/facebookresearch/ParlAI/7930#tests/containers/2

This PR should fix it.

**Testing steps**
See the CI


stephenroller:
To answer another offline question, if you add `[long]` or `[gpu]` to your commit message, then you will also run nightly gpu tests in circleci with your commit.

PR #1890, opened 2019-08-02
[WIP] Adding BERT variant of the Transformer

EricMichaelSmith:
**Patch description**
Allows the Transformer to more closely resemble the architecture of the HuggingFace BERT model as wrapped by ParlAI. Can be called with `--variant bert`. Will be used to run inference on an EmpatheticDialogues model pre-trained/finetuned with HuggingFace BERT.

Features of this BERT variant of the Transformer:
- Like the XLM variant, performs layer norm on the summed token/position/segment embeddings. HuggingFace performs this in its `BertEmbeddings` class (https://github.com/huggingface/pytorch-transformers/blob/master/pytorch_transformers/modeling_bert.py#L269 )
- Adds an additional linear layer to the end of the encoder after reduction, analogous to in ParlAI's `BertWrapper` (https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/bert_ranker/helpers.py#L94 )

**Testing steps**
This PR is a work in progress: tests will be run soon.

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -498,7 +498,7 @@ class TransformerEncoder(nn.Module):
             or self.reduction_type is None
             and self.variant != 'bert'
         ):
-            # TODO: figure out: is the code that passes back the mask still usable? From
+            # TODO: is the code that passes back the mask still usable? From
             #  looking at the models that use this encoder, the output of the encoder is
             #  always assumed to be a Tensor, not a 2-ple of Tensors
             return output, mask
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -611,7 +611,7 @@ class TransformerDecoder(nn.Module):
 
         self.embeddings = embedding
 
-        if self.variant == ['bert', 'xlm']:
+        if self.variant in ['bert', 'xlm']:
             self.norm_embeddings = LayerNorm(self.dim, eps=LAYER_NORM_EPS)
         elif self.variant == 'aiayn':
             pass
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -493,14 +493,7 @@ class TransformerEncoder(nn.Module):
         if self.variant in 'bert':
             output = self.additional_linear_layer(output)
 
-        if (
-            self.reduction_type == 'none'
-            or self.reduction_type is None
-            and self.variant != 'bert'
-        ):
-            # TODO: is the code that passes back the mask still usable? From looking at
-            #  the models that use this encoder, the output of the encoder is always
-            #  assumed to be a Tensor, not a 2-ple of Tensors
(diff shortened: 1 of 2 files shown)


EricMichaelSmith:
@emilydinan @stephenroller @klshuster CI tests are failing with a `ValueError: Can't handle --variant xlm` message. Is there some way to register this flag value during testing?

parlai/agents/transformer/modules.py
```diff
@@ -470,21 +477,34 @@ def forward(self, input, positions=None, segments=None):
...
 
+        if self.variant in 'bert':
+            output = self.additional_linear_layer(output)
+
+        if (
+            self.reduction_type == 'none'
+            or self.reduction_type is None
+            and self.variant != 'bert'
+        ):
+            # TODO: is the code that passes back the mask still usable? From looking at
```
klshuster: to answer your TODO - yes. E.g. the Polyencoder uses the mask (https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/transformer/polyencoder.py#L264)

parlai/agents/transformer/modules.py
```diff
@@ -591,7 +611,7 @@ def __init__(
 
         self.embeddings = embedding
 
-        if self.variant == 'xlm':
+        if self.variant == ['bert', 'xlm']:
```
emilydinan: your tests are failing because of this `==`. you need to replace it with `in`

EricMichaelSmith: -____- Ha, you can really tell this is a WIP. Fixed - thanks!

parlai/agents/transformer/modules.py
EricMichaelSmith: Ah, that's great to know - thanks! I've removed the TODO and fixed the conditional. I see that reduction_type == None is not presented as an option in the help string, which makes sense given that using it would cause an error when not using the polyencoder

emilydinan:
is this still WIP @EricMichaelSmith? can it be mergeD?

EricMichaelSmith:
I haven't been using it recently, which is why I haven't merged it in yet - as far as I know, it works well given the evaluation routines I've run with it so far, but lemme know if you think there are specific other tests I should run before merging it in :)

parlai/agents/transformer/modules.py
```diff
@@ -470,21 +477,27 @@ def forward(self, input, positions=None, segments=None):
...
-            return output
         elif self.reduction_type == 'none' or self.reduction_type is None:
             output = tensor
-            return output, mask
         else:
             raise ValueError(
                 "Can't handle --reduction-type {}".format(self.reduction_type)
             )
 
+        if self.variant in 'bert':
```
stephenroller: should be ==

EricMichaelSmith:
> Why wasn't this merged? If you don't need the functionality anymore (as indicated by staleness), then let's abandon it to avoid the maintenance costs. But it looks like it should've been merged already.
Good point - I'll abandon because we don't need the functionality anymore

PR #1893, opened 2019-08-06
only print number of threads initialized, not init per thread

alexholdenmiller:
**Patch description**
Small change changing the logging of thread initialisation. We could add it back in if we had logging levels, but for now it just makes the logs much harder to read.

**Logs**
```
$ python {anyscript.py} -nt 20
...
[ 20 threads initialized ]
```


alexholdenmiller:
can we use f-strings yet?

PR #1894, opened 2019-08-06
fixed bug in fixed candidate ranking and add speedup option for ranking eval

alexholdenmiller:
**Patch description**
1) fix error setting up fixed candidates
~~2) separate loss instantiation into separate function~~ -> moved to #1899
3) add rank_top_k option to do top-k selection instead of sorting all candidates by score. if you're just looking for hits@ metrics this is much faster, though this makes mean_rank / mrr calculations a bit wonky (it doesn't sort anything beyond the k). I'm open to suggestions on how to handle this--maybe good to zero them out?
4) switch pad handling to be a bit faster when using inline candidates

**Testing steps**
The following run uses a relatively large 75k candidate set with fixed encodings during evaluation. I added a timer around valid/test in train_mode.py.
```
$ python examples/train_model.py -t wikimovies --dict-file /tmp/dict_wikimovies -bs 128 -nt 10 -eps 5 -m memnn --no-cuda -cands batch -ecands fixed -fcp /private/home/ahm/ParlAI/data/WikiMovies/movieqa/knowledge_source/entities.txt -dt train:ordered --ignore-bad-candidates True -mf /tmp/run
```
master: `~48s training / IndexError on eval`
master + error fix: `~48s training / ~49s valid + test`
rank_speed: `~46s training / ~47s valid + test` (slightly faster training + slightly faster eval)
rank_top_k=100: `~46s training / ~36s valid + test` (faster eval if you just want hits@100)
rank_top_k=1: `~47s training / ~32s valid + test` (faster eval if you just want hits@1)

The biggest speedup you'll see is when using inline candidates due to the change in the "ordering" code.
```
python examples/train_model.py -t wikimovies --dict-file /tmp/dict_wikimovies -bs 4 -esz 1 -hops 0 --memsize 0 -tf False -nt 10 -eps 1 -m memnn --no-cuda -cands batch -ecands inline -dt train:ordered -vme 1000 --short-final-eval True -mf /tmp/run
```
master: `467s valid + test`
rank_speed: `402s valid + test`
rank_top_k=100: `388s valid + test`

You might notice that this is a no-hop memnn with esz of only 1, and it still takes forever to do just 1000 examples here. Precomputing the candidate vectors helps a lot. (Or even better, the candidate vectors.)

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -156,7 +156,7 @@ class TorchRankerAgent(TorchAgent):
                 states = {}
 
         self.rank_top_k = opt.get('rank_top_k', -1)
-        self.rank_loss = self.build_loss(opt)
+        self.rank_loss = nn.CrossEntropyLoss(reduce=True, size_average=False)
         if self.use_cuda:
             self.model.cuda()
             self.rank_loss.cuda()
@@ -180,9 +180,6 @@ class TorchRankerAgent(TorchAgent):
             self.model = torch.nn.parallel.DistributedDataParallel(
                 self.model, device_ids=[self.opt['gpu']], broadcast_buffers=False
             )
-
-    def build_loss(self, opt):
-        return nn.CrossEntropyLoss(reduce=True, size_average=False)
 
     def set_interactive_mode(self, mode, shared=False):
         self.candidates = self.opt['candidates']
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -119,6 +119,13 @@ class TorchRankerAgent(TorchAgent):
             help='Ignore examples for which the label is not present in the '
             'label candidates. Default behavior results in RuntimeError. ',
         )
+        agent.add_argument(
+            '--rank-top-k',
+            type=int,
+            default=-1,
+            help='Ranking returns the top k results of k > 0, otherwise sorts every '
+            'single candidate according to the ranking.',
+        )
 
     def __init__(self, opt, shared=None):
         # Must call _get_init_model() first so that paths are updated if necessary
@@ -148,7 +155,8 @@ class TorchRankerAgent(TorchAgent):
             else:
                 states = {}
 
-        self.rank_loss = nn.CrossEntropyLoss(reduce=True, size_average=False)
+        self.rank_top_k = opt.get('rank_top_k', -1)
+        self.rank_loss = self.build_loss(opt)
         if self.use_cuda:
             self.model.cuda()
             self.rank_loss.cuda()
@@ -172,6 +180,9 @@ class TorchRankerAgent(TorchAgent):
             self.model = torch.nn.parallel.DistributedDataParallel(
                 self.model, device_ids=[self.opt['gpu']], broadcast_buffers=False
             )
+
+    def build_loss(self, opt):
+        return nn.CrossEntropyLoss(reduce=True, size_average=False)
 
     def set_interactive_mode(self, mode, shared=False):
         self.candidates = self.opt['candidates']
@@ -257,7 +268,10 @@ class TorchRankerAgent(TorchAgent):
         """Return predictions from training."""
         # TODO: speed these calculations up
         batchsize = scores.size(0)
-        _, ranks = scores.sort(1, descending=True)
+        if self.rank_top_k > 0:
+            _, ranks = scores.topk(self.rank_top_k, 1, largest=True)
+        else:
+            _, ranks = scores.sort(1, descending=True)
         for b in range(batchsize):
             rank = (ranks[b] == label_inds[b]).nonzero().item()
             self.metrics['rank'] += 1 + rank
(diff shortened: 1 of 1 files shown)


parlai/core/torch_ranker_agent.py
```diff
@@ -257,7 +268,12 @@ def _get_train_preds(self, scores, label_inds, cands, cand_vecs):
         """Return predictions from training."""
         # TODO: speed these calculations up
         batchsize = scores.size(0)
-        _, ranks = scores.sort(1, descending=True)
+        if self.rank_top_k > 0:
+            _, ranks = scores.topk(
+                min(self.rank_top_k, scores.size(1)), 1, largest=True
```
alexholdenmiller: the min here is because e.g. babi only has 6 candidates!

parlai/core/torch_ranker_agent.py
```diff
@@ -362,15 +378,21 @@ def eval_step(self, batch):
...
 
         # Update metrics
         if label_inds is not None:
             loss = self.rank_loss(scores, label_inds)
             self.metrics['loss'] += loss.item()
             self.metrics['examples'] += batchsize
             for b in range(batchsize):
-                rank = (ranks[b] == label_inds[b]).nonzero().item()
+                rank = (ranks[b] == label_inds[b]).nonzero()
+                rank = rank.item() if len(rank) == 1 else scores.size(1)
```
alexholdenmiller: if you use rank_top_k, you may not have the true answer in the top k

parlai/core/torch_ranker_agent.py
```diff
@@ -382,13 +404,11 @@ def eval_step(self, batch):
...
-            if len(ordering) != len(cand_list):
-                # ignore padding
-                true_ordering = [x for x in ordering if x < len(cand_list)]
-                ordering = true_ordering
             # using a generator instead of a list comprehension allows
             # to cap the number of elements.
-            cand_preds_generator = (cand_list[rank] for rank in ordering)
+            cand_preds_generator = (
+                cand_list[rank] for rank in ordering if rank < len(cand_list)
+            )
```
alexholdenmiller: saves memory and actually faster

parlai/core/torch_ranker_agent.py
```diff
@@ -586,31 +606,33 @@ def _build_candidates(self, batch, source, mode):
...
                 raise ValueError(
                     "If using candidate source 'fixed', then you must provide the path "
-                    "to a file of candidates with the flag --fixed-candidates-path"
+                    "to a file of candidates with the flag --fixed-candidates-path or "
+                    "the name of a task with --fixed-candidates-task."
                 )
+            warn_once(
+                "[ Executing {} mode with a common set of fixed candidates "
+                "(n = {}). ]".format(mode, len(self.fixed_candidates))
+            )
```
alexholdenmiller: having the warning before causes a "can't get `len` of `None`" error

parlai/core/torch_ranker_agent.py
```diff
@@ -586,31 +606,33 @@ def _build_candidates(self, batch, source, mode):
...
 
             cands = self.fixed_candidates
             cand_vecs = self.fixed_candidate_vecs
 
             if label_vecs is not None:
                 label_inds = label_vecs.new_empty((batchsize))
                 bad_batch = False
-                for i, label_vec in enumerate(label_vecs):
-                    label_vec_pad = label_vec.new_zeros(cand_vecs[i].size(0)).fill_(
```
alexholdenmiller: cand_vecs here are fixed and not built in a batch but the `i` here was trying to get "batch row i" from the candidate vectors

alexholdenmiller: the other changes are primarily just renaming to variables that are more clear (better for us to use "batch_idx" rather than "i")

alexholdenmiller:
So this passes all existing tests but it would be good to know: what are some other fixed-candidate-ranking scenarios with either different models or different tasks?

I switched to a transformer and tried this:
```
python examples/train_model.py -t wikimovies --dict-file /tmp/dict_wikimovies -bs 256 -nt 1 -eps 1 -cands batch -ecands fixed -fcp /private/home/ahm/ParlAI/data/WikiMovies/movieqa/knowledge_source/entities.txt -dt train:ordered --ignore-bad-candidates True -m transformer/ranker -mf /tmp/run
```
master: `~33s train / IndexError on eval`
rank_speed: `~33s train / no error on eval`

parlai/core/torch_ranker_agent.py
```diff
@@ -173,6 +181,9 @@ def __init__(self, opt, shared=None):
                 self.model, device_ids=[self.opt['gpu']], broadcast_buffers=False
             )
 
+    def build_loss(self, opt):
+        return nn.CrossEntropyLoss(reduce=True, size_average=False)
```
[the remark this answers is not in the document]
stephenroller: it would be vaguely nice if this were made more similar to `compute_loss` in TGA, which is nice and easy to override with custom losses. If not that, can we make this `_build_loss` for the time being so it's less likely someone overrides it? I'm hoping we can unify those two interfaces.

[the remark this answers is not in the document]
alexholdenmiller: reverting this change and moving to new PR

parlai/core/torch_ranker_agent.py
```diff
@@ -362,15 +378,21 @@ def eval_step(self, batch):
                 cand_encs = self.vocab_candidate_encs
 
         scores = self.score_candidates(batch, cand_vecs, cand_encs=cand_encs)
-        _, ranks = scores.sort(1, descending=True)
+        if self.rank_top_k > 0:
+            _, ranks = scores.topk(
+                min(self.rank_top_k, scores.size(1)), 1, largest=True
```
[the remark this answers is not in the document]
alexholdenmiller: worked for bs=1, thanks!

PR #1900, opened 2019-08-07
Move _build_encoder to class function

klshuster:
**Patch description**
Makes `_build_encoder` a class function of `TransformerGeneratorModel`, which allows subclasses to easily modify the encoder. 

**Testing steps**
Tested internally


klshuster:
Abandoning, this is not necessary

PR #1911, opened 2019-08-11
Enable fairseq unit test.

stephenroller:
**Patch description**
CircleCI has supported GPU for a while now, so we could've been testing the fairseq module using the test we *already have*. However, this test is skipped because fairseq isn't installed.

This patch adds fairseq installation to the GPU tests, so that test also runs.

**Testing steps**
CircleCI.


stephenroller:
Abandoning for #2226.

PR #1915, opened 2019-08-13
Quiet the rouge warning.

stephenroller:
**Patch description**
The py-rouge warning happens a lot (on import of metrics). We get rid of that warning.

**Testing steps**
test_torch_agent.py used to throw the warning, now it doesn't.

**Logs**
```
devfair0237 parlai quietrouge $ python tests/test_torch_agent.py
.............
Ran 13 tests in 3.938s

OK
```

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -33,8 +33,8 @@ except ImportError:
 try:
     import rouge
 except ImportError:
-    # User doesn't have rouge installed, so we can't use it for rouge
-    # We'll just turn off things, but we might want to warn the user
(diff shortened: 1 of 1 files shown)


parlai/core/metrics.py
```diff
@@ -31,11 +31,10 @@
     nltkbleu = None
 
 try:
-    import rouge as rouge
+    import rouge
 except ImportError:
     # User doesn't have rouge installed, so we can't use it for rouge
     # We'll just turn off things, but we might want to warn the user
```
emilydinan: prob should remove "we might want to warn the user" since we are no longer warning them 🌵

Issue #1914, opened 2019-08-13
Can't use train-predict=True with inline candidates on QAngaroo


james-assiene:
**Bug description**
When I use --train-predict = True and --candidates inline I obtain an error

```
/root/ParlAI/parlai/core/torch_ranker_agent.py:577: UserWarning: [ Executing train mode with provided inline set of candidates ]
  ''.format(mode)
Traceback (most recent call last):
  File "/root/ParlAI/parlai/scripts/train_model.py", line 798, in <module>
    TrainLoop(setup_args().parse_args()).train()
  File "/root/ParlAI/parlai/scripts/train_model.py", line 717, in train
    world.parley()
  File "/root/ParlAI/parlai/core/worlds.py", line 734, in parley
    batch_act = self.batch_act(agent_idx, batch_observations[agent_idx])
  File "/root/ParlAI/parlai/core/worlds.py", line 702, in batch_act
    batch_actions = a.batch_act(batch_observation)
  File "/root/ParlAI/parlai/core/torch_agent.py", line 1698, in batch_act
    output = self.train_step(batch)
  File "/root/ParlAI/parlai/core/torch_ranker_agent.py", line 359, in train_step
    return self._get_train_preds(scores, label_inds, cands, cand_vecs)
  File "/root/ParlAI/parlai/core/torch_ranker_agent.py", line 283, in _get_train_preds
    preds = [cands[i][ordering[0]] for i, ordering in enumerate(ranks)]
  File "/root/ParlAI/parlai/core/torch_ranker_agent.py", line 283, in <listcomp>
    preds = [cands[i][ordering[0]] for i, ordering in enumerate(ranks)]
IndexError: list index out of range
```

**Reproduction steps**

A Colab notebook is available here : https://drive.google.com/open?id=1hX095tE3bA61DTnpcHxt6-V6kyh0CL0T

**Additional context**

I found a weird trick to "bypass" this issue : I train without --train-predict True, save the model (checkpoint) , train with --train-predict True using the model from the checkpoint and then it works
When I don't use --candidates inline everything works fine with --train-predict True

emilydinan:
Hi @james-assiene! Thank you for filing this detailed issue -- was very helpful in debugging. You should find a fix in #1916.

emilydinan:
Closing, but feel free to re-open if you have further questions!

PR #1916, opened 2019-08-13
[TRA] Train predictions fix

emilydinan:
**Patch description**
Issue flagged thanks to #1914. Need to filter out padded candidates from the ordering, otherwise we get an index error.

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -272,25 +272,32 @@ class TorchRankerAgent(TorchAgent):
         else:
             _, ranks = scores.sort(1, descending=True)
         for b in range(batchsize):
-            rank = (ranks[b] == label_inds[b]).nonzero().item()
+            rank = (ranks[b] == label_inds[b]).nonzero()
+            rank = rank.item() if len(rank) == 1 else scores.size(1)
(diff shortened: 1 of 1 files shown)


alexholdenmiller:
hm this might be related to the diff I landed

emilydinan:
> hm this might be related to the diff I landed

i actually think it was broken before too!

alexholdenmiller:
#1894 changed the similar code in eval_step but not in _get_train_preds

parlai/core/torch_ranker_agent.py
```diff
@@ -277,11 +277,20 @@ def _get_train_preds(self, scores, label_inds, cands, cand_vecs):
...
+            elif cand_vecs.dim() == 3:
+                cand_list = cands[i]
+            if len(ordering) != len(cand_list):
+                # ignore padding
+                for x in ordering:
+                    if x < len(cand_list):
+                        preds.append(cand_list[x])
+                        break
+            else:
+                preds.append(cand_list[ordering[0]])
```
alexholdenmiller: return Output

stephenroller:
File an issue for testing a couple TRA models with every variant for -cands?

PR #1919, opened 2019-08-13
Pytorch version tests.

stephenroller:
**Patch description**
We decided to support a rolling pytorch release (past 2 versions). So let's explicitly test them.

**Testing steps**
CircleCI should run with two versions of pytorch.


mwillwork:
Pretty slick! Are we going to give an alert/announcement saying we are doing this though? B/c I imagine people use ParlAI without running the tests...

stephenroller:
Discussed offline. Due to intern season we won't force upgrades yet, but mentioning in the README that we support 1.1 and newer.

PR #1925, opened 2019-08-14
[TGA] refactor beam search

stephenroller:
**Patch description**
This is a refactor of beam search, with the primary goal of making it more readable; improving comments, documentation & API names; and creating a better building block for implementing top-k and nucleus sampling

**Reduced behavior**
- Deleted `--beam-block-ngram`; its behavior was counterintuitive (blocking 4-grams also meant blocking 1, 2, and 3 grams, meaning this arg was really binary). Given that this is nonstandard behavior that we tend not to use much presently, I've opted to remove it.
- Deleted `--beam-dot-log`. Something similar may be brought back at a later date, but for now we don't need it.
- Deleted `PerplexityEvaluatorAgent`. Its API is difficult to use (you need to figure out and maintain the target dictionary yourself), and the few places we do use it already have copies that exist from the seq2seq_v0/v1 forks. If we need it back in the future, it's in the git history.

**Updated API**
- Introduce a `TreeSearch` class, which WILL be the root abstract class for Beam Search, Top-K, Top-P, and whatever else we might think about implementing in the future.
- Rename `TorchGeneratorAgent.beam_seach` to `TorchGeneratorAgent._generate`, which uses `BeamSearch` objects to conduct its generation. Move much of its bookkeeping code into private methods in the `TreeSearch` so this is more readable.
- Simplify some of the API around beam search. This sacrifices the default behavior of getting access to the other generated beams (besides the highest weighted one), but that was previously unused and causing the aggregation code to be run twice.

**Testing steps**
- Existing seq2seq & transformer/generator tests
- Confirm on a pretrained convai2 model that it gets the same results for a few choices of `--beam-size`.
- New unit tests conforming to specifications of `upgrade_opt`.


stephenroller:
Explicitly depends on #1935 now (due to unit test).

stephenroller:
Well this diff is impossible to read.

parlai/core/torch_generator_agent.py
```diff
@@ -680,41 +601,42 @@ def eval_step(self, batch):
...
+                min_n_best=self.beam_min_n_best,
+                padding_token=self.NULL_IDX,
+                bos_token=self.START_IDX,
+                eos_token=self.END_IDX,
+                device=device,
+            )
+        else:
+            raise ValueError(f"Can't use inference method {method}")
+
+    def _generate(self, batch, beam_size, max_ts=40):
```
klshuster: curious why here the `max_ts` is 40 whereas earlier in the code you default to `256`

parlai/core/torch_generator_agent.py
```diff
@@ -783,148 +698,61 @@ def beam_search(
...
-            Set to 2 as usual in ParlAI
+            end of sentence token ID
+        :param min_length:
+            minimum length of the predicted sequence
         :param min_n_best:
-            Beam will not be done unless this amount of finished hypothesis
-            (with EOS) is done
+            Search will not be finished until a minimum number of utterances are
+            completed.
         :param cuda:
```
klshuster: nit: param change (it's now `device` not `cuda`)

parlai/core/torch_generator_agent.py
```diff
@@ -1090,163 +891,96 @@ def get_hyp_from_finished(self, hypothesis_tail):
...
+        if self.beam_size != 1:
+            raise ValueError('Greedy search can only be run with beam size 1.')
 
-        :param outputs:
-            self.outputs from the beam
+    def select_paths(self, logprobs, prior_scores):
+        s, i = logprobs.max(1)
+        best_scores = s + prior_scores.unsqueeze(1)
+        hyp_ids = torch.arange(logprobs.size(0)).to(logprobs.device)
+        return (hyp_ids, i, best_scores)
```
klshuster: is `i` a list here? in the `select_paths` docstring above it mentions that this will be a list, just curious if this is/ if this will break things if not

parlai/core/torch_generator_agent.py
```diff
@@ -261,22 +211,17 @@ def forward(
             - encoder_states are the output of model.encoder. Model specific types.
               Feed this back in to skip encoding on the next call.
         """
-        if ys is not None:
-            # TODO: get rid of longest_label
-            # keep track of longest label we've ever seen
-            # we'll never produce longer ones than that during prediction
-            self.longest_label = max(self.longest_label, ys.size(1))
+        assert ys is not None, "Greedy decoding in TGModel.forward no longer supported."
```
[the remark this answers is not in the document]
stephenroller: I didn't like the dual abstraction with beam search code over here and greedy decoding over there. They're functionally equivalent.

parlai/core/torch_generator_agent.py
stephenroller: This should be updated.

parlai/core/torch_generator_agent.py
stephenroller: better variable names, yes (y)

PR #1932, opened 2019-08-16
Messages fixes for teachers.py

hadasah:
**Patch description**
Messages broke two lines in teachers.py that I've encountered. Changed them to call force_set()


parlai/core/teachers.py
```diff
@@ -1255,5 +1255,5 @@ def _setup_data(self, path):
                         eps = []
         if len(eps) > 0:
             # add last episode
-            eps[-1]['episode_done'] = True
```
emilydinan: nice catch

parlai/core/teachers.py
```diff
@@ -408,7 +408,7 @@ def act(self):
             action = action.copy()
             labels = action.pop('labels')
             if not self.opt.get('hide_labels', False):
-                action['eval_labels'] = labels
+                action.force_set('eval_labels', labels)
```
emilydinan: hmm, so this would happen if the action already has eval_labels? 

this is confusing to me because the above `if` statement checks if there are `labels` in the action. in which case would we have both `labels` and `eval_labels`?

hadasah: oh yeah I didn't look at the context at all. This is only triggered by the temp fromfile task I was using, shouldn't be affecting people usually. I'll get rid of it.

PR #1934, opened 2019-08-18
Top-K and Nucleus Sampling

### LINKED ISSUE #1652: Add Top-k sampling (and stochastic sampling) to beam search
Currently we only support beam search in Torch Generator Agent. Top-K sampling is a popular strategy introduced by [Fan et al., 2018](https://arxiv.org/abs/1805.04833) which should be added.

We should aim to implement this as a somewhat clean API so as to decouple it from the Torch Generator Agent nicely.

stephenroller:
**Patch description**
Implements Top-K (Fan et al., 2018) and Nucleus (Holtzman et al., 2019) sampling for TorchGeneratorAgent.

Depends on the refactor done in #1925. Closes #1652.

Changes:
- [x] Adds `--inference` argument. New behavior is a true case for `upgrade_opt`!
- [x] Add explicit Greedy search.
- [x] Add Top-K sampling
- [x] Add Nucleus/top-p sampling.
- [ ] Delete old greedy search code
- [ ] Add tests for each of the methods (suggestions?)

**Testing steps**
Manual launching with a pretrained model.


parlai/core/torch_generator_agent.py
```diff
@@ -990,3 +1052,60 @@ def select_paths(self, logprobs, prior_scores):
...
+        scores = topk.values[hyp_ids, choices]
+        best_scores = prior_scores.expand_as(scores) + scores
+        return (hyp_ids, tok_ids, best_scores)
+
+
+class NucleusSampling(TreeSearch):
+    """
+    Nucelus, aka top-p sampling (Holtzman et al., 2019).
+
+    Samples from a truncated distribution which covers a fixed CDF porition
```
stephenroller: spelling

parlai/core/torch_generator_agent.py
```diff
@@ -303,7 +319,7 @@ def add_cmdline_args(cls, argparser):
             '--beam-min-length',
             type=int,
             default=1,
-            help='Minimum length of prediction to be generated by ' 'the beam search',
```
emilydinan: lol

parlai/core/torch_generator_agent.py
```diff
@@ -313,6 +329,18 @@ def add_cmdline_args(cls, argparser):
             help='Skip beam search. Useful for speeding up training, '
             'if perplexity is the validation metric.',
         )
+        agent.add_argument(
+            '--inference',
+            choices={'beam', 'greedy', 'topk', 'nucleus', 'topp'},
```
emilydinan: why do we have `topp` AND `nucleus` in here?

stephenroller: Since people call it both, I figured it'd be nice to give it the alias. You suggest only one I take it?

emilydinan: I found it confusing, I think people reading this would wonder what the difference was. Maybe add a comment saying they are the same

stephenroller: I'm just going to only keep one.

stephenroller:
Okay this diff is pretty straightforward given #1925.

parlai/core/torch_generator_agent.py
```diff
@@ -995,3 +1023,60 @@ def select_paths(self, logprobs, prior_scores):
...
+    def select_paths(self, logprobs, prior_scores):
+        probs = torch.softmax(logprobs, dim=-1)
+        sprobs, sinds = probs.sort(dim=-1, descending=True)
+        mask = (sprobs.cumsum(dim=-1) - sprobs[:, :1]) > self.p
+        sprobs[mask] = 0
+        sprobs.div_(sprobs.sum(dim=-1).unsqueeze(1))
+        choices = torch.multinomial(sprobs, 1)[:, 0]
+        hyp_ids = torch.arange(logprobs.size(0)).to(logprobs.device)
+        tok_ids = sinds[hyp_ids, choices]
+        scores = sprobs[hyp_ids, choices].log()
```
emilydinan: question, why do you take the log here?

stephenroller: Because `best_scores` keeps track of the probability of the full utterance. It's much better to add probabilities in logspace than to multiply them in linspace.

stephenroller:
I'll merge this directly into master after the other PR merges.

parlai/core/torch_generator_agent.py
emilydinan: then why don't you do the same thing above in top k?

stephenroller: top-k never leaves logspace. The top-p had to be converted to linspace to do the truncation. (Can't compute the CDF in logspace...)

emilydinan: oh i see now, thanks

stephenroller: I’ll leave a comment

parlai/core/torch_generator_agent.py
```diff
@@ -984,3 +1012,66 @@ def select_paths(self, logprobs, prior_scores):
...
+    are considered at each time.
+
+    Typical values of k are 2, 10, 50.
+
+    See https://arxiv.org/abs/1805.04833 for details.
+    """
+
+    def __init__(self, k, *args, **kwargs):
+        self.k = k
+        super().__init__(*args, **kwargs)
```
pietern: Is setting `self.k` done before calling base class `__init__` for some reason? I've only seen this flipped around where first the base class is initialized and then local initialization is done. I think this opens up the possibility of the base class setting `self.k` as well and overriding what you want to do here (although I know it doesn't here, just commenting on the pattern).

parlai/core/torch_generator_agent.py
```diff
@@ -984,3 +1012,66 @@ def select_paths(self, logprobs, prior_scores):
...
+
+    See https://arxiv.org/abs/1904.09751 for details.
+    """
+
+    def __init__(self, p, *args, **kwargs):
+        self.p = p
+        super().__init__(*args, **kwargs)
+
+    def select_paths(self, logprobs, prior_scores):
+        # unlike the other treesearch methods, we have to switch to linspace
```
beauby: Nitpick: When comments are longer than a word or two, consider making them full sentences (capital letters/punctuation). 🧐

PR #1937, opened 2019-08-20
fix rouge setup errors

alexholdenmiller:
**Patch description**
Rouge needs nltk punkt tokenizer set up in order to work, so unit tests fail if user does not have nltk installed and the punkt tokenizer downloaded.

Now, rouge will be set to None if the punkt tokenizer is not set up.

After fixing that, unit tests still fail because rouge is still added to the metrics list even if rouge is not set up due to the `else` block catching when `rouge is None`.

**Testing steps**
`python tests/test_eval_model.py`

**Results on master branch**
If you don't happen to have the punk tokenizer set up, you get the following:
```
ERROR: test_metrics_select (__main__.TestEvalModel)
Test output of running eval_model
  File "tests/test_eval_model.py", line 91, in test_metrics_select
    str_output, valid, test = testing_utils.eval_model(opt)
  File "/Users/ahm/ParlAI/parlai/core/testing_utils.py", line 272, in eval_model
    valid = None if skip_valid else ems.eval_model(popt)
  File "/Users/ahm/ParlAI/parlai/scripts/eval_model.py", line 117, in eval_model
    task_report = _eval_single_world(opt, agent, task)
  File "/Users/ahm/ParlAI/parlai/scripts/eval_model.py", line 83, in _eval_single_world
    world.parley()
  File "/Users/ahm/ParlAI/parlai/core/worlds.py", line 275, in parley
    agents[0].observe(validate(acts[1]))
  File "/Users/ahm/ParlAI/parlai/core/teachers.py", line 351, in observe
    self.metrics.update(observation, self.lastY)
  File "/Users/ahm/ParlAI/parlai/core/metrics.py", line 348, in update
    self.metrics['rouge-1'] += rouge1
TypeError: unsupported operand type(s) for +=: 'float' and 'NoneType
```

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -32,7 +32,10 @@ except ImportError:
 
 try:
     import rouge
-except ImportError:
+    import nltk
+
+    nltk.tokenize.load('tokenizers/punkt/{0}.pickle'.format('en'))
+except (ImportError, LookupError):
     # User doesn't have py-rouge installed, so we can't use it.
     # We'll just turn off rouge computations
     rouge = None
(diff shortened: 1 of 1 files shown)


stephenroller:
I don’t think this is what we want. Every import of metrics.py is going to cause I/O. The protection was supposed to be happening in the actual `_rouge` calculation.

stephenroller:
https://github.com/facebookresearch/ParlAI/blob/539fd4e20a7bdef950c9a4df4fd88a4e4a2df817/parlai/core/metrics.py#L178-L183

Yeah, this is supposed to be already happening. Why isn't it triggering is the real question?

alexholdenmiller:
okay added a check that the rouge metrics aren't None instead which should do it

alexholdenmiller:
The previous check for the punkt tokenizer was working, so it was returning None which was being added to the float metrics and triggering the error. Added the error to the PR description and switched to a different fix that just checks the metric returned is not None.

stephenroller:
Uh, wait for dexter's approval. I want him to review this. I was just trying to undo my "needs changes"

PR #1940, opened 2019-08-20
Cleanup of tensorboard integration.

stephenroller:
**Patch description**
Small cleanup of tensorboard integration. Makes it so there's only one flag, `--tensorboard-log true` to enable it. It always gets saved to `model_file.tensorboard`. All numeric metrics are logged on every valid and every train log.

**Testing steps**
Manual testing. Will upload a screenshot later.

### DIFF
--- a/parlai/core/logs.py
+++ b/parlai/core/logs.py
@@ -9,24 +9,23 @@ Log metrics to tensorboard.
 This file provides interface to log any metrics in tensorboard, could be
 extended to any other tool like visdom.
 
-If you use tensorboard logging, all event folders will be stored in
-``PARLAI_DATA/tensorboard`` folder. In order to Open it with TB, launch
-tensorboard as:
-
 .. code-block: none
 
    tensorboard --logdir <PARLAI_DATA/tensorboard> --port 8888.
 """
 
-# TODO: update this with pytorch 1.1's tensorboard API
-
 import os
+import json
+import numbers
+
+try:
+    from tensorboardX import SummaryWriter
+except ImportError:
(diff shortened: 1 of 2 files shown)


parlai/core/logs.py
```diff
@@ -38,94 +37,32 @@ def add_cmdline_args(argparser):
...
-        if opt['tensorboard_tag'] is None:
-            tensorboard_tag = opt['starttime']
-        else:
-            tensorboard_tag = opt['starttime'] + '__'.join(
-                [i + '-' + str(opt[i]) for i in opt['tensorboard_tag'].split(',')]
-            )
-        if opt['tensorboard_comment']:
-            tensorboard_tag += '__' + opt['tensorboard_comment']
+        if SummaryWriter is None:
+            raise RuntimeError('Please run `pip install tensorboard tensorboardX`.')
```
stephenroller: change to ImportError

parlai/core/logs.py
```diff
@@ -9,25 +9,24 @@
...
-If you use tensorboard logging, all event folders will be stored in
-``PARLAI_DATA/tensorboard`` folder. In order to Open it with TB, launch
-tensorboard as:
-
 .. code-block: none
 
    tensorboard --logdir <PARLAI_DATA/tensorboard> --port 8888.
 """
 
-# TODO: update this with pytorch 1.1's tensorboard API
```
stephenroller: This is intentional. The pytorch 1.1/1.2 API is incomplete and tensorboardX works better.

PR #1941, opened 2019-08-20
Minor cleanups

stephenroller:
**Patch description**
- Move a helper method to the module where it's being used, rather than having it be imported. This has a few minor usages internally that need to be updated.
- Clean up some docstrings.
- Drop another private import that doesn't really need to be there.

**Testing steps**
CI


parlai/core/worlds.py
```diff
@@ -1051,6 +1054,44 @@ def shutdown(self):
 ################################################################################
 # Functions for creating tasks/worlds given options.
 ################################################################################
+def _create_task_agents(opt):
```
mwillwork: This seems fine ... i.e. we no longer have an import to a method with _ but I think it begs the question of why is this a top level function? (Same as the _path functions.) ParlAI is overall quite functional (in the programming sense). Especially here where we are building objects, why not use a builder design pattern? or factory?

stephenroller: Why though? The logic could be improved, but a factory method is still a factory. It doesn't need to be wrapped as a classmethod or something.

mwillwork: Fine... :) (Were there no other instances of this btw?)

stephenroller:
```
$ ag 'import.* _'
parlai/core/params.py
13:import sys as _sys

parlai/mturk/tasks/image_chat/image_chat_collection/worlds.py
20:from parlai.core.metrics import _exact_match

parlai/mturk/tasks/personality_captions/personality_captions/worlds.py
19:from parlai.core.metrics import _exact_match

parlai/mturk/webapp/server.py
36:from parlai import __path__ as parlai_path
41:    from parlai_internal import __path__ as parlai_int_path

projects/controllable_dialogue/Analysis_n_Graphs.ipynb
74:    "from parlai.core.metrics import _f1_score, _bleu\n",
```

So some false positives, a a couple obscure project_specific spots that aren't important.

PR #1942, opened 2019-08-20
[TransformerRanker] Allow passing context segments into MemNet forward

emilydinan:
**Patch description**
Allow passing segments for the context encoder through the forward for the TransformerMemNet module (which is used by the bi-ranker). cc @hadasah

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -220,7 +220,7 @@ class TransformerMemNetModel(nn.Module):
 
         return encoded
 
-    def encode_context_memory(self, context_w, memories_w):
+    def encode_context_memory(self, context_w, memories_w, context_segments=None):
         """Encode the memories."""
         # [batch, d]
         if context_w is None:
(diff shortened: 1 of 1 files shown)


parlai/agents/transformer/modules.py
```diff
@@ -243,9 +243,11 @@ def encode_context_memory(self, context_w, memories_w):
 
         return weights, context_h
 
-    def forward(self, xs, mems, cands):
+    def forward(self, xs, mems, cands, context_segments=None):
         """Forward pass."""
```
stephenroller: can we add parameter explanations?

emilydinan: sure

emilydinan:
> Where's the sub-thing that uses this? anyway it's fine.

nothing is using this right now. @hadasah has a use case in which she'd like to use different segment IDs for the context input in the transformer biranker. the transformer encoder forward takes `segments` as an argument, but it is not exposed when it is invoked through the TransformerMemNet module (which is what the biranker is using). this simply exposes it as an option

PR #1944, opened 2019-08-21
Task feedback

hadasah:
**Patch description**
If you set get_task_feedback: True in an mturk task, this will show a screen for collecting feedback after the last task. Default value for get_task_feedback is False.

Putting this in core_components since it seems like a pretty general feature, but I can also probably move this into just my task if this doesn't seem useful.

**Testing steps**
Run any of the internal eval scripts I wrote which sets get_task_feedback to True.


parlai/mturk/core/mturk_manager.py
```diff
@@ -582,7 +582,7 @@ def _on_alive(self, pkt):
             completed_assignments = worker_state.completed_assignments()
             max_hits = self.max_hits_per_worker
             if (self.is_unique and completed_assignments > 0) or (
-                max_hits != 0 and completed_assignments > max_hits
+                max_hits != 0 and completed_assignments >= max_hits
```
JackUrb: Good catch

parlai/mturk/core/mturk_manager.py
```diff
@@ -1986,10 +1985,13 @@ def _on_alive(self, pkt):
...
+                worker_state.completed_assignments()
+                + worker_state.active_conversation_count()
+            )
             max_hits = self.max_hits_per_worker
             if (self.is_unique and completed_assignments > 0) or (
-                max_hits != 0 and completed_assignments > max_hits
+                max_hits != 0 and completed_assignments >= max_hits
             ):
                 text = (
                     'You have already participated in this HIT the maximum '
```
JackUrb: I'd add to this description that they may just be currently participating in a hit that puts them over the max

stephenroller:
Let's finish this up please.

hadasah:
Woah. I completely forgot about this. Will update tmw or Monday.

stephenroller:
ya after you sleep lol

hadasah:
@JackUrb turns out copying these changes to dev is kinda involved because I haven't been copying any of the changes I made to dev (e.g. subtasks). I might land this first and then make all of the copied changes at a later time.

PR #1949, opened 2019-08-23
Bert cross fix

shrimai:
fixes the padding issue for bert cross-encoder.


parlai/agents/bert_ranker/cross_encoder_ranker.py
```diff
@@ -67,12 +67,41 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
             fp16=self.opt.get('fp16'),
         )
 
+    def concat_without_padding(self, text_idx, cand_idx, null_idx=0):
```
emilydinan: if this is the exact same functino in transformer/crossencoder (which i believe it is, correct me if i'm wrong) then it should be refactored into a utility function and imported here

emilydinan: thank you for fixing this 😄

shrimai: done.

PR #1950, opened 2019-08-23
[Tests] add a comprehensive TRA test

emilydinan:
**Patch description**
Add a test that tests a torch ranker agent with various candidate options, both training and evaluating.

Also fixed a bug in the torch_ranker_agent. We shouldnt' automatically set `encode_candidate_vecs` to True in interactive mode, as agents like the cross encoder should not pre-encode candidates (as the representations are dependent on context).


stephenroller:
YESSSS

emilydinan:
Right now this only tests the transformer/ranker. We can have it test every TRA by running through a list, but the test will become quite long. Thoughts @stephenroller ?

PR #1969, opened 2019-08-27
Continued battle with test_distributed

stephenroller:
**Patch description**
We're finding its very difficult to control for random seed exactly across all workers, causing models to initialize slightly differently. Previously, we just checked that the norm of the model parameters was the same, and raised an exception when they weren't. This test was failing quite frequently.

Now instead, we just explicitly use the primary worker's copy of the parameters, and force a sync at the beginning of the program. This is much easier than trying to control the random seed, and is relatively cheap compared to the full training run.

OLD MESSAGE: @gchanan and @mrshenli suggested we actually check all the parameters rather than an aggregation, possibly improving the reliability of distributed.

**Testing steps**
We're going to run CircleCI 20 times.

### DIFF
--- a/parlai/core/build_data.py
+++ b/parlai/core/build_data.py
@@ -21,7 +21,11 @@ import hashlib
 import tqdm
 import math
 import zipfile
-from multiprocessing import Pool
+
+try:
+    from torch.multiprocessing import Pool
+except ImportError:
+    from multiprocessing import Pool
 
 
 class DownloadableFile:
--- a/parlai/utils/distributed.py
+++ b/parlai/utils/distributed.py
@@ -243,14 +243,17 @@ def sync_parameters(model: torch.nn.Module) -> bool:
         # if things aren't distributed, of course things are in sync
         return True
 
-    # syncrhonize all the parameters
+    # sync all the parameters
     with torch.no_grad():
         for p in model.parameters():
             if not is_primary_worker():
+                # zero out parameters on all workers EXCEPT the primary worker
                 p.data.zero_()
-            dist.all_reduce(p.data)
+            # sum the parameters across all workers, resulting in everyone having
+            # the parameters of the primary worker
+            dist.all_reduce(p.data, dist.ReduceOp.SUM)
 
-    # double check everything synced
+    # double check everything synced correctly
     norm2 = sum((p.data ** 2).sum().float() for p in model.parameters()).item()
     all_versions = all_gather_list(norm2)
     if not all(n == norm2 for n in all_versions):
--- a/parlai/utils/distributed.py
+++ b/parlai/utils/distributed.py
@@ -17,6 +17,7 @@ import pickle
 import contextlib
 
 try:
+    import torch.nn
     import torch.version
     import torch.distributed as dist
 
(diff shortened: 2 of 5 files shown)


stephenroller:
Ready for reviww

parlai/core/build_data.py
```diff
@@ -21,7 +21,11 @@
 import tqdm
 import math
 import zipfile
-from multiprocessing import Pool
+
+try:
+    from torch.multiprocessing import Pool
```
stephenroller: using torch.multiprocessing fixes the issues around spawn that I was originally fighting, hence the changes in the import method.

parlai/utils/distributed.py
```diff
@@ -229,21 +230,27 @@ def sync_object(data, max_size=16384):
...
+    Sync all parameters across all workers are the same.
 
-    Always returns True, or raises an AssertionError if they are not
-    synchronized.
+    Always returns True, or raises an AssertionError if there was a failure.
 
-    :param torch.nn.Module model: A pytorch model.
-    :return: True
+    :param model: A pytorch model.
+    :return: always True
```
klshuster: why?

parlai/utils/distributed.py
```diff
@@ -229,21 +230,27 @@ def sync_object(data, max_size=16384):
...
     """
     if not is_distributed():
         # if things aren't distributed, of course things are in sync
         return True
 
-    # compute the local norm:
+    # syncrhonize all the parameters
```
klshuster: nit: synchronize

parlai/utils/distributed.py
```diff
@@ -229,21 +230,27 @@ def sync_object(data, max_size=16384):
...
+    with torch.no_grad():
+        for p in model.parameters():
+            if not is_primary_worker():
+                p.data.zero_()
+            dist.all_reduce(p.data)
```
klshuster: does all_reduce need to be called with primary worker or is it ok for any of the workers to call it? seems like this would make all params 0?

parlai/utils/distributed.py
stephenroller: how do i manage to write any code correctly?

parlai/utils/distributed.py
stephenroller: it could return None. it just felt like a nice thing to me.

parlai/utils/distributed.py
stephenroller: all_reduce sums the result across all workers, including the primary worker. Hence we want to zero everyone out, EXCEPT the primary worker, and then take the sum. The result will be everyone has the primary worker's parameters.

parlai/utils/distributed.py
klshuster: ahh i am ok with `True` I just wasn't sure if it was being used

parlai/utils/distributed.py
klshuster: ok, thank you, didn't realize the implicit op was sum but that makes sense

stephenroller: it has an optional reduce arg which I'll add for clarity

PR #1972, opened 2019-08-28
fixed bug for force_set of _set_text_vec with start/end tokens and mi…

mwillwork:
…sspelling of criterion in torch classifier agent


emilydinan:
@mwillwork is this good to merge?

PR #1985, opened 2019-09-04
Dialogue safety: transformer classifier agent and teachers

emilydinan:
**Patch description**
This patch adds:
- A transformer classifier agent
- Dialogue safety teachers including: adversarial teacher, standard teacher, and multiturn teacher, as well as a WikiToxicComments teacher (note: the WikiToxicComments teacher does not automatically download the data as one must agree to competition rules first.

To come (in a future PR):
- Updated project README
- Pretrained models


parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+
+# Constants
+OK_CLASS = '__ok__'
+NOT_OK_CLASS = '__notok__'
+
+SINGLE_TURN_DATA = 'single_turn_safety.json'
+MULTI_TURN_DATA = 'multi_turn_safety.json'
+
+
+def build(datapath):
```
klshuster: guess we're doing this now instead of a separate `build.py` file?

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+
+    def num_episodes(self):
+        return len(self.data)
+
+    def num_examples(self):
+        return len(self.data)
+
+    def _setup_data(self, datatype):
+        # load data
+        self.data_dump = json.load(open(self.data_path, 'rb'))['standard']
```
klshuster: nit: could you use `with open(...` logic here? looks like you're leaving a file open

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+            for i in range(self.round - 1):
+                rnd = str(i + 1)
+                for x in ['good', 'bad']:
+                    self.data += self.data_dump[d_type][rnd][x]
+
+        # add data from current round
+        for x in ['good', 'bad']:
+            self.data += self.data_dump[d_type][str(self.round)][x]
+
+        self.fixed_random.shuffle(self.data)
```
klshuster: why do we need this? won't training shuffle it anyway?

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+    """
+    Data from the adversarial collection described in the paper
+    `Build it Break it Fix it for Dialogue Safety: Robustness from
+    Adversarial Human Attack` (<https://arxiv.org/abs/1908.06083>)
+
+    To see data from rounds 1, 2, and 3, try running:
+    `python examples/display_data.py -t dialogue_safety:adversarial --round 3`
+
+    To see data from round 2 only, try running:
+    `python examples/display_data.py -t dialgoue_safety:adversarial --round 2
```
klshuster: nit: dialogue*

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+        self.fixed_random = random.Random(42)
+        self.round = opt['round']
+        self.label_candidates = [NOT_OK_CLASS, OK_CLASS]
+
+        if shared and 'data' in shared:
+            self.data = shared['data']
+        else:
+            self._setup_data(opt['datatype'])
+
+        opt = copy.deepcopy(opt)
```
klshuster: why the deepcopy here?

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+        return shared
+
+
+class MultiturnTeacher(FixedDialogTeacher):
+    """
+    Data from the multi-turn adversarial collection described in the paper
+    `Build it Break it Fix it for Dialogue Safety: Robustness from
+    Adversarial Human Attack` (<https://arxiv.org/abs/1908.06083>)
+
+    To see data containing mulit-turn conversations, try running
```
klshuster: nit: multi*

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+    def get(self, episode_idx, entry_idx):
+        return Message(self.data[episode_idx])
+
+    def share(self):
+        shared = super().share()
+        shared['data'] = self.data
+        return shared
+
+
+class AdversarialTeacher(FixedDialogTeacher):
```
klshuster: this looks identical to the `StandardTeacher` except for one line in `_setup_data`, perhaps this can subclass `Standard` and just change that one line?

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+        BLUE = '\033[1;96m'
+        CYAN = '\033[1;94m'
+        MAGENTA = '\033[1;95m'
+
+        # only use colors if we're outputting to a terminal
+        USE_COLORS = _sys.stdout.isatty()
+        if not USE_COLORS:
+            RESET = RED = YELLOW = GREEN = BLUE = CYAN = MAGENTA = ''
+
+        # generate the rainbow stars
```
klshuster: assuming this is just for aesthetic?

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+        with open(os.path.join(self.data_path, file_name), 'w') as f:
+            f.write(json.dumps(response, indent=4))
+
+    def build(self, opt):
+        self._get_data()
+        version = 'v2.0'
+        read_path = self.data_path
+        if not build_data.built(self.data_path, version):
+            print('[building data from : ' + read_path + ']')
+            if build_data.built(self.data_path):
```
klshuster: not sure we need this check - if the data is built, then this will just remove the data (since you are not redownloading anything in this function)

parlai/tasks/dialogue_safety/agents.py
```diff
@@ -0,0 +1,533 @@
...
+        else:
+            self.data = data
+
+    def get(self, episode_idx, entry_idx):
+        return Message(self.data[episode_idx])
+
+
+class WikiToxicCommentsTeacher(FixedDialogTeacher):
+    """
+    Dataset of comments from Wikipedia's Talk page edits. Taken from
```
klshuster: might be good to also specify that you're converting it to a binary classification task

parlai/agents/transformer_classifier/transformer_classifier.py
```diff
@@ -0,0 +1,73 @@
...
+        num_classes = len(self.class_list)
+        self.base_model = TransformerMemNetModel(self.opt, self.dict)
+        return TransformerLinearWrapper(self.base_model.context_encoder, num_classes)
+
+    def vectorize(self, *args, **kwargs):
+        """ Add the start and end token to the text.
+        """
+        kwargs['add_start'] = True
+        kwargs['add_end'] = True
+        obs = TorchRankerAgent.vectorize(self, *args, **kwargs)
```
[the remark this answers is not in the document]
emilydinan: good call, because i wasn't actually inheriting from TRA when i first wrote it (was inheriting from transformer ranker)

parlai/tasks/dialogue_safety/agents.py
emilydinan: i don't really know. what do you prefer?

parlai/tasks/dialogue_safety/agents.py
emilydinan: yes 😄

parlai/tasks/dialogue_safety/agents.py
klshuster: i don't personally care, though maybe good to move to a `build.py` file as that is what we specify in the official docs to do (so might as well keep that consistent)

parlai/tasks/dialogue_safety/agents.py
emilydinan: good catch, not necessary

parlai/tasks/dialogue_safety/agents.py
emilydinan: this checks if the data exists but NOT the specific version of the data (i.e. the above checks if it has the right version)

parlai/tasks/dialogue_safety/agents.py
emilydinan: otherwise it's all 'good' data followed by all 'bad' data, so it will look weird when you display

emilydinan:
this probably deserves another quick glance due to substantial changes cc @klshuster

parlai/tasks/dialogue_safety/agents.py
klshuster: yeah, i meant for this specific teacher, since it seems that the downloading of this wiki dataset happens outside of ParlAI right?

parlai/tasks/dialogue_safety/base_agent.py
```diff
@@ -0,0 +1,100 @@
...
+        if shared and 'data' in shared:
+            self.data = shared['data']
+        else:
+            self._setup_data(opt['datatype'])
+
+        super().__init__(opt, shared)
+        self.reset()
+
+    @abstractmethod
+    def _load_data_dump(self):
```
klshuster: perhaps `raise NotImplementedError()` here as this is mandatory

parlai/tasks/dialogue_safety/agents.py
emilydinan: ohhh gotcha yeah. i'll delete

parlai/tasks/dialogue_safety/base_agent.py
emilydinan: the decorator abstractmethod throws an error if a subclass is instantiated without this function implemented

PR #1993, opened 2019-09-13
[Poly-encoders] [GPU] Updates to Poly-encoder (and some Bi-encoder)

klshuster:
**Patch description**
This patch does the following:

1. Bi-encoder:
    - updates to work with a fixed-candidate-set
2. TransformerEncoder:
    - updates to allow returning the position embeddings if reduction type is none
3. BasicAttention:
    - updates to allow an optional `values` arg, which can be separate from the `ys` (which are assumed to be the keys in the attention)
4. Poly-encoder:
    - Fixes an issue when loading a pretrained model when using `--polyencoder-type codes`
    - Allows for using the position embeddings as keys into the attention
    - override `BasicAttention`, as there is a squeeze issue when using `--polyencoder-type codes --poly-n-codes 1`
     

**Testing steps**
Thoroughly tested via sweeps. 

**Logs**
Previously, when running the poly-encoder training script from https://parl.ai/projects/polyencoder/ with `--polyencoder-type codes`, one gets the following error:
```
RuntimeError: Error(s) in loading state_dict for PolyEncoderModule:
	Missing key(s) in state_dict: "codes".
```
Now, we no longer get this issue

Additionally, when running with `--polyencoder-type codes --poly-n-codes 1`, one gets the following error: 
```
  File "examples/train_model.py", line 16, in <module>
    TrainLoop(opt).train()
  File "/private/home/kshuster/TestParlAI/parlai/scripts/train_model.py", line 719, in train
    world.parley()
  File "/private/home/kshuster/TestParlAI/parlai/core/worlds.py", line 737, in parley
    batch_act = self.batch_act(agent_idx, batch_observations[agent_idx])
  File "/private/home/kshuster/TestParlAI/parlai/core/worlds.py", line 705, in batch_act
    batch_actions = a.batch_act(batch_observation)
  File "/private/home/kshuster/TestParlAI/parlai/core/torch_agent.py", line 1720, in batch_act
    output = self.train_step(batch)
  File "/private/home/kshuster/TestParlAI/parlai/core/torch_ranker_agent.py", line 352, in train_step
    scores = self.score_candidates(batch, cand_vecs)
  File "/private/home/kshuster/TestParlAI/parlai/agents/transformer/polyencoder.py", line 149, in score_candidates
    ctxt_rep=ctxt_rep, ctxt_rep_mask=ctxt_rep_mask, cand_rep=cand_rep
  File "/private/home/kshuster/miniconda3/lib/python3.6/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "/private/home/kshuster/miniconda3/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 152, in forward
    outputs = self.parallel_apply(replicas, inputs, kwargs)
  File "/private/home/kshuster/miniconda3/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 162, in parallel_apply
    return parallel_apply(replicas, inputs, kwargs, self.device_ids[:len(replicas)])
  File "/private/home/kshuster/miniconda3/lib/python3.6/site-packages/torch/nn/parallel/parallel_apply.py", line 83, in parallel_apply
    raise output
  File "/private/home/kshuster/miniconda3/lib/python3.6/site-packages/torch/nn/parallel/parallel_apply.py", line 59, in _worker
    output = module(*input, **kwargs)
  File "/private/home/kshuster/miniconda3/lib/python3.6/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "/private/home/kshuster/TestParlAI/parlai/agents/transformer/polyencoder.py", line 339, in forward
    return self.score(ctxt_rep, ctxt_rep_mask, cand_rep)
  File "/private/home/kshuster/TestParlAI/parlai/agents/transformer/polyencoder.py", line 316, in score
    self.attention, cand_embed, ctxt_rep, ctxt_rep_mask
  File "/private/home/kshuster/TestParlAI/parlai/agents/transformer/polyencoder.py", line 241, in attend
    return attention_layer(queries, keys, mask)
  File "/private/home/kshuster/miniconda3/lib/python3.6/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "/private/home/kshuster/TestParlAI/parlai/agents/transformer/modules.py", line 854, in forward
    l1 = torch.bmm(xs, ys.transpose(1, 2))
IndexError: Dimension out of range (expected to be in range of [-2, 1], but got 2)
```
With the fix, this no longer happens

### DIFF
--- a/parlai/agents/transformer/biencoder.py
+++ b/parlai/agents/transformer/biencoder.py
@@ -21,6 +21,13 @@ class BiencoderAgent(TransformerRankerAgent):
         obs = TorchRankerAgent.vectorize(self, *args, **kwargs)
         return obs
 
+    def _vectorize_text(self, *args, **kwargs):
+        """override to add start end tokens. necessary for fixed cands"""
+        if 'add_start' in kwargs:
+            kwargs['add_start'] = True
+            kwargs['add_end'] = True
+        return super()._vectorize_text(*args, **kwargs)
+
     def _set_text_vec(self, *args, **kwargs):
         """ Add the start and end token to the text.
         """
--- a/parlai/agents/transformer/biencoder.py
+++ b/parlai/agents/transformer/biencoder.py
@@ -1,6 +1,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+"""Bi-encoder Agent."""
 
 from .transformer import TransformerRankerAgent
 from parlai.core.torch_ranker_agent import TorchRankerAgent
@@ -8,29 +9,28 @@ import torch
 
 
 class BiencoderAgent(TransformerRankerAgent):
-    """
+    """Bi-encoder Transformer Agent.
+
     Equivalent of bert_ranker/biencoder but does not rely on an external
     library (hugging face).
     """
 
     def vectorize(self, *args, **kwargs):
-        """ Add the start and end token to the text.
-        """
+        """Add the start and end token to the text."""
         kwargs['add_start'] = True
         kwargs['add_end'] = True
         obs = TorchRankerAgent.vectorize(self, *args, **kwargs)
         return obs
 
     def _vectorize_text(self, *args, **kwargs):
-        """override to add start end tokens. necessary for fixed cands"""
+        """Override to add start end tokens. necessary for fixed cands."""
         if 'add_start' in kwargs:
             kwargs['add_start'] = True
             kwargs['add_end'] = True
         return super()._vectorize_text(*args, **kwargs)
 
     def _set_text_vec(self, *args, **kwargs):
-        """ Add the start and end token to the text.
-        """
+        """Add the start and end token to the text."""
         obs = super()._set_text_vec(*args, **kwargs)
         if 'text_vec' in obs and 'added_start_end_tokens' not in obs:
             obs.force_set(
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -77,6 +77,7 @@ def _build_encoder(
         activation=opt['activation'],
         variant=opt['variant'],
         output_scaling=opt['output_scaling'],
+        get_pos_embs=opt['get_pos_embs']
     )
 
 
@@ -356,6 +357,8 @@ class TransformerEncoder(nn.Module):
         Future versions may support things like GPT-2, ...
     :param output_scaling:
         Scale the outputs by a given scalar
+    :param bool get_pos_embs:
+        If true, return the position embeddings when reduction type is none
     """
 
     def __init__(
@@ -378,6 +381,7 @@ class TransformerEncoder(nn.Module):
         variant='aiayn',
         n_segments=0,
         output_scaling=1.0,
+        get_pos_embs=False
     ):
         super(TransformerEncoder, self).__init__()
 
@@ -454,6 +458,7 @@ class TransformerEncoder(nn.Module):
                 )
             )
         self.output_scaling = output_scaling
+        self.get_pos_embs = get_pos_embs
 
     def forward(self, input, positions=None, segments=None):
         """
@@ -480,7 +485,8 @@ class TransformerEncoder(nn.Module):
                     x=positions.max().item(), y=self.n_positions
                 )
             )
-        tensor = tensor + self.position_embeddings(positions).expand_as(tensor)
+        position_embs = self.position_embeddings(positions).expand_as(tensor)
+        tensor = tensor + position_embs
 
         if self.n_segments >= 1:
             if segments is None:
@@ -508,7 +514,10 @@ class TransformerEncoder(nn.Module):
             return output
         elif self.reduction_type == 'none' or self.reduction_type is None:
             output = tensor
-            return output, mask
+            ret = (output, mask)
+            if self.get_pos_embs:
+                ret = (output, mask, position_embs)
+            return ret
         else:
             raise ValueError(
                 "Can't handle --reduction-type {}".format(self.reduction_type)
@@ -856,9 +865,11 @@ class BasicAttention(nn.Module):
         self.get_weights = get_weights
         self.residual = residual
 
-    def forward(self, xs, ys, mask_ys=None):
-        """ xs: B x query_len x dim
-            ys: B x key_len x dim
+    def forward(self, xs, ys, mask_ys=None, values=None):
+        """ xs: B x query_len x dim (queries)
+            ys: B x key_len x dim (keys)
+            values: B x value_len x dim (values)
+                    if None, default to ys
             TODO: Document this
         """
         bsz = xs.size(0)
@@ -876,7 +887,9 @@ class BasicAttention(nn.Module):
             attn_mask = attn_mask.repeat(1, x_len, 1)
             l1.masked_fill_(attn_mask, -float('inf'))
         l2 = self.softmax(l1)
-        lhs_emb = torch.bmm(l2, ys)
+        if values is None:
+            values = ys
+        lhs_emb = torch.bmm(l2, values)
 
         # # add back the query
         if self.residual:
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -510,7 +510,7 @@ class TransformerEncoder(nn.Module):
             divisor = mask.float().sum(dim=1).unsqueeze(-1).clamp(min=1).type_as(tensor)
             output = tensor.sum(dim=1) / divisor
             return output
-        elif 'none' in self.reduction_type or self.reduction_type is None:
+        elif self.reduction_type is None or 'none' in self.reduction_type:
             output = tensor
             ret = (output, mask)
             if self.reduction_type == 'none_with_pos_embs':
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -77,7 +77,6 @@ def _build_encoder(
         activation=opt['activation'],
         variant=opt['variant'],
         output_scaling=opt['output_scaling'],
-        get_pos_embs=opt['get_pos_embs']
     )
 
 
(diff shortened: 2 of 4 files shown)


parlai/agents/transformer/polyencoder.py
```diff
@@ -267,16 +289,19 @@ def encode(self, ctxt_tokens, cand_tokens):
...
                     self.code_attention,
-                    self.codes.repeat(bsz, 1, 1),
-                    ctxt_out,
-                    ctxt_mask,
+                    queries=self.codes.repeat(bsz, 1, 1),
+                    keys=att_keys,
+                    values=ctxt_out,
+                    mask=ctxt_mask,
                 )
+                ctxt_pos = None  # we don't need this anymore
```
stephenroller: why don't we need this anymore?

parlai/agents/transformer/transformer.py
```diff
@@ -91,6 +91,13 @@ def add_common_cmdline_args(argparser):
         help='Share word embeddings table for candidate and context'
         'in the memory network',
     )
+    argparser.add_argument(
+        '--get-pos-embs',
+        type='bool',
+        default=False,
+        help='If true, return the position embeddings from the encoder '
```
stephenroller: This feels like it shouldn't be a user option, or at least could be rolled into the reduction_type option?

parlai/agents/transformer/polyencoder.py
klshuster: ctxt position embs are used as attention keys only when the values are the transformer outputs. for the `codes` case, we do not need them for the secondary attention over the `codes`

parlai/agents/transformer/biencoder.py
```diff
@@ -21,6 +21,13 @@ def vectorize(self, *args, **kwargs):
         obs = TorchRankerAgent.vectorize(self, *args, **kwargs)
         return obs
 
+    def _vectorize_text(self, *args, **kwargs):
+        """override to add start end tokens. necessary for fixed cands"""
+        if 'add_start' in kwargs:
+            kwargs['add_start'] = True
+            kwargs['add_end'] = True
+        return super()._vectorize_text(*args, **kwargs)
```
emilydinan: a few questions:
1) do we still need these `_set_text_vec` and `vectorize` overrides? 
2) why do we have this check here for `if 'add_start' in kwargs`?
3) we may experience the problem where the observation is cached and we continually add start/end tokesn to the labels at each epoch. how do we get around this? the way i handled it for the text case is to check for the presence of an `added_start_end_tokens` key in the `_set_text_vec` function (see below). this probably requires some careful thought/may be cleaner with a refactor. cc @stephenroller.

parlai/agents/transformer/modules.py
```diff
@@ -856,9 +865,11 @@ def __init__(self, dim=1, attn='cosine', residual=False, get_weights=True):
...
         self.residual = residual
 
-    def forward(self, xs, ys, mask_ys=None):
-        """ xs: B x query_len x dim
-            ys: B x key_len x dim
+    def forward(self, xs, ys, mask_ys=None, values=None):
+        """ xs: B x query_len x dim (queries)
+            ys: B x key_len x dim (keys)
+            values: B x value_len x dim (values)
+                    if None, default to ys
```
emilydinan: can you get rid of this TODO?

parlai/agents/transformer/polyencoder.py
```diff
@@ -86,7 +96,8 @@ def __init__(self, opt, shared=None):
             self.model = torch.nn.DataParallel(self.model)
 
     def build_model(self, states=None):
-        return PolyEncoderModule(self.opt, self.dict, self.NULL_IDX)
+        self.model = PolyEncoderModule(self.opt, self.dict, self.NULL_IDX)
+        return self.model
```
emilydinan: i don't think you need to set self.model here, since the call to build_model is like `self.model = self.build_model(...)`, so this could just be `return  PolyEncoderModule(self.opt, self.dict, self.NULL_IDX)`

parlai/agents/transformer/polyencoder.py
```diff
@@ -139,17 +150,24 @@ def score_candidates(self, batch, cand_vecs, cand_encs=None):
...
         scores = self.model(
-            ctxt_rep=ctxt_rep, ctxt_rep_mask=ctxt_rep_mask, cand_rep=cand_rep
+            ctxt_rep=ctxt_rep, ctxt_rep_mask=ctxt_rep_mask, cand_rep=cand_rep, ctxt_pos=ctxt_pos
         )
         return scores
 
+    def load_state_dict(self, state_dict):
+        """Override to account for codes"""
+        if self.model.type == 'codes' and 'codes' not in state_dict:
+            state_dict['codes'] = self.model.codes
```
emilydinan: interesting solution, we could just do a `load_state_dict(..., strict=False)` but that might not capture other errors/keys missing

parlai/agents/transformer/polyencoder.py
```diff
@@ -227,14 +246,16 @@ def get_encoder(self, opt, dict, null_idx, reduction_type):
...
+    def attend(self, attention_layer, queries, keys, values, mask):
         """ Unify the API of MultiHeadAttention and
-            BasicAttention that are slighlty different
+            BasicAttention that are slightly different
         """
-        if isinstance(attention_layer, BasicAttention):
-            return attention_layer(queries, keys, mask)
+        if keys is None:
+            keys = values
+        if isinstance(attention_layer, PolyBasicAttention):
```
emilydinan: i see that it was like this before, but i find this a bit strange. can we set something like `self.attention_type` so we don't have to check the module name every time?

parlai/agents/transformer/polyencoder.py
```diff
@@ -285,29 +310,32 @@ def encode(self, ctxt_tokens, cand_tokens):
...
             :param ctxt_rep_mask: 2D byte tensor, bsz x ctxt_len, in case
                 there are some elements of the ctxt that we should not take into
                 account.
+            :param ctx_pos: 3D float tensor, bsz x sent_len x dim
             :param cand_embed: 3D float tensor, bsz x num_cands x dim
 
             :returns: scores, 2D float tensor: bsz x num_cands
         """
-
+        keys = ctxt_rep if self.attention_keys == 'context' else ctxt_pos
```
emilydinan: can you add a comment in the code here explaining this

parlai/agents/transformer/polyencoder.py
```diff
@@ -330,5 +359,19 @@ def forward(
...
+class PolyBasicAttention(BasicAttention):
+    def __init__(self, poly_type, n_codes, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.poly_type = poly_type
+        self.n_codes = n_codes
+
+    def forward(self, *args, **kwargs):
+        lhs_emb = super().forward(*args, **kwargs)
+        if self.poly_type == 'codes' and self.n_codes == 1 and len(lhs_emb.shape) == 2:
+            lhs_emb = lhs_emb.unsqueeze(self.dim - 1)
```
emilydinan: can you explain this?

parlai/agents/transformer/transformer.py
klshuster: i think this might be the way to go; at first I had just turned it on by default but figured people would want to have a way to turn it off. but changing to work with reduction type of, e.g., `none_with_pos` might be better

parlai/agents/transformer/biencoder.py
klshuster: 1. I *think* so - this relates to the issue where the fixed candidate labels are vectorized directly via the `_vectorize_text` function rather than the `vectorize` function. 
2. Sometimes, this function is called with `add_start` as a positional arg, rather than a keyword arg; i am overriding this specifically for the case of building fixed cands, which calls this with `_vectorize_text(..., add_start=...)`, so i check for that case. Also, if i do add it in when it's called as a positional arg, I'll get an error where the function is passed two values for `add_start`
3. Yeah, this is actually handled below in the `_set_text_vec` function. Ideally it is handled in the `_set_label_vec` function in the torch agent as well.

parlai/agents/transformer/modules.py
klshuster: yeah i'll try to add some documentation

parlai/agents/transformer/polyencoder.py
klshuster: I thought about that too, but I agree that it's better that we fail loudly when we pass in incorrect state dicts

parlai/agents/transformer/polyencoder.py
klshuster: I think the reason it is like this is that the codes version has two levels of attention, which can be different (i.e. there is a `self.codes_attention_type` as well), so we'd have to add another flag in here somewhere anyway

parlai/agents/transformer/polyencoder.py
klshuster: This is a bit of a tricky situation. When we are using the `codes` polyencoder, the output of the first level of attention is `bsz x n_codes x hsz`; since there is only 1 code, the `BasicAttention` squeezes to make it `bsz x hsz`; thus, we need to unsqueeze to have the appropriate dimensionality for use later in the pipeline.

klshuster:
I've addressed most of your comments @stephenroller @emilydinan, also i blacked/linted so tests should pass

parlai/agents/transformer/biencoder.py
emilydinan: re: 3 -- after looking into it a bit more, ti should be handled in set_label_vec because we do not call _vectorize_text again if the label_vec is inside the obs

emilydinan:
it's not clear to me why all of the transformer generator tests failed

klshuster:
it's because i broke something oops

parlai/agents/transformer/biencoder.py
salama4: so sorry, i'm junior, how to download the whole of these models to replace the old ones in my pc, cause the aren't work properly

PR #2000, opened 2019-09-15
Solve opt key indexing issues using get

vedantpuri:
**Patch description**
Argument parsing without **explicitly** using the parser (by passing in an argument dictionary) causes issues with certain fields not being set for the model resulting in a key error when trying to access ```opt['keyname']```. This was fixed in this patch by swapping ```[]``` with the get method and setting defaults.

**Testing steps**
- Navigate to /parlai/scripts/eval_model.py
- Comment out everything in main
- Add the following to main:
```
opt = {"datapath": "../../data/", "datatype": "test"}
opt.update({"task": "convai2", "model_file": "models:controllable_dialogue/convai2_finetuned_baseline", 'log_every_n_secs': 2})
eval_model(opt)
```
**Logs**
Truncated Output:
```
[loading fbdialog data:../../data/ConvAI2/valid_self_original.txt]
2s elapsed: {'exs': 1, '%done': '0.01%', 'time_left': '16792s', 'accuracy': 0, 'f1': 0.2609, 'bleu': 1.018e-07, 'token_acc': 0.2778, 'loss': 2.567, 'ppl': 13.03}
4s elapsed: {'exs': 2, '%done': '0.03%', 'time_left': '17108s', 'accuracy': 0, 'f1': 0.2257, 'bleu': 5.098e-08, 'token_acc': 0.3548, 'loss': 2.757, 'ppl': 15.75}
```

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -690,7 +690,7 @@ class TorchAgent(ABC, Agent):
             self.model = shared['model']
             self.criterion = shared['criterion']
             self.metrics = shared['metrics']
-            if self.opt['batchsize'] == 1 or self.opt['interactive_mode']:
+            if self.opt['batchsize'] == 1 or self.opt.get('interactive_mode', False):
                 # if we're not using batching (e.g. mturk), then replies really need
                 # to stay separated
                 self.replies = {}
@@ -734,7 +734,7 @@ class TorchAgent(ABC, Agent):
         self.rank_candidates = opt['rank_candidates']
         self.add_person_tokens = opt.get('person_tokens', False)
         # set interactive mode or not according to options.
-        self.set_interactive_mode(opt['interactive_mode'], shared)
+        self.set_interactive_mode(opt.get('interactive_mode', False), shared)
 
     def build_history(self):
         """Return the constructed history object."""
--- a/parlai/scripts/eval_model.py
+++ b/parlai/scripts/eval_model.py
@@ -75,13 +75,14 @@ def _eval_single_world(opt, agent, task):
     log_time = TimeLogger()
 
     # max number of examples to evaluate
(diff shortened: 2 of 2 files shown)


stephenroller:
Do you mind elaborating when we expect this to be necessary?

vedantpuri:
Hi @stephenroller ,

So this was done to resolve errors me and @klshuster encountered yesterday at our CS5152 Hackathon while trying to make eval_model.py work from within google collab (without calling it from a python shell). The method we were trying to use didn't create a parser object and hence some of the fields were not set per say (eg. interactive-mode). The code I was trying to run in collab to do this is shown below:
```
from parlai.scripts import eval_model
opt = {"datapath": "/root/ParlAI/data", "datatype": "test"}
opt.update({"task": "convai2", "model_file": "models:controllable_dialogue/convai2_finetuned_baseline"})
eval_model .eval_model(opt)
```
However later one of our team found a way to get around the use of `get' by using the code below:
```
from parlai.scripts import eval_model
arg_dict = {"datapath": "/root/ParlAI/data", "datatype": "test"}
arg_dict.update({"task": "babi:task10k:1", "model": "memnn","model_file": "/tmp/babi_memnn", "eval_candidates": "vocab"})
cl_args = []
for arg, val in arg_dict.items():
  arg = '--' + arg.replace('_', '-')
  cl_args = cl_args + [arg, str(val)]
  opt = parser.parse_args(cl_args)
eval_model.eval_model(opt)
```

Now, even though we got around that problem using the code above, I was curious about shouldn't using `get' be the safer way to go about it even if it is not necessary ?

klshuster:
Per offline discussion with @vedantpuri, we feel this change is something we would perhaps want in the future, though maybe in the context of a larger change; closing out for now, will address at a future time.

stephenroller:
Let's discuss it tomorrow.

PR #2003, opened 2019-09-16
Check for shared before dataparallel call

hadasah:
**Patch description**
DataParallel objects get nested in at least these 5 files if __init__ is called with shared dict and data parallel is set to True. This isn't a problem is these nested DataParallel objects only observe, but it does become problematic if we want to call batch_act.

Fix is just to check for shared is None first

### DIFF
--- a/parlai/agents/bert_ranker/bi_encoder_ranker.py
+++ b/parlai/agents/bert_ranker/bi_encoder_ranker.py
@@ -46,7 +46,7 @@ class BiEncoderRankerAgent(TorchRankerAgent):
         super().__init__(opt, shared)
         # it's easier for now to use DataParallel when
         self.data_parallel = opt.get('data_parallel') and self.use_cuda
-        if self.data_parallel:
+        if self.data_parallel and shared is None:
             self.model = torch.nn.DataParallel(self.model)
         if is_distributed():
             raise ValueError('Cannot combine --data-parallel and distributed mode')
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -147,7 +147,8 @@ class TorchClassifierAgent(TorchAgent):
                     raise ValueError(
                         'Cannot combine --data-parallel and distributed mode'
                     )
-                self.model = torch.nn.DataParallel(self.model)
+                if shared is None:
+                    self.model = torch.nn.DataParallel(self.model)
         if shared:
             # We don't use get here because hasattr is used on optimizer later.
             if 'optimizer' in shared:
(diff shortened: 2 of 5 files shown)


parlai/agents/bert_ranker/bi_encoder_ranker.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, opt, shared=None):
         super().__init__(opt, shared)
         # it's easier for now to use DataParallel when
         self.data_parallel = opt.get('data_parallel') and self.use_cuda
-        if self.data_parallel:
+        if self.data_parallel and shared is None:
```
emilydinan: this is good but i'm just noticing that the `self.rank_loss` below is not correct. we should override `build_criterion` instead.

parlai/agents/bert_ranker/bi_encoder_ranker.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, opt, shared=None):
...
             self.model = torch.nn.DataParallel(self.model)
         if is_distributed():
```
emilydinan: this should probably go under the check for `if self.data_parallel`..

parlai/core/torch_classifier_agent.py
```diff
@@ -147,7 +147,8 @@ def __init__(self, opt, shared=None):
                     raise ValueError(
                         'Cannot combine --data-parallel and distributed mode'
                     )
-                self.model = torch.nn.DataParallel(self.model)
+                if shared is None:
```
emilydinan: let's get rid of this `if` here and put this entire `if self.use_cuda` block under the `else:` statement above

parlai/agents/bert_ranker/bi_encoder_ranker.py
stephenroller: ?

emilydinan: the `if_distributed()` call raises an error and says distributed and data_parallel aren't compatible. but it doesn't make sense to raise this error `if not self.data_parallel`...

emilydinan: (it is not related to the change that @hadasah made, it's an error i just noticed)

parlai/agents/bert_ranker/bi_encoder_ranker.py
stephenroller: Created a new issue: #2019.

parlai/agents/bert_ranker/bi_encoder_ranker.py
stephenroller: Created a new issue #2020.

parlai/core/torch_classifier_agent.py
stephenroller: Let's actually do this one in this PR.

Issue #1921, opened 2019-08-13
Mark Pytorch <1.1 as a hard failure


stephenroller:
With the release last week of PyTorch 1.2, we've decided to officially support PyTorch 1.1 and newer. This will be a rolling release, with only the last 2 versions of PyTorch supported at any time.

We are also marking PyTorch as a *mandatory* requirement of ParlAI. This has effectively been the case for a long time, but we still vaguely pretended otherwise.

There are still lots of places that have try-excepts around the import of torch. We should

1. Filter all those so they are just regular imports.
2. Check the pytorch version in somewhere relatively central. The top of torch_agent.py is probably reasonable.

Examples of places we use it:

https://github.com/facebookresearch/ParlAI/blob/ecb0240d87e1589cc7a406944db2925da3323915/parlai/agents/tfidf_retriever/utils.py#L15-L18

Others:

```
$ ag 'install pytorch'
parlai/agents/drqa/drqa.py
24:    raise ImportError('Need to install pytorch: go to pytorch.org')

parlai/agents/legacy_agents/seq2seq/torch_agent_v1.py
28:    raise ImportError('Need to install Pytorch: go to pytorch.org')

parlai/agents/tfidf_retriever/utils.py
18:    raise ImportError('Need to install Pytorch: go to pytorch.org')

parlai/agents/bert_ranker/README.md
11:In order to use those agents you need to install pytorch-pretrained-bert (https://github.com/huggingface/pytorch-pretrained-BERT). If you have not installed, running the model will prompt you to run:
12:```pip install pytorch-pretrained-bert```

parlai/agents/bert_ranker/helpers.py
16:        '`pip install pytorchr-pretrained-bert`.'

parlai/agents/bert_ranker/bert_dictionary.py
14:        'pip install pytorch-pretrained-bert'

parlai/agents/bert_classifier/bert_classifier.py
31:            "pip install pytorch-pretrained-bert"

parlai/core/image_featurizers.py
51:            raise ImportError('Need to install Pytorch: go to pytorch.org')

parlai/core/pytorch_data_teacher.py
30:    raise ImportError('Need to install Pytorch: go to pytorch.org')

parlai/core/torch_agent.py
47:    raise ImportError('Need to install Pytorch: go to pytorch.org')

parlai/scripts/extract_image_feature.py
123:            raise ImportError('Need to install Pytorch: go to pytorch.org')

parlai/tasks/coco_caption/agents.py
[...]

PR #2010, opened 2019-09-21
 Deprecate Torch < 1.1

vedantpuri:
**Patch description**
 Deprecate Torch < 1.1. Fixes  [#1921 ](https://github.com/facebookresearch/ParlAI/issues/1921).

**Testing steps**
Follow instructions to install ParlAI. It should install a pytorch version >= 1.1 (since I added that in requirements.txt). The test case ```python tests/datatests/test_new_tasks.py``` should run without failure.

**Logs**
Truncated Output:
```
$ python setup.py test -s tests.suites.unittests -v
…
test_generation_multi (test_seq2seq.TestHogwildSeq2seq)
This test uses a multi-turn task and multithreading. ... ok
…
Ran 127 tests in 658.747s

OK (skipped=6)
```
**A note on currently failing CircleCI tests**
- datatests: Couldn't run it locally since one (suspecting coco caption) or more of the tasks changed takes an enormous amount of time to download and train which my laptop isn't able to handle. 
- os_x tests: Strangely enough, on Circle CI the test hangs and never completes (even tried running locally for over 30 mins) for a pip installation of pytorch. Whereas if I install torch using conda, the tests pass locally (as can be seen in logs). I have added a comment above the test to mention this strange behavior.

To remedy os_x tests so that they do not spoil the current build when merged, I have modified circleci so that os_x tests use conda to install torch instead of pip

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,10 @@ installtorchcpuosx: &installtorchcpuosx
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.1.0'
+      source /Users/distiller/miniconda/etc/profile.d/conda.sh
+      conda config --set always_yes yes
+      conda activate base
+      conda install --quiet 'pytorch>=1.1.0' -c pytorch
 
 installtorchcpu36: &installtorchcpu36
   run:
@@ -106,6 +109,15 @@ setupcuda: &setupcuda
       sudo apt-key add /var/cuda-repo-10-1-local-10.1.105-418.39/7fa2af80.pub
       nvidia-smi
       pyenv global 3.6.5
+
+installconda: &installconda
+  run:
+    name: Install Anaconda
+    working_directory: ~/
+    command: |
+      brew install wget --with-libressl
+      wget -q https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh
+      bash miniconda.sh -b -p "$HOME"/miniconda
 
 buildwebsite: &buildwebsite
   run:
@@ -136,12 +148,16 @@ jobs:
     steps:
       - checkout
       - <<: *fixgit
+      - <<: *installconda
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchcpuosx
       - run:
           name: Unit tests (OSX)
-          command: python setup.py test -s tests.suites.unittests -v
+          command: |
+            source /Users/distiller/miniconda/etc/profile.d/conda.sh
+            conda activate base
+            python setup.py test -s tests.suites.unittests -v
 
   unittests_36:
     <<: *standard_cpu36
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ installtorchgpu11: &installtorchgpu11
   run:
     name: Install torch GPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.1.0'
+      pip3 install --progress-bar off 'torch>=1.3.0'
       pip3 install --progress-bar off 'git+https://github.com/rsennrich/subword-nmt.git#egg=subword-nmt' # bpe support
       pip3 install --progress-bar off pytorch-pretrained-bert
       pip3 install --progress-bar off torchtext
@@ -69,7 +69,7 @@ installtorchgpu12: &installtorchgpu12
   run:
     name: Install torch GPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.1.0'
+      pip3 install --progress-bar off 'torch>=1.3.0'
       pip3 install --progress-bar off 'git+https://github.com/rsennrich/subword-nmt.git#egg=subword-nmt' # bpe support
       pip3 install --progress-bar off pytorch-pretrained-bert
       pip3 install --progress-bar off torchtext
@@ -79,22 +79,19 @@ installtorchcpuosx: &installtorchcpuosx
   run:
     name: Install torch CPU and dependencies
     command: |
-      source /Users/distiller/miniconda/etc/profile.d/conda.sh
-      conda config --set always_yes yes
-      conda activate base
-      conda install --quiet 'pytorch>=1.1.0' -c pytorch
+      pip3 install --progress-bar off 'torch>=1.3.0'
 
 installtorchcpu36: &installtorchcpu36
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.1.0'
+      pip3 install --progress-bar off 'torch>=1.3.0'
 
 installtorchcpu37: &installtorchcpu37
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.1.0'
+      pip3 install --progress-bar off 'torch>=1.3.0'
 
 setupcuda: &setupcuda
   run:
@@ -109,15 +106,6 @@ setupcuda: &setupcuda
       sudo apt-key add /var/cuda-repo-10-1-local-10.1.105-418.39/7fa2af80.pub
       nvidia-smi
       pyenv global 3.6.5
-
-installconda: &installconda
-  run:
-    name: Install Anaconda
-    working_directory: ~/
-    command: |
-      brew install wget --with-libressl
-      wget -q https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh
-      bash miniconda.sh -b -p "$HOME"/miniconda
 
 buildwebsite: &buildwebsite
   run:
@@ -148,7 +136,6 @@ jobs:
     steps:
       - checkout
       - <<: *fixgit
-      - <<: *installconda
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchcpuosx
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ installtorchgpu11: &installtorchgpu11
   run:
     name: Install torch GPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch<1.2.0'
+      pip3 install --progress-bar off 'torch>=1.2.0'
       pip3 install --progress-bar off 'git+https://github.com/rsennrich/subword-nmt.git#egg=subword-nmt' # bpe support
       pip3 install --progress-bar off pytorch-pretrained-bert
       pip3 install --progress-bar off torchtext
@@ -79,19 +79,19 @@ installtorchcpuosx: &installtorchcpuosx
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.0.0-cp37-none-macosx_10_7_x86_64.whl
+      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.2.0-cp37-none-macosx_10_7_x86_64.whl
 
 installtorchcpu36: &installtorchcpu36
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.0.1.post2-cp36-cp36m-linux_x86_64.whl
+      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.2.0.post2-cp36-cp36m-linux_x86_64.whl
 
 installtorchcpu37: &installtorchcpu37
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.0.1.post2-cp37-cp37m-linux_x86_64.whl
+      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.2.0.post2-cp37-cp37m-linux_x86_64.whl
 
 setupcuda: &setupcuda
   run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,8 +142,6 @@ jobs:
       - run:
           name: Unit tests (OSX)
           command: |
-            source /Users/distiller/miniconda/etc/profile.d/conda.sh
-            conda activate base
             python setup.py test -s tests.suites.unittests -v
 
   unittests_36:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ installtorchgpu11: &installtorchgpu11
   run:
     name: Install torch GPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.2.0'
+      pip3 install --progress-bar off 'torch>=1.1.0'
       pip3 install --progress-bar off 'git+https://github.com/rsennrich/subword-nmt.git#egg=subword-nmt' # bpe support
       pip3 install --progress-bar off pytorch-pretrained-bert
       pip3 install --progress-bar off torchtext
@@ -69,7 +69,7 @@ installtorchgpu12: &installtorchgpu12
   run:
     name: Install torch GPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.2.0'
+      pip3 install --progress-bar off 'torch>=1.1.0'
       pip3 install --progress-bar off 'git+https://github.com/rsennrich/subword-nmt.git#egg=subword-nmt' # bpe support
       pip3 install --progress-bar off pytorch-pretrained-bert
       pip3 install --progress-bar off torchtext
@@ -79,19 +79,19 @@ installtorchcpuosx: &installtorchcpuosx
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.2.0-cp37-none-macosx_10_7_x86_64.whl
+      pip3 install --progress-bar off 'torch>=1.1.0'
 
 installtorchcpu36: &installtorchcpu36
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.2.0.post2-cp36-cp36m-linux_x86_64.whl
+      pip3 install --progress-bar off 'torch>=1.1.0'
 
 installtorchcpu37: &installtorchcpu37
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.2.0.post2-cp37-cp37m-linux_x86_64.whl
+      pip3 install --progress-bar off 'torch>=1.1.0'
 
 setupcuda: &setupcuda
   run:
--- a/parlai/core/distributed_utils.py
+++ b/parlai/core/distributed_utils.py
@@ -15,6 +15,7 @@ in non-distributed mode.
 import builtins
 import pickle
 import contextlib
+from parlai.core.utils import check_torch_version
 
 try:
     import torch.version
@@ -31,11 +32,7 @@ def validate_params(opt):
 
     Raises exceptions if anything is wrong, otherwise returns None.
     """
-    if torch.__version__ < "1.1.0":
-        raise ImportError(
-            "Please upgrade to PyTorch >=1.1; "
-            "visit https://pytorch.org for instructions."
-        )
+    check_torch_version()
 
     if opt.get('no_cuda', False):
         raise ValueError('Distributed mode only makes sense when using GPUs.')
--- a/parlai/core/distributed_utils.py
+++ b/parlai/core/distributed_utils.py
@@ -17,13 +17,8 @@ import pickle
 import contextlib
 from parlai.core.utils import check_torch_version
 
-try:
-    import torch.version
-    import torch.distributed as dist
-
-    TORCH_AVAILABLE = True
-except ImportError:
-    TORCH_AVAILABLE = False
+check_torch_version()
+import torch.distributed as dist
 
 
 def validate_params(opt):
@@ -50,7 +45,7 @@ def validate_params(opt):
 
 def is_distributed():
     """Return if we are in distributed mode."""
-    return TORCH_AVAILABLE and dist.is_available() and dist.is_initialized()
+    return dist.is_available() and dist.is_initialized()
 
 
 def num_workers():
--- a/parlai/core/distributed_utils.py
+++ b/parlai/core/distributed_utils.py
@@ -31,9 +31,9 @@ def validate_params(opt):
 
     Raises exceptions if anything is wrong, otherwise returns None.
     """
-    if torch.version.__version__.startswith('0.'):
+    if torch.__version__ < "1.1.0":
         raise ImportError(
-            "Please upgrade to PyTorch >=1.0; "
+            "Please upgrade to PyTorch >=1.1; "
             "visit https://pytorch.org for instructions."
         )
 
--- a/parlai/core/distributed_utils.py
+++ b/parlai/core/distributed_utils.py
@@ -18,7 +18,8 @@ import contextlib
 from parlai.core.utils import check_torch_version
 
 check_torch_version()
-import torch.distributed as dist
+import torch  # noqa: E402
+import torch.distributed as dist  # noqa: E402
 
 
 def validate_params(opt):
--- a/parlai/core/pytorch_data_teacher.py
+++ b/parlai/core/pytorch_data_teacher.py
@@ -24,14 +24,14 @@ import os
 from functools import wraps
 import importlib
 from functools import lru_cache
-import torch  # noqa: F401
-from torch.utils.data import ConcatDataset, Dataset, DataLoader, sampler
-from torch.multiprocessing import Lock, Value
 import ctypes
 from threading import Thread, Condition, RLock
 
 # Perform Torch version check
 check_torch_version()
+import torch  # noqa: E402
+from torch.utils.data import ConcatDataset, Dataset, DataLoader, sampler  # noqa: E402
+from torch.multiprocessing import Lock, Value  # noqa: E402
 
 
 class BatchSortCache(object):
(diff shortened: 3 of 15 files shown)


requirements.txt
```diff
@@ -21,6 +21,7 @@ scipy
 sh
 sphinx
 sphinx_rtd_theme
+torch>=1.1
```
klshuster: I'm not sure we want to add this here specifically, since there are many alternative ways of installing pytorch - this will use pip, while one might want to use conda and also might want to install with CUDA etc. @stephenroller do you have any thoughts on this?

parlai/core/distributed_utils.py
```diff
@@ -31,9 +31,9 @@ def validate_params(opt):
 
     Raises exceptions if anything is wrong, otherwise returns None.
     """
-    if torch.version.__version__.startswith('0.'):
+    if torch.__version__ < "1.1.0":
```
klshuster: i think this check and the one in `pytorch_data_teacher.py` are fine for now

vedantpuri: Do you mean I roll back to the previous check(```if torch.version.__version__.startswith('0.'):```) 
or keep the one I made (```if torch.__version__ < "1.1.0":```) ?

stephenroller: Use your new one, but can we unify them so they're all one call?

utils.check_torch_version()

so we don't have to upgrade in many places

parlai/tasks/coco_caption/agents.py
```diff
@@ -11,10 +11,8 @@
 from .build_2017 import build as build_2017
 from .build_2017 import buildImage as buildImage_2017
 
-try:
-    import torch  # noqa: F401
-except ImportError:
-    raise ImportError('Need to install Pytorch: go to pytorch.org')
+
+import torch  # noqa: F401
```
stephenroller: I think you can drop anything leftover with a `# noqa: F401`. They were imports only for checking versioning.

requirements.txt
stephenroller: Yes, what Kurt said.

vedantpuri: Okay I can remove this from here, but certain Circle CI tests like os_x tests specifically install pytorch 1.0.0 and hence those tests will fail as they are doing now.

Also, I am not sure how to deal with the beam search failing on this PR, Any thoughts ?

vedantpuri:
I made the needed changes but the os_x test failing is ditrubing because it seems to fail on multithreading tasks. Specifically: test_seq2seq.TestHogwildSeq2seq and test_memnn.TestMemnn (not in this commit but in a previous one). In this commit the se12seq failed since it took too long and the Time Limit Exceeded. In an earlier run of Circle CI, the two aforementioned tests failed due to:
```
RuntimeError: error executing torch_shm_manager at "/Users/distiller/ParlAI/.eggs/torch-1.2.0-py3.7-macosx-10.13-x86_64.egg/torch/bin/torch_shm_manager" at ../torch/lib/libshm/core.cpp:99
```
@stephenroller, @klshuster  I would be grateful if you guys could help out on this

Also, I have updated the Circle CI config.yaml so that it installs pytorch>=1.1 for every system build we test on, please have a look at that too!

klshuster:
if you try running those specific tests locally, can you reproduce the errors?

vedantpuri:
So @klshuster, following your comment, I tried running the tests on my mac and it has been stuck on (test_seq2seq.TestHogwildSeq2seq) for the past > 30 minutes without any output and hence I cut it.
Is there any other way to go about this? (since my machine might be too slow for this)

parlai/core/utils.py
```diff
@@ -33,6 +33,14 @@
     __TORCH_AVAILABLE = False
 
 
+def check_torch_version():
```
klshuster: i think we need to get rid of the try/except above as well, and perhaps move into this function as well?

parlai/core/torch_agent.py
```diff
@@ -46,6 +47,9 @@
 except ImportError:
```
klshuster: i think if we have this `check_torch_version` perhaps we can just move this try/except to there totally, and just have `import torch` here

parlai/tasks/coco_caption/agents.py
```diff
@@ -11,10 +11,6 @@
 from .build_2017 import build as build_2017
 from .build_2017 import buildImage as buildImage_2017
 
-try:
-    import torch  # noqa: F401
```
klshuster: so we don't need torch here anymore right?

parlai/tasks/flickr30k/agents.py
```diff
@@ -8,10 +8,6 @@
 from parlai.core.image_featurizers import ImageLoader
 from .build import build
 
-try:
```
klshuster: same q as for `coco_caption`

parlai/tasks/coco_caption/agents.py
vedantpuri: Yes, as stephen mentioned, I removed the import torch statements ending in ```# noqa: F401``` other than in one place (where it was needed)

parlai/tasks/flickr30k/agents.py
vedantpuri: Same explanation as above

parlai/core/torch_agent.py
vedantpuri: Yes, I should probably remove the try catch here and add the try catch within the ```check_torch_version```

vedantpuri: So the isse mentioned that we get rid of all try-catch around the imports since torch is now a requirement. So I can remove the try catch here and replace it with a regular import, but I am not sure if I should have the try-catch in our function, since I thought the purpose of this PR was to remove ll try-catches. Do let me know if I am wrong and you would like it otherwise!

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ installtorchgpu12: &installtorchgpu12
   run:
     name: Install torch GPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.2.0'
+      pip3 install --progress-bar off 'torch>=1.1.0'
```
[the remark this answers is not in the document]
vedantpuri: Both options install 1.2.0 regardless

parlai/core/torch_agent.py
klshuster: I think we want to keep *1* try catch so that we can fail on our own terms (we still want to fail) - i.e. fail with appropriate instructions, rather than "Module Not Found"

parlai/core/utils.py
vedantpuri: Made some changes in the new commit.

vedantpuri:
Refer to PR description for passing log of osx_tests on my local machine (MBP mid-2012 2.5 GHz Intel Core i5, 8GB RAM)

vedantpuri:
Updated Circle CI to use conda installation of pytorch in os_x tests instead of pip in order to resolve the issue we were facing earlier with hogwild in seq2seq

parlai/core/pytorch_data_teacher.py
```diff
@@ -23,22 +24,14 @@
...
 from threading import Thread, Condition, RLock
 
-
-if torch.version.__version__.startswith('0.'):
-    raise ImportError(
-        "Please upgrade to PyTorch >=1.0; "
-        "visit https://pytorch.org for instructions."
-    )
+# Perform Torch version check
+check_torch_version()
```
klshuster: we should perform check prior to importing, as you have done in other files

vedantpuri: Done so in latest commit!

stephenroller:
Wait WHAT?

.circleci/config.yml
```diff
@@ -79,19 +79,22 @@ installtorchcpuosx: &installtorchcpuosx
...
+      conda config --set always_yes yes
+      conda activate base
+      conda install --quiet 'pytorch>=1.1.0' -c pytorch
 
 installtorchcpu36: &installtorchcpu36
   run:
     name: Install torch CPU and dependencies
     command: |
-      pip3 install --progress-bar off https://download.pytorch.org/whl/cpu/torch-1.0.1.post2-cp36-cp36m-linux_x86_64.whl
+      pip3 install --progress-bar off 'torch>=1.1.0'
```
stephenroller: Oh that's nice, they used to have separate CPU/CUDA distributions but I guess they unified?

.circleci/config.yml
```diff
@@ -107,6 +110,15 @@ setupcuda: &setupcuda
       nvidia-smi
       pyenv global 3.6.5
 
+installconda: &installconda
+  run:
+    name: Install Anaconda
+    working_directory: ~/
+    command: |
+      brew install wget --with-libressl
+      wget -q https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh
```
stephenroller: We use full conda within FAIR, and I've historically witnessed some problems with some people using miniconda and others using conda.

I don't understand what is with the conda issue though. What makes the distribution different? Is conda python compiled differently? We should flag some pytorch people and ask what's going on.

Especially because seq2seq hogwild is tested on every commit.

parlai/core/distributed_utils.py
```diff
@@ -15,14 +15,11 @@
...
 
-try:
-    import torch.version
-    import torch.distributed as dist
-
-    TORCH_AVAILABLE = True
-except ImportError:
-    TORCH_AVAILABLE = False
+check_torch_version()
+import torch  # noqa: E402
```
stephenroller: you can remove the ones that are marked `noqa: E402`. They were there only for the import test.

This dist one shouldn't be E402 though.

parlai/core/utils.py
```diff
@@ -20,19 +20,27 @@
...
+        raise ImportError(
+            "Please upgrade to PyTorch >=1.1; "
+            "visit https://pytorch.org for instructions."
+        )
 
 
+check_torch_version()
+import torch  # noqa: E402
+
+TORCH_LONG = torch.long
```
stephenroller: you can drop this and change the constant to `torch.long` in padded3d.

tests/test_seq2seq.py
```diff
@@ -130,6 +130,7 @@ def test_badinput(self):
 
 
 class TestHogwildSeq2seq(unittest.TestCase):
+    # os_x tests: Hangs if using pip installation of pytorch; Succeeds if using conda installation
```
stephenroller: I'm so confused. If this is the case then people need to be using conda to install parlai on os x, and that's a bit annoying and undesirable.

Let's try to trace back the issue.

parlai/core/distributed_utils.py
vedantpuri: I had removed the previous torch statements having ***```# noqa: F401```***.
The ```noqa: E402``` was added by me in order to ignore the lint error that says module level import not at top of module. Also, we need torch in this module in areas like ```torch.cuda.ByteTensor(buffer_size)```

tests/test_seq2seq.py
vedantpuri: Yes this has been bothering me for the past few days! It is really strange that such a thing is causing an issue. Luckily @klshuster observed that it was working locally using conda and when I tried the same it was true!
Strangely enough, this test worked on an earlier version of torch installed using pip.

How would you suggest we solve this ?

.circleci/config.yml
vedantpuri: So for linux distributions they still have a separate specific command for a non-CUDA installation, but regardless if a GPU isn't found those tests are skipped and it is essentially run on the CPU. I used these commands since I felt they were easier to interpret and had no dammage of using them. If you would like me to change these back to the long urls with specific cpu only installation I can do that too!

.circleci/config.yml
vedantpuri: I guess that could be changed to full conda. I used miniconda since it doesn't install a bunch of unnecessary stuff like conda does. Using conda also worked on my system locally.
Would you want me to try and change this to full conda ?
I am not sure what makes the two distributions different and yeah it would be great to ask the pyTorch people what the issue could be.

tests/test_seq2seq.py
stephenroller: I'm going to harass some of the pytorch devs and see what they say. Can you give specific me commands which fail/succeed?

vedantpuri: Sounds like the right plan!
So using a pip installation ```test_generation_multi (__main__.TestHogwildSeq2seq)``` just hangs and doesn't complete (tried locally once and no output for > 30 minutes)

Using conda the entire test file (that test and 5 more) run in about ```36.956s``` on my machine!
The command is mentioned below: 

Running from ParlAI:
```python tests/test_seq2seq.py -v```

File: .circleci/config.yml (hunk 11 above)
Comment (vedantpuri) [reply .circleci/config.yml]: Also I was looking around other Facebook Research repos and I notice that 
[here](https://github.com/facebookresearch/habitat-api/blob/2df02c088a8cfdedc051f107eccf65078dbcbfca/.circleci/config.yml#L85) too they used miniconda and installed torch using conda.
Do let me know though if you would prefer full conda instead!

File: tests/test_seq2seq.py (hunk 14 above)
Comment (klshuster) [reply tests/test_seq2seq.py]: jumping in here to reiterate that `python tests/test_seq2seq.py` hangs with pip but completes successfully with conda (i.e. that is the command to run)

Comment (stephenroller) [reply tests/test_seq2seq.py]: I'd like to isolate a test case we could provide the pytorch devs to file a bug report. If @vedantpuri wants to do this, that would be awesome. We can also have him file the official report so that he gets that sweet pytorch contributor credit.

Comment (vedantpuri) [reply tests/test_seq2seq.py]: Sure I can try doing that. Just to be clear what do you mean specifically by "isolate a test case". 
I believe that the train loop is involved in this bug so I would probably need to create a seq2seq trainer to reproduce the bug. Any thoughts how to go about it exactly ?

Comment (stephenroller) [reply tests/test_seq2seq.py]: We need a self-contained .py script that only imports torch, etc. The script should hang or fail when run on os x pip and succeed on os x conda. The whole thing should be very small, no more than 50 lines or so.

If we're unable to create this isolated test case, then it means we have a bug in the hogwild code in ParlAI, and we're violating some contract.

Comment (klshuster) [reply tests/test_seq2seq.py]: To add on to this - we would want to find where in the seq2seq model perhaps (or the trainloop) that the code is hanging

Comment (stephenroller) [reply tests/test_seq2seq.py]: Do we know which pytorch version is hanging? Maybe `pip install` vs conda are causing different versions to be installed?

Comment (vedantpuri) [reply tests/test_seq2seq.py]: Both are torch 1.2.0

Comment (vedantpuri) [reply tests/test_seq2seq.py]: So I have been trying to get to the bottom of this for the past few hours and even though I haven't figured it out completely, I have some updates and could use some help:
- Following the trace of our test I reached ```ParlAI/parlai/core/testing_utils```.  Now within ```train_model(opt)``` everything works normal in both distributions upto: 
```python
valid, test = tl.train()
```
at which point the pip distribution hangs and conda passes through. Naturally, I dug deeper to the next layer: ```ParlAI/parlai/scripts/train_model```
***Note:*** At this point I would suggest you guys comment out ```with capture_output() as output:``` and reindent the code in order to reproduce the steps I have added below. It would also be useful to comment out all other tests other than ```TestHogwildSeq2seq```
- Now within the train function I inspected to see at which point the pip installation hangs and started adding before and after style print statements within the loop. After a while it was narrowed down to the call: ```world.parley()```. I created a counter variable outside the loop and incremented it at the bottom of the loop. Around the ```world.parley()``` statement I printed the counter, looking something like this:
```python
        ctr = 0
        with world:
            while True:
                print(ctr)
                # do one example / batch of examples
                world.parley()
                print(ctr)
                self.parleys += 1

                # get the total training examples done, compute epochs
                self._total_epochs = (
                    self._preempted_epochs
                    + num_workers() * self.world.get_total_epochs()
                )
                exs_per_epoch = self.world.num_examples()
                self._total_exs = int(np.round(self._total_epochs * exs_per_epoch))
                # and use the primary worker's timings for everything
                train_time, log_time, validate_time = sync_object(
                    (
                        self.train_time.time(),
                        self.log_time.time(),
                        self.validate_time.time(),
                    )
                )
                # check counters and timers
                if self._total_epochs >= self.max_num_epochs:
                    self.log()
                    print(
                        '[ num_epochs completed:{} time elapsed:{}s ]'.format(
                            self.max_num_epochs, train_time
                        )
                    )
                    break
                if train_time > self.max_train_time:
                    print('[ max_train_time elapsed:{}s ]'.format(train_time))
                    break
                if log_time > self.log_every_n_secs:
                    self.log()
                if (
                    validate_time > self.val_every_n_secs
                    or self._total_epochs - self.last_valid_epoch
                    >= self.val_every_n_epochs
                ):
                    stop_training = self.validate()
                    self.last_valid_epoch = self._total_epochs
                    if stop_training:
                        break
                if (
                    self.save_time.time() > self.save_every_n_secs
                    and opt.get('model_file')
                    and is_primary_worker()
                ):
                    print(
                        "[ saving model checkpoint: {}.checkpoint".format(
                            opt['model_file']
                        )
                    )
                    self.save_model('.checkpoint')
                    self.save_time.reset()
                ctr += 1
```
Surprisingly, in the **pip** installation *6* (7th iteration) is the last counter printed and it hangs after that, whereas conda runs smoothly till the end.

- Currently I have been looking at the ```parley()``` function in ```class HogwildWorld``` in ```ParlAI/parlai/core/worlds.py```.  For debugging purposes I added a few print statements to the function to make it look like:
```python
def parley(self):
        """Queue one item to be processed."""
        # schedule an example
        self.sync['queued_sem'].release()
        print("release")
        # keep main process from getting too far ahead of the threads
        # this way it can only queue up to numthreads unprocessed examples
        self.sync['threads_sem'].acquire()
        print("acquire")
        self.update_counters()
```
and observe that for every before after print of ```ctr`` , there is a ```release``` and ```acquire`` printed between them except for the ctr = 6.

A sample log of my pip installation debugging has been pasted here:
```
0
release
acquire
0
1
release
acquire
1
2
release
acquire
2
3
release
acquire
3
4
release
acquire
4
5
release
acquire
5
6
release

```
I tried debugging by adding other print statements as well but it only got more confusing due the unorderly fashion printing of multiple threads. I would want to say that it reaches a deadlock situation trying to acquire a lock that hasn't been released but the fact that the same thing works in conda stops me from saying that. I have been trying to come up with an isolated test case but have not been able to replicate a similar code flow to reproduce the bug yet. I could really use your help on this since you guys might know better than me about what's going on with the multithreading using the information I gave you. @stephenroller @klshuster

stephenroller:
Interesting. Good digging, @vendantpuri.

It's been a long time since I read the Hogwild code super carefully, and it's complicated enough that could see a deadlock appearing via platform differences (there's at least a couple sketchy sleeps in there...). And multiprocessing is always a nightmare in Python!

Next debugging steps I suggest:
1. Yank out the unit test into a script (Removing the capture output). Run the script. Do you observe the pip/conda failure? Run the test several times to see if it's consistent. If you find this isn't happening, then we seem to have some weirdness around a tests mixing. Perhaps a resource isn't being freed properly.
2. Replace the seq2seq with a `repeat_query` agent. Do you see it? What about with a `transformer/generator` agent? Repeat both of these several times to check consistency. If `repeat_query` also has the bug, then we know it's 100% parlai and not pytorch. If it happens only with seq2seq, try different `--rnn-class` (gru, lstm, rnn) and see if you observe a difference there.

Report back with what you find. This will inform our next steps around whether we think it's a deadlock, a bug in pytorch, etc.

vedantpuri:
> 1. Yank out the unit test into a script

By this I assume you mean to create a new script with only the ```test_generation_multi``` (from TestHogwildSeq2seq) function in it. I can import other stuff in that script right ? (like testing_utils) and then call those functions from imported modules (with capture output commented out etc.)

vedantpuri:
This is what I have so far:

| **Model**        | **Train**           |  **Valid**  |  **Test**  |
| ------------- |:-------------:| :-----:| :-----:|
| seq2seq gru     | *Conda*: runs <br> *Pip* : hangs at 6     |   *Conda*: runs <br> *Pip*: never reached | *Conda*: runs <br> *Pip*: never reached |
| seq2seq lstm     | *Conda*: runs <br> *Pip*: hangs at ctr=6     |   *Conda*: runs <br> *Pip*: never reached | *Conda*: runs  <br> *Pip*: never reached |
| seq2seq rnn | Both run |   Both run | Both run |
| transformer/generator | *Conda*: runs always <br> *Pip* : hangs after ctr = 2 | *Conda*: hangs sometiems <br> *Pip*: never tried | *Conda*: doesn't reach if valid had hanged <br> *Pip*: never tried |
| repeat_query | Both run | Both run| Both run|

stephenroller:
Cool thanks for compiling the information.  That's a lot of confusing inconsistencies. Let's sit on this a bit and I'll think about the next steps.

vedantpuri:
Any updates/ideas on how to go about this ?
@stephenroller @klshuster

stephenroller:
https://github.com/pytorch/pytorch/releases pytorch just hit 1.3, so we should bump the versions here and see if anything is improved.

vedantpuri:
@stephenroller, I tried using ">=1.3" for torch. OS_X fails on two tests. Trace shown below:
```
ERROR: test_labelcands_multi (test_memnn.TestMemnn)
This test uses a multi-turn task and multithreading.
  File "/Users/distiller/ParlAI/tests/test_memnn.py", line 67, in test_labelcands_multi
    rank_candidates=True,
  File "/Users/distiller/ParlAI/parlai/core/testing_utils.py", line 236, in train_model
    tl = tms.TrainLoop(popt)
  File "/Users/distiller/ParlAI/parlai/scripts/train_model.py", line 376, in __init__
    self.world = create_task(opt, self.agent)
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 1195, in create_task
    world = HogwildWorld(opt, world)
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 935, in __init__
    self.threads.append(HogwildProcess(i, opt, world.share(), self.sync))
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 141, in share
    shared_data['agents'] = self._share_agents()
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 153, in _share_agents
    shared_agents = [a.share() for a in self.agents]
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 153, in <listcomp>

    shared_agents = [a.share() for a in self.agents]
  File "/Users/distiller/ParlAI/parlai/core/torch_ranker_agent.py", line 700, in share
    shared = super().share()
  File "/Users/distiller/ParlAI/parlai/core/torch_agent.py", line 1172, in share
    self.model.share_memory()
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1111, in share_memory
    return self._apply(lambda t: t.share_memory_())
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 202, in _apply
    module._apply(fn)
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 224, in _apply
    param_applied = fn(param)
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1111, in <lambda>
    return self._apply(lambda t: t.share_memory_())
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/tensor.py", line 279, in share_memory_
    self.storage().share_memory_()
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/storage.py", line 116, in share_memory_
    self._share_filename_()
RuntimeError: error executing torch_shm_manager at "/Users/distiller/venv/lib/python3.7/site-packages/torch/bin/torch_shm_manager" at ../torch/lib/libshm/core.cpp:99

ERROR: test_generation_multi (test_seq2seq.TestHogwildSeq2seq)
This test uses a multi-turn task and multithreading.
  File "/Users/distiller/ParlAI/tests/test_seq2seq.py", line 152, in test_generation_multi
    lookuptable='all',
  File "/Users/distiller/ParlAI/parlai/core/testing_utils.py", line 236, in train_model
    tl = tms.TrainLoop(popt)
  File "/Users/distiller/ParlAI/parlai/scripts/train_model.py", line 376, in __init__
    self.world = create_task(opt, self.agent)
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 1195, in create_task
    world = HogwildWorld(opt, world)
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 935, in __init__
    self.threads.append(HogwildProcess(i, opt, world.share(), self.sync))
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 141, in share
    shared_data['agents'] = self._share_agents()
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 153, in _share_agents
    shared_agents = [a.share() for a in self.agents]
  File "/Users/distiller/ParlAI/parlai/core/worlds.py", line 153, in <listcomp>
    shared_agents = [a.share() for a in self.agents]
  File "/Users/distiller/ParlAI/parlai/core/torch_generator_agent.py", line 451, in share
    shared = super().share()
  File "/Users/distiller/ParlAI/parlai/core/torch_agent.py", line 1172, in share
    self.model.share_memory()
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1111, in share_memory
    return self._apply(lambda t: t.share_memory_())
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 202, in _apply
    module._apply(fn)
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 202, in _apply
    module._apply(fn)
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 224, in _apply
    param_applied = fn(param)
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1111, in <lambda>
    return self._apply(lambda t: t.share_memory_())
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/tensor.py", line 279, in share_memory_
    self.storage().share_memory_()
  File "/Users/distiller/venv/lib/python3.7/site-packages/torch/storage.py", line 116, in share_memory_
    self._share_filename_()
RuntimeError: error executing torch_shm_manager at "/Users/distiller/venv/lib/python3.7/site-packages/torch/bin/torch_shm_manager" at ../torch/lib/libshm/core.cpp:99
```

I remember seeing this earlier once in torch 1.2 too!

stephenroller:
Abandoning, this will be redone soon with the torch 1.4 release.

PR #2026, opened 2019-09-27
[TRA] Torch Ranker Agent Edits

klshuster:
**Patch description**
***Edit: The below description is no longer entirely valid. There is now a new class that implements the `--add-label-to-fixed-cands` functionality; Bi-encoder and Poly-encoder now subclass this agent. Additionally, `--reencode-candidate-vecs` is now on by default.***

I needed to add some functionality to the TRA when doing experiments with the Poly-encoder on the wikipedia IR task. These might be useful changes for others as well, so I'm putting up this PR

1. `--add-label-to-fixed-cands` - Evaluation with the Wikipedia IR task involves building a candidate set of 10k documents **PLUS** the original label, yielding 10k+1 candidates. This option will catch the error that a label is not in the fixed candidate set, and add the label in for that specific example. 
2. `--reencode-candidate-vecs` - Another thing I wanted to do with the Wikipedia IR task was use a fixed candidate set during evaluation for early stopping; however, the candidate encodings need to change between eval runs, as the candidate encoder weights have changed. Thus, i've added an option to reencode-candidate-vecs for every eval. 
3. `--encode-candidate-vecs-batchsize` - the candidate vectors for the Wikipedia IR task are very large - paragraphs from wikipedia articles. We run into GPU memory issues with the default encoding batchsize of 256, so i've added an option to reduce this. 

**Testing steps**
Updated TRA tests to test out `--add-label-to-fixed-cands`. Others tested via appropriate training scripts, e.g. the below:

```
python -u examples/train_model.py --init-model \
    zoo:pretrained_transformers/bi_model_huge_reddit/model --batchsize 256 -t <wiki_task> \
    --shuffle True --model transformer/biencoder --eval-batchsize 1 --warmup_updates 100 \
    --lr-scheduler-patience 0 --lr-scheduler-decay 0.4 -lr 0.005 --data-parallel True \
    --history-size 20 --label-truncate 360 --text-truncate 360 -vp 5 -veps 0.1 -vme 1000 \
    --validation-metric accuracy --validation-metric-mode max --save-after-valid True \
    --log_every_n_secs 20 --candidates batch --dict-tokenizer bpe \
    --dict-file /checkpoint/parlai/dicts/20190222_bpelower_ost+toronto+wiki/dict \
    --dict-lower True --optimizer adamax --output-scaling 0.06 --variant xlm \
    --reduction_type mean --share-encoders False --learn-positional-embeddings True \
    --n-layers 12 --n-heads 12 --ffn-size 3072 --attention-dropout 0.1 --dropout 0.1 \
    --relu-dropout 0.0 --n-positions 1024 --embedding-size 768 --activation gelu \
    --embeddings-scale False --n-segments 2 --learn-embeddings True \
    --share-word-embeddings False --dict-endtoken __start__ --fp16 True \
    --train-predict False --num-cands 10 --encode-candidate-vecs True \
    -ecands fixed -fcp <path/to/wiki/labels> --encode-candidate-vecs-batchsize 128 \
    --reencode-candidate-vecs True --model-file /tmp/wikitemp123

```


stephenroller:
These all seem fine, but man, something feels out of control here lol.

parlai/core/torch_ranker_agent.py
```diff
@@ -335,6 +357,8 @@ def is_valid(self, obs):
 
     def train_step(self, batch):
         """Train on a single batch of examples."""
+        if self.candidates != 'fixed' and self.opt.get('reencode_candidate_vecs'):
+            self.fixed_candidate_encs = None
```
emilydinan: i'm not sure if this is doing anything here. if we pre-encode the candidates, they are must be passed in to the function `score_candidates(.., cand_encs=...)`. we are not passing them in during the train step (see line 376)

parlai/core/torch_ranker_agent.py
```diff
@@ -58,6 +58,13 @@ def add_cmdline_args(cls, argparser):
             help='The source of candidates during evaluation (defaults to the same'
             'value as --candidates if no flag is given)',
         )
+        agent.add_argument(
+            '--add-label-to-fixed-cands',
```
emilydinan: do we see this as generally useful?

parlai/core/torch_ranker_agent.py
```diff
@@ -916,10 +980,11 @@ def _make_candidate_encs(self, vecs):
         """
 
         cand_encs = []
-        vec_batches = [vecs[i : i + 256] for i in range(0, len(vecs), 256)]
+        bsz = self.opt.get('encode_candidate_vecs_batchsize', 256)
```
emilydinan: nice

parlai/core/torch_ranker_agent.py
```diff
@@ -883,7 +945,9 @@ def _make_candidate_vecs(self, cands):
         cand_vecs = []
         for batch in tqdm(cand_batches):
             cand_vecs.extend(self.vectorize_fixed_candidates(batch))
-        return padded_3d([cand_vecs], dtype=cand_vecs[0].dtype).squeeze(0)
+        return padded_3d(
```
emilydinan: 😮 good catch

parlai/core/torch_ranker_agent.py
```diff
@@ -446,6 +488,10 @@ def eval_step(self, batch):
             cand_preds = self.block_repeats(cand_preds)
 
         preds = [cand_preds[i][0] for i in range(batchsize)]
+
+        # clean up candidates
+        if self.eval_candidates == 'fixed':
+            self.fixed_candidates = self.fixed_candidates[: self.num_fixed_candidates]
```
emilydinan: question, do we really need to track the number of fixed candidates here? aren't you always adding exactly one candidate here?

parlai/core/torch_ranker_agent.py
```diff
@@ -663,7 +709,19 @@ def _build_candidates(self, batch, source, mode):
...
-                        bad_batch = True
+                        if (
+                            self.opt.get('add_label_to_fixed_cands')
+                            and not self.is_training
+                        ):
+                            cand_vecs = torch.cat(
+                                (cand_vecs, label_vec_pad.unsqueeze(0))
+                            )
+                            cands.append(batch.labels[batch_idx])
+                            label_inds[batch_idx] = self._find_match(
```
emilydinan: i don't think we need to `_find_match` here, this seems like wasted computation as we know exactly which index the label should be at

parlai/core/torch_ranker_agent.py
klshuster: I am not sure; this was specifically useful for the Wikipeida IR task from the Starspace paper (which we used for the poly-encoder/bi-encoder experiments). I've made the option `hidden` to keep from the user, and only allow in certain circumstances.

parlai/core/torch_ranker_agent.py
klshuster: we get `cand_encs` from `self.fixed_candidate_encs` during eval_step; i'm just deleting them here so that we know that we need to re-encode them during eval

parlai/core/torch_ranker_agent.py
klshuster: it's possible that with a larger eval-batchsize, one could be adding more than one candidate; this just ensures that we don't assume it's always exactly 1

parlai/core/torch_ranker_agent.py
klshuster: credit for this goes to @dexterju who had a bug with this (and has probably fixed in his local checkout)

parlai/core/torch_ranker_agent.py
klshuster: yeah good call, i'll change that

klshuster:
@stephenroller we can talk offline if you want more clarification

parlai/core/torch_ranker_agent.py
emilydinan: it seems a little weird to put this here, although i can't think of a better place. maybe just add a comment explaining?

@stephenroller  do you have any other ideas?

klshuster:
Notes:

1. move the first arg to a separate agent `Wikipedia...TA`
2. Second arg should be on...by default
3. land bsz arg

klshuster:
I've made the following changes (will update PR description as well):

1. There is now a new agent subclassing TRA, which handles adding a label to fixed cands. Bi-encoder and Poly-encoder now both subclass it.
2. The `reencode_candidate_vecs` option is turned on by default

parlai/core/torch_ranker_agent.py
stephenroller: Yes, pleaseeee add a comment. You could alternatively define a silly helper function:

```
def _invalidate_candidate_vec_cache(self):
    self.fixed_candidate_encs = None
```

so that the call is descriptive.

tests/test_tra.py
```diff
@@ -271,6 +223,54 @@ def _get_args(self):
     def _get_threshold(self):
         return 0.6
 
+    @testing_utils.retry(ntries=3)
```
stephenroller: 😍

klshuster:
moved the invalidation to a function as stephen suggested:

```python
def _maybe_invalidate_fixed_encs_cache(self):
        if self.candidates != 'fixed':
            self.fixed_candidate_encs = None
```

parlai/core/torch_ranker_agent.py
stephenroller: Sounds like a comment could help

klshuster: i moved this functionality into the new class btw

parlai/agents/transformer/biencoder.py
```diff
@@ -8,7 +8,76 @@
 import torch
 
 
-class BiencoderAgent(TransformerRankerAgent):
+class AddLabelFixedCandsTRA(TorchRankerAgent):
```
stephenroller: Typically call it a Trait, like `AddLabelFixedCandsTrait`, but in this case it looks like you could just have it be an abstract class and lose the multiple inheritance.

stephenroller: I was imagining though that you'd have `Polyencoder(TorchRankerAgent)` and `IRFriendlyPolyencoder(AddLabelFixedCandsTRA, TorchRankerAgent)`

klshuster:
@stephenroller I agree that the IRFriendly stuff should be put in a separate subclass; i've just moved the inheritance of `AddLabelFixedCandsTRA` to `IRFriendly<Bi/Poly>encoderAgent`

parlai/agents/transformer/biencoder.py
```diff
@@ -107,3 +107,9 @@ def _set_text_vec(self, *args, **kwargs):
             )
             obs['added_start_end_tokens'] = True
         return obs
+
+
+class IRFriendlyBiencoderAgent(AddLabelFixedCandsTRA):
```
stephenroller: Then wouldn't this need `IRFriendlyBiencoderAgent(AddLabelFixedCandsTRA, BiEncoder)`?

klshuster: 🤦‍♂

klshuster:
Re-requesting review so that we can get this encode cand vecs batchsize fix in

parlai/agents/transformer/biencoder.py
```diff
@@ -38,3 +107,13 @@ def _set_text_vec(self, *args, **kwargs):
...
         return obs
+
+
+class IRFriendlyBiencoderAgent(AddLabelFixedCandsTRA, BiencoderAgent):
+    """Bi-encoder agent that allows for adding label to fixed cands."""
+
+    @classmethod
+    def add_cmdline_args(cls, argparser):
+        """Add cmd line args."""
+        super(AddLabelFixedCandsTRA, cls).add_cmdline_args(argparser)
```
stephenroller: Want this flagged in #2115.

tests/test_tra.py
```diff
@@ -223,6 +223,57 @@ def _get_args(self):
...
+                f.write(train_val_cands_str)
+            args['fixed_candidates_path'] = tmp_cands_file
+            args['encode_candidate_vecs'] = False  # don't encode before training
+            args['ignore_bad_candidates'] = False
+            args['model_file'] = os.path.join(tmpdir, 'model')
+            args['dict_file'] = os.path.join(tmpdir, 'model.dict')
+            args['num_epochs'] = 20
+            # Train model where it has access to the candidate in labels
+            stdout, valid, test = testing_utils.train_model(args)
+            self.assertGreaterEqual(
```
stephenroller: Is our goal to test that it performs at a certain level with the fixed candidates, or that the fixed candidate can be loaded/encoded correctly, or what? Because if you dropped this greaterEqual, then you could drop the retry, and I think this test would still be adequate for what we want, no?

stephenroller:
merge in master and rerun tests ofc

emilydinan:
cc @dexterju

tests/test_tra.py
klshuster: Yeah we just want to check that the fixed cands can be loaded/encoded correctly. I will keep greaterEqual test but i will drop it to 0.0 and get rid of retry

stephenroller: You can just drop the adsert then. It will still run

Issue #1886, opened 2019-07-30
Switch from print to logging


klshuster:
Currently, ParlAI communicates with a user via print statements, for all sorts of situations: warnings ([example](https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/agents.py#L459)), tracking training progress ([example](https://github.com/facebookresearch/ParlAI/blob/master/parlai/scripts/train_model.py#L682)), general model info ([example](https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/language_model/language_model.py#L206)), etc. This is not ideal, as it places the burden on a user to capture the output in order to log any sort of progress.

To remedy this, we'd like to incorporate logging throughout ParlAI using Python's logging module ([here](https://docs.python.org/3/library/logging.html)) as it provides numerous abstractions/handling for various types of things we would like to log (warnings, info, etc). The logging could include a prefix, e.g. including some combination of `<timestamp>|<module>|<filename>|<line>|<log level>`.

An additional ask for this is a unit test that ensures future PRs do not use print statements.

vedantpuri:
Haven't done much work on this yet but I was thinking of having a script in parlai.core that initializes a logger variable and importing this script to any file that prints something and then using the logger variable there. I was thinking of having the logger print to stdout as well as saving the log of that day to a particular file.
The  things I am confused about as of now are:
- How does this work in areas where ```capture_output()``` has been called. I haven't tried this yet but hoping to have the output written to the log file atleast in this case if not stdout (I suspect that is what would happen since it blocks stdout). Any thoughts ?
- Also not sure about how to handle the usage of ```override_print()``` with logging. (```distributed_utils.py```)
- For the testing phase, I was simply hoping to do a regex match on the string in a program like this:
```python
with open(fname) as myFile:
    lines = myFile.readlines()
    for line in lines:
        tokens = line.split()
        for tok in tokens:
            if regex.match(tok):
                break
```
This seems to work but would you know if there would be a better way ?

klshuster:
I think including a logger definition in `parlai.core.utils` would be helpful for this, and you can import/instantiate that logger in appropriate places. I like the idea of printing to stdout and additionally logging to a file (perhaps related to a model file if specified? default log file could be something we need to think about)

To answer your questions:

- `capture_output()` is primarily used in testing, so that our tests do not print to stdout. this should work as is, without you having to change anything (it just captures anything printed to stdout so that our script outputs do not clutter test outputs)
[...]

PR #2030, opened 2019-09-29
Implement Logging

vedantpuri:
**Patch description**
Implements a logging mechanism. Partially Solves #1886. As of now only a few print statements have been replaced to provide as a sample before changing all print statements. 
To completely solve the issue (replace all print statements and re-work existing loggers) some follow-up tasks have been added as a new issue in #2044 which can have a separate PR

**Testing steps**
Everything should work as before.

**Other information**
Some important implementation info and questions:
- The inspiration for the logging mechanism being implemented was taken from facebookresearch/habitat-api. Their logging module can be found [here](https://github.com/facebookresearch/habitat-api/blob/263dc63db2628a1876c01909c5cf0bab7ee7882b/habitat/core/logging.py)
- It was suggested that capture_output be re-written and the logger be disabled during that. That would cause problems since in many cases we want to return ```output.getvalue()``` from capture_output and use it in our tests and if logging were disabled the result of that would always be empty. I dealt with that by adding a handler in the logger for capture_output which mutes stdout, still writes to the file (I felt this could be useful), and preserves output in the TeeStringIO stream. The captured output can then be accessed via ```output.getvalue()``` as it was before. Currently, I haven't replaced all print statements and hence capture_output still uses ```contextlib.redirect_stdout(sio)```. Later once all print statements have been removed, the line containing ```contextlib.redirect_stdout(sio)``` and the following line may be replaced with just ```yield sio``` [tried this on a small example and it works]. I personally feel this is a better and necessary alternative to disabling the logger.
- override_print has been modified to disable the logger (this is ok since it doesn't require the output later on after the context) and add prefix to future log statements (both file and stdout) if a prefix is specified [also tried on a small example and worked]
- It was discussed in the issue that global methods be defined such as log, info etc. I argue that this might spoil things since this way eventually the call would be made from ```logging_utils``` and hence we wouldn't be able to track the file where actually the log was made from. Also the log function requires a level to be specified and as of now I have used INFO as the default.
- Finally, as a sample I have switched all the print statements in "ParlAI/examples/" which you can review and let me know if they are okay. [Marked them as INFO level for now]

### DIFF
--- a/examples/base_train.py
+++ b/examples/base_train.py
@@ -23,7 +23,6 @@ from parlai.core.logging_utils import logger
 import time
 
 
-
 def main():
     # Get command line arguments
     parser = ParlaiParser()
--- a/examples/base_train.py
+++ b/examples/base_train.py
@@ -43,21 +43,21 @@ def main():
     start = time.time()
     # train / valid loop
     for _ in range(1):
-        logger.log(logger.INFO, '[ training ]')
+        logger.info('[ training ]')
         for _ in range(opt['num_iters']):  # train for a bit
             world_train.parley()
 
-        logger.log(logger.INFO, '[ training summary. ]')
-        logger.log(logger.INFO, world_train.report())
+        logger.info('[ training summary. ]')
+        logger.info(world_train.report())
 
-        logger.log(logger.INFO, '[ validating ]')
+        logger.info('[ validating ]')
         for _ in range(1):  # check valid accuracy
             world_valid.parley()
 
-        logger.log(logger.INFO, '[ validation summary. ]')
-        logger.log(logger.INFO, world_valid.report())
+        logger.info('[ validation summary. ]')
+        logger.info(world_valid.report())
 
-    logger.log(logger.INFO, 'finished in {} s'.format(round(time.time() - start, 2)))
+    logger.info('finished in {} s'.format(round(time.time() - start, 2)))
 
 
 if __name__ == '__main__':
--- a/examples/base_train.py
+++ b/examples/base_train.py
@@ -19,7 +19,9 @@ processing the data.
 from parlai.core.params import ParlaiParser
 from parlai.core.agents import Agent
 from parlai.core.worlds import create_task
+from parlai.core.logging_utils import logger
 import time
+
 
 
 def main():
@@ -42,21 +44,21 @@ def main():
     start = time.time()
     # train / valid loop
     for _ in range(1):
-        print('[ training ]')
+        logger.log(logger.INFO, '[ training ]')
         for _ in range(opt['num_iters']):  # train for a bit
             world_train.parley()
 
-        print('[ training summary. ]')
-        print(world_train.report())
+        logger.log(logger.INFO, '[ training summary. ]')
+        logger.log(logger.INFO, world_train.report())
 
-        print('[ validating ]')
+        logger.log(logger.INFO, '[ validating ]')
         for _ in range(1):  # check valid accuracy
             world_valid.parley()
 
-        print('[ validation summary. ]')
-        print(world_valid.report())
+        logger.log(logger.INFO, '[ validation summary. ]')
+        logger.log(logger.INFO, world_valid.report())
 
-    print('finished in {} s'.format(round(time.time() - start, 2)))
+    logger.log(logger.INFO, 'finished in {} s'.format(round(time.time() - start, 2)))
 
 
 if __name__ == '__main__':
--- a/parlai/agents/tfidf_retriever/build_db.py
+++ b/parlai/agents/tfidf_retriever/build_db.py
@@ -10,14 +10,19 @@ Adapted from Adam Fisch's work at github.com/facebookresearch/DrQA/
 
 import sqlite3
 import os
-import logging
+import logging  # Used only for constants
 
 from tqdm import tqdm
 
 from collections import deque
 import random
 from parlai.core.agents import create_task_agent_from_taskname
-from parlai.core.logging_utils import logger
+from parlai.core.logging_utils import ParlaiLogger
+
+fmt = '%(asctime)s: [ %(message)s ]', '%m/%d/%Y %I:%M:%S %p'
+logger = ParlaiLogger(
+    name=__name__, console_level=logging.INFO, console_format=fmt, file_format=fmt
+)
 
 # ------------------------------------------------------------------------------
 # Store corpus.
--- a/parlai/agents/tfidf_retriever/build_db.py
+++ b/parlai/agents/tfidf_retriever/build_db.py
@@ -17,13 +17,7 @@ from tqdm import tqdm
 from collections import deque
 import random
 from parlai.core.agents import create_task_agent_from_taskname
-
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
-fmt = logging.Formatter('%(asctime)s: [ %(message)s ]', '%m/%d/%Y %I:%M:%S %p')
-console = logging.StreamHandler()
-console.setFormatter(fmt)
-logger.addHandler(console)
+from parlai.core.logging_utils import logger
 
 # ------------------------------------------------------------------------------
 # Store corpus.
--- a/parlai/agents/tfidf_retriever/build_db.py
+++ b/parlai/agents/tfidf_retriever/build_db.py
@@ -10,18 +10,17 @@ Adapted from Adam Fisch's work at github.com/facebookresearch/DrQA/
 
 import sqlite3
 import os
-import logging  # Used only for constants
 
 from tqdm import tqdm
 
 from collections import deque
 import random
 from parlai.core.agents import create_task_agent_from_taskname
-from parlai.core.logging_utils import ParlaiLogger
+from parlai.core.logging_utils import ParlaiLogger, INFO_LEVEL
 
 fmt = '%(asctime)s: [ %(message)s ]', '%m/%d/%Y %I:%M:%S %p'
 logger = ParlaiLogger(
-    name=__name__, console_level=logging.INFO, console_format=fmt, file_format=fmt
+    name=__name__, console_level=INFO_LEVEL, console_format=fmt, file_format=fmt
 )
 
 # ------------------------------------------------------------------------------
--- a/parlai/agents/tfidf_retriever/build_db.py
+++ b/parlai/agents/tfidf_retriever/build_db.py
@@ -16,11 +16,11 @@ from tqdm import tqdm
 from collections import deque
 import random
 from parlai.core.agents import create_task_agent_from_taskname
-from parlai.core.logging_utils import ParlaiLogger, INFO_LEVEL
+from parlai.core.logging_utils import ParlaiLogger, INFO
 
-fmt = '%(asctime)s: [ %(message)s ]', '%m/%d/%Y %I:%M:%S %p'
+fmt = '%(asctime)s: [ %(message)s ]'
 logger = ParlaiLogger(
-    name=__name__, console_level=INFO_LEVEL, console_format=fmt, file_format=fmt
+    name=__name__, console_level=INFO, console_format=fmt, file_format=fmt
 )
 
 # ------------------------------------------------------------------------------
--- a/parlai/core/logging_utils.py
+++ b/parlai/core/logging_utils.py
@@ -16,25 +16,35 @@ WARN_LEVEL = logging.WARNING
 ERROR = logging.ERROR
 CRITICAL = logging.CRITICAL
 
+DEFAULT_FILE_FORMAT = '%(asctime)s : %(levelname)s : %(message)s'
 
+# Some functions in this class assume that ':' will be the separator used in
+# the logging formats setup for this class
 class ParlaiLogger(logging.Logger):
     def __init__(
         self,
         name,
-        console_level,
+        console_level=DEBUG,
         console_format=None,
         file_format=None,
         file_level=INFO,
         filename=None,
     ):
         """
-        Initialize the logger object
-        :param name: Name of the logger
-        :param console_level: min. Level of messages logged to console
-        :param console_format: The format of messages logged to the console. Simple stdout is used if None specified
-        :param file_format: The format of messages logged to the file (A default is used if None specified)
-        :param file_level: min. Level of messages logged to the file
-        :param filename: The file the logs are written to
+        Initialize the logger object.
+        :param name:
+            Name of the logger
+        :param console_level:
+            min. Level of messages logged to console
+        :param console_format:
+            The format of messages logged to the console.
+            Simple stdout is used if None specified
+        :param file_format:
+            The format of messages logged to the file
+        :param file_level:
+            min. Level of messages logged to the file
+        :param filename:
+            The file the logs are written to
         """
         super().__init__(name, console_level)  # can be initialized with any level
 
@@ -42,83 +52,92 @@ class ParlaiLogger(logging.Logger):
         if filename:
             # Default format used if no file format provided
             if file_format is None:
-                file_format = '%(asctime)s :: %(levelname)s :: %(message)s'
+                file_format = DEFAULT_FILE_FORMAT
             self.fileHandler = logging.FileHandler(filename)
-            self.fileHandler.level = (
-                file_level
-            )  # Log to file levels: file_level and above
+            # Log to file levels: file_level and above
+            self.fileHandler.level = file_level
             self.fileFormatter = logging.Formatter(file_format)
             self.fileHandler.setFormatter(self.fileFormatter)
             super().addHandler(self.fileHandler)
 
         # Logging to stdout
         self.streamHandler = logging.StreamHandler(sys.stdout)
-        self.streamHandler.level = (
-            console_level
-        )  # Log to stdout levels: console_level and above
-        if console_format is not None:
-            self.consoleFormatter = logging.Formatter(console_format)
-            self.streamHandler.setFormatter(self.consoleFormatter)
+        # Log to stdout levels: console_level and above
+        self.streamHandler.level = console_level
+        self.consoleFormatter = logging.Formatter(console_format)
+        self.streamHandler.setFormatter(self.consoleFormatter)
         super().addHandler(self.streamHandler)
 
         # To be used with testing_utils.capture_output()
         self.altStream = None
 
     def log(self, msg, level=INFO):
-        """ Default Logging function """
+        """Default Logging function."""
         super().log(level, msg)
 
     def add_file_handler(self, filename, level=INFO, format=None):
         """
-        Add a file handler to the logger object
+        Add a file handler to the logger object.
         Use case: When logging using the logger object instead of instantiating a new ParlaiLogger
-                  this function might  be useful to add a filehandler on the go
-        Only does so if there is no file handler existing
+                  this function might  be useful to add a filehandler on the go.
+        Only does so if there is no file handler existing.
         """
         if not hasattr(self, 'fileHandler'):
             if format is None:
-                file_format = '%(asctime)s :: %(levelname)s :: %(message)s'
+                file_format = DEFAULT_FILE_FORMAT
             self.fileHandler = logging.FileHandler(filename)
             self.fileHandler.level = level  # Log to file levels: level and above
             self.fileFormatter = logging.Formatter(file_format)
             self.fileHandler.setFormatter(self.fileFormatter)
             super().addHandler(self.fileHandler)
         else:
-            self.info("ParlaiLogger: A filehandler already exists")
+            raise Exception("ParlaiLogger: A filehandler already exists")
 
     def add_format_prefix(self, prefix):
-        """Include `prefix` in all future logging statements"""
+        """Include `prefix` in all future logging statements."""
         # change both handler formatters to add a prefix
         new_str = prefix + " " + '%(message)s'
-        self.streamHandler.setFormatter(logging.Formatter(new_str))
+
+        prevConsoleFormat = self.consoleFormatter._fmt.split(':')[:-1]
+        # Check if there was a format before this
+        if prevConsoleFormat:
+            # If so append prefix neatly after last divider
+            prevConsoleFormat += [' ' + new_str]
+            updatedConsoleFormat = ':'.join(prevConsoleFormat)
+        else:
+            updatedConsoleFormat = new_str
+        self.streamHandler.setFormatter(logging.Formatter(updatedConsoleFormat))
+
         if hasattr(self, 'fileHandler'):
-            prevFileFormat = self.fileFormatter._fmt.split('::')[:-1]
+            prevFileFormat = self.fileFormatter._fmt.split(':')[:-1]
+            # A space before the previous divider because a format always exists
             prevFileFormat += [' ' + new_str]
-            updatedFileFormat = '::'.join(prevFileFormat)
+            updatedFileFormat = ':'.join(prevFileFormat)
             self.fileHandler.setFormatter(logging.Formatter(updatedFileFormat))
 
     def reset_formatters(self):
-        """Resort back to initial formatting"""
+        """Resort back to initial formatting."""
         if hasattr(self, 'fileHandler'):
             self.fileHandler.setFormatter(self.fileFormatter)
-        self.streamHandler.setFormatter(logging.Formatter('%(message)s'))
+        self.streamHandler.setFormatter(self.consoleFormatter)
 
     def mute(self):
-        """Stop logging to stdout"""
+        """Stop logging to stdout."""
         self.streamHandler.level = float('inf')
 
     def unmute(self):
-        """Resume logging to stdout"""
-        self.streamHandler.level = logging.DEBUG
+        """Resume logging to stdout."""
+        print(self.console_level)
+        self.streamHandler.level = DEBUG
 
     def redirect_out(self, stream):
-        """Redirect all logging output to `stream`"""
+        """Redirect all logging output to `stream`."""
         self.altStream = stream
         self.altStreamHandler = logging.StreamHandler(self.altStream)
         super().addHandler(self.altStreamHandler)
 
     def stop_redirect_out(self):
-        """Stop redirecting output to alternate stream"""
+        """Stop redirecting output to alternate stream."""
         if self.altStream is None:
             raise Exception('No existing redirection.')
         else:
@@ -129,4 +148,4 @@ class ParlaiLogger(logging.Logger):
 # -----------------------------------
 # Forming the logger                #
 # -----------------------------------
-logger = ParlaiLogger(name=__name__, console_level=DEBUG)
+logger = ParlaiLogger(name=__name__)
--- a/parlai/core/logging_utils.py
+++ b/parlai/core/logging_utils.py
@@ -10,6 +10,12 @@ import logging
 from pathlib import Path
 from datetime import date
 
+INFO_LEVEL = logging.INFO
+DEBUG_LEVEL = logging.DEBUG
+WARN_LEVEL = logging.WARNING
+ERROR_LEVEL = logging.ERROR
+CRITICAL_LEVEL = logging.CRITICAL
+
 
 class ParlaiLogger(logging.Logger):
     def __init__(
@@ -25,7 +31,7 @@ class ParlaiLogger(logging.Logger):
         Initialize the logger object
         :param name: Name of the logger
         :param console_level: min. Level of messages logged to console
-        :param console_format: The format of messages logget to the console. Simple stdout is used if None specified
+        :param console_format: The format of messages logged to the console. Simple stdout is used if None specified
         :param file_format: The format of messages logged to the file (A default is used if None specified)
         :param file_level: min. Level of messages logged to the file
         :param filename: The file the logs are written to
@@ -45,7 +51,7 @@ class ParlaiLogger(logging.Logger):
             filename = os.path.join(proj_root, log_dir_path, log_fname)
         # Default format used if no file format provided
         if file_format is None:
-            file_format = '%(asctime)s :: %(pathname)s :: %(funcName)s :: %(lineno)d :: %(levelname)s :: %(message)s'
+            file_format = '%(asctime)s :: %(levelname)s :: %(message)s'
 
         super().__init__(name, console_level)  # can be initialized with any level
 
--- a/parlai/core/logging_utils.py
+++ b/parlai/core/logging_utils.py
@@ -63,9 +63,6 @@ class ParlaiLogger(logging.Logger):
         self.altStreamHandler = logging.StreamHandler(self.altStream)
         super().addHandler(self.altStreamHandler)
 
-    def get_supressed_output(self):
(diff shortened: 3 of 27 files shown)


parlai/core/logging_utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def unmute_stdout(self):
+        self.streamHandler.level = logging.DEBUG
+
+    def redirect_out(self, stream):
+        self.altStream = stream
+        self.altStreamHandler = logging.StreamHandler(self.altStream)
+        super().addHandler(self.altStreamHandler)
+
+    def stop_redirect_out(self):
```
klshuster: could you please add a check here for `self.altStreamHandler` being `None` - i.e. if this is called prior to the function above, we might run into an error.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,89 @@
...
+        self.fileHandler.setFormatter(self.fileFormatter)
+
+        # Logging to stdout
+        self.streamHandler = logging.StreamHandler(sys.stdout)
+        self.streamHandler.level = (
+            logging.DEBUG
+        )  # Log to stdout levels: debug and above
+        super().addHandler(self.streamHandler)
+
+    def change_formatters(self, prefix):
```
klshuster: could you please add a little bit more documentation to each of these functions (i.e. docstrings)

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,89 @@
...
+        self.altStreamHandler.flush()
+        super().removeHandler(self.altStreamHandler)
+
+
+# -----------------------------------
+# Forming the logger "The specifics" #
+# -----------------------------------
+
+# Background work
+today = date.today().strftime("%d/%m/%Y").split("/")
```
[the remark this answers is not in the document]
vedantpuri: The background work portion you mean ?
Can things in ```__init__``` then be accessed as normal ?

parlai/core/testing_utils.py
```diff
@@ -289,17 +291,20 @@ def display_data(opt):
...
         dd.display_data(popt)
-    with capture_output() as test_output:
+    valid_output = output.getvalue()
+    with capture_output() as output:
         popt['datatype'] = 'test:stream'
         dd.display_data(popt)
+    test_output = output.getvalue()
 
-    return (train_output.getvalue(), valid_output.getvalue(), test_output.getvalue())
+    return (train_output, valid_output, test_output)
```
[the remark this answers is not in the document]
vedantpuri: So in between my commits I was using something called logger.get_suppressed_value() (and hence accessing the get value immediately) and later realised that doing that was not necessary. While changing back I only changed logger.get_suppressed_value() back to output.getvalue() and I guess this was one of the cases where getvalue was called at the bottom and not when initializing train_output etc.
If you would like I could change it back but won't affect anything either way.

parlai/core/logging_utils.py
vedantpuri: Sure should it be in a try catch ?

examples/base_train.py
```diff
@@ -42,21 +43,21 @@ def main():
     start = time.time()
     # train / valid loop
     for _ in range(1):
-        print('[ training ]')
+        logger.log(logger.INFO, '[ training ]')
```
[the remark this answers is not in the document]
vedantpuri: Yes I was also going to do that, I thought this way might be easier to change the levels later. i.e. If we make all print statements logger.log(logger.INFO...  then later on we just need to update the level.
The other alternative is to use logger.info everywhere now and then while changing the level change accordingly after the ```.```

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,89 @@
...
+proj_root = str(Path(__file__).parent.parent.parent)
+log_dir_path = proj_root + "/" + "Logs/" + year + "/" + month
+log_fname = year + "-" + month + "-" + day + ".log"
+if not os.path.exists(log_dir_path):
+    os.makedirs(log_dir_path)
+log_loc = os.path.join(proj_root, log_dir_path, log_fname)
+
+# ParlaiLogger obj creation
+parlai_format = '%(asctime)s :: %(pathname)s :: %(funcName)s :: %(lineno)d :: %(levelname)s :: %(message)s'
+logger = ParlaiLogger(
```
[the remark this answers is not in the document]
vedantpuri: I was also thinking about that but I saw that habitat-api followed this mechanism and I just went with it. 
Also another reasoning that came to mind was that almost every file has some sort of print statements and imports other modules as well, I was worried that there might be some conflict if we instantiate say two logger objects in two files, won't that lead to problems ?
Or what you are trying to say that every class has its own logger object (which will be a lot of logger objects and then confusion on which file to write etc.)

I was just a bit confused about instantiating it in multiple places, so if you could elaborate or provide an example or two how that might be used that would be great.

examples/base_train.py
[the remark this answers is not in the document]
klshuster: i think it's better to do `logger.info()` for now and then we can go in and change the functions afterward

parlai/core/logging_utils.py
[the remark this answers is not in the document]
klshuster: not sure what you mean by `accessed as normal` but yeah, the background work can be moved into `__init__`

parlai/core/logging_utils.py
klshuster: yeah

parlai/core/logging_utils.py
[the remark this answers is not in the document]
klshuster: ahh ok, if this is what they are doing then that works too. but yeah i would like to at least have the background stuff moved into the `__init__` function

parlai/core/testing_utils.py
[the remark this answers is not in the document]
klshuster: yeah, let's not change things unnecessarily

parlai/core/logging_utils.py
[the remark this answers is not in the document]
vedantpuri: For some reason I was thinking about the __init__ file instead of the constructor. My bad. I can move that work to the constructor.

vedantpuri:
Some places in the project were already using logging. I tried to modify them so that they use ParlaiLogger instead and I think I incorporated all the changes requested

parlai/agents/tfidf_retriever/build_db.py
```diff
@@ -17,13 +17,7 @@
...
 import random
 from parlai.core.agents import create_task_agent_from_taskname
-
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
-fmt = logging.Formatter('%(asctime)s: [ %(message)s ]', '%m/%d/%Y %I:%M:%S %p')
-console = logging.StreamHandler()
-console.setFormatter(fmt)
-logger.addHandler(console)
+from parlai.core.logging_utils import logger
```
klshuster: thanks for removing/replacing these!

parlai/core/logging_utils.py
```diff
@@ -12,32 +12,74 @@
...
+        console_format=None,
+        file_format=None,
+        file_level=logging.INFO,
+        filename=None,
+    ):
+        """
+        Initialize the logger object
+        :param name: Name of the logger
+        :param console_level: min. Level of messages logged to console
+        :param console_format: The format of messages logget to the console. Simple stdout is used if None specified
```
klshuster: nit: logged

parlai/core/logging_utils.py
```diff
@@ -12,32 +12,74 @@
...
+            day, month, year = today[0], today[1], today[2]
+            proj_root = str(Path(__file__).parent.parent.parent)
+            log_dir_path = proj_root + "/" + "Logs/" + year + "/" + month
+            log_fname = year + "-" + month + "-" + day + ".log"
+            if not os.path.exists(log_dir_path):
+                os.makedirs(log_dir_path)
+            filename = os.path.join(proj_root, log_dir_path, log_fname)
+        # Default format used if no file format provided
+        if file_format is None:
+            file_format = '%(asctime)s :: %(pathname)s :: %(funcName)s :: %(lineno)d :: %(levelname)s :: %(message)s'
```
klshuster: this `file_format` might be a little bit verbose, could we possibly have the default include less?

or, perhaps we can have an optional `verbose` flag for each log function that makes the prefix like this.

parlai/agents/tfidf_retriever/build_db.py
```diff
@@ -10,20 +10,19 @@
 
 import sqlite3
 import os
-import logging
+import logging  # Used only for constants
```
klshuster: perhaps we can make these class constants within `ParlaiLogger`, or perhaps just redefine them in `parlai.core.logging_utils` as module constants, so you could do something like this:

`from parlai.core.logging_utils import ParlaiLogger, INFO_LEVEL, ...`

klshuster:
Also, it would be great if you could include some screenshots of these scripts running to show that the logging is working 😄

parlai/core/logging_utils.py
vedantpuri: The flag idea might not work since to implement that we need to write functions such as log,info,debug etc in the ParlaiLogger class and when we do that it is saying that we call those functions from within this file, thus making tracking the pathname, funcname and linno of no use since it will always be the same (wherever it is called in logging.utils). But I can reduce the default format to just ```time :: level ::  message```

vedantpuri:
Most changes to existing loggers were in functions in classes and it is hard to run them since they require a whole lot of instantiating etc. The ones I could run (to some extent) were:
projects/controllable_dialogue/mturk/run.py
Screen Shot 2019-10-01 at 6 20 32 PM
After this point ^ it asked me for an mturk key which I din't have so had to abort at this point.

Note: fb and ab are just random arguments since I didn't have a database file at hand to actually run this, so I ran it till it realized the input was garbage and that spit out one log statement:
Screen Shot 2019-10-01 at 6 31 56 PM

I was hoping to run ```projects/wizard_of_wikipedia/mturk_evaluation_task/run.py``` but its taking forever to run on my machine!

Now while doing all this came up with a few other loggers that are already in place:
Most of them exist in 
```parlai/mturk``` and ```parlai/messenger``` and use a logger that is instantiated in ```parlai/mturk/core/shared_utils.py```. Would probably want to deal with those next.

Also this task turned out to be larger than I anticipated with many nitty-gritties to be taken care of, would it be possible to assign another teammate  in our group that could help with this, I feel it would speed up things in the process and we can contribute to the master branch soon.

**Comment by vedantpuri** [comment]
In this last commit I saw that the log folders were being created in areas where my project didn't exist and hence substituted the $HOME logic with what was in wizard_of_wikipedia. Also used wizard_of_wikipedia precaution check in controllable since I thought that should be done there.

I just went on a tangent on this commit but I think it's useful. Let me know if otherwise and I can rollback to the commit before this! @klshuster

File: parlai/core/logging_utils.py [hunk 11]
```diff
@@ -0,0 +1,129 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import os
+import sys
+import logging
+from pathlib import Path
+from datetime import date
+
+INFO_LEVEL = logging.INFO
```
stephenroller: I suggest we drop the `_LEVEL` and keep it consistent with the logging constants. Then you can also change to `from logging import INFO, DEBUG, ...`

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class ParlaiLogger(logging.Logger):
+    def __init__(
+        self,
+        name,
+        console_level,
+        console_format=None,
+        file_format=None,
+        file_level=logging.INFO,
```
stephenroller: use the constant you've defined.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        updatedFileFormat = '::'.join(prevFileFormat)
+        self.fileHandler.setFormatter(logging.Formatter(updatedFileFormat))
+        self.streamHandler.setFormatter(logging.Formatter(new_str))
+
+    def reset_formatters(self):
+        """Resort back to initial formatting"""
+        self.fileHandler.setFormatter(self.fileFormatter)
+        self.streamHandler.setFormatter(logging.Formatter('%(message)s'))
+
+    def mute_stdout(self):
```
stephenroller: would `mute` and `unmute` make more sense?

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        :param console_format: The format of messages logged to the console. Simple stdout is used if None specified
+        :param file_format: The format of messages logged to the file (A default is used if None specified)
+        :param file_level: min. Level of messages logged to the file
+        :param filename: The file the logs are written to
+
+        A default file is initialized if no filename was provided. It can be
+        found in ParlAI/Logs/*year*/*month*/*YYYY-MM-DD*.log
+        """
+        # Setup logging location if none specified
+        if filename is None:
```
stephenroller: I'm not sure I'm ready for default logging-to-file always. This will clobber the NFS we use and anger our admins. Let's just have if filename is None do nothing, and you can document where the right way place to do this is.

Additionally I run probably dozens of commands every day, they'd all overwrite each other without a microsecond stamp.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,129 @@
...
+            today = date.today().strftime("%d/%m/%Y").split("/")
+            day, month, year = today[0], today[1], today[2]
+            proj_root = str(Path(__file__).parent.parent.parent)
```
stephenroller: Even though this code will probably go: better to just `import parlai` to get the root path and not have this triple parent

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,129 @@
...
+            log_dir_path = proj_root + "/" + "Logs/" + year + "/" + month
+            log_fname = year + "-" + month + "-" + day + ".log"
+            if not os.path.exists(log_dir_path):
+                os.makedirs(log_dir_path)
+            filename = os.path.join(proj_root, log_dir_path, log_fname)
+        # Default format used if no file format provided
+        if file_format is None:
+            file_format = '%(asctime)s :: %(levelname)s :: %(message)s'
```
stephenroller: We might be able to amp this up, no?

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,129 @@
...
+        self.fileHandler.setFormatter(self.fileFormatter)
+        super().addHandler(self.fileHandler)
+
+        # Logging to stdout
+        self.streamHandler = logging.StreamHandler(sys.stdout)
+        self.streamHandler.level = (
+            console_level
+        )  # Log to stdout levels: console_level and above
+        if console_format is not None:
+            if len(console_format) == 2:  # If a tuple is passed in
```
stephenroller: hmm, I'd rather check `isinstance(console_format, tuple)`. But it's unclear to me what the two formats indicate? Do you mind describing?

tests/test_torch_agent.py
```diff
@@ -12,6 +12,7 @@
 from parlai.core.testing_utils import capture_output, tempdir
 from parlai.core.utils import Message
 import parlai.core.testing_utils as testing_utils
+from parlai.core.logging_utils import logger
```
stephenroller: Do we need the import in these places that won't use it?

parlai/core/logging_utils.py
vedantpuri: https://github.com/facebookresearch/ParlAI/pull/2030#discussion_r330243127

parlai/core/logging_utils.py
vedantpuri: Usually the format passed in is a simple string. But in some instances of existing loggers in parlai there were formats having the asctime as the first part and the format of the time as the second part of the format. This check was to handle that case.

tests/test_torch_agent.py
vedantpuri: Not sure why I imported it there, you're right it's not being used and I will remove that!

parlai/core/logging_utils.py
stephenroller: Seems like Kurt and I disagree. He gets to make the call.

parlai/core/logging_utils.py
stephenroller: I'm sorry, can you point to an example?

stephenroller: (I'd rather just get rid of what those old things were doing and have this code be simpler.)

vedantpuri:
@stephenroller  Made the changes you had requested. Also since we were discussing of using the importing the logger object as opposed to instantiating it everywhere (https://github.com/facebookresearch/ParlAI/pull/2030#discussion_r329799537 )I thought it might be useful to have this method: ```add_file_handler``` and hence I implemented it so that one can add a file handler to capture it (since the purpose of making this logger was to capture output). 
Removed tuple handling and hence removed areas that passed in formats as tuples and simply passed in strings.

Also, you mentioned about the mass regex swap https://github.com/facebookresearch/ParlAI/pull/2030#pullrequestreview-295988181. Yes! that would be great in a separate PR, otherwise this PR will take forever to complete. Also I feel adding ParlaiLogger.print (I assume you meant logger.print since none of the methods are static in that class) might not be the best approach since that would mess with the task of checking that there are no print statements in future PRs which has been designated as a part of this issue.
Also, As I mentioned in a comment above
'Now while doing all this came up with a few other loggers that are already in place:
Most of them exist in
parlai/mturk and parlai/messenger and use a logger that is instantiated in parlai/mturk/core/shared_utils.py. Would probably want to deal with those next.'
So we need a way to deal with that also (so that they use ParlaiLogger) but I would suggest that that also be done in a separate PR if possible.

@klshuster tagging you as well to notify you of the progress

stephenroller:
We're just going to grep for `^\s*print\(` right? `logger.print` would be okay.

vedantpuri:
Yes I guess that works. Also for the mass replacement: We need to import the logger object to that file as well ``` from parlai.core.logging_utils import logger``` and any other info levels that it might need. But as you mentioned that is more suitable for a separate PR.

So all that remains to  implement is a print function within ParlaiLogger (Would probably have to name it slightly different to avoid conflict with the built in print) but do we need anything apart from that?
@stephenroller @klshuster

projects/controllable_dialogue/mturk/run.py
```diff
@@ -118,7 +118,7 @@ def main():
     argparser.add_argument(
         '--mturk-log',
         type=str,
-        default=('$HOME/ParlAI/data/mturklogs/controllable/{}.log'.format(start_time)),
+        default=('data/mturklogs/controllable/{}.log'.format(start_time)),
```
klshuster: i'm not sure we should change this?

projects/wizard_of_wikipedia/mturk_evaluation_task/run.py
```diff
@@ -90,17 +92,22 @@ def inject_override(opt, override_dict):
...
-        logger.addHandler(console)
+        fmt = '%(asctime)s: [ %(message)s ]'
+        logfile = None
         if 'mturk_log' in opt:
-            logfile = logging.FileHandler(opt['mturk_log'], 'a')
-            logfile.setFormatter(fmt)
-            logger.addHandler(logfile)
+            logfile = opt['mturk_log']
+            if not os.path.isdir(os.path.dirname(logfile)):
+                raise OSError(
```
klshuster: rather than raise here, you can just run `os.makedirs(dir, exist_ok=True)` which will make the dir (`exist_ok` means that it will not error if the directory already exists)

projects/controllable_dialogue/mturk/run.py
vedantpuri: Since commands are run from within parlai we want these logs to be created within the parlai project folder I assumed. For eg. For me ParlAI was not placed in HOME. So this created an unnecessary directory somewhere my project didn't exist. Also I conformed to the way this was done in wizards of wikipedia to keep things consistent, thought it was weird to have different implementations in different files (controllable vs wizard of wikipedia).

Could change it back if you would like, but I feel both files should be consistent with their location.

klshuster: ahh no don't worry about it, that change is fine then

vedantpuri:
@stephenroller In my latest commit I added a log function that uses INFO as the default level and calls the parent class' (logging) log method with the passed in level. I think this might make the migration from print to logging easier. I will wait for you to approve before I merge!
@klshuster I made the os.dir creation change you requested! I shall create a new issue referencing:
this PR and current logging issue
and describe what all needs to be changed:
primarily: conforming existing loggers in parlai/messenger and parlai/mturk to use ParlaiLogger and
switch current print statements to use logger (regex replacement and adding import statement)
test for no print statements in future PRs.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # Logging to a file
+        if filename:
+            # Default format used if no file format provided
+            if file_format is None:
+                file_format = '%(asctime)s :: %(levelname)s :: %(message)s'
+            self.fileHandler = logging.FileHandler(filename)
+            self.fileHandler.level = (
+                file_level
+            )  # Log to file levels: file_level and above
```
stephenroller: ditto above

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        :param file_level: min. Level of messages logged to the file
+        :param filename: The file the logs are written to
+        """
+        super().__init__(name, console_level)  # can be initialized with any level
+
+        # Logging to a file
+        if filename:
+            # Default format used if no file format provided
+            if file_format is None:
+                file_format = '%(asctime)s :: %(levelname)s :: %(message)s'
```
stephenroller: Not necessary but one thing you could consider:

pull this out into a constant at the top, like

```python
DEFAULT_FILE_FORMAT = '%(asctime)s :: %(levelname)s :: %(message)s'
```
and then in the definition of __init__ you'd have as a default parameter, `file_format=DEFAULT_FILE_FORMAT`.

What you have is also good.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        if console_format is not None:
+            self.consoleFormatter = logging.Formatter(console_format)
+            self.streamHandler.setFormatter(self.consoleFormatter)
+        super().addHandler(self.streamHandler)
+
+        # To be used with testing_utils.capture_output()
+        self.altStream = None
+
+    def log(self, msg, level=INFO):
+        """ Default Logging function """
```
stephenroller: nit: remove the spaces, add a period.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        super().log(level, msg)
+
+    def add_file_handler(self, filename, level=INFO, format=None):
+        """
+        Add a file handler to the logger object
```
stephenroller: nit, add a period at the end.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        if not hasattr(self, 'fileHandler'):
+            if format is None:
+                file_format = '%(asctime)s :: %(levelname)s :: %(message)s'
+            self.fileHandler = logging.FileHandler(filename)
+            self.fileHandler.level = level  # Log to file levels: level and above
+            self.fileFormatter = logging.Formatter(file_format)
+            self.fileHandler.setFormatter(self.fileFormatter)
+            super().addHandler(self.fileHandler)
+        else:
+            self.info("ParlaiLogger: A filehandler already exists")
```
stephenroller: do you think this should raise an exception instead?

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    def add_file_handler(self, filename, level=INFO, format=None):
+        """
+        Add a file handler to the logger object
+        Use case: When logging using the logger object instead of instantiating a new ParlaiLogger
+                  this function might  be useful to add a filehandler on the go
+        Only does so if there is no file handler existing
+        """
+        if not hasattr(self, 'fileHandler'):
+            if format is None:
+                file_format = '%(asctime)s :: %(levelname)s :: %(message)s'
```
stephenroller: note that we could reuse the DEFAULT_FILE_FORMAT here (or in the definition.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+            self.fileFormatter = logging.Formatter(file_format)
+            self.fileHandler.setFormatter(self.fileFormatter)
+            super().addHandler(self.fileHandler)
+        else:
+            self.info("ParlaiLogger: A filehandler already exists")
+
+    def add_format_prefix(self, prefix):
+        """Include `prefix` in all future logging statements"""
+        # change both handler formatters to add a prefix
+        new_str = prefix + " " + '%(message)s'
```
stephenroller: is the assumption that the console logger only has '%(message)s' ever as its format? Maybe we can at least document that limitation at the top of the class.

parlai/core/testing_utils.py
```diff
@@ -173,21 +173,23 @@ def __str__(self):
...
-
     >>> with capture_output() as output:
-    ...     print('hello')
+    ...     logger.info('hello')
     >>> output.getvalue()
     'hello'
     """
     sio = TeeStringIO()
+    logger.mute()  # Stop logging to stdout
+    logger.redirect_out(sio)  # Instead log to sio (to preserve output for later)
```
stephenroller: I want to revisit this. later. We might just leave this as is.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def reset_formatters(self):
+        """Resort back to initial formatting"""
+        if hasattr(self, 'fileHandler'):
+            self.fileHandler.setFormatter(self.fileFormatter)
+        self.streamHandler.setFormatter(logging.Formatter('%(message)s'))
+
+    def mute(self):
+        """Stop logging to stdout"""
+        self.streamHandler.level = float('inf')
```
stephenroller: is it okay if level isn't an int? I reckon it is but just wondering

projects/wizard_of_wikipedia/mturk_evaluation_task/run.py
```diff
@@ -81,7 +81,9 @@ def main():
         help='Each worker must be unique',
     )
     argparser.add_argument(
-        '--mturk-log', type=str, default='data/mturklogs/{}.log'.format(start_time)
+        '--mturk-log',
```
stephenroller: i'm quite impressed how deep you've gone to upgrade this old stuff.

tests/nightly/gpu/test_controllable.py
```diff
@@ -6,6 +6,7 @@
 
 import unittest
 import parlai.core.testing_utils as testing_utils
+from parlai.core.logging_utils import logger
```
stephenroller: can drop this one right?

tests/nightly/gpu/test_convai2.py
```diff
@@ -26,15 +26,17 @@ class TestConvai2Seq2Seq(unittest.TestCase):
     def test_seq2seq_hits1(self):
         import projects.convai2.baselines.seq2seq.eval_hits as eval_hits
 
-        with testing_utils.capture_output() as stdout:
+        with testing_utils.capture_output() as output:
```
stephenroller: did this break? my instinct is we don't need this to change?

tests/nightly/gpu/test_transresnet.py
```diff
@@ -8,6 +8,7 @@
 import unittest
 import parlai.scripts.display_data as display_data
 import parlai.core.testing_utils as testing_utils
+from parlai.core.logging_utils import logger
```
stephenroller: yeah let's just get rid of all the extra unnecessay logger imports

tests/test_build_data.py
```diff
@@ -64,11 +64,11 @@ def test_download_multiprocess_chunks(self):
             'https://parl.ai/downloads/mnist/mnist.tar.gz.BAD',
         ]
 
-        with testing_utils.capture_output() as stdout:
+        with testing_utils.capture_output() as output:
```
stephenroller: Let's leave these bits alone, even if the name is a misnomer a bit.

tests/test_params.py
```diff
@@ -34,7 +35,7 @@ class TestParlaiParser(unittest.TestCase):
     def test_upgrade_opt(self):
         """Test whether upgrade_opt works."""
         with testing_utils.tempdir() as tmp:
-            with testing_utils.capture_output() as _:
```
stephenroller: Same here, let's just leave this alone since it wasn't broken.

parlai/core/logging_utils.py
vedantpuri: Tried it out and it seemed to work so I guess it's okay. Got this solution off of SO.

tests/nightly/gpu/test_transresnet.py
stephenroller: Why isn't the linter complaining about these actually?

parlai/core/logging_utils.py
vedantpuri: That's a good catch. Fixed it in my next commit to remove that assumption. Now the only assumption is that the format is divided by ```:``` because that's what I split on. Mentioning that on top of the class.

parlai/core/logging_utils.py
```diff
@@ -0,0 +1,132 @@
...
+        console_level,
+        console_format=None,
+        file_format=None,
+        file_level=INFO,
+        filename=None,
+    ):
+        """
+        Initialize the logger object
+        :param name: Name of the logger
+        :param console_level: min. Level of messages logged to console
```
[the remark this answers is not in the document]
stephenroller: Just to clarify above: I intended to compliment. You wrote a good docstring.

tests/nightly/gpu/test_transresnet.py
vedantpuri: Not sure about that

vedantpuri: Just figured this out. Capture output uses the logger object, so whichever file calls capture output needs to have the logger object imported there. For now removed all imports and restored capture output to its original form (can be dealt in the next PR where print is switched to logging)

parlai/core/testing_utils.py
```diff
@@ -179,18 +179,36 @@ def capture_output():
...
-    # logger.mute()  # Stop logging to stdout
-    # logger.redirect_out(sio)  # Instead log to sio (to preserve output for later)
     with contextlib.redirect_stdout(sio), contextlib.redirect_stderr(sio):
         yield sio
-    # yield sio  # Uncomment this once all print statements removed and comment the above two statements instead
-    # logger.stop_redirect_out()  # Stop redirecting [Removes handler]
-    # logger.unmute()  # From now on log to stdout
+
+
+# # TODO: Replace capture_output with this version before completing #2044
```
stephenroller: In the future i'd prefer if you didn't land commented out code. Please make an appropriate branch.

vedantpuri: Discussed this aspect with @klshuster and this is the solution we came up with. But can keep that in mind for the future.

PR #2032, opened 2019-10-01
Function for accessing task agent in world

klshuster:
**Patch description**
1. Add function to `World` class for accessing the task agent
2. Change `eval` in `train_model` to use the task agent ID when reporting metrics, rather than the task specified on the command line.

**Testing steps**
Relying mainly on unit tests here...

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -159,7 +159,7 @@ class World(object):
 
     def get_task_agent(self):
         """Return task agent, if applicable."""
-        raise RuntimeError('Implement in subworld')
+        raise NotImplementedError('Implement in subworld')
 
     def get_acts(self):
         """Return the last act of each agent."""
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -158,7 +158,7 @@ class World(object):
         return self.agents
 
     def get_task_agent(self):
-        """Return task agent, if applicable"""
+        """Return task agent, if applicable."""
         raise RuntimeError('Implement in subworld')
 
     def get_acts(self):
@@ -270,10 +270,10 @@ class DialogPartnerWorld(World):
         self.acts = [None] * len(self.agents)
         if self.agents is not None and len(self.agents) > 0:
             # Name the world after the first agent.
-            self.id = self.agents[0].getID()
+            self.id = self.get_task_agent().getID()
 
     def get_task_agent(self):
-        """Return task agent"""
+        """Return task agent."""
         return self.get_agents()[0]
 
     def parley(self):
@@ -295,7 +295,7 @@ class DialogPartnerWorld(World):
 
     def epoch_done(self):
         """Only the first agent indicates when the epoch is done."""
-        return self.agents[0].epoch_done()
+        return self.get_task_agent().epoch_done()
 
     def report(self):
         """Report all metrics of all subagents."""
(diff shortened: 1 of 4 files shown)


klshuster:
Re-requesting a review from @stephenroller and/or @emilydinan to confirm that the rest of these changes look good

parlai/core/worlds.py
```diff
@@ -157,6 +157,10 @@ def get_agents(self):
         """Return the list of agents."""
         return self.agents
 
+    def get_task_agent(self):
+        """Return task agent, if applicable."""
+        raise RuntimeError('Implement in subworld')
```
stephenroller: is `NotImplementedError` better?

parlai/core/worlds.py
```diff
@@ -525,6 +537,10 @@ def get_agents(self):
         """Return the agents in the *current* subworld."""
         return self.worlds[self.world_idx].get_agents()
 
+    def get_task_agent(self):
+        """Return the task agent in the *current* subworld."""
+        return self.worlds[self.world_idx].get_task_agent()
```
stephenroller: This is a really funny definition, because there could be many task agents... Maybe this one should just raise a RuntimeError? If we do that, do unit tests fail?

Do we have unit tests covering multitask training? I don't think we do. We need those asap.

parlai/scripts/train_model.py
```diff
@@ -308,7 +308,8 @@ def run_eval(valid_worlds, opt, datatype, max_exs=-1, write_log=False):
         task_report = _run_single_eval(opt, v_world, max_exs / len(valid_worlds))
         reports.append(task_report)
 
-    tasks = [world.opt['task'] for world in valid_worlds]
+    # tasks = [world.opt['task'] for world in valid_worlds]
```
stephenroller: cleanup

parlai/core/worlds.py
klshuster: yeah will do

parlai/core/worlds.py
klshuster: i was trying to match the functionality of the function before, i.e. `def get_agents(self)`. I agree that this is a weird definition for a teacher where there are multiple `task_agents`

we actually test multitasking in two places: `test_train_model` (tests the multitask metrics) and `test_pytorch_data_teacher` (tests multitasking with that). I can try seeing if raising a `RuntimeError` will break those tests

klshuster:
I've changed it to a `RuntimeError` in `Multiworld`, and I pass tests locally (well, our two multi-tasking tests)

jaseweston:
seems this does the job anyway btw?:
                    tasks = [world.getID() for world in valid_worlds]
that is, the world has a name which seems to come from the teacher anyway
also i think get_agents() itself isnt implemented properly in batchWorld
it should be doing something like  def get_agents():  return self.world.get_agents()
but isn't .. maybe u can add to your PR?

stephenroller:
yeah `world.getID()` does seem like a better choice, though i like the `get_task_agent` anyway.

stephenroller:
Jason was multitasking on some things that ended up with the same name with just looking via the `-t` flag.

klshuster:
Ok, fixed `get_agents` in `BatchWorld` and changed `train_model` to use `world.getID()`. Will merge after tests pass

PR #2040, opened 2019-10-02
[wip] Update example seq2seq to be a full torch agent.

stephenroller:
**Patch description**
Begin rewriting the seq2seq tutorial as a minimal implementation of a TorchAgent. I'm not sure I like it, given that we typically recommend people use `TorchGeneratorAgent`.


stephenroller:
Abandoning in favor of #1766.

PR #2041, opened 2019-10-02
Add WandbLogger

annirudh:
**Patch description**
This PR adds a new logger implementation that tracks metrics using https://www.wandb.com/. I found the [dashboards](https://app.wandb.ai/annirudh/parlai) and automatic parameter tracking useful when training models.

I intended `WandbLogger` to be a drop-in replacement the `Tensorboard` logger.

**Testing steps**
- Running `train_model.py` does not require `wandb` pip package when `-wblog` is not specified.
- Running `train_model.py` with `-wblog` correctly sends metrics to W&B
- Using both `-wblog` and `tblog` works, with metrics tracked in both Tensorboard and W&B.

**Logs**
```
python examples/train_model.py -t babi:task10k:1 -mf /tmp/babi_memnn -bs 1 -nt 4 -eps 5 -m memnn -wblog true
```

### DIFF
--- a/parlai/core/logs.py
+++ b/parlai/core/logs.py
@@ -23,6 +23,11 @@ try:
 except ImportError:
     SummaryWriter = None
 
+try:
+    import wandb
+except ImportError:
+    wandb = None
+
 
 class TensorboardLogger(object):
     """Log objects to tensorboard."""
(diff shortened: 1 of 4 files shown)


stephenroller:
Thanks for your contribution. We weren't really planning on adding this. Are there other users besides yourself who would find this useful?

annirudh:
I believe the automatic tracking of hyperparameters, training metrics, and system metrics would be generically useful to anyone training models.

Screen Shot 2019-10-03 at 4 45 41 PM

Screen Shot 2019-10-03 at 4 48 32 PM

For me, a big benefit of W&B over loggers like TensorBoard is cloud persistence. I have a history of all the experiments I've run, which is particularly useful when I'm using W&B to run sweeps as I can query my run history to find the parameters that yield the best performing models.

stephenroller:
I've decided not to accept this patch at this time.

I don't want to add another dependency into our main train loop right now, particularly one that depends a proprietary third party service. Thank you very much for your contribution. We will revisit this decision if we receive significant user requests for it.

PR #2043, opened 2019-10-02
self_observe

stephenroller:
**Patch description**
Enable `self_observe` as part of a `BatchWorld`, which tells `BatchWorld` to relay the batch observation back to the agent.

This is implemented as discussed in *Parley* a few weeks ago, with act() handling itself, and Batch world doing the broadcast.

Notable benefits:
- History doesn't have to think about resetting on an episode done. Self-observe does this now.
- We don't have the weird logic around observe where it tries to find the last reply.
- Shared replies are no longer a thing, ensuring we don't have to deal with the protection logic around turkers clobbering each other's histories.
- `update_history` focuses on just the one observation now. This could be further simplified.
- Will enable dynamic batching.

**Testing steps**
God bless CI, this is a bold one. I kinda wish a few more tests existed, like a simulated turker and a simulated interactive.


stephenroller:
Should update unit tests in test_torch_agent.py.

parlai/core/worlds.py
```diff
@@ -679,16 +679,23 @@ def batch_observe(self, index, batch_actions, index_acting):
...
             else:
-                if index == index_acting:
-                    return None  # don't observe yourself talking
                 observation = validate(batch_actions[i])
-            observation = agents[index].observe(observation)
+
+            if index == index_acting:
+                # self_observe is distinguished from a normal observe
+                if hasattr(agents[index], 'self_observe'):
+                    agents[index].self_observe(observation)
```
emilydinan: interesting that batch world was already kind of set up to do this lol

parlai/core/torch_agent.py
```diff
@@ -1551,47 +1520,59 @@ def match_batch(self, batch_reply, valid_inds, output=None):
...
-            use_reply == 'none'
-            or not self.observation
-            or self.observation['episode_done']
-        ):
-            return None
+        use_reply = self.opt.get('use_reply', 'label')
+
+        if self.last_observation is None:
+            raise RuntimeError(
+                "You're self_observing without having observed something. Check if "
```
emilydinan: idk that we should always throw a RuntimeError here? what if in interactive mode we want the model to talk first? can you double check that this doesn't break?

parlai/core/torch_agent.py
```diff
@@ -1551,47 +1520,59 @@ def match_batch(self, batch_reply, valid_inds, output=None):
...
+            # additionally mark the last observation as invalid
+            self.last_observation = None
+            return
 
-        if use_reply == 'label':
-            # first look for the true label, if we aren't on a new episode
+        if use_reply == 'none':
+            # we're not including our own responses anyway.
+            return
+        elif use_reply == 'label':
```
emilydinan: wasn't there an option like 'label else model'?

parlai/core/worlds.py
stephenroller: Rather, it specifically had carved out a case to avoid observing itself!

parlai/core/torch_agent.py
stephenroller: We never have supported that. Going first is always a `__silence__` input, or just the personas, or whatever.

Yes, I'll need to check interactive.py.

parlai/core/torch_agent.py
```diff
@@ -1716,12 +1696,11 @@ def reset_metrics(self):
 
     def act(self):
         """Call batch_act with the singleton batch."""
-        if self._replies_are_shared:
-            raise RuntimeError(
-                'act() will misbehave in batching mode. Set batchsize to 1, or '
-                '--interactive-mode true'
-            )
-        return self.batch_act([self.observation])[0]
+        # BatchWorld handles calling self_observe, but we're in a Hogwild or Interactive
```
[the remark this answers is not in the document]
stephenroller: Well, there's no reason that `act()` can't handle the self-observation. And there's a ton of places where we nakedly call `act()`. We don't want to have to update all of those. (MTurk for example).

We talked about this in Parley. IIRC no one was particularly enthusiastic about my decision here, but no one also fought me very hard on it.

parlai/core/torch_agent.py
stephenroller: That's really what label is.

stephenroller: Looks like `label_else_model` is what we used pre TorchAgent:

```
devfair0237 parlai selfobserve $ ag label_else_model
parlai/agents/legacy_agents/memnn/memnn_v0.py
87:            default='label_else_model',
89:            choices=['none', 'model', 'label', 'label_else_model'],

parlai/agents/legacy_agents/seq2seq/seq2seq_v0.py
305:            default='label_else_model',
307:            choices=['none', 'model', 'label', 'label_else_model'],

parlai/agents/legacy_agents/seq2seq/utils_v0.py
324:    useReplies='label_else_model',
355:            useReplies == 'label_else_model' and len(history['labels']) == 0

parlai/agents/starspace/starspace.py
162:            default='label_else_model',
164:            choices=['none', 'model', 'label', 'label_else_model'],

parlai/core/utils.py
70:    useReplies='label_else_model',
107:            useReplies == 'label_else_model' and len(history['labels']) == 0
```

parlai/core/torch_agent.py
```diff
@@ -1551,30 +1524,93 @@ def match_batch(self, batch_reply, valid_inds, output=None):
...
+        # update the history using the observation
+        self.history.update_history(observation)
+        return self.vectorize(
+            observation,
+            self.history,
+            text_truncate=self.text_truncate,
+            label_truncate=self.label_truncate,
+        )
+
+    def self_observe(self, self_message: Message) -> None:
```
klshuster: first typed function??

parlai/core/torch_agent.py
klshuster: i can imagine this coming up in e.g. an mturk task where we want the model to come first, but i'm fine with just observing `__silence__` at that point

tests/test_torch_agent.py
```diff
@@ -897,14 +864,51 @@ def test_interactive_mode(self):
...
+        shared.observe(Message({'text': 'bar', 'episode_done': True}))
+        response = shared.act()
+        self.assertIn(
+            'Evaluating 0', response['text'], 'Incorrect output in single act()'
+        )
+
+    def test_use_reply(self):
+        """Check that self-observe is correctly acting on labels."""
+        # default is hybrid label-model, which uses the label if it's available, and
+        # otherwise the label
```
klshuster: `uses the label if it's available, and otherwise the label` what does this mean?

stephenroller:
@jaseweston you might want to be aware of these.

tests/test_torch_agent.py
```diff
@@ -897,14 +864,51 @@ def test_interactive_mode(self):
...
+            agent.history.get_history_str(), 'Call\nEvaluating 0 (responding to Call)!'
+        )
+        # now some of the other possible values of --use-reply
+        # --use-reply model. even if there is a label, we should see the model's out
+        agent = get_agent(use_reply='model')
+        obs = Message({'text': 'Call', 'labels': ['Response'], 'episode_done': False})
+        agent.observe(obs)
+        _ = agent.act()
+        self.assertEqual(agent.history.get_history_str(), 'Call\nTraining 0!')
+        # --use-reply none never thinks of itself
```
emilydinan: this is kind of a weird comment lol

parlai/core/torch_agent.py
```diff
@@ -1551,30 +1524,93 @@ def match_batch(self, batch_reply, valid_inds, output=None):
...
+                "self_observe to clear the history, probably because you missed an "
+                "act(). This was changed in #2043. File a GitHub issue if you require "
+                "assistance."
+            )
+
+        if observation.get('episode_done'):
+            self.__expecting_clear_history = True
+        elif 'labels' in observation or 'eval_labels' in observation:
+            # make sure we note that we're expecting a reply in the future
+            self.__expecting_to_reply = True
```
emilydinan: so this won't catch errors in interactive mode?

parlai/core/torch_agent.py
```diff
@@ -1551,30 +1524,93 @@ def match_batch(self, batch_reply, valid_inds, output=None):
...
+        if self.observation is None:
+            raise RuntimeError(
+                "You're self_observing without having observed something. Check if "
+                "you're missing a step in your observe/act/self_observe loop."
+            )
+
+        if self.observation['episode_done']:
+            if not self.__expecting_clear_history:
+                raise RuntimeError(
+                    "You probably overrode observe() without implementing calling "
```
emilydinan: this is a bit restrictive?

parlai/core/torch_agent.py
stephenroller: The old case of interactive bugs isn’t possible because agents never use shared state to handle reply handling now.

stephenroller: Unless you’re saying we wouldn’t catch a missing call to self observe in interactive, in which case I think you’re right.

stephenroller: I think my concern about an assert without a label is we couldn’t handle multiparty correctly then.

kestasjk:
FYI in case anyone else ends up here I got this error after a pytorch memory exception happened but the exception was caught and retried, so it may not be that the code is missing an act() but the program flow was interrupted without cleaning up.

PR #2049, opened 2019-10-03
GPT-2 tokenizer

dexterju:
**Patch description**
This will add a gpt2 stype bpe tokenizer inside Parlai
**Testing steps**
Added test in test_dict.py

### DIFF
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -9,6 +9,7 @@ from parlai.core.build_data import modelzoo_path
 from .agents import Agent
 from .build_data import make_dir
 from collections import defaultdict
+from .gpt2_helper import Gpt2BpeHelper
 import codecs
 import copy
 import numpy as np
@@ -189,7 +190,7 @@ class DictionaryAgent(Agent):
             default=DictionaryAgent.default_tok,
             help='Which tokenizer to use. Defaults to "split", which splits '
             'on whitespace as well as recognizing basic punctuation. '
-            'Other options include nltk and spacy.',
+            'Other options include nltk, spacy and gpt2.',
             hidden=True,
         )
         dictionary.add_argument(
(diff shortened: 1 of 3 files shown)


dexterju:
@stephenroller  I am not super sure about the test, https://github.com/facebookresearch/ParlAI/blob/master/tests/test_dict.py doesn't test any tokenizer which requires parameters config? Like nltk ones, where self.nltk will not be set unless this parameter is given at init time?

dexterju:
Testing example: 
'Hello, ParlAI! 😀'
['Hello', ',', 'ĠPar', 'l', 'AI', '!', 'ĠðŁĺ', 'Ģ']

stephenroller:
Does vec2text correctly reverse it?

parlai/core/dict.py
```diff
@@ -447,6 +449,12 @@ def nltk_tokenize(self, text, building=False):
             for token in self.word_tok.tokenize(sent)
         )
 
+    def gpt2_bpe_tokenize(self, text):
```
stephenroller: i think this has to be named `gpt2_tokenize`

dexterju:
Should be good for review @stephenroller

parlai/core/gpt2_helper.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python3
```
klshuster: should we consider a `utils` folder (perhaps not even in `core`)? it feels a little weird to me to have `gpt2_helper` in core

parlai/core/gpt2_helper.py
```diff
@@ -0,0 +1,161 @@
...
+        self.byte_encoder = bytes_to_unicode()
+        self.byte_decoder = {v: k for k, v in self.byte_encoder.items()}
+        self.bpe_ranks = dict(zip(bpe_merges, range(len(bpe_merges))))
+        self.cache = {}
+        try:
+            import regex as re
+
+            self.re = re
+        except ImportError:
+            raise ImportError('Please install regex with: pip install regex')
```
klshuster: is this not a requirement in parlai already? seems like it should be if not

stephenroller: This was copy pasta from fairseq.

parlai/core/gpt2_helper.py
stephenroller: I agree. We'll move things in another PR though.

dexterju:
Merge : )

PR #2052, opened 2019-10-04
Detect bad imports with the linter.

stephenroller:
**Patch description**
#2030 had some imports that were unused. I expected the linter to detect these, but they weren't. Turns out, I had marked them as explicit ignores in the config file. Who knows? Maybe past Stephen but not current Stephen.

**Testing steps**
flaked parlai/core and found some unused imports. Let's see who gets them first ;)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -9,7 +9,6 @@ extend-ignore =
     E266
     W503
     F403
-    F401
     D202
 select = C,E,F,W,B,B950,D,RST
 


vedantpuri:
Figured out those bad imports just now. I was calling logger functions in capture_output. So whichever file used capture output needed to import logger so that, the function call won't fail. For now I have removed all import statements and the working of capture output is as it was before.

PR #2054, opened 2019-10-04
Cap max examples in eval_model

klshuster:
**Patch description**
This is probably a bug with a world somewhere, but figured I'd add this in as I've had this change locally for a long time now.

I ran into an issue where eval would not stop after all the examples because a world's `epoch_done()` method was not correctly implemented (I don't remember the command unfortunately). When we're evaluating a model, I don't think it makes sense to evaluate beyond the number of examples in a world; so why not cap `max_cnt` at that instead of `float('inf')`? 

**Testing steps**
Run `eval_model.py` for any model. Also unit tests

### DIFF
--- a/parlai/scripts/eval_model.py
+++ b/parlai/scripts/eval_model.py
@@ -75,7 +75,7 @@ def _eval_single_world(opt, agent, task):
     log_time = TimeLogger()
 
     # max number of examples to evaluate
-    max_cnt = opt['num_examples'] if opt['num_examples'] > 0 else float('inf')
+    max_cnt = opt['num_examples'] if opt['num_examples'] > 0 else world.num_examples()
     cnt = 0
 
     while not world.epoch_done() and cnt < max_cnt:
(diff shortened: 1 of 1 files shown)


stephenroller:
Hmm, interesting that this broke so many tests...

stephenroller:
Sounds like verify_data.py should assert that num_examples and num_episodes is correct...

stephenroller:
num_examples must be lying in the multiworld?

klshuster:
i believe this is related to the `train:evalmode` bug, so i think we can abandon this; will do so once that fix is in

stephenroller:
Will you dot this i, @klshuster?

stephenroller:
I think the root bug is that batchworld with training like, never hits epoch done maybe. We've gotten several complaints about this recently.

klshuster:
can you elaborate on the complaints we've gotten? I was sure that this was a `train:evalmode` bug specifically (i.e. that we never reach `epoch_done` with that datatype)

stephenroller:
yeah so if -bs 1, I think we do hit epoch_done eventually, but if -bs > 1, then it loops forever. But only in train:evalmode. Which means there's probably some if condition somewhere that checks `if train in datatype` instead of `if train in datatype and evalmode not in datatype`. Probably around batchworld or teachers.

stephenroller:
We can abandon this with #2239, yes?

klshuster:
yes

Issue #1881, opened 2019-07-26
MemNN backwards compatibility integration test


stephenroller:
Right now MemNN has some tests, but it it lacks a backwards compatibility test like some of the other models:

https://github.com/facebookresearch/ParlAI/blob/48187b7aaacea5f910719074fe78d13c409e6776/tests/test_transformers.py#L127-L168

The task is to train a memnn that solves the `-t integration_test` task and produce the model. We will then upload it to the unit test files and ensure it's run in CircleCI always.

domrigoglioso:
It seems like the pre-trained models get downloaded from an internal website in dowload_unittest_models(). After the MemNN model is made, how would I go about uploading that?

stephenroller:
Kurt can upload them.

PR #2056, opened 2019-10-04
Backwards compatibility test for MemNN (#1881)

domrigoglioso:
**Patch description**
Related Issue: #1881
 Trained a MemNN model on `-t integration_test`
(Full command: 
`python examples/train_model.py -m memnn -t integration_tests -mf /tmp/integ_memnn -bs 1 -sval True -vtim 1000 -ttim 7200`)

**Testing steps**
Ran test_memnn.py and expected to pass test cases (with one skip). Test cases passed.

**Logs**
```
$ python tests/test_memnn.py

.s.
Ran 3 tests in 16.772s

OK (skipped=1)
```


tests/test_memnn.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+# !/usr/bin/env python3
```
stephenroller: No space here.

tests/test_memnn.py
```diff
@@ -77,6 +77,39 @@ def test_labelcands_multi(self):
             "test hits@1 = {}\nLOG:\n{}".format(test['hits@1'], stdout),
         )
 
+    def test_backcomp(self):
+        '''
```
stephenroller: use " instead of ' for docstring quotes.

tests/test_memnn.py
```diff
@@ -77,6 +77,39 @@ def test_labelcands_multi(self):
...
+                model='memnn',
+                model_file='zoo:unittest/memnn/model',
+                dict_file='zoo:unittest/memnn/model.dict',
+                batch_size=16,
+            )
+        )
+
+        self.assertGreaterEqual(
+            valid['accuracy'],
+            0.88,
```
stephenroller: was hoping for a higher number but alright.

domrigoglioso: I could try re-training the model with a different learning rate/other parameters or for a longer time if necessary. Also Kurt has already uploaded the model.

stephenroller: dope, this is perfect then.

PR #2035, opened 2019-10-02
Implement conversation rendering

vedantpuri:
**Patch description**
A script to render conversations provided in ACUTE-Eval format. Implements the feature requested in #2021 

**Testing steps**
To create the conversation:
```
python projects/controllable_dialogue/tasks/build.py
```

Generating CLI output:
```
python convo_render.py data/controllable_dialogue/evaluation_logs/inquisitive_model_ct_setting10.jsonl -ne 10
```

Generating an HTML file:
```
python convo_render.py data/controllable_dialogue/evaluation_logs/inquisitive_model_ct_setting10.jsonl -o web_convo.html -ne 10
```

Generating a PDF file:
```
python convo_render.py data/controllable_dialogue/evaluation_logs/inquisitive_model_ct_setting10.jsonl -o web_convo.pdf -ne 10
```

Generating a PNG file:
```
python convo_render.py data/controllable_dialogue/evaluation_logs/inquisitive_model_ct_setting10.jsonl -o web_convo.png -ne 10
```
**Logs**
Generated CLI Output:
Screen Shot 2019-10-01 at 10 44 26 PM

Generated HTML File: (Removed several ```li``` statements manually to keep it concise.)
```HTML
<html>
<head>
	<meta http-equiv="content-type" content="text/html; charset=utf-8">
	<title>Rendered HTML</title>
	<style type="text/css">
		@media print {
			@page { margin: 0; size:8in 9.5in;}
		}
		ul{
			list-style: none;
			margin: 0;
			padding: 0;
		}
		ul li{
			display:inline-block;
			clear: both;
			padding: 20px;
			border-radius: 30px;
			margin-bottom: 2px;
			font-family: Helvetica, Arial, sans-serif;
		}
		.inquisitive_model_ct_setting10{
			background: #eee;
			float: left;
		}
		.human{
			background: #0084ff;
			color: #fff;
			float: right;
		}
		.breaker{
			color: #bec3c9;
			display: block;
			height: 20px;
			margin: 20px 20px 20px 20px;
			text-align: center;
			text-transform: uppercase;
		}

	</style>
</head>
<body>
	<ul>
		<li class="human">How are you doing today?</li>
		<li class="inquisitive_model_ct_setting10">I am good, how are you?</li>
		<li class="human">Pretty good, just making plans to go to business school next year.</li>
		<li class="inquisitive_model_ct_setting10">What kind of business school?</li>
		<li class="human">Just regular business school I guess.  I have a degree already in finance.</li>
[...]

Issue #2021, opened 2019-09-26
Conversation Renderer


stephenroller:
# Feature Description

We could use a script (in `parlai.scripts`) that is a *conversation renderer*. That is, it can be used to generate figures that can be included in papers, blog posts, etc; or used for inspecting logs in a friendly manner.

It should take as input one our conversation formats. Look in the CLI section of this Issue for examples of that format. These are the same inputs to the [ACUTE-Eval](https://arxiv.org/abs/1909.03087), and are likely going to be our standard going forward.

---

Ideally such a renderer would support multiple modes:
- [x] Command line rendering (see below)
- [x] HTML rendering (good for jupyter notebooks)
- [x] PDF rendering (best for papers)
- [x] PNG rendering (would use the same as PDF, just different options)

Other desired features:
- [x] Roughly mirror [Messenger style](https://www.socialmediaexaminer.com/wp-content/uploads/2018/05/facebook-messenger-chat-conversation.png) with rounded corners, blues/grays, etc.
- [ ] Icons to use for both side of the conversation. See [ACUTE-Eval](https://arxiv.org/abs/1909.03087) figures to see how this sort of works.
- [ ] PDF maybe have transparent background if possible?
- [x] For PNG/PDF, command line options to determine the size.
- [ ] Mute stdout from headless chrome when creating a PDF/PNG 

## HTML

A function that outputs a string of HTML.

This will be one of the root ones, and the PDF/PNG ones should just output a temporary HTML file and then call headless chrome correctly.

## PDF Howto

I've managed to get nice PDF rendering with Headless Chrome, and got the admins to install chrome on all of our cluster machines. This command works on other OSes but heres OS X:

`/Applications/Google\ Chrome.app/Contents/MacOS//Google\ Chrome --headless --print-to-pdf 'file:///tmp/test.html'`

Where /tmp/test.html contains:
```html
<html>
<head>
    <meta http-equiv="content-type" content="text/html; charset=utf-8">

    <title>Test</title>
    <style type="text/css">
@media print {
  @page { margin: 0; }
 }
    </style>
</head>
<body>
    <h1> oy whos down for to parley</h1>
    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed auctor
    vestibulum sem id luctus. Class aptent taciti sociosqu ad litora torquent
    per conubia nostra, per inceptos himenaeos. Aliquam facilisis sit amet quam
[...]

PR #2059, opened 2019-10-05
Enhance conversation renderring

vedantpuri:
**Patch description**
Builds upon #2035 in order to complete other tasks remaining in #2021 
In this PR:
- Add images to the conversation
- Add CL argument to control size for screenshots
- Mute headless logging

**Testing steps**
Assuming the jsonl file is created already, the following command can be used to generate a screenshot of dimensions 800 X 590
```bash
python convo_render.py inquisitive_model_ct_setting10.jsonl -o out.png -ws 800 590
```

**Logs**
out.png from the previous command:
amb

**Other information**
Following up on other tasks:
- I tried researching but couldn't find a way to make the background transparent using headless chrome, if you know of a way please let me know!
- Window size seems to be useful only for screenshots, the PDF size is predetermined by the height and width parameter of the HTML page (already have args for these)
- Currently in a screenshot one can see the scroller on the top right. Minor issue but not sure how to get rid of that, just the way headless works.
- @klshuster could you elaborate on the image and captioning task and how you want it to look like ? Maybe even provide a sample dataset that can be considered to be the standard going ahead.

### DIFF
--- a/parlai/scripts/convo_render.py
+++ b/parlai/scripts/convo_render.py
@@ -4,12 +4,12 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-import os
 import sys
 import json
 import random
 import argparse
 import tempfile
+import subprocess
 
 # Constants
 END_OF_CONVO = "EOC"
@@ -17,6 +17,7 @@ CHROME_PATH = r'/Applications/Google\ Chrome.app/Contents/MacOS//Google\ Chrome'
 
 ALT_EMOJI_IMG = "https://pbs.twimg.com/media/DUzY3TpWkAAOi34.png"
 HUMAN_EMOJI_IMG = "https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/160/apple/76/woman_1f469.png"
+
 
 def gen_convo_ul(conversations):
     """
@@ -250,5 +251,7 @@ if __name__ == "__main__":
                     cmd = f"{CHROME_PATH} --headless --crash-dumps-dir=/tmp --print-to-pdf=\"{output_file}\" {fname}"
                 else:
                     cmd = f"{CHROME_PATH} --headless --crash-dumps-dir=/tmp --screenshot=\"{output_file}\" {fname}"
-                os.system(cmd)
+                subprocess.run(
+                    cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True
+                )
                 file_handle.close()
--- a/parlai/scripts/convo_render.py
+++ b/parlai/scripts/convo_render.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import os
 import sys
 import json
 import random
@@ -15,7 +16,7 @@ import subprocess
 END_OF_CONVO = "EOC"
 CHROME_PATH = r'/Applications/Google\ Chrome.app/Contents/MacOS//Google\ Chrome'
 
-ALT_EMOJI_IMG = "https://pbs.twimg.com/media/DUzY3TpWkAAOi34.png"
+ALT_EMOJI_IMG = "https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/160/facebook/230/parrot_1f99c.png"
 HUMAN_EMOJI_IMG = "https://emojipedia-us.s3.dualstack.us-west-1.amazonaws.com/thumbs/160/apple/76/woman_1f469.png"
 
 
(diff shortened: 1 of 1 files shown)


parlai/scripts/convo_render.py
```diff
@@ -175,6 +216,13 @@ def display_cli(conversations, alt_speaker, human_speaker):
         type=int,
         default=10,
     )
+    parser.add_argument(
+        "--window-size",
+        "-ws",
+        help="Window Size for screenshot",
+        nargs='+',
+        default=[800, 480],
+    )
```
stephenroller: Use the `--width` and `--height` arguments you already have. Just multiply them by 300 (dpi).

parlai/scripts/convo_render.py
```diff
@@ -209,6 +257,17 @@ def display_cli(conversations, alt_speaker, human_speaker):
...
+                    if len(args.window_size) != 2:
+                        raise Exception("Invalid window size provided")
+                    if (
+                        not args.window_size[0].isdigit()
+                        or not args.window_size[1].isdigit()
+                    ):
+                        raise ValueError(
+                            "Please provide integer values for window size"
+                        )
+                    cmd = f"{CHROME_PATH} --headless --crash-dumps-dir=/tmp --window-size={args.window_size[0]},{args.window_size[1]} --screenshot=\"{output_file}\" {fname}"
```
stephenroller: wrap multiple lines

parlai/scripts/convo_render.py
```diff
@@ -4,17 +4,20 @@
...
 import random
 import argparse
 import tempfile
+import subprocess
 
 # Constants
 END_OF_CONVO = "EOC"
 CHROME_PATH = r'/Applications/Google\ Chrome.app/Contents/MacOS//Google\ Chrome'
 
+ALT_EMOJI_IMG = "https://pbs.twimg.com/media/DUzY3TpWkAAOi34.png"
```
stephenroller: the user should be able to specify the icons as arguments. These are fine as defaults though. Maybe the alt could be the [facebook parrot emoji](https://emojipedia.org/facebook/4.0/parrot/) by default though :D

vedantpuri: Sure I can add that, but they need to be URL's because local image files don't work well when converting from html to pdf/png.

parlai/scripts/convo_render.py
vedantpuri: Could you elaborate a bit on that please?

parlai/scripts/convo_render.py
stephenroller: Really? Even `file:///path/to/filename.png`?

parlai/scripts/convo_render.py
stephenroller: Line length should be under 88 characters. 

```python
cmd = (
    f"{CHROME_PATH} --headless --crash-dumps-dir=/tmp "
    f"--window-size={args.window_size[0]},{args.window_size[1]} "
    f"--screenshot=\"{output_file}\" {fname}"
)
```

parlai/scripts/convo_render.py
vedantpuri: Tried multiplying them by 300 but that results in a very large conversation and hence everything looks very small. Example:
b

Using 100 as the mulltiplier seems better:
b

parlai/scripts/convo_render.py
vedantpuri: Hadn't tried the file approach. Thanks for that suggestion! Works with that!

parlai/scripts/convo_render.py
vedantpuri: Used 100 for now!

vedantpuri:
@klshuster I was thinking that it might be better to have a separate script to generate the images and captions. Adding logic to this would just clutter the code more and make the script harder to read. Just my opinion, any thoughts ?

klshuster:
@vedantpuri that's not a bad idea, perhaps that would be something like `image_captions_renderer.py`; might be good to file a new issue requesting this functionality

parlai/scripts/convo_render.py
```diff
@@ -176,24 +240,75 @@ def display_cli(conversations, alt_speaker, human_speaker):
...
+    :return: src (possibly pre-pended with "file://")
+    """
+    if src != default:
+        # check if URl
+        if not src.startswith('https://') and not src.startswith('http://'):
+            # Either a file or incorrect input
+            if os.path.isabs(src):
+                src = "file://" + src
+            else:
+                raise Exception(
```
stephenroller: use an `IOError`; in general you should try to avoid base class exceptions.

vedantpuri:
Once merged, can the issue be closed ?
I moved the image captioning rendering to another issue as discussed with Kurt above!

klshuster:
yes, please close the corresponding issue if you have not already done so

PR #2060, opened 2019-10-05
(just for debug)testing my_agent 

cloudygoose:
I'm trying to test a fairseq model for interactive mturk HIT.

### DIFF
--- a/parlai/agents/my_agent/my_agent.py
+++ b/parlai/agents/my_agent/my_agent.py
@@ -23,10 +23,12 @@ class MyAgentAgent(Agent):
         # put any other special reusing stuff in shared too
         shared = super().share()
         shared['model'] = self.model
+        return shared
 
     def observe(self, observation):
         # your goal is to build up the string input to the model here
         self.dialogue_history += observation['text']
+        return observation
 
     def act(self):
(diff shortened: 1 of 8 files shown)


parlai/agents/my_agent/my_agent.py
```diff
@@ -0,0 +1,36 @@
...
+        self.dialogue_history = ''
+
+    def reset(self):
+        super().reset()
+        self.dialogue_history = ''
+    
+    def share(self):
+        # put any other special reusing stuff in shared too
+        shared = super().share()
+        shared['model'] = self.model
```
stephenroller: I messed up here. You need to `return shared`

parlai/agents/my_agent/my_agent.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def observe(self, observation):
+        # your goal is to build up the string input to the model here
+        self.dialogue_history += observation['text']
```
stephenroller: I messed up, you need to return observation here.

cloudygoose: Thanks so much! This works!

Issue #2058, opened 2019-10-04
[TGA] beam search continues after EOS


stephenroller:
**Bug description**
This bug was found by @uralik. Beam search continues after EOS is produced, which is effectively "shrinking" the beam size after the first EOS. This is likely making beam search slightly weaker.

**Steps to resolution**
- [x] Add an assert in `_generate` that all final hypothesis should end on an EOS and have exactly one EOS. Verify the assert is failing on a standard setup.
- [x] Make `TreeSearch.advance()` stop exploring that beam once EOS is done (fix the bug).

The assert should be left in with a note explaining that the user messed up their implementation some how.

uralik:
# Reproduce:

1. Add line:
`print('\n'.join([str(i[0].tolist()) for i in beams[0]._get_rescored_finished(n_best=30)]))`

in the end of `self._generate()` function of TGA. The idea is to check n_best list after running a beam search.

2. Run the following command:
`python ParlAI/parlai/scripts/display_model.py -t convai2:self:none -mf <YOUR_MODEL> -ne 1 --inference beam --beam-size 10 -bs 1 --skip-generation False --beam-min-n-best 30`

The output will look like this:
```
[1, 14, 8, 58, 1880, 1390, 1089, 12, 169, 5, 102, 72, 15, 20, 2, 2, 2, 2]
[1, 14, 8, 58, 1880, 1390, 1089, 12, 169, 5, 15, 20, 2, 2, 2, 2, 2, 2]
[1, 14, 8, 58, 1880, 1390, 1089, 12, 169, 5, 102, 72, 15, 20, 2, 2, 2]
[1, 14, 8, 58, 1880, 1390, 1089, 12, 169, 5, 15, 20, 2, 2, 2, 2, 2]
[1, 14, 8, 58, 1880, 1390, 1089, 12, 169, 5, 102, 72, 15, 20, 2, 2]
[1, 14, 8, 58, 1880, 1390, 1089, 12, 169, 5, 15, 20, 2, 2, 2, 2]
[1, 14, 8, 58, 1880, 465, 1089, 12, 169, 5, 15, 20, 2, 2, 2, 2, 2, 2]
[1, 14, 8, 58, 1880, 1390, 1089, 12, 169, 5, 102, 72, 15, 20, 2]
[1, 14, 8, 58, 1880, 1390, 1089, 12, 169, 6, 102, 72, 15, 20, 2, 2]
[...]

PR #2064, opened 2019-10-07
[TGA] Beam EOS score fix

uralik:
Fix for https://github.com/facebookresearch/ParlAI/issues/2058

Steps to reproduce and show how it differs with the fix are shown in the issue linked above.

The first commit adds EOS penalty in the advance func.

The second commit changes the scores returned by the nucleus and topk. Since after the first EOS sequence will have a neginf score (since we panelize it in the base class), we change it to always output score of the currently selected token instead. Since there is no notion of a hypothesis or sequence in topk/nucleus, this should be fine. 

In case if one needs a seq-level score, it can be easily obtained by summing over needed scores from `self.all_scores` (no backtracking is needed as in beam).


parlai/core/torch_generator_agent.py
```diff
@@ -1043,8 +1049,7 @@ def select_paths(self, logprobs, prior_scores):
         hyp_ids = torch.arange(logprobs.size(0)).to(logprobs.device)
         tok_ids = indices[hyp_ids, choices]
         scores = values[hyp_ids, choices]
-        best_scores = prior_scores.expand_as(scores) + scores
-        return (hyp_ids, tok_ids, best_scores)
+        return (hyp_ids, tok_ids, scores)
```
stephenroller: Why are you changing this? The contract of tree search is the logprob of the full utterance is contained in the third return value.

parlai/core/torch_generator_agent.py
```diff
@@ -853,8 +852,15 @@ def advance(self, logprobs):
...
 
+        # penalize hypotheses ending in EOS on the prior scores (self.scores) level
+        # this is only related to search which uses prior scores (self.scores) (e.g. beam)
+        for hyp_id, token in enumerate(self.outputs[-1]):
+            if token == self.eos:
+                self.scores[hyp_id] = neginf(self.scores.dtype)
+
         hyp_ids, tok_ids, self.scores = self.select_paths(logprobs, self.scores)
-        self.all_scores.append(self.scores)
+        # use clone() here to ensure that self.all_scores will not be changed later due to any penalties to self.scores
```
stephenroller: wrap on 2 lines please

parlai/core/torch_generator_agent.py
```diff
@@ -870,7 +876,7 @@ def advance(self, logprobs):
                 eostail = _HypothesisTail(
                     timestep=len(self.outputs) - 1,
                     hypid=hypid,
-                    score=self.scores[hypid],
+                    score=self.all_scores[-1][hypid],
```
stephenroller: does this mean we're excluding the probability of generating the EOS token?

parlai/core/torch_generator_agent.py
```diff
@@ -8,8 +8,7 @@
 """
 Generic PyTorch-based Generator agent.
 
-Implements quite a bit of boilerplate, including forced-decoding loss and beam
-search.
+Implements quite a bit of boilerplate, including forced-decoding loss and a tree search.
```
stephenroller: does this still fit on 88 characters?

parlai/core/torch_generator_agent.py
```diff
@@ -853,8 +852,15 @@ def advance(self, logprobs):
         if self.scores is None:
             self.scores = torch.zeros(1).type_as(logprobs).to(logprobs.device)
 
+        # penalize hypotheses ending in EOS on the prior scores (self.scores) level
+        # this is only related to search which uses prior scores (self.scores) (e.g. beam)
+        for hyp_id, token in enumerate(self.outputs[-1]):
+            if token == self.eos:
+                self.scores[hyp_id] = neginf(self.scores.dtype)
```
stephenroller: correct me if i'm wrong: if you modified `logprobs[hyp_id]` here such that EVERYTHING was neginf except padding, then the assert I propose would be okay. (Nucleus and topk search would start only outputting padding)

parlai/core/torch_generator_agent.py
uralik: we want to put the value from all_scores here (copied there at line 863), because self.scores is an active buffer which we will penalize on the next time step. Then here it will be a -inf score in the end (this is bad).

parlai/core/torch_generator_agent.py
uralik: I changed this to avoid having -inf scores here after the first EOS was sampled. But actually thats fine if all finalized hypos after the first one will have -inf score, I will change this back. Then only the first sampled EOS will make a non -inf finalized hypo per each beam id

parlai/core/torch_generator_agent.py
uralik: discussed offline another way

parlai/core/torch_generator_agent.py
uralik: yes

parlai/core/torch_generator_agent.py
uralik: ok

uralik:
Several changes in the latest commit:

1.  Only EOS pointers with legit (non-inf) scores are added to `.finished`, this way to make sure that in case of stochastic decoding all finished predictions have only a single EOS.
2. To check point 1 above there is a loop with assert over finalized candidates
3. Since this loop in point 2 are done, there is no need to call `.get_top_hyp`, it is a waste of time, we change this too.
4. Now important: with such assert, default value of `beam_min_n_best` (3) is not good anymore. For example, in such case any decoding with beam size < 3 will crash with this assert. To address this, we add:
a. default argparse value now is 0. If it was not given form saved opt, then we substitute beam_size value there in the code (safe option if a user has no idea about it).
b. if we load the model and value there will be 3 (old default), then we put a warning saying that in some cases this may lead to a crash. In the comment we also point out that user may override this with a CLI argument.

Perhaps there are better ways of handling this `beam_min_n_best`, lets discuss if you have smth in mind.

parlai/core/torch_generator_agent.py
```diff
@@ -305,9 +297,8 @@ def __init__(self, opt, shared=None):
         init_model, is_finetune = self._get_init_model(opt, shared)
         super().__init__(opt, shared)
 
-        self.beam_size = opt.get('beam_size', 1)
-        self.beam_min_n_best = opt.get('beam_min_n_best', 3)
-        self.beam_min_length = opt.get('beam_min_length', 3)
+        self.beam_size = opt.get('beam_size')
```
stephenroller: These changes should be undone. We don’t want none values

parlai/core/torch_generator_agent.py
```diff
@@ -264,17 +263,10 @@ def add_cmdline_args(cls, argparser):
...
-            type=int,
-            default=3,
-            help='Minimum number of nbest candidates to achieve '
-            'during the beam search',
-        )
         agent.add_argument(
             '--beam-min-length',
             type=int,
-            default=1,
+            default=3,
```
stephenroller: Undo this

parlai/core/torch_generator_agent.py
```diff
@@ -728,8 +715,21 @@ def _generate(self, batch, beam_size, max_ts):
             ).unsqueeze(-1)
             decoder_input = torch.cat([decoder_input, selection], dim=-1)
 
-        # get the top prediction for each
-        beam_preds_scores = [b.get_top_hyp() for b in beams]
+        # check that there is at least one finished candidate
+        # and assert that each of them contains only one EOS
+        n_best_beam_preds_scores = [b._get_rescored_finished() for b in beams]
```
stephenroller: So we don’t want to call methods with a leading _, because they’re supposed to be protected. So can we upgrade the behavior of get_top_hyp instead of having this logic here?

parlai/core/torch_generator_agent.py
```diff
@@ -942,12 +947,13 @@ def _get_rescored_finished(self, n_best=None):
...
         # if we never actually finished, force one
         if not self.finished:
+            self.outputs[-1][0] = self.eos
             self.finished.append(
                 _HypothesisTail(
                     timestep=len(self.outputs) - 1,
                     hypid=0,
                     score=self.all_scores[-1][0],
-                    tokenid=self.eos,
+                    tokenid=self.outputs[-1][0],
```
stephenroller: Do we need these changes?

uralik: basically we can keep self.eos, but using outputs (previously updated) makes a clear reason why we have updated it, so I would keep it like this

parlai/core/torch_generator_agent.py
uralik: yeah, we get rid of get_top_hyp since we actually do not use it at all , since we do this EOS assert check for each candidate. Now get_rescored_finished will be not protected.

PR #2068, opened 2019-10-08
Escape emoji.

stephenroller:
**Patch description**
Unfortunately, due to a local configuration issue, my environment does not render the emoji, which is a little obnoxious. Rather than spend a day trying to fix the configuration of the jumphost, I'm requesting we just escape the unicode here.

**Testing steps**
tests/test_dict.py


dexterju:
:)

PR #2070, opened 2019-10-08
Add missing args in create_agent even when datapath is present

ideanl:
**Patch description**
This patch updates `create_agent` to fill in missing fields even when `datapath` is specified. Previously, if `datapath` was specified, the code to add these model args would not run.

**Testing steps**
Create an agent and specify `datapath`, check to see if missing model subargs are added.

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -656,20 +656,19 @@ def create_agent(opt, requireModelExists=False):
     opt['model_file'] + '.opt' must exist and contain a pickled or json dict
     containing the model's options).
     """
-    if opt.get('datapath', None) is None:
-        # add datapath, it is missing
-        from parlai.core.params import ParlaiParser, get_model_name
+    from parlai.core.params import ParlaiParser, get_model_name
 
-        parser = ParlaiParser(add_parlai_args=False)
-        parser.add_parlai_data_path()
-        # add model args if they are missing
-        model = get_model_name(opt)
-        if model is not None:
-            parser.add_model_subargs(model)
-        opt_parser = parser.parse_args("", print_args=False)
-        for k, v in opt_parser.items():
-            if k not in opt:
-                opt[k] = v
+    parser = ParlaiParser(add_parlai_args=False)
+    # add datapath, if it is missing
+    parser.add_parlai_data_path()
+    # add model args if they are missing
+    model = get_model_name(opt)
+    if model is not None:
+        parser.add_model_subargs(model)
+    opt_parser = parser.parse_args("", print_args=False)
+    for k, v in opt_parser.items():
+        if k not in opt:
+            opt[k] = v
 
     if opt.get('model_file'):
         opt['model_file'] = modelzoo_path(opt.get('datapath'), opt['model_file'])


stephenroller:
I'm like 99% sure this patch will have a negative consequence that isn't being caught by the tests here. (I'm honestly not convinced the code we have already is correct). You should instead ensure what you're passing to create_agent is a full, complete opt.

If you want to tell me about how you ended up discovering this, I can potentially help you more.

stephenroller:
Then again, it.... might be okay?

ideanl:
I discovered this when calling `create_agent` with an internal model that had additional required args that weren't being populated in the `opts` (causing exceptions). I eventually realized this was only an issue when the `datapath` was specified, and it doesn't really make sense that model_subargs would be added to the parser only if `datapath` is specified. Ran this by @klshuster.

stephenroller:
Why weren't the args being included via `add_cmdline_args`? Can you make a minimal example to demonstrate the issue?

ideanl:
The args are included in `add_cmdline_args`, but `add_cmdline_args` is never being called since it is called by `add_model_subargs` on a `ParlaiParser`. 

As an example, consider an agent subclassing `TorchAgent`. `TorchAgent.add_cmdline_args` includes the `interactive-mode` arg and sets it to False by default. Now suppose I try to create an agent like the following:

```
opts = {'model_file': ..., 'model': ..., 'datapath': ...}
agent = create_agent(opts)
```

Because `datapath` is specified, `add_cmdline_args` is never run since a `ParlaiParser` would not be instantiated and `add_model_subargs` is never called. Thus, `interactive-mode` is never added.

However, the following example would successfully lead to `interactive-mode` being added and it would default to False (because `datapath` isn't specified in `opts`):
```
opts = {'model_file': ..., 'model': ...}
agent = create_agent(opts)
```

This patch would equalize the behavior in both scenarios so that both cases would instantiate the `ParlaiParser` and add any relevant args with their default values.

stephenroller:
Yes, the semantics makes sense to me here. Can you explain a bit more about how you discovered the bug? I’m curious what your use case was that you were instantiating an agent via a partial opt that had datapath, instead of say, using a full blown parser in your script or similar.

stephenroller:
To be clear, I’m just trying to understand how we hit this. TBH I didn’t even know we had this strange special case in here. This change might be the right thing; deleting the entire special block and mandating fully instantiated opts might be the right thing

ideanl:
More specifically, the `opts` are being populated with defaults at the entry point of a service via instantiating a `ParlaiParser` and running `parse_args` on it. However, at a later point when we retrieve models, we set `opt['model']` to a different value (previously was None) which means the model_subargs are not populated in the `opt`. The `datapath` was being set to a mounted directory where the datasets have already been downloaded.

I could resolve this issue in my code by instantiating another `ParlaiParser` and calling `parse_args` just before calling `create_agent`, but I discovered this is already being done here when a `datapath` isn't set. In theory it is possible that deleting the entire special block may make sense, but it may break any systems that currently depend on that behavior of parsing args in `create_agent`.

emilydinan:
My only concern is that if the model args are already parsed then we might be overriding things by re-parsing. I'm a bit surprised that all tests have passed. Can you add a `[long]` in you commit message?

stephenroller:
I don't think the [long] trick works for forked repos sadly.

ideanl:
Even though we're reparsing, it looks like the only time we actually modify the original `opt` is:
```
if k not in opt:
  opt[k] = v
```
This shouldn't override anything already specified in `opt` but could add new entries.

PR #2073, opened 2019-10-09
Suppress CI outputs and Make minor nits

vedantpuri:
**Patch description**
While working on another branch I realized that Circle CI tests were being verbose in some cases. This PR is to fix that and other stuff needed to do that:
- tfidf_retriever test had some issue related to logging which was resulting in it not properly suppressing output during the circle CI run, fixed that
- legacy versioning was also quite verbose, added a capture_output around that. I assumed here that we do not want any output in the tests and hence did that, let me know if I was wrong in doing so.
- A gpt-2 test was also spitting out output, so silenced that as well
- Since I was making a PR regarding these small nits, I thought might as well remove the comments you asked me not to be there @stephenroller in testing_utils and distributed_utils.

**Testing steps**
Run the unit tests. None of them should be verbose, i.e. all contain just the name of the test and an 'ok' once completed.

**Logs**
N/A

**Other information**
One thing that **still remains** that I found was that the quickstart test recently added by @young-k is also quite verbose. I tried fixing it by redirecting it to /dev/null but that results in a test failure due to the case of (10mins no output) arising. I think we should try and fix that since currently it is quite verbose. Another suggestion I had for that test was to surround each command with echo message of what its running and a completion echo. A solution to this could be to echo a "." every 8 - 10 mins so that the test doesn't fail and we safely redirect the output. Again, not sure if it was intended to have the output but assuming not. @young-k you might know more about this and might be able to fix it if need be!

N/A


vedantpuri:
Recreating from beginning to include changes in master!

Issue #2075, opened 2019-10-09
Shift quickstart.sh to nightly tests


vedantpuri:
As discussed with @klshuster, it might be better to shift quickstart.sh to a nightly test since it takes a very long time to run and it would suffice to check it once a day.

Also, the test is currently very verbose (which adds to the run-time since IO increases the time it takes to run), we should try and reduce it. Suggestions I had to improve this test:
- Surround each command with a printf, echo message specifying what test is being run and that it has completed.
- For commands taking a long time to run, one could use the ```log every n secs``` argument and crank that up. We still need some verbosity because circle CI by default fails if there is no output for 10mins or more. An alternative to this could be to redirect all output to /dev/null/ and  figure out a way to echo a symbol say a "." every few seconds indicating progress and circumventing the circle CI default failute of 10mins.

**Notes:** 
- If quickstart.sh is run with ```sh tests/quickstart.sh >/dev/null/ ``` circle CI fails due to the 10min limit.
- If each python command within quickstart.sh is redirected to /dev/null/ also fails on training and has some issues while running.

@young-k maybe you could try working on this or providing insights about how to go about it!

stephenroller:
I suspect it's spending a long amount of time on downloading data. It goes for 3 models and 3 tasks.

We could also cap this with a `--num-examples`:
https://github.com/facebookresearch/ParlAI/blob/dc0dd1b97f0ce010c4fb40e666255fea31621d21/tests/test_quickstart.sh#L17

We could make this faster:
https://github.com/facebookresearch/ParlAI/blob/dc0dd1b97f0ce010c4fb40e666255fea31621d21/tests/test_quickstart.sh#L16
with `-nl 1 --dict-tokenizer split --emb 32 --ffn-size 128`

young-k:
Could the `twitter` model be switched for a different one?

stephenroller:
Sure. If we do that we should update the corresponding quickstart.

vedantpuri:
@young-k Could you close this issue if completed ?

PR #2076, opened 2019-10-09
Suppress verbose tests and Make minor nits

vedantpuri:
**Patch description**
While working on another branch I realized that Circle CI tests were being verbose in some cases. This PR is to fix that and other stuff needed to do that:
- tfidf_retriever test had some issue related to logging which was resulting in it not properly suppressing output during the circle CI run, fixed that
- legacy versioning was also quite verbose, added a capture_output around that. I assumed here that we do not want any output in the tests and hence did that, let me know if I was wrong in doing so.
- A gpt-2 test was also spitting out output, so silenced that as well
- Since I was making a PR regarding these small nits, I thought might as well remove the comments you asked me not to be there @stephenroller in testing_utils and distributed_utils.

**Testing steps**
Run the unit tests. None of them should be verbose, i.e. all contain just the name of the test and an 'ok' once completed.

**Logs**
N/A

**Other information**
One thing that **still remains** that I found was that the quickstart test recently added by is also quite verbose. Refer to #2075 for further details.

N/A

### DIFF
--- a/tests/test_dict.py
+++ b/tests/test_dict.py
@@ -22,8 +22,9 @@ class TestDictionary(unittest.TestCase):
     """Basic tests on the built-in parlai Dictionary."""
 
     def test_gpt2_bpe_tokenize(self):
-        opt = Opt({'dict_tokenizer': 'gpt2', 'datapath': './data'})
-        agent = DictionaryAgent(opt)
+        with testing_utils.capture_output():
+            opt = Opt({'dict_tokenizer': 'gpt2', 'datapath': './data'})
+            agent = DictionaryAgent(opt)
         self.assertEqual(
             agent.gpt2_tokenize(u'Hello, ParlAI! 😀'),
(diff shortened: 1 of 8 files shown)


tests/test_dict.py
```diff
@@ -22,8 +22,9 @@ class TestDictionary(unittest.TestCase):
     """Basic tests on the built-in parlai Dictionary."""
 
     def test_gpt2_bpe_tokenize(self):
-        opt = Opt({'dict_tokenizer': 'gpt2', 'datapath': './data'})
-        agent = DictionaryAgent(opt)
+        with testing_utils.capture_output():
+            opt = Opt({'dict_tokenizer': 'gpt2', 'datapath': './data'})
+            agent = DictionaryAgent(opt)
```
stephenroller: huh, didn't expect that to be printing something but OKAY

PR #2077, opened 2019-10-09
Decrease runtime of quickstart test and move to nightly

young-k:
**Patch description**
As noted in #2075, `quickstart.sh` currently takes around [40 minutes](https://app.circleci.com/jobs/github/facebookresearch/ParlAI/14942) to run. 

This PR lowers the time to [20 minutes](https://app.circleci.com/jobs/github/facebookresearch/ParlAI/15206), a majority of which is accomplished by setting `num_epochs` to `0.01`.

**Testing steps**

CircleCI

**Logs**

N/A

**Other information**

The majority of time is now spent downloading the `twitter` dataset, so I think the main speedup would be from switching to a smaller dataset.

N/A

### DIFF
--- a/tests/test_quickstart.sh
+++ b/tests/test_quickstart.sh
@@ -13,7 +13,7 @@ python examples/display_model.py -t babi:task10k:1 -mf /tmp/babi_memnn -ecands v
 # train a transformer on twitter
 pip3 install emoji unidecode
 python examples/display_data.py -t twitter
-python examples/train_model.py -t twitter -mf /tmp/tr_twitter -m transformer/ranker -bs 10 -vtim 3600 -cands batch -ecands batch --data-parallel True --num-epochs 0.001 -nl 1 --dict-tokenizer split -emb fasttext --ffn-size 128
+python examples/train_model.py -t twitter -mf /tmp/tr_twitter -m transformer/ranker -bs 10 -vtim 3600 -cands batch -ecands batch --data-parallel True --max-train-time 20 -nl 1 --dict-tokenizer split -emb fasttext --ffn-size 128
 python examples/eval_model.py -t twitter -bs 50 -m legacy:seq2seq:0 -mf models:convai2/seq2seq/convai2_self_seq2seq_model --num-examples 1
 python examples/display_model.py -t twitter -mf /tmp/tr_twitter -ecands batch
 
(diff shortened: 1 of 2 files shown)


stephenroller:
instead of `--num-epochs .001` you can do `--max-train-time seconds`

tests/test_quickstart.sh
```diff
@@ -13,8 +13,8 @@ python examples/display_model.py -t babi:task10k:1 -mf /tmp/babi_memnn -ecands v
 # train a transformer on twitter
 pip3 install emoji unidecode
 python examples/display_data.py -t twitter
-python examples/train_model.py -t twitter -mf /tmp/tr_twitter -m transformer/ranker -bs 10 -vtim 3600 -cands batch -ecands batch --data-parallel True --num-epochs 0.01
-python examples/eval_model.py -t twitter -bs 50 -m legacy:seq2seq:0 -mf models:convai2/seq2seq/convai2_self_seq2seq_model
+python examples/train_model.py -t twitter -mf /tmp/tr_twitter -m transformer/ranker -bs 10 -vtim 3600 -cands batch -ecands batch --data-parallel True --max-train-time seconds 20 -nl 1 --dict-tokenizer split -emb fasttext --ffn-size 128
```
klshuster: by `seconds` stephen meant a number, e.g. `--max-train-time 20`

stephenroller:
It seems to spend an alarming amount of time downloading fasttext embeddings. We could add `-emb random` to drop that, which would help a lot.

Also it looks like it's just predicting unk unk unk. Can you verify that the results are reasonable?

stephenroller:
After you try axing that download, let's look at the runtime and then consider moving to nightly as y'all previously suggested.

young-k:
I think the `unk` tokens are expected, since that's the expected behavior as noted in the [quickstart](https://parl.ai/docs/tutorial_quick.html#add-a-simple-model).

young-k:
It's down to [11 minutes](https://app.circleci.com/jobs/github/facebookresearch/ParlAI/15607), does this suffice as enough of an improvement, or should it still be moved to a nightly test?

stephenroller:
Also i think this actually depends on the GPU tests being fixed from the pytorch 1.3 upgrade. @vedantpuri and @EricMichaelSmith should have a fix for that asap, so hold off until we know we can sort that out.

Make sure to add `[long]` to your commit message so that the nightly tests run.

Issue #2039, opened 2019-10-02
Proper versioning in requirements.txt


stephenroller:
Look into introducing proper versioning in `requirements.txt`. This is a step in installation, and an important step toward our milestone of proper versioning in ParlAI. See #2013 for an example of where this was problematic.

As a secondary step, this would enable us to use [circleci caching](https://circleci.com/docs/2.0/caching/) to speed up the runtime of unit tests. This should also be introduced.

One rough approach would be to get a shell on a circleci instance and run pip freeze. This should be done *without* the torch installation, unless we manage to resolve issues with #2010 such that `pip install torch` works well.

Assigning to @klshuster to delegate.

vedantpuri:
So I understand that we want to introduce versions in requirements.txt just like you did so recently for nltk. So when you mention pip freeze, is that a way to get all the required versions ? and then update requirements.txt with a ```>=``` of that version.
So running pip freeze would be a one time thing just to get the version right ? Slightly confused about that.
Looking into caching so that once setup.py is run every subsequent job uses the cached libraries, hope I understand that correctly.

stephenroller:
Pip freeze might be enough. It would need to happen from a clean machine and something about torch would need to be removed to ensure the user still can choose which version they want (cpu vs gpu) but this may not be as necessary anymore? Idk. This one needs someone to just dive in and figure out the complications I discovered 4 months ago that caused me not to do this.

vedantpuri:
@stephenroller Can this be closed or do we want to resolve torch versioning first?

PR #2078, opened 2019-10-10
Add versions to requirements and Implement dependency caching

vedantpuri:
**Patch description**
Add specific dependency versions to requirements.txt (used pip freeze on a fresh conda environment to get these numbers). Also implements dependency caching (Decreases dependency installation time by ~45x). Addresses #2039 

**Testing steps**
All tests should work as they did earlier

**Logs**
N/A

**Other information**
***Notes:***
* Currently I have implemented dependency caching only for tests that I was able to run in this pr, i.e. unittests_36,unittests_37, osx_tests, gpu11_tests and gpu12_tests.
* After several rounds of trying, it seemed like venv/lib was the right location since the libraries are loaded from there.
* Each of them have their own dependency cache. The reasoning is split into two sections:
   * For os_x tests, it doesn't workout if we save a cache on a linux machine and try to restore on os_x, the restoration just hangs, so we do need to save the cache on os_x and load it from os_x, hence a separate cache for os_x
   * For others, if we just have one cache key we get the following error:
   ```python: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory```. Probably because different builds are trying to access the same libraries and are needing access at the same time and we do not have libpython3.6m.so.1.0 installed on those images. It was easiest to have separate cache keys for each. I would say this is fine since the cache is just saved once and subsequent builds use that cache, but let me know if otherwise.

N/A


stephenroller:
Corroborating evidence. Notice the change in time for "installing dependencies"

before:
Screen Shot 2019-10-10 at 5 17 03 PM

After:
Screen Shot 2019-10-10 at 5 17 09 PM


My only concern is that you've done `>=` for all of the dependencies, but that means the cache will not catch new releases. What do you think we should do about that?

vedantpuri:
Yes, I saw the speedup and that has made me super excited about this PR!
For your concern:
I was thinking that if for ParlAI we later need to set a minimum threshold to be higher for a particular library , we would need to go change that in requirements.txt , which would in turn change the checksum that I have used in the cache key (checksum {{ requirements.txt }} and hence eventually create a new cache for the new requirements. Not sure if that answers your question/addresses your concern though.

Also, as I mentioned in the PR description, I haven't been able to do this for other tests like mturk and datatest etc. Is there any way I can test caching on those as well on this branch ? I don't want to implement caching there without testing before merging to master.

stephenroller:
But it wouldn’t detect if a new version of a dependency breaks parlai

stephenroller:
Also if you add [long] into any commit message the other tests will run

vedantpuri:
> But it wouldn’t detect if a new version of a dependency breaks parlai

Assuming changes are only made as PRs in this project, won't the breakage be detected in the PR. In other words, say we need to change the requirement for lib xyz from >=1.0.0 to  >=2.0.0 (in requirements.txt), but we realize that 2.0.0 breaks parlai. In the PR for this change, (the cache would contain 2.0.0 )some tests would have failed to show that changing to 2.0.0 didn't work out and we would only merge once that is fixed right ? 
If you're saying that within the PR too the master cache would be used, I am not sure about that but there is a way to include the branch name in the cache key which I can implement which might solve this.
Still not sure if we are on the same page, maybe I am missing something or not realizing something obvious, if possible could you give me a detailed example of a possible situation that may cause problems?

stephenroller:
I’m saying that because of the cache, it will behave differently relative to a user installing a fresh copy. If we specify “lib>=1.0” and leave it that way, and lib gets updated to 2.0, our cache won’t change (and our builds will be good), but new users will install 2.0 because 2.0>=1.0.

So we should either find more clever ways to validate the cache, or we should specify *exact* versions and manually upgrade everything ourselves.

vedantpuri:
Ahh! I see what you're saying. I would say specifying exact versions and manual updates should be fine since I suspect that requirement version changes are not made very frequently. I think that's better also because this way we would never have weird installation issues with any new user. 
I can work on making the versions exact and trying out caching in other tests this weekend, once we decide if that's what we want to do!

stephenroller:
That sounds good to me.

vedantpuri:
@stephenroller I think this seems to work, **except**, I am not sure if [this](https://circleci.com/gh/facebookresearch/ParlAI/15412) is the way nightly GPU should behave, i.e. In all containers except 2, the tests are skipped because they needed a GPU and probably couldn't find one (should that happen ?). This has happened earlier also in the master branch as you can see [here](https://circleci.com/gh/facebookresearch/ParlAI/15361?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

stephenroller:
Your PR looks good but I’m glad you noticed the GPU stuff. That looks broken and I have no idea why. Please file a new issue.

stephenroller:
Interestingly https://circleci.com/gh/facebookresearch/ParlAI/15195#tests/containers/4 looks fine

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ installtorchgpu12: &installtorchgpu12
   run:
     name: Install torch GPU and dependencies
     command: |
-      pip3 install --progress-bar off 'torch>=1.2.0'
+      pip3 install torch>=1.2.0+cu101 -f https://download.pytorch.org/whl/torch_stable.html
```
stephenroller: What is this? I don't know this syntax/option/etc.

vedantpuri: I extrapolated that command from here:
Screen Shot 2019-10-14 at 2 22 59 PM

stephenroller: Okay, let's use `torch==1.2.0+cu100` and verify nightly gpu runs fine. Then we can land this, and update #2080 to be about the upgrade to 1.3.0, and roping it in with the deprecate 1.0 issue.

vedantpuri: So I'm not sure what happened but something blew up with pip and now most tests are failing. I was trying a few things but I reset to a previous commit and now that's failing too. Could you help out a bit ?

vedantpuri: A couple of updates:

- It appears that a new version of pip was released today 2019-10-14 ([link](https://pip.pypa.io/en/stable/news/)) which somehow caused issues and went on to fail all tests using pip (eg. Commit id: 5c2ac44 (the commit before the comments) was working just fine. I tried changing a few things today and it seemed to broke so I reset to that id and yet it still broke).So in circle CI's config I have specified the previous version of pip for now! 

- I found the specific URL to install torch 1.3 with CUDA and have used it in my latest commit and it seems to be working.

vedantpuri:
@stephenroller 
I tried caching the installed drivers (multiple commits that have now been reset back and hence not visible) but there are many complications related to the binaries being installed and keys being added and hence was unable to implement driver caching properly. Perhaps that can be a further enhancement (in a new issue).
Besides that, I think the dependency caching works and I have rebased with master to include the GPU fix.

stephenroller:
I’d recommend just caching the wgets, not the installs. If it doesn’t work we can skip it this PR

stephenroller:
The nightly failure looks transient. Hitting rebuild.

vedantpuri:
> I’d recommend just caching the wgets, not the installs. If it doesn’t work we can skip it this PR

So I indeed did try caching the downloads (in commits that have been reset now) and turns out the downloading process is very minimal and results in a very minimal speedup, which I personally is not worth clutterring the code with restore_cache and save_cache statements. Most of the time is taken in uncompressing and installing if I recall correctly. I believe we should find a clever workaround for **that** in a separate issue. Let me know what you think!

Also, @klshuster would wait for your review too before merging!

Issue #2001, opened 2019-09-15
Create Agent Tutorial Command is Broken


klshuster:
**Bug description**
The command at the very end of the ["Creating an Agent"](https://parl.ai/docs/tutorial_seq2seq.html) yields the error below.

**Reproduction steps**
`python examples/train_model.py -t babi:task10k:1 --dict-file /tmp/dict_babi:task10k:1 -bs 32 -vtim 30 -vcut 0.95 -m example_seq2seq`

**Expected behavior**
The model should train accordingly.

**Logs**

```

$ python examples/train_model.py -t babi:task10k:1 --dict-file /tmp/dict_babi:task10k:1 -bs 32 -vtim 30 -vcut 0.95 -m example_seq2seq
...
Traceback (most recent call last):
  File "examples/train_model.py", line 16, in <module>
    TrainLoop(opt).train()
  File "/Users/kshuster/ParlAI/parlai/scripts/train_model.py", line 376, in __init__
    self.world = create_task(opt, self.agent)
  File "/Users/kshuster/ParlAI/parlai/core/worlds.py", line 1198, in create_task
    world = BatchWorld(opt, world)
  File "/Users/kshuster/ParlAI/parlai/core/worlds.py", line 662, in __init__
    shared = world.share()
  File "/Users/kshuster/ParlAI/parlai/core/worlds.py", line 141, in share
    shared_data['agents'] = self._share_agents()
  File "/Users/kshuster/ParlAI/parlai/core/worlds.py", line 153, in _share_agents
    shared_agents = [a.share() for a in self.agents]
  File "/Users/kshuster/ParlAI/parlai/core/worlds.py", line 153, in <listcomp>
    shared_agents = [a.share() for a in self.agents]
  File "/Users/kshuster/ParlAI/parlai/agents/example_seq2seq/example_seq2seq.py", line 191, in share
    shared = super().share()
  File "/Users/kshuster/ParlAI/parlai/core/torch_agent.py", line 1176, in share
    shared['model'] = self.model
AttributeError: 'ExampleSeq2seqAgent' object has no attribute 'model'
```

stephenroller:
I believe we agreed in the weekly meeting that we should just delete the tutorial. It's not helpful anymore.

stephenroller:
@young-k just make a PR deleting the tutorial, and the example_seq2seq folder please.

PR #2084, opened 2019-10-15
Remove Create Agent tutorial and all references

young-k:
**Patch description**
As discussed in #2001, the `Create Agent` tutorial is unnecessary. I removed the `example_seq2seq` folder, the tutorial, and references to the tutorial as well.

**Testing steps**
Unittests? Not certain what the best way to test this is!

**Logs**
N/A

**Other information**
N/A


stephenroller:
[docker is down](https://status.docker.com) so we can rerun this later and then land.

I'm not sure if we should take #2040 further to bring this back or not.

Issue #2080, opened 2019-10-12
Nightly GPU tests are broken


vedantpuri:
As discussed [here](https://github.com/facebookresearch/ParlAI/pull/2078#issuecomment-541279469) it seems like the nightly GPU tests are broken. It can be seen in [this build](https://circleci.com/gh/facebookresearch/ParlAI/15361?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) from master that the tests are skipped since a GPU could not be found.
The last successful build for these tests on master was [here](https://circleci.com/gh/facebookresearch/ParlAI/15195#tests/containers/4) where all tests used the GPU to run to completion

stephenroller:
It appears that the last successful one ran pytorch 1.2 and the first broken one ran pytorch 1.3. Though this may not be the problem, I estimate there's a high probability.

vedantpuri:
That's a good catch. I looked into the command that installs torch for nightly, it is the same as that for gpu12. Upon checking circle CI, I found that consequently gpu12 also doesn't run tests that requires the GPU. Maybe for this reason in ```installtorchgpu11``` the command is ```pip3 install --progress-bar off 'torch<1.2.0'```. Should I replicate this command in gpu12 ?

stephenroller:
Our policy is supposed to be “we support the last two versions of torch” so I want to figure out why 1.3 doesn’t work

vedantpuri:
Yes that makes sense. I updated the installtorchgpu12 command to :```pip3 install torch>=1.2.0+cu101 -f https://download.pytorch.org/whl/torch_stable.html``` and it seems to work as you can see [here](https://circleci.com/gh/facebookresearch/ParlAI/15432#config/containers/0). The odd thing is that in the section where torch is being installed, I see no stdout of such and was confused about that. Could you look into that and help me solve this confusion ?

PR #2085, opened 2019-10-15
Update NVIDIA drivers

vedantpuri:
**Patch description**
As mentioned in #2080 CUDA is not being properly installed with the existing ```installtorchgpu12``` command. This has occurred since the release of pytorch 1.3.

~~In this PR I have included a fix/workaround for that by specifying the exact link for the cuda installation file rather than just pip install torch. But maybe there is a better solution to that.~~

Further discussion on this PR revealed that stale GPU drivers maybe the problem and turns out that indeed that was the case. Updated the drivers to install the latest one available for a linux distribution (430.40) and looks like things are working as normal again.

 Looping in @EricMichaelSmith on this PR as requested by stephen.

**Testing steps**
gpu12 and nightly tests should work as expected

**Logs**
N/A

**Other information**
N/A

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,9 +99,9 @@ setupcuda: &setupcuda
     working_directory: ~/
     command: |
       # download and install nvidia drivers, cuda, etc
-      wget -q 'https://s3.amazonaws.com/ossci-linux/nvidia_driver/NVIDIA-Linux-x86_64-430.50.run'
-      sudo /bin/bash ./NVIDIA-Linux-x86_64-430.50.run -s --no-drm
-      wget -q https://developer.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-ubuntu1404-10-1-local-10.1.243-418.87.00_1.0-1_amd64.deb
+      wget -q 'https://s3.amazonaws.com/ossci-linux/nvidia_driver/NVIDIA-Linux-x86_64-430.40.run'
+      sudo /bin/bash ./NVIDIA-Linux-x86_64-430.40.run -s --no-drm
+      wget -q https://developer.download.nvidia.com/compute/cuda/10.1/Prod/local_installers/cuda-repo-ubuntu1404-10-1-local-10.1.243-418.87.00_1.0-1_amd64.deb
       sudo dpkg -i cuda-repo-ubuntu1404-10-1-local-10.1.243-418.87.00_1.0-1_amd64.deb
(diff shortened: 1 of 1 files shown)


stephenroller:
From an internal post:

> I am getting a cuda error on my devfair0225. 
> Yesterday afternoon I was running some scripts which used cuda but now `torch.cuda.is_available()` is returning False and I am getting errors that `No CUDA runtime is found` and also : 
> "The NVIDIA driver on your system is too old (found version 10000).
> Please update your GPU driver by downloading and installing a new
version from the URL: http://www.nvidia.com/Download/index.aspx
Alternatively, go to: https://pytorch.org to install
a PyTorch version that has been compiled with your version of the CUDA driver."

So maybe these are the problem?

https://github.com/facebookresearch/ParlAI/blob/969e5e773236ca03e729b4d24fab2b9da0927bb2/.circleci/config.yml#L102-L106

vedantpuri:
Would it be possible for you to let me know what to replace with what and maybe I can try that! I am not sure of specifications if any and don't want to do specify an incorrect URL.

stephenroller:
https://www.nvidia.com/en-us/drivers/unix/ looks like nvidia 430.50 is the latest stable so we could just change the 410.79 for the first command to those and cross our fingers.

If you go to [this page](https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&target_distro=Ubuntu&target_version=1404&target_type=deblocal) you can get the updated commands for the bottom 3

stephenroller:
The easiest way here though, is to go into the circleci GPU test (after logging into CircleCI using your github account) and click "rerun with ssh" and actually try out commands.

vedantpuri:
I am unable to rerun using SSH. I get the following error:
```
Failed to enable SSH
No SSH key is found. Please make sure you've added at least one SSH key in your VCS account.
```

Is there any SSH key I need to add and if so how.

Also, the build currently fails with exit code 8 (floating point error ?) not sure why

stephenroller:
You can read this tutorial on setting up ssh keys with your github account:

https://help.github.com/en/enterprise/2.15/user/articles/adding-a-new-ssh-key-to-your-github-account

vedantpuri:
After SSH-ing in and digging around the latest version available for linux seems to be 430.40 which I have used and looks like everything is working smoothly!

stephenroller:
(Also if you didn’t already, maybe you can cache these large downloads from Nvidia too)

PR #2086, opened 2019-10-16
[Bug FIx]Rouge metrics are computed but not reported.

dexterju:
**Patch description**
As stating in the title.
**Test**
Run any eval_model with --metrics all


stephenroller:
This was supposed to be done by the block below which iterates through everything in metrics_list.

stephenroller:
I still feel like the root cause is off, not the actual code below that is supposed to be doing this computation.

stephenroller:
Let's revisit this.

klshuster:
This was handled in #2116, closing

PR #2088, opened 2019-10-16
[Long] Update NLTK Version in Requirements

klshuster:
**Patch description**
Per https://nvd.nist.gov/vuln/detail/CVE-2019-14751, there is a vulnerability in NLTK versions lower than 3.4.5.

**Testing steps**
Relying on unit tests here


dexterju:
Thanks for doing this!

PR #2096, opened 2019-10-23
Adding Doxygen Doc for ParlAI : all graphs of class inheritance and dependencies added

liloup1789:
…ependencies added.

**Patch description**
I parsed the ParlAI repertory with doxygen in order to generate all the graphs and a "java" style doc, in my opinion helpfull to navigate through all the framework easily.

**Testing steps**
No code no test

**Logs**
```
```

**Other information**


liloup1789:
Hello : 
test_copyright (test_code.TestCopyright) ... FAIL

What should I do to pass this test ?

liloup1789:
Hello Stephen, 
Thanks you for your reply. 
I agree that it is a bit heavy to have two different docs. The advantage of doxygen was the graphs generation that I could not find in Sphinx. I found this : https://www.sphinx-doc.org/en/master/usage/extensions/inheritance.html I don't know if it is included in your doc... 

Thank you for the cprt information ! 

Best

stephenroller:
Hi @liloup1789, I don't think I can accept a 1.5 million LOC change, and I would still prefer we didn't have two documentation systems. If you wanted to add the sphinx extension I would like to see what the output looks like, but we can add the dependency if it's good.

liloup1789:
Hello,
Sorry it was an automatic pull request. I did not mean to. Won't happen again

Issue #1946, opened 2019-08-21
Change image_mode choice 'none' to 'no_image_model' across ParlAI


mwillwork:
Right now many teachers for image related tasks use image_mode == 'none', but this is bad practice. One should not use string-like versions of other types (i.e. None). We would like to change this to the constant 'no_image_model' and confirm that image teachers still work properly.

Run each image related task via the command at the top of the teacher if present or modify the documentation, with help from a team member, to insert a run command and then proceed.

domrigoglioso:
I tried running `python examples/train_model.py -t flickr30k -mf /tmp/flickr_test -m memnn` without any changes to test what the training of an image related task would look like and I get the following error traceback:
```
Traceback (most recent call last):
  File "examples/train_model.py", line 16, in <module>
    TrainLoop(opt).train()
  File "/home/dom/Documents/ParlAI/parlai/scripts/train_model.py", line 706, in train
    world.parley()
  File "/home/dom/Documents/ParlAI/parlai/core/worlds.py", line 285, in parley
    acts[1] = agents[1].act()
  File "/home/dom/Documents/ParlAI/parlai/core/torch_agent.py", line 1723, in act
    return self.batch_act([self.observation])[0]
  File "/home/dom/Documents/ParlAI/parlai/core/torch_agent.py", line 1747, in batch_act
    output = self.train_step(batch)
  File "/home/dom/Documents/ParlAI/parlai/core/torch_ranker_agent.py", line 342, in train_step
    else batch.image.size(0)
AttributeError: 'list' object has no attribute 'size'
```
I also tried the same command but with `-t mnist_qa` which runs fine, so I think it could just be an issue with flickr30k, although I haven't tried it on any other visual tasks yet.

PR #2112, opened 2019-10-29
Change image_mode from none to no_image_model

domrigoglioso:
**Patch description**
As in #1946, changes constant in teachers for image related tasks that use image_mode == 'none' to 'no_image_model'.

**Testing steps**
Ran on taks: flickr30k, mnist_qa, hopefully some vqa task later
Task: flickr30k
`python3 examples/display_data.py -t flickr30k` after changes, expected to see conversations. Did see conversations.
`python3 examples/train_model.py -t flickr30k -m memnn -mf /tmp/test_flickr30k --candidates fixed -vme 20 --image_mode no_image_model -vtim 200`
after changes, and same command but with `--image_mode none`. Expected outputs between both to be the same, got same output. 
Tried `--image_mode none` after changes, expected `ModuleNotFoundError: No module named 'torchvision'` (as that's what happens when you input models that don't exist)
 error and got the error.

Task: mnist_qa
 `python3 examples/train_model.py -t mnist_qa -m memnn -mf /tmp/test_mnistqa --candidates fixed -vme 20 --image_mode no_image_model -vtim 200 -dt train`
 Testing expectation similar as above.

**Logs**
(I can add some lines from the start of training / log of the error when there's an invalid image model later)

**Other information**
N/A

`python tests/datatests/test_new_tasks.py`
Got output:
```
There are 591753 missing_text in coco_caption:DefaultTeacher.
There are 82783 missing_text in coco_caption:V2014Teacher.
There are 591753 missing_text in coco_caption:V2017Teacher.
There are 29000 missing_text in flickr30k:DefaultTeacher.
```
and stopped execution early (it had been running for a few hours).


stephenroller:
Looks like the test needs to be updated

domrigoglioso:
@stephenroller 
When I was running `python tests/datatests/test_new_tasks.py` on my changes, I got a bunch of missing_text, like I added to the original description. Are those ok to have, since `flickr30k` also has 29000 missing_text before the changes, from running `verify_data` on that.

stephenroller:
@klshuster can you chime in.

klshuster:
Yeah that seems right since those are all captioning tasks so there is no task input (and those numbers pretty much correspond directly to the number of examples in the training sets). we should probably modify `verify_data` to have something like `'missing_text_and_image'` and check for that in each example rather than just a "text" field.

domrigoglioso:
Maybe? I haven't run verify_data on any tasks that use the YFCC100M dataset, but if this current amount of testing is ok then it's good to go.

stephenroller:
Yeah it looks good to me.

klshuster:
Agreed with Stephen, wouldn't worry about the YFCC100M-related datasets for now

PR #2113, opened 2019-10-29
Read classes from file in torch classifier agent

dgonzalez-ri:
**Patch description**
New flag to allow for the list of classes to be read from a file. Useful when the list is too long. It defaults to `None` and the value should be the path to a `.txt` file containing the classes, one class per line, without single/double quotes. Just like the following:

```
class_1
class_2
class_3
...
```

**Logs**
```
$ python examples/train_model.py -t empathetic_dialogues:emotionClassification -m bert_classifier --classes-from-file classes.txt -mf /tmp/bert_classification --validation-metric accuracy --validation-metric-mode max --dict-tokenizer bpe --dict-lower True -lr 3e-04 --shuffle True -bs 64 -veps 0.5 -vp 10 -vme 100 --lr-scheduler fixed --lr-scheduler-patience 3 --lr-scheduler-decay 0.3 --get-all-metrics False
```


parlai/core/torch_classifier_agent.py
```diff
@@ -81,19 +81,29 @@ def add_cmdline_args(parser):
...
+        )
 
     def __init__(self, opt, shared=None):
         init_model, self.is_finetune = self._get_init_model(opt, shared)
         super().__init__(opt, shared)
 
         # set up classes
-        if opt.get('classes') is None:
-            raise RuntimeError('Must specify --classes argument.')
+        if opt.get('classes') is None and opt.get('classes_from_file') is None:
```
emilydinan: nice!

parlai/core/torch_classifier_agent.py
```diff
@@ -81,19 +81,29 @@ def add_cmdline_args(parser):
...
+            raise RuntimeError('Must specify --classes or --classes-from-file argument.')
         if not shared:
-            self.class_list = opt['classes']
+            if opt["classes_from_file"] is not None:
```
emilydinan: nit: can you use single quotation marks here?

dgonzalez-ri: yes! done :)

PR #2117, opened 2019-10-30
Initial acute-eval open source commit

hadasah:
**Patch description**
Open sourcing ACUTE-EVAL 

worlds.py needs cleanup (as does run.py still) but putting out PR for comments


stephenroller:
That was a lot less code than it felt like at the time :D

hadasah:
I mean, I've deleted *many* lines over the many iterations lol.

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -0,0 +1,54 @@
...
+    The model names here should match the names of files in the folder
+'onboarding_tasks': List of tuples in format (id1, id2, name of matchup) where
+    id1, id2 are conversation ids where id2 is the correct conversation's id
+Help strings for the other arguments can be found in run.py
+"""
+
+
+def set_args():
+    args = add_args()
+    args['dialogs_path'] = '/home/ParlAI/parlai/mturk/tasks/acute_eval/example'
```
emilydinan: can we set this from the person's parlai home variable? it might not be in /home/ParlAI

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -0,0 +1,54 @@
+from parlai.mturk.tasks.pairwise_dialogue_eval.run import main as run_main, add_args
+
+"""
+Example script for running ACUTE-EVAL.
+The only arguments that need to be modified for this to be run are:
+'dialogs_path':  Path to folder containing logs of all model conversations.
```
emilydinan: might be a bit more natural if we made these arguments to `set_args` instead of having to hard code the values in this script

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+        elif os.path.isdir(full_fn):
+            for sfn in list_files(full_fn):
+                yield sfn
+
+
+def setup_task_queue(opt):
+    """ Initialize task queue to contain the specified number of instances of
+    each pairing
+    """
+    # hacky fix for the parlai parser hacky fix
```
emilydinan: hmm... why aren't the hyphens already replaced?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+    annotations_per_pair = opt['annotations_per_pair']
+    all_conv_data = {}
+    conv_ids_by_model = {}
+    internal_id = 0
+
+    # read in all conversation data
+    for data_fn in os.listdir(data_folder):
+        full_data_fn = os.path.join(data_folder, data_fn)
+        with open(full_data_fn, 'r') as dialog_data_file:
+            model_name = data_fn.split('.')[0]
```
emilydinan: so the data is expected to be in the data folder with specific file formatting? can we document this somewhere?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+                    subtask_data['task_specs']['internal_id']
+                )
+                workers_to_conversations_seen[worker_id].remove(
+                    subtask_data['conversations'][0]
+                )
+                workers_to_conversations_seen[worker_id].remove(
+                    subtask_data['conversations'][1]
+                )
+            except ValueError():
+                print("WARNING: couldn't remove task from worker's history")
```
emilydinan: when would this happen?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+def get_onboarding_tasks(worker_id, tasks_per_hit):
+    """ Get the next onboarding task for this worker id. If the worker has never
```
emilydinan: can you add the argument descriptors to your docstrings?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+        '--annotations_per_pair',
+        type=int,
+        default=1,
+        help='Number of annotations per conversation comparison pair',
+    )
+    argparser.add_argument(
+        '--pair_data',
+        type=list,
+        default=None,
+        help='list of (conv1, conv2, hit, desc) (for pre-chosen pairs, e.g. for replicating previous experiments)',
```
emilydinan: also, what is hit and desc ion this argument?

parlai/mturk/tasks/acute_eval/example_script.py
hadasah: this was only meant as an example - I expect people to copy this script and modify all of the args

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -0,0 +1,54 @@
...
+        * args['pairs_per_matchup']
+        / (args['task_description']['num_subtasks'] - 1)
+    )  # release enough hits to finish all annotations requested
+    args['block_qualification'] = 'onboarding_qual_name'
+    args['assignment_duration_in_seconds'] = 600
+    args['reward'] = 0.5
+    args['max_hits_per_worker'] = 1
+
+    # Additional args that can be set - here we show the default values.
+    # For a full list, refer to run.py & the parlai/params.py
```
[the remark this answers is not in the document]
hadasah: No, they don't need to be, but since I used all of the default values, there's no difference in the function. I can uncomment if that looks better.

parlai/mturk/tasks/acute_eval/example_script.py
hadasah: I set it up to force users to write a script because we have a lot of really long strings and lists in the arguments, and it would be really ugly to try to parse it from commandline, but open to suggestions

emilydinan: I meant you could set as function args in `set_args()` not as CLI args. that way you can just import this function and switch the args instead of having to re-write the body of the function every time.

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+
+    ## Fill task queue
+    for i in range(annotations_per_pair):
+        all_task_keys = list(desired_tasks.keys())
+        np.random.shuffle(all_task_keys)
+        for internal_id in all_task_keys:
+            task_queue.put(desired_tasks[internal_id])
+    # limit number of hits worker can do by default
+    if opt['max_hits_per_worker'] == 0:
+        opt['max_hits_per_worker'] = (len(desired_tasks) + len(onboarding_tasks)) / opt[
```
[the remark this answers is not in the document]
hadasah: yes, good catch.

parlai/mturk/tasks/acute_eval/run.py
hadasah: yes, I need to write a good README

parlai/mturk/tasks/acute_eval/run.py
hadasah: it wasn't at the time I wrote this - I think the hyphen replacing was happening before my args were being added, but then the underscores in my args got converted to hyphens - maybe it's been fixed since.

parlai/mturk/tasks/acute_eval/run.py
hadasah: in theory, never, and I never saw this triggered.  But apparently it's technically possible for weird race conditions to happen with MTurk, so I left some of these in just in case.

JackUrb:
Took a quick pass over the react, nothing seems out of place to me

parlai/mturk/tasks/acute_eval/run.py
emilydinan: what if instead we made this *not* dependent on the datafile name? shouldn't the model name be specified in the json anyway? can't we just provide some root dir to search through for json's with that model name?

hadasah: I would be ok with instead specifying a dict with model_name: file_name - is that sort of what you had in mind?

parlai/mturk/tasks/acute_eval/example_script.py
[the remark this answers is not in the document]
klshuster: yes please 😄

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -0,0 +1,54 @@
+from parlai.mturk.tasks.pairwise_dialogue_eval.run import main as run_main, add_args
```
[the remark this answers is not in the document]
klshuster: also a quick note that this is the reason the PR is failing all tests currently lol

parlai/mturk/tasks/acute_eval/example_script.py
mwillwork: Could this be another command line param? (And put in the README the format it expects the file to be in?) It would be nice also to specify what it does if you give it a directory versus what it does when you get it a file and have the default behavior do the right. In addition, one nice thing would be if someone just has the ability to implement a single function that parses their file. (So pull this out into its own function that can be overridden:  https://github.com/facebookresearch/ParlAI/blob/8a3018dc8e2f4f5f6992c2a7058c27fb267b1524/parlai/mturk/tasks/acute_eval/run.py#L163 ) The frontend expects certain fields to be there such as "dialog",  so the format should be specified to.

mwillwork:
A few suggestions:
1.  One nice thing would be if someone just has the ability to implement a single function that parses their file and then everything else just works. (So pull this out into its own function that can be overridden:  https://github.com/facebookresearch/ParlAI/blob/8a3018dc8e2f4f5f6992c2a7058c27fb267b1524/parlai/mturk/tasks/acute_eval/run.py#L163 ) The frontend expects certain fields to be there such as "dialog",  so the format should be specified to in documentation.
2. As discussed in person, for checking the onboarding tasks it would be nice if it didn't always assume the 2nd answer is correct. (i.e. have the ability to give an expected answer in the JSON.) 
3. I don't want to use any of my existing conversations as my onboarding tasks, so I'm going to put some custom JSON and create a function to build the onboarding tasks from that. Would be good to support this.
4. Can you put in the README that you are using "task" intentionally - and what a comparison is versus a pair versus a task versus a conversation versus a dialog. This would make the code easier to figure out.  
5. I think this functionality probably belongs in a class (call it ACUTEEvalBuilder?): you have a bunch of functions and some global variables, so basically this functionality all is related and shares state...hence class.
6. Would be good to have tests for this if it's going to be our standard eval method. I did a related MTurk task where I heavily modified this code (so it could have been my new code), but I somehow ended up doing more annotations per task than I wanted and it's not easy to figure out why. A test testing various flags, workers getting blocked, how the tasks are selected, incomplete tasks being returned to the task queue, replicating a previous run, etc would be great.

parlai/mturk/tasks/acute_eval/run.py
emilydinan: I was thinking more along the lines of: each JSON conversation has the key `model_name`, right? Can we just specify the root directory, search all conversations, and select by `model_name` key?

parlai/mturk/tasks/acute_eval/example_script.py
mwillwork: Could this be another command line param? (And put in the README the format it expects the file to be in?)

parlai/mturk/tasks/acute_eval/run.py
mwillwork: It would be nice to have default functionality that does the right thing if you give it a directory and also if you give it a file. In addition, one nice thing would be if someone just has the ability to implement a single function that parses their file. (So pull this out into its own function that can be overridden.) I have to parse a CSV file someone else gave me for example. The frontend expects certain fields to be there such as "dialog",  so the format should be specified too as per emily's comment above.

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+        mturk_manager.create_hits()
+
+        def check_worker_eligibility(worker):
+            return True
+
+        def assign_worker_roles(workers):
+            workers[0].id = display_agent_name
+
+        def run_conversation(mturk_manager, opt, workers):
+            task_data = get_new_task_data(workers[0], opt['comparisons_per_hit'])
```
emilydinan: `comparisons_per_hit` is not an argument above, so this causes an error. did you mean `subtasks` per hit?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+    task_data.extend([desired_tasks[id] for id in tasks_chosen])
+    return task_data
+
+
+def return_task_data(worker_id, task_data):
+    """ When worker doesn't complete a task, return it to the queue or
+    change their onboarding status depending on the task"""
+    for subtask_data in task_data:
+        if subtask_data['task_specs'].get('is_onboarding', False):
+            worker_onboarding_tasks_todo[worker_id].append(
```
emilydinan: this is not defined. i think you meant `workers_to_onboarding_tasks_todo`. causes an error.

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+def get_new_task_data(worker, tasks_per_hit):
+    """ Get next task for worker. Returns the next onboarding task if worker
+    hasn't finished them all, or finds a task from the queue they haven't seen
+    If they've seen everything in the queue, spin up an extra task (one that
+    was in the queue and is now saturated)
+    """
+    worker_id = worker.worker_id
+    task_data = get_onboarding_tasks(worker_id, tasks_per_hit)
+    if len(task_data) == tasks_per_hit:
+        return onboarding_tasks
```
emilydinan: this will also break here -- did you mean to return `task_data`

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+
+def set_args():
+    args = add_args()
+    args['dialogs_path'] = '/home/ParlAI/parlai/mturk/tasks/acute_eval/example'
+    args['model_comparisons'] = [('modela', 'modelb')]
+    args['onboarding_tasks'] = [('ZYX', 'AGHIJK', 'example_qual')]
+    args['task_description'] = {
+        'num_subtasks': 2,
```
emilydinan: i think this should be the existing arg, `subtasks_per_hit`, instead of `num_subtasks`.

emilydinan: EDIT: this is why i found that issue with the Optional feedback not submitting.

parlai/mturk/tasks/acute_eval/run.py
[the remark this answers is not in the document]
stephenroller: use // to avoid the float

parlai/mturk/tasks/acute_eval/frontend/components/custom.jsx
```diff
@@ -0,0 +1,514 @@
...
+              >
+              <div
+                className="container"
+                style={{'width': 'auto',}}>
+                <ControlLabel> {form_question} </ControlLabel>
+                <FormGroup>
+                    <Col sm={6}>
+                        <Radio
+                          name="speakerChoice"
+                          value="1"
```
stephenroller: it'd be better if these are the model names, so what we dump to disk later is more easily analyzed.

parlai/mturk/tasks/acute_eval/frontend/components/custom.jsx
```diff
@@ -0,0 +1,514 @@
...
+                          style={{'width': '100%'}}
+                          checked={this.state.speakerChoice == "1"}
```
stephenroller: this would have to be changed too

parlai/mturk/tasks/acute_eval/frontend/components/custom.jsx
```diff
@@ -0,0 +1,514 @@
...
+                          onChange={this.handleInputChange}
+                        >
+                          {choice1}
+                        </Radio>
+                      </Col>
+                      <Col sm={6}>
+                        <Radio
+                          name="speakerChoice"
+                          value="2"
```
stephenroller: ditto

parlai/mturk/tasks/acute_eval/frontend/components/custom.jsx
```diff
@@ -0,0 +1,514 @@
...
+                          style={{'width': '100%'}}
+                          checked={this.state.speakerChoice == "2"}
```
stephenroller: ditto

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+    tries = 0
+    completed_tasks = workers_tasks_completed.get(worker_id, [])
+    seen_conversations = workers_to_conversations_seen.get(worker_id, [])
+    while (not task_queue.empty()) and tries < task_queue.qsize():
+        try:
+            next_task = task_queue.get()
+        except Queue.Empty:
```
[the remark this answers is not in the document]
hadasah: queues technically aren't thread safe. In theory, you could have some weird cases where the queue is waiting for some items to be put back, or where qsize is still.

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+    id1,
+    id2,
+    internal_id,
+    all_conv_data,
+    s1_choice,
+    s2_choice,
+    question,
+    is_flipped,
+    hitid='',
+    matchup='regular',
```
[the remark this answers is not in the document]
hadasah: I don't really need the matchup field to exist at all, 'regular' was just a placeholder value. I was mostly using 'matchup' to mark qual comparisons

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+                opt['s1_choice'],
+                opt['s2_choice'],
+                opt['question'],
+                opt['correctness_is_flipped'],
+                matchup=matchup,
+                is_qual=True,
+            )
+            internal_id += 1
+    else:
+        raise NotImplementedError("--onboarding_tasks cannot be empty or None")
```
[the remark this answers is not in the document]
hadasah: changed this to warn_once

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,544 @@
...
+        '--block_on_onboarding_fail',
+        type=bool,
+        default=True,
+        help='whether to block on onboarding failure',
+    )
+    argparser.add_argument(
+        '--onboarding_tasks',
+        type=list,
+        default=None,
+        help='onboarding tasks to screen workers with, list of (conv1id, conv2id, matchup) tuples',
```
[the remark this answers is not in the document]
hadasah: matchup is actually never used for anything in run.py. I mostly use it to track qual annotations and what models are being compared. Not sure whether I want to keep matchup here but if I do, I'll make a note in the readme

parlai/mturk/tasks/acute_eval/example_script.py
hadasah: I see. I don't particularly have a problem with that, but it would be no different from what currently exists, just in a slightly different place. In that sense, I feel like it's actually cleaner to have it this way if you do want to set a lot of these args to custom (non-default) values.

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -0,0 +1,54 @@
...
+
+"""
+Example script for running ACUTE-EVAL.
+The only arguments that need to be modified for this to be run are:
+'dialogs_path':  Path to folder containing logs of all model conversations.
+    Each model should have one file named '<modelname>.jsonl'
+'model_comparisons': List of tuples indicating pairs of models to be compared.
+    The model names here should match the names of files in the folder
+'onboarding_tasks': List of tuples in format (id1, id2, name of matchup) where
+    id1, id2 are conversation ids where id2 is the correct conversation's id
```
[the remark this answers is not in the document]
hadasah: added something to run.py, will add something to readme as well

parlai/mturk/tasks/acute_eval/example_script.py
stephenroller: It actually is a CLI arg, Margaret chose to set it via a script.

parlai/mturk/tasks/acute_eval/README.md
```diff
@@ -0,0 +1,108 @@
...
+Margaret Li, Jason Weston, Stephen Roller.
+_[ACUTE-EVAL: Improved Dialogue Evaluation with Optimized Questions and Multi-turn Comparisons](https://arxiv.org/abs/1909.03087)_.
+
+## Citation
+
+If you use this evaluation method in your own work, please cite with the
+following BibTex entry:
+
+  @misc{li2019acuteeval,
+    title={ACUTE-EVAL: Improved Dialogue Evaluation with Optimized Questions and Multi-turn Comparisons},
```
stephenroller: Why not the workshop?

hadasah:
Addressing some comments:

I could potentially move the next button, but we iterated a few times to settle on the current UI, and I didn't see an unusual number of returns.

Pairing logic has been removed from run.py. I'll add a separate file that enables random pairing creation. The new run.py requires a jsonl files that explicitly defines one pairing per line.

The answer recorded is now the name of the model selected.

parlai/mturk/tasks/acute_eval/README.md
hadasah: It's not archival? Does it still have proceedings that can be cited?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,413 @@
...
+    argparser.add_parlai_data_path()
+    argparser.add_mturk_args()
+    argparser.add_argument(
+        '--annotations-per-pair',
+        type=int,
+        default=1,
+        help='Number of annotations per conversation comparison pair',
+    )
+    argparser.add_argument(
+        '--pairings-filename', type=str, default='pairings.jsonl', help=''
```
emilydinan: can you add a help string here? and elsewhere where it's missing

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def read_task_from_jsonl(
+    pairing_dict, internal_pair_id, s1_choice, s2_choice, question
+):
+    """
+    Build task dict according to expected format.
+    """
+    conv_orders = [[0, 1], [1, 0]]
+    conv_order = conv_orders[np.random.choice([0, 1])]
```
emilydinan: nit: can get rid of these two lines and just do `conv_order = random.choice([[0, 1], [1, 0]])`

emilydinan: alternatively, it seems you only really need a bool to track this conversation order (i.e. you only care about the first index)

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,413 @@
...
+        worker_response = response_data[i]['speakerChoice']
+        expected_response = all_task_data[i]['pairing_dict']['correct_answer']
+        num_onboarding_tasks += 1
+        if worker_response == expected_response:
+            num_correct += 1
+    if num_onboarding_tasks == 0:
+        if worker_id in onboarding_failed_workers:
+            return_task_data(worker_id, all_task_data)
+        return
+    if (num_correct / num_onboarding_tasks) >= threshold:
```
emilydinan: can you add some  more comments in the code explaining the flow through this funciton?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,413 @@
...
+        'question': opt['question'],
+    }
+    # append the contents of task_config.py to the configuration
+    opt['frontend_version'] = 1
+    try:
+        import sys
+
+        sys.path.append(opt['task_folder'])
+        from task_config import task_config
+    except Exception:
```
emilydinan: when does this happen? should we add a warning?

hadasah: We have a default task_config with the title, description, tags of the task to be shown on MTurk, but if the user writes one, I try to load it. I decided to go with having them write it in a .py file instead of .json (which would be easier to deal with) because I kept Jack's task_config.py template. There are a few lines on this in the README.
In other words, nothing is wrong if there's an exception, it just means the user didn't write a new task_config. But I should add a warning.

hadasah: We have a default task_config with the title, description, tags of the task to be shown on MTurk, but if the user writes one, I try to load it. I decided to go with having them write it in a .py file instead of .json (which would be easier to deal with) because I kept Jack's task_config.py template. There are a few lines on this in the README.
In other words, nothing is wrong if there's an exception, it just means the user didn't write a new task_config. But I should add a warning.

stephenroller:
We don't need to address all of those comments. we can chat irl about this

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,413 @@
...
+    task_data['task_specs'] = specs
+    task_data['pairing_dict'] = pairing_dict
+    specs['conversation_order'] = conv_order
+    specs['internal_pair_id'] = internal_pair_id
+    specs['s1_choice'] = s1_choice
+    specs['s2_choice'] = s2_choice
+    specs['question'] = question
+    if pairing_dict['is_onboarding']:
+        specs['is_onboarding'] = True
+        onboarding_tasks[internal_pair_id] = task_data
```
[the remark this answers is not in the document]
stephenroller: Let’s file a new task to follow up on this

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,413 @@
...
+    seen_conversations = workers_to_conversations_seen.get(worker_id, [])
+    while (not task_queue.empty()) and tries < task_queue.qsize():
+        try:
+            next_task = task_queue.get()
+        except Queue.Empty:
+            break
+        tries += 1
+
+        internal_pair_id = next_task['task_specs']['internal_pair_id']
+        dialogue0_id = next_task['pairing_dict']['dialogue_dicts'][0]['id']
```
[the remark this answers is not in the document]
stephenroller: In the follow up task

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -0,0 +1,413 @@
...
+    opt['task_description'] = {
+        'num_subtasks': opt['subtasks_per_hit'],
+        'question': opt['question'],
+    }
+    # append the contents of task_config.py to the configuration
+    opt['frontend_version'] = 1
+    try:
+        import sys
+
+        sys.path.append(opt['task_folder'])
```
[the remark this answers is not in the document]
stephenroller: Explain this plz

stephenroller:
Okay I explicitly marked follow ups with 👀 and never minds with 👎

parlai/mturk/tasks/acute_eval/run.py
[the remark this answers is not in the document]
hadasah: lol I'll just copy what I said to emily:

"We have a default task_config with the title, description, tags of the task to be shown on MTurk, but if the user writes one, I try to load it. I decided to go with having them write it in a .py file instead of .json (which would be easier to deal with) because I kept Jack's task_config.py template. There are a few lines on this in the README.
In other words, nothing is wrong if there's an exception, it just means the user didn't write a new task_config. But I should add a warning."

I guess this is super nonstandard though based on your reaction. Maybe I should switch to a .json task_config file.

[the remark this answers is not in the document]
hadasah: discussed offline, this is going into a command line arg

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -0,0 +1,66 @@
...
+    args['block_on_onboarding'] = True
+    args['block_qualification'] = 'onboarding_qual_name'
+
+    # general ParlAI mturk settings
+    args['assignment_duration_in_seconds'] = 600
+    args['reward'] = 0.5  # amount to pay workers per hit
+    args['max_hits_per_worker'] = 1  # max # hits a worker may complete
+    args['is_sandbox'] = True  # set to False to release real hits
+
+    args['annotations_per_pair'] = 1  # #times to use the same conversation pair
```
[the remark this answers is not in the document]
hadasah: it was meant to be #times as in "number of times" but yeah that looks confusing at a glance

parlai/mturk/tasks/acute_eval/frontend/components/custom.jsx
```diff
@@ -0,0 +1,576 @@
...
+      paddingRight: "20px",
+      paddingBottom: "20px",
+      overflowY: "scroll"
+    };
+
+    window.setTimeout(() => {
+      this.handleResize();
+    }, 10);
+
+    top_pane_style["height"] = this.state.chat_height + "px";
```
pringshia: any reason why this line isn't just part of the `top_pane_style` object above?

hadasah: I don't remember there being a good reason, I can probalby move it next PR

PR #2125, opened 2019-11-01
Auto-format docstrings.

stephenroller:
**Patch description**
This is the follow on to #2124, where I actually perform the autoformatting.

**Testing steps**
Codemod, nothing to test.

(Data tests skipped because this doesn't change behavior).


stephenroller:
Landing despite lint; these were not caused by these diffs.

PR #2126, opened 2019-11-01
Prevent -dt train in eval_model without evalmode.

stephenroller:
**Patch description**
Prevent `-dt train` in `eval_model.py` by raising an exception.

**Testing steps**
CircleCI new unit test.


dexterju:
What is that Codecov thing? interesting

PR #2127, opened 2019-11-01
[codecov] Set flags to track changes better.

stephenroller:
**Patch description**
From the [codecov docs](https://docs.codecov.io/docs/flags):
> Flags allow you to isolate coverage reports for different tests. This is particularly helpful if you have multiple types of tests (e.g., unit, integration, frontend, backend, etc), or you're employing a monorepo setup where you'd like to encapsulate each project's test coverage independently.

And from the CircleCI docs:
>  CIRCLE_JOB | String | The name of the current job.

**Testing steps**
Let's see what codecov says here!


vedantpuri:
Also, while we're at it, was thinking of adding the [codecov badge](https://codecov.io/gh/facebookresearch/ParlAI/settings/badge) to the README. Though since coverage is low at this point was not sure if it were a great idea to put it out there. What is your opinion?

stephenroller:
Yeah I would love the badge but it’s far too embarrassing right now 🙈😅

vedantpuri:
I see what you were trying, I myself tried using the include flag to accomplish that but upon researching SO, found that using source was the best way to track files that weren't hit. Include doesn't do that.
I wanted to go along the adding the __init__.py to the folders path but thought using source solved the issue. But I'd be interested to see if using that approach gives us something different. Also, could you fill me in on what our main objective is here ?

vedantpuri:
Also, should we be getting rid of [coverage.sh](https://github.com/facebookresearch/ParlAI/blob/d80573384feb589f4246bba39e6c2ec85cefb97b/tests/coverage.sh) ?

PR #2128, opened 2019-11-03
[TorchAgent] Add unks to metrics.

stephenroller:
**Patch description**
Start counting unks of labels. Seeking debate as to whether it would be more helpful if we did unks of text, or combined, and whether it should be a mean?

**Testing steps**
CircleCI.


stephenroller:
Just a diagnostic metric, makes it very obvious if you have any serious vectorization issues.

stephenroller:
To revisit after metrics rewrite.

PR #2129, opened 2019-11-04
Fix for printing rouge metrics

hadasah:
**Patch description**
--metrics all and --metrics rouge weren't printing rouge metrics, turns out metrics.report() iterates over self.metrics_list, which doesn't include 'rouge-1,' 'rouge-2,' etc, which is what we want. I store the metrics we want in another attribute.

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -256,12 +256,12 @@ class Metrics(object):
                 pass
             else:
                 self.metrics_list.add(each_m)
-        self.print_metrics_list = (
+        self._print_metrics_list = (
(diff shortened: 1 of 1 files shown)


parlai/core/metrics.py
```diff
@@ -256,12 +256,12 @@ def __init__(self, opt):
                 pass
             else:
                 self.metrics_list.add(each_m)
-        metrics_list = (
+        self.print_metrics_list = (
```
stephenroller: can you make it a private variable, `_print_metrics_list`?

PR #2132, opened 2019-11-05
Add initial WebSocket based chat service

ideanl:
**Patch description**
This PR is the initial implementation of a WebSocket based chat service. Much of this code is identical or very similar to the messenger chat service, which will be generalized in #2079. This service is simplified from the messenger service, and does not yet include display names, attachments (stickers, images, etc.), or quick replies.

**Testing steps**
Ran the service using `run.py` similar to the Messenger service and provided first the `chatbot` example task and then the `overworld_demo`. Then connected via the internal WebSocket based web-app and began chatting. Tested pairing in `overworld_demo` as well as singleplayer in `chatbot`.

Example server run command (for chatbot task): `./parlai/chat_service/services/websocket/run.py --config-path parlai/chat_service/tasks/chatbot/config.yml`
To customize the listening port, specify the `--port` flag (defaults to 35496).

Connect via WebSocket at `ws://<YOUR DOMAIN>:<PORT>/websocket`. Example websocket frontend JS:
```
var ws = new WebSocket("ws://localhost:35496/websocket");
ws.onmessage = function(msg) {
    console.log("Received Message: " + msg.data)
}
ws.send("Hey");
```
Sample output from above: 
```
Received Message: Welcome to the overworld for the ParlAI messenger chatbot demo. Please type "begin" to start.
```

Components that could be generalized in #2079:

**In agents.py:**
- `set_stored_data`
- `get_new_act_message`
- `act_blocking`
- `episode_done`
- `mark_inactive`
- `act`, `observe`, and `put_data` could be generalized with small changes such as ensuring messenger specific fields such as quick_replies are not required.

**In websocket_manager.py**:
- `MessengerWorldRunner` import
- `AgentState` import
- `shared_utils` import
- most of `parse_config` is identical, the rest of it can be overridden (such as page_id for messenger)
- `check_timeout_in_pool`
- most of `remove_agent_from_pool` with some post-removal done by messenger
- `add_agent_to_pool`
- `_expire_all_conversations`
- `_get_unique_pool`
- `get_agent_state`
- `_get_agent`
- part of `_complete_setup`
- `_load_model`
-  In, `start_task` the loop body and the `_done_callback` are mostly identical. However, the `while self.running` part is not identical because the WebSocket implementation uses a periodic tornado callback rather than a while loop. The following line is only relevant to messenger:
      `agents = [a for a in agents if a.disp_id is not None]`
- `_on_new_message`, except the message format would need to be consistent. Messenger also has `typing_on` and `handle_bot_read` calls which are irrelevant to WebSockets.
- `_log_debug`

### DIFF
--- a/parlai/chat_service/services/websocket/agents.py
+++ b/parlai/chat_service/services/websocket/agents.py
@@ -38,10 +38,7 @@ class WebsocketAgent(Agent):
         if 'payload' in act:
             raise ValueError("Payload not supported yet by websockets")
         else:
-            if act['id'] != '':
-                msg = '{}: {}'.format(act['id'], act['text'])
-            else:
-                msg = act['text']
+            msg = act['text']
             resp = self.manager.observe_message(
                 self.id,
                 msg,
--- a/parlai/chat_service/services/websocket/agents.py
+++ b/parlai/chat_service/services/websocket/agents.py
@@ -39,18 +39,12 @@ class WebsocketAgent(Agent):
             raise ValueError("Payload not supported yet by websockets")
         else:
             msg = act['text']
-            resp = self.manager.observe_message(
-                self.id,
-                msg,
-            )
+            self.manager.observe_message(self.id, msg)
 
     def put_data(self, message):
         """Put data into the message queue if it hasn't already been seen"""
         logging.info(f"Received new message: {message}")
-        action = {
-            'episode_done': False,
-            'text': message,
-        }
+        action = {'episode_done': False, 'text': message}
         self.msg_queue.put(action)
 
     def get_new_act_message(self):
@@ -85,7 +79,6 @@ class WebsocketAgent(Agent):
             if msg is not None and self.message_request_time is not None:
                 self.message_request_time = None
         return msg
-
 
     def act_blocking(self, timeout=None):
         """Repeatedly loop until we retrieve a message from the queue"""
--- a/parlai/chat_service/services/websocket/agents.py
+++ b/parlai/chat_service/services/websocket/agents.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import time
+from queue import Queue
+
+from parlai.core.agents import Agent
+
+
+class WebsocketAgent(Agent):
+    """Class for a person that can act in a ParlAI world via websockets"""
+
+    def __init__(self, opt, manager, task_id, socketID):
+        super().__init__(opt)
+        self.manager = manager
+        self.id = socketID
+
+        self.active = True
+        self.msg_queue = Queue()
+        self.stored_data = {}
+        self.set_stored_data()
+
+    def set_stored_data(self):
+        """Gets agent state data from manager"""
+        agent_state = self.manager.get_agent_state(self.id)
+        if agent_state is not None and hasattr(agent_state, 'stored_data'):
+            self.stored_data = agent_state.stored_data
--- a/parlai/chat_service/services/websocket/agents.py
+++ b/parlai/chat_service/services/websocket/agents.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import time
+import logging
 from queue import Queue
 
 from parlai.core.agents import Agent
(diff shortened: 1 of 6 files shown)


parlai/chat_service/services/websocket/agents.py
```diff
@@ -0,0 +1,94 @@
...
+        """Send an agent a message through the websocket manager"""
+        logging.info(f"Sending new message: {act}")
+        if 'payload' in act:
+            raise ValueError("Payload not supported yet by websockets")
+        else:
+            msg = act['text']
+            self.manager.observe_message(self.id, msg)
+
+    def put_data(self, message):
+        """Put data into the message queue if it hasn't already been seen"""
```
klshuster: should we be checking that the message has not already been seen?

parlai/chat_service/services/websocket/agents.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        # if this is the first act since last sent message start timing
+        if self.message_request_time is None:
+            self.message_request_time = time.time()
+
+        # If checking timeouts
+        if timeout:
+            # If time is exceeded, timeout
+            if time.time() - self.message_request_time > timeout:
+                return self.mark_inactive()
```
klshuster: should this be defined somewhere?

parlai/chat_service/services/websocket/sockets.py
```diff
@@ -0,0 +1,44 @@
...
+    return str(uuid.uuid4())
+
+
+class MessageSocketHandler(WebSocketHandler):
+    subs = {}
+
+    def __init__(self, *args, **kwargs):
+        self.message_callback = kwargs.pop('message_callback', None)
+        self.sid = get_rand_id()
+        super(MessageSocketHandler, self).__init__(*args, **kwargs)
```
klshuster: nit: can just call `super().__init__...`

parlai/chat_service/services/websocket/sockets.py
```diff
@@ -0,0 +1,44 @@
...
+            logging.info(f"Current subscribers: {self.subs}")
+
+    def on_close(self):
+        """Runs when a socket is closed"""
+        del self.subs[self.sid]
+
+    def on_message(self, message):
+        """Callback that runs when a new message is received from a client"""
+        logging.info('websocket message from client: {}'.format(message))
+        self.message_callback(message, self.sid)
```
klshuster: will this fail if there is no `self.message_callback`? should we provide a default one?

parlai/chat_service/services/websocket/websocket_manager.py
```diff
@@ -0,0 +1,458 @@
...
+
+# TODO: Use a generalized World Runner module (Issue #2079)
+from parlai.chat_service.services.messenger.world_runner import MessengerWorldRunner
+import parlai.chat_service.services.messenger.shared_utils as shared_utils
+from parlai.chat_service.services.websocket.sockets import MessageSocketHandler
+from agents import WebsocketAgent
+import tornado
+from tornado.options import options
+
+PORT = 35496
```
klshuster: just a question - is this an arbitrary number?

parlai/chat_service/services/websocket/agents.py
ideanl: Updated the docs to reflect that we don't do the check. 

`get_new_act_message` removes the message from the queue and the `on_new_message` callback for WebSockets is only be called once per message, so duplicate messages shouldn't be an issue.

parlai/chat_service/services/websocket/agents.py
ideanl: Defined it as a `pass` as in the messenger implementation. Included this in the PR description as a function that can be generalized out.

parlai/chat_service/services/websocket/sockets.py
ideanl: Defined a default that outputs a warning

parlai/chat_service/services/websocket/websocket_manager.py
ideanl: Yeah arbitrary default port number that can be specified via `--port`. I found this port number in other uses of WebSockets internally

ideanl:
Updated the description with more information about running/using the service. Also included more details regarding what can be generalized w.r.t the messenger implementation (for #2079).

parlai/chat_service/services/websocket/websocket_manager.py
ideanl: Removed the constant from the manager and instead defined it as a default in the argument parser

PR #2135, opened 2019-11-06
Fixed a pre-attention bug for seq2seq

ShaojieJiang:
**Patch description**
Previously when using `pre` attention for seq2seq, the hidden state for calculating attention is not updated in each step.

**Testing steps**
Tested on DailyDialog, all metrics look slightly better.

**Logs**
Not applicable

**Other information**
No other information

Not applicable


stephenroller:
I'm a little confused. I'm slightly worried this patch just makes it post attention, which I do expect to be better...

@alexholdenmiller can you look at this?

stephenroller:
@ShaojieJiang ultimately I'm not comfortable altering legacy behavior like this, especially something that radically changes the semantics of the operation. If you defined a new kind of attention (instead of `pre`, `post`, you'd have like `hidden_pre` or something) which implements the behavior you're interested in.

It's worth noting that your implementation is very similar to post attention, which is generally more standard I believe.

stephenroller:
Closing due to inactivity. Feel free to reopen.

PR #2146, opened 2019-11-10
dialogue unlikelihood project link

jaseweston:


lemon234071:
Thank you for your efforts. When will the code be released?

stephenroller:
The code is currently living in #2791. It has a few broken tests and lint issues that are preventing it from being merged.

lemon234071:
> The code is currently living in #2791. It has a few broken tests and lint issues that are preventing it from being merged.

Thank you very much!

PR #2150, opened 2019-11-11
Make websockets chat service use the new chat service base classes

ideanl:
**Patch description**
This PR changes the websockets chat service to use the newly made base classes for chat services. The message format has been changed to conform to the base class requirement of having a dict with a sender key, which is now the case. `_parse_config` and `_complete_setup` are no longer being called from init in the base chat service manager vs. the messenger implementation, so I had to call them when overriding init.

**Testing steps**
See testing steps in #2132

### DIFF
--- a/parlai/chat_service/services/websocket/websocket_manager.py
+++ b/parlai/chat_service/services/websocket/websocket_manager.py
@@ -16,12 +16,8 @@ import threading
 import traceback
 import sys
 from parlai.core.agents import create_agent
+from parlai.chat_service.core.chat_service_manager import ChatServiceManager
 
-# TODO: Use generalized module (Issue #2079)
-from parlai.chat_service.services.messenger.messenger_manager import AgentState
-
-# TODO: Use a generalized World Runner module (Issue #2079)
-from parlai.chat_service.services.messenger.world_runner import MessengerWorldRunner
 import parlai.chat_service.services.messenger.shared_utils as shared_utils
 from parlai.chat_service.services.websocket.sockets import MessageSocketHandler
 from agents import WebsocketAgent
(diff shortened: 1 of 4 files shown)


parlai/core/params.py
```diff
@@ -458,6 +458,13 @@ def add_websockets_args(self):
         websockets.add_argument(
             '--port', default=35496, type=int, help='Port to run the websocket handler'
         )
+        websockets.add_argument(
+            '--password',
```
klshuster: i'm not sure if it's there yet, but this should be eventually wrapped up in some `add_chat_service_args` function

parlai/chat_service/services/websocket/websocket_manager.py
```diff
@@ -9,234 +9,68 @@
...
+        self.agent_id_to_overworld_future = {}
+        self.task_to_agent_ids = {}
         self._load_model()
 
     def _load_model(self):
         """Load model if necessary"""
         if 'model_file' in self.opt or 'model' in self.opt:
             self.opt['shared_bot_params'] = create_agent(self.opt).share()
 
+    def _handle_message_read(self, event):
```
emilydinan: what is this for? can you add a docstring?

parlai/chat_service/services/websocket/websocket_manager.py
```diff
@@ -418,7 +191,7 @@ def _create_agent(self, task_id, socketID):
         :param agent_id:
             int agent id
         """
-        return WebsocketAgent(self.opt, self, task_id, socketID)
+        return WebsocketAgent(self.opt, self, socketID, task_id)
```
emilydinan: ooo tricky that we swapped this

parlai/chat_service/services/websocket/websocket_manager.py
```diff
@@ -451,6 +224,17 @@ def observe_message(self, socket_id, text):
         asyncio.set_event_loop(asyncio.new_event_loop())
         return MessageSocketHandler.subs[socket_id].write_message(text)
 
-    def _log_debug(self, text):
-        time = datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')
-        shared_utils.print_and_log(logging.DEBUG, f'{time}: {text}', should_print=True)
+    def restructure_message():
```
emilydinan: can we add docstrings here too?

parlai/chat_service/services/websocket/websocket_manager.py
ideanl: Added a docstring

parlai/chat_service/services/websocket/websocket_manager.py
ideanl: Yeah I had it in the wrong order before the base class was created. Now it's consistent with the other agents

parlai/chat_service/services/websocket/websocket_manager.py
ideanl: Added a docstring

parlai/core/params.py
ideanl: Created the function and put common args between the services in it

PR #2155, opened 2019-11-14
[WoW] Add apprentice only teacher.

emilydinan:
**Patch description**
Add a flag to allow only to train on apprentice utterances as labels for wizard of wikipedia.

To test:
```
python examples/display_data.py -t wizard_of_wikipedia:basicdialog --speaker-label apprentice
```

### DIFF
--- a/parlai/tasks/wizard_of_wikipedia/agents.py
+++ b/parlai/tasks/wizard_of_wikipedia/agents.py
@@ -362,12 +362,9 @@ class BasicdialogTeacher(WizardOfWikipediaTeacher):
 
     def len_episode(self, ep):
         d = self.data[ep]
-        if self.speaker_label != 'both':
-            first_speaker = d['dialog'][0]['speaker']
-            if self.speaker_label == 'wizard' and 'Wizard' in first_speaker:
-                return (len(d['dialog']) - 1) // 2
-            elif self.speaker_label == 'apprentice' and 'Apprentice' in first_speaker:
-                return (len(d['dialog']) - 1) // 2
+        first_speaker = d['dialog'][0]['speaker'].lower()
+        if self.speaker_label != 'both' and self.speaker_label in first_speaker:
+            return (len(d['dialog']) - 1) // 2
         return len(d['dialog']) // 2
 
     def get(self, episode_idx, entry_idx=0):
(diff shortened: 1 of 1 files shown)


parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -344,33 +344,41 @@ class BasicdialogTeacher(WizardOfWikipediaTeacher):
...
     def num_examples(self):
         return self.num_exs
 
     def len_episode(self, ep):
         d = self.data[ep]
-        if self.wizard_dialog and ('Wizard' in d['dialog'][0]['speaker']):
-            return (len(d['dialog']) - 1) // 2
+        if self.speaker_label != 'both':
+            first_speaker = d['dialog'][0]['speaker']
+            if self.speaker_label == 'wizard' and 'Wizard' in first_speaker:
```
klshuster: nit: since these two conditions do the same thing I feel like we could probably combine them more elegantly but it's also not gonna look too great if we just do "A and B or C and D" so up to you

parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -344,33 +344,41 @@ class BasicdialogTeacher(WizardOfWikipediaTeacher):
...
         return len(d['dialog']) // 2
 
     def get(self, episode_idx, entry_idx=0):
         d = self.data[episode_idx]
         episode_done = entry_idx == (self.len_episode(episode_idx) - 1)
 
         idx = entry_idx * 2
-        if self.wizard_dialog and ('Wizard' in d['dialog'][0]['speaker']):
+        first_speaker = d['dialog'][0]['speaker']
+        if self.speaker_label == 'wizard' and 'Wizard' in first_speaker:
```
klshuster: ditto

parlai/tasks/wizard_of_wikipedia/agents.py
emilydinan: yeah I thought the line was gonna be too long and it would look bad lol

emilydinan: although...maybe not, i'll try

stephenroller:
y'all wouldn't consider making a new teacher instead of a flag, would you? :D

emilydinan:
> y'all wouldn't consider making a new teacher instead of a flag, would you? :D

do you feel that this is better?

Issue #1877, opened 2019-07-26
Specialized add_cmdline_args calls for recommended hyperparameters


stephenroller:
We have a *lot* of command line parameters in ParlAI. We also have a lot of institutional knowledge of what are good recommended values, but we don't usually write them down anywhere for the public.

We should hack ParlaiParser to additionally accept a `recommended` parameter as part of its options. It should take a string as input. The `--help` string should then automatically print these recommended values for all the arguments. Something like `(default: 5, recommended: yadyayda)`.

Refer to:

https://github.com/facebookresearch/ParlAI/blob/48187b7aaacea5f910719074fe78d13c409e6776/parlai/core/params.py#L1014

stephenroller:
@klshuster will begin suggesting some initial values for some agent.

vedantpuri:
Any tips on how to go about this ?

stephenroller:
It will need to be supported by `ParlaiParser.add_argument`, and maybe for now it will just do something like store it in the help message.

PR #2158, opened 2019-11-16
Hack add_argument to allow recommender parameter

vedantpuri:
**Patch description**
As mentioned in #1877 this PR attempts to hack ParlaiParser in order to allow us to add a recommended value for certain tasks. This PR essentially allows one to specify recommended values which will be appended to the help message!
This is what I interpreted of the task and thought of approaching it! Let me know if you think there is a better way for this and I'd be happy to change it!

**Testing steps**
Run
```bash
python tests/test_params.py
```

**Logs**
```
...
Ran 3 tests in 0.113s

OK
```

**Other information**
NA

NA


parlai/core/params.py
```diff
@@ -1071,8 +1071,16 @@ def _handle_hidden_args(self, kwargs):
                 kwargs['help'] = argparse.SUPPRESS
         return kwargs
 
+    def add_recommended_to_help(self, kwargs):
+        """Add recommended value to help string if recommended exists"""
+        if 'recommended' in kwargs:
+            kwargs['help'] += " Recommended: " + kwargs['recommended']
```
stephenroller: Make sure you handle the case of help doesn’t exist. I would also prefer recommended to be able to take any type so do a format here please

vedantpuri: I had actually added that but I assumed that every arg would have a help kw and removed that just before committing, I'll re-add it!
For the formatting, yes good point, will casting as a str suffice ?

vedantpuri:
Just pushed some changes you had requested. Not sure what is a clean way of testing this. Once I create a parser object and try adding some arguments, how do I access those individual arguments and assert their values ?

parlai/core/params.py
```diff
@@ -1071,8 +1071,16 @@ def _handle_hidden_args(self, kwargs):
                 kwargs['help'] = argparse.SUPPRESS
         return kwargs
 
+    def add_recommended_to_help(self, kwargs):
+        """Add recommended value to help string if recommended exists"""
+        if 'help' in kwargs and 'recommended' in kwargs:
+            kwargs['help'] += " Recommended: " + str(kwargs['recommended'])
```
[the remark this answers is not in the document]
stephenroller: I think a key thing we’ll probably want is a range of values too, but this could be done as a set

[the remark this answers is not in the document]
stephenroller: ie recommended {1e-4, 1e-3}

stephenroller:
I’ll add to it. Transformer has a default of —variant aiayn  as default but recommended should be xlm. And gelu for activation

vedantpuri:
So @klshuster it turns out that that our ```CustomHelpFormatter``` already includes the default values in the help message when printing it, so me specifically trying to add that creates a double default message displayed: eg.
Screen Shot 2019-11-18 at 6 11 57 PM
Hence I won't be adding default to the help message, and would be following a similar format for recommended, i.e. adding it within parenthesis as default is shown.

vedantpuri:
@stephenroller @klshuster I did try adding a test but that doesn't seem like the best way to test it, any suggestions ?
If you think the current way is ok, I can test other added recommended values as well.

parlai/core/params.py
```diff
@@ -1071,8 +1071,17 @@ def _handle_hidden_args(self, kwargs):
                 kwargs['help'] = argparse.SUPPRESS
         return kwargs
 
+    def augment_help_msg(self, kwargs):
```
stephenroller: please make a `_private_method`

tests/test_parsing.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestRecommended(unittest.TestCase):
+    """Basic tests on the recommended values of parser."""
+
+    def test_help(self):
+        """Test help message of parser"""
+        parser = setup_args()
+        parser.set_defaults(
+            task='integration_tests',
+            model='transformer/polyencoder',
```
stephenroller: So this thing has a lot of dependencies, and we really only need to test one small behavior. So rather than an integration test, we'd prefer a smaller unit test.

So probably better would be to just:
- create a parlai parser
- add an argument with a recommended value
- add an argument group, and in that add an argument with a recommended value
- then do the format_help stuff like what you have below. The string code you've written is a LITTLE messy, maybe we can find to have a better matcher? A regex might be the way to go. With some flag, you can have them support newlines. something like `r'--my-option[^\n]*\n[^\n]*(recommended: myrecommendation)'`

I'm also pretty sure we have a `test_params.py` which is where this test can really go

stephenroller: buried deep in ArgumentParser there might be a way to get the help message for just one option too, which would also work.

parlai/agents/transformer/transformer.py
```diff
@@ -71,6 +70,7 @@ def add_common_cmdline_args(argparser):
         choices={'aiayn', 'xlm'},
         default='aiayn',
         help='Chooses locations of layer norms, etc.',
+        recommended='xlm',
     )
     argparser.add_argument(
         '--activation',
```
stephenroller: this is the other one i was pointing to

tests/test_parsing.py
vedantpuri: Tried looking for it could't find one, the current check(one I just pushed) seems concise and neat, what do you think

parlai/agents/transformer/transformer.py
```diff
@@ -3,7 +3,6 @@
 # LICENSE file in the root directory of this source tree.
 """Transformer Agents."""
 from parlai.core.agents import Agent
-from parlai.utils.misc import warn_once
```
klshuster: is this no longer needed in this file?

vedantpuri: Yeah in a previous commit in this PR, ```lint_changed.sh``` threw an error that it wasn't being used, so I got rid of it!

PR #2159, opened 2019-11-17
Implement local terminal chat service

vedantpuri:
**Patch description**
Subtask of #2079 (Moved to new issue: #2207) . In this PR we(@domrigoglioso and I) attempt to implement a simple chat service that one can run on their terminal. 

**Testing steps**
1. Run: `python parlai/chat_service/services/terminal_chat/run.py --config-path parlai/chat_service/tasks/chatbot/config.yml`
2. Run: `python client.py`
3. Interact through client.
Also repeated the above with the argument `--port 10001` in each command, and with two clients at once.

**Logs**
N/A

**Other information**
N/A

N/A


parlai/chat_service/services/terminal_chat/playground/server.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import socket
```
stephenroller: I don't like that all of this is living as just global state in a module. It seems more appropriate to write this into a class with like a serve_forever() function. You can look at python's builtin `http.server` for inspiration.

vedantpuri: @stephenroller Thanks for your suggestion and feedback we will look into it. However, this code was just for me and my teammate to experiment and learn about how the socket communication works and hence is essentially "rough-work". As I mentioned in the README in this dir, it's just a toy example and would be removed before this PR is merged.

vedantpuri:
A couple of notes/questions for @domrigoglioso and @klshuster:
- After looking up the answers above, I felt it might be easier to just extend websocket and use it locally for our terminal chat, and I was able to accomplish a working version in my latest commit with minimal added code! Let me know what you think (there are some minor issues though discussed in the following points)

- Currently, one has to run the server (```run.py``` file) and the client(```client.py```), was wondering if there was a way to send the former to the background and have only the latter running, was running into some issues regarding threading and would appreciate advice on that!

- One issue currently is that the server itself doesn't shut down if the client side has exited, so we need some fix for that I guess. I am assuming this would persist in the original websocket chat implementation from which we directly extend.

- Currently in ```client.py``` the port number is hard-coded, might be a good idea to derive it as a variable after the manager etc. has been instantiated.

Display related:
- Currently, the sleep time is also hardcoded(to wait for the bot's reply), was hoping to get some advice to have a signal type system where I display the next line only after the bot has replied.

- I had added some flair to the visual by following what I did in ```convo_render.py``` but an issue is that the blue background is sometimes hard to control with the input, specially when the screen scrolls.

To experience this first hand:
- STEP 1: RUN(from root dir): 
```python parlai/chat_service/services/terminal_chat/run.py --config-path parlai/chat_service/tasks/chatbot/config.yml```
- STEP 2: RUN(from parlai/chat_service/services/terminal_chat/): 
```python client.py```
- STEP 3: Interact

For reference here is a screenshot of the working terminal bot:
Screen Shot 2019-11-28 at 10 19 29 PM

File: parlai/chat_service/services/terminal_chat/changes_to_core.txt [hunk 2]
```diff
@@ -0,0 +1,5 @@
+Core Agents.py:
```
klshuster: what is this file?

parlai/chat_service/services/terminal_chat/client.py
```diff
@@ -0,0 +1,78 @@
...
+import uuid
+import websocket
+
+try:
+    import thread
+except ImportError:
+    import _thread as thread
+import time
+
+# STEP 1: RUN: python parlai/chat_service/services/terminal_chat/run.py --config-path parlai/chat_service/tasks/chatbot/config.yml
```
klshuster: could these instructions go in a `README.md` file?

parlai/chat_service/services/terminal_chat/terminal_manager.py
```diff
@@ -0,0 +1,20 @@
...
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from parlai.chat_service.services.websocket.websocket_manager import WebsocketManager
+from parlai.chat_service.core.chat_service_manager import ChatServiceManager
+
+
+class TerminalManager(WebsocketManager):
+    class TerminalMessageSender(ChatServiceManager.ChatServiceMessageSender):
```
klshuster: does this work if we just `pass` and not even include these definitions?

parlai/chat_service/services/terminal_chat/changes_to_core.txt
domrigoglioso: This is a file to keep of track of things we should change in the `chat_services/core` folder, the file will be deleted before merging. So for example, in `chat_services/core/agents.py`, we there should be a `self.message_partners` class variable but currently there isn't.

klshuster: ok, please indicate when this is ready for full review

vedantpuri:
@domrigoglioso Had a couple of concerns before we submit for full review. Mentioning them here as notes:
- I am assuming you have tested the new variable port implementation 
- One thing I didn’t test was having one server and multiple clients connecting to it, so maybe you could test that 
- Minor nit: instead of add_terminal_args maybe name it to add_terminal_chat args (otherwise sounds similar to cmdline args). And maybe add comments about the :param: 
- Documentation: I think we should document the funcitons (:param: :return: and description format), classes (one docstring)  and files (one doc string describing a high level view).

- Terminal chat readme: Could reformat it to:
```
  # Terminal Chat
 ** some info what this is and motivation and how we implemented (specify we extend from websocket thanks to @ideanl’s implementation) **  
## Setup 
steps to set it up 
```
- We should get rid of this in main chat service README, since we didn’t fill it in
￼
Image

domrigoglioso:
@klshuster It's ready for full review

parlai/chat_service/services/terminal_chat/client.py
```diff
@@ -0,0 +1,113 @@
...
+
+import json
+import uuid
+import websocket
+from parlai.core.params import ParlaiParser
+
+try:
+    import thread
+except ImportError:
+    import _thread as thread
```
stephenroller: what is this? can you leave a comment?

parlai/chat_service/services/terminal_chat/client.py
```diff
@@ -0,0 +1,113 @@
...
+
+def on_open(ws):
+    """
+    Starts a new thread that loops, taking user input and sending it to the websocket.
+
+    :param ws: websocket.WebSocketApp that sends messages to a terminal_manager
+    """
+    id = get_rand_id()
+
+    def run(*args):
```
stephenroller: i'm not a huge fan of closures, would you be willing to refactor to a global helper function like `_run`, or a class altogether?

parlai/chat_service/services/terminal_chat/client.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def setup_args():
+    """
+    Set up args, specifically for the port number.
+
+    :return: A parser that parses the port from commandline arguments.
+    """
+    parser = ParlaiParser(False, False)
+    parser.add_terminal_chat_args(is_client=True)
```
stephenroller: Does this really need to be a global variable? I feel like this is local to just this script, and so the args should be as so.

parlai/chat_service/services/terminal_chat/run.py
```diff
@@ -0,0 +1,50 @@
...
+
+def run(opt):
+    """
+    Run TerminalManager.
+    """
+    opt['service'] = SERVICE_NAME
+    manager = TerminalManager(opt)
+    try:
+        manager.start_task()
+    except BaseException:
```
stephenroller: please drop the catchall, you only need try/finally

parlai/chat_service/services/terminal_chat/run.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python3
```
stephenroller: i'm unclear why this script exists and how it differs from the client? is it the server? Can you document it as such?

parlai/chat_service/services/terminal_chat/run.py
```diff
@@ -0,0 +1,50 @@
...
+
+def setup_args():
+    """
+    Set up args.
+
+    :return: A parser that takes in command line arguments for chat services (debug, config-path, password), and a port.
+    """
+    parser = ParlaiParser(False, False)
+    parser.add_parlai_data_path()
+    parser.add_terminal_chat_args()
```
stephenroller: again i feel better about not leaking these args into global parlai parser

parlai/chat_service/services/terminal_chat/terminal_manager.py
```diff
@@ -0,0 +1,16 @@
...
+from parlai.chat_service.services.websocket.websocket_manager import WebsocketManager
+
+
+class TerminalManager(WebsocketManager):
+    """
+    Chat Service manager that runs chat service tasks using terminal to send and receive
+    messages.
+    """
+
+    pass
```
stephenroller: lol wow really?

parlai/chat_service/services/terminal_chat/README.md
```diff
@@ -0,0 +1,18 @@
+## Terminal Chat
+This allows you to participate in a ParlAI world as an agent using the terminal.
+This extends the `websocket` chat service implementation to run a server locally,
+which you can send and receive messages from using the terminal.
+
+## Setup
+1. Run:
+
+ `python parlai/chat_service/services/terminal_chat/run.py --config-path path/to/config.yml --port PORT_NUMBER`
```
stephenroller: maybe server_run.py instead?

What is the config yaml file?

parlai/chat_service/services/terminal_chat/client.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def get_rand_id():
+    """
+    :return: The string of a random id using uuid4
+    """
+    return str(uuid.uuid4())
+
+
+def prBlueBG(text):
```
stephenroller: make these and the other helper methods `_private_methods`

stephenroller:
(also i know it's exam season so don't take my comments too seriously)

parlai/chat_service/services/terminal_chat/run.py
domrigoglioso: `chat_services/services/websocket/run.py` does a similar thing but with `parser.add_websocket_args()` so we just followed that. Should we just be using `add_chatservice_args` and getting the port number some other way?

parlai/chat_service/services/terminal_chat/README.md
klshuster: this is the new way of running these services - it's where all the info is specified (rather than having a separate `run.py` file for each task). It's generally been the convention that `run.py` starts the server

parlai/chat_service/services/terminal_chat/client.py
```diff
@@ -33,43 +29,83 @@ def prBlueBG(text):
...
+    parser.add_terminal_chat_args(is_client=True)
+    return parser.parse_args()
 
 
 if __name__ == "__main__":
+    opt = setup_args()
+    port = opt.get('port', 34596)
     ws = websocket.WebSocketApp(
-        "ws://localhost:35496/websocket",
+        "ws://localhost:{}/websocket".format(port),
```
klshuster: just a clarification is this the same port that the server uses?

parlai/chat_service/services/terminal_chat/run.py
klshuster: @stephenroller is saying you can add args locally here rather than define the function `add_terminal_chat_args` here, i suppose we could have done that as well with the websockets

parlai/chat_service/services/terminal_chat/README.md
stephenroller: that's fine; where is an example config file? It's not provided in the documentation here.

why did we go with yaml? Now we've got json and yaml and pickle?

parlai/chat_service/services/terminal_chat/client.py
domrigoglioso: Yeah it should be. It's whatever port you pass into `--port`, and you'll get an error if there's no server running on the port you pass.

parlai/chat_service/services/terminal_chat/README.md
klshuster: `yaml` is something we had been using internally, there is no "reason" specifically we're using yaml, it looks kinda nice 😄 

Also there is an example config in line 13 of this file

stephenroller: alright good enough for me

parlai/chat_service/services/terminal_chat/client.py
```diff
@@ -96,13 +97,17 @@ def setup_args():
...
+    parser_grp.add_argument(
+        '--port', default=35496, type=int, help='Port to run the terminal chat server'
+    )
     return parser.parse_args()
 
 
 if __name__ == "__main__":
     opt = setup_args()
     port = opt.get('port', 34596)
+    print("port IS: ", port)
```
klshuster: nit: if we're going to print, could we make it a little more descriptive?

vedantpuri: Just changed that! Was using this for debugging but I feel indeed having the port printed might be useful. Not sure why circle CI didn't run on the latest commit thoug!

klshuster:
are there any pending changes? or can we merge this?

vedantpuri:
@klshuster We were just waiting on @stephenroller to also approve. He said he would review it in a few days. Otherwise I think we are good to go

klshuster:
Nah we don't need @stephenroller i'll go ahead and merge

PR #2171, opened 2019-11-18
Add support for images to WebSockets chat service

ideanl:
**Patch description**
This PR adds support for images to the WebSockets chat service. Given this, there are changes to the message structure while attempting to follow the chat service message structure defined in the `chat_service` README as close as possible. Additionally, this adds support for a `module_initialize` function in the worlds file.

The `text` field in the messages is always a string of text, which makes it easy to compare the text in a message to a string (or to support a single message with text + image in the future).
The `attachment` field is for images, where the following fields are expected:
- type (the only supported type right now is `image`)
- text  (For images, this is the base 64 encoded image)
- mime_type (This is the Mime type for the images, this is necessary for an HTML UI to render an image correctly)

WebSocket messages are now stringified JSON objects that conform to the message structure above.

**Testing steps**
See #2132 for running a websocket server for testing. Messages must be sent as stringifed JSON now using the structure above. To send images, provide the `attachment` object with a base64 encoded image for the text.

Example text message:
'{"text": "Example text message"}'
Example image message:
'{"attachment": {"type": "image", "text": "<IMAGE BASE64>", "mime_type": "image/jpg"}}'

### DIFF
--- a/parlai/chat_service/core/agents.py
+++ b/parlai/chat_service/core/agents.py
@@ -10,7 +10,7 @@ from queue import Queue
 from parlai.core.agents import Agent
 
 
-class ChatServiceAgent(ABC, Agent):
+class ChatServiceAgent(Agent, ABC):
     """Base class for a person on a chat serivce that can act in a ParlAI world"""
 
     def __init__(self, opt, manager, receiver_id, task_id):
--- a/parlai/chat_service/core/agents.py
+++ b/parlai/chat_service/core/agents.py
@@ -14,6 +14,7 @@ class ChatServiceAgent(Agent, ABC):
     """Base class for a person on a chat serivce that can act in a ParlAI world"""
 
     def __init__(self, opt, manager, receiver_id, task_id):
+        self.opt = opt
         super().__init__(opt)
         self.manager = manager
         self.id = receiver_id
--- a/parlai/chat_service/core/agents.py
+++ b/parlai/chat_service/core/agents.py
@@ -14,7 +14,6 @@ class ChatServiceAgent(Agent, ABC):
     """Base class for a person on a chat serivce that can act in a ParlAI world"""
 
     def __init__(self, opt, manager, receiver_id, task_id):
-        self.opt = opt
         super().__init__(opt)
         self.manager = manager
         self.id = receiver_id
--- a/parlai/chat_service/core/chat_service_manager.py
+++ b/parlai/chat_service/core/chat_service_manager.py
@@ -316,6 +316,12 @@ class ChatServiceManager(ABC):
             message to put on queue
         """
         agent_id = message['sender']['id']
+        if not self.world_runner.initialized:
+            self.observe_message(
+                agent_id, 'Please wait while the worlds are initializing...'
+            )
+            return
+
         if agent_id not in self.messenger_agent_states:
             self._on_first_message(message)
             return
--- a/parlai/chat_service/core/chat_service_manager.py
+++ b/parlai/chat_service/core/chat_service_manager.py
@@ -310,6 +310,18 @@ class ChatServiceManager(ABC):
         future = self.world_runner.launch_overworld(
             task_id, self.overworld, self.onboard_map, agent
         )
+
+        def _done_callback(fut):
+            self.observe_message(agent_id, 'See you later!')
+            for world_type in self.agent_pool:
+                agent_state = self.get_agent_state(agent_id)
+                if agent_state in self.agent_pool[world_type]:
+                    self.agent_pool[world_type].remove(agent_state)
+                    self.remove_agent_from_pool(agent_state, world_type=world_type)
+            del self.messenger_agent_states[agent_id]
+            del self.agent_id_to_overworld_future[agent_id]
+
+        future.add_done_callback(_done_callback)
         self.agent_id_to_overworld_future[agent_id] = future
 
     def _on_new_message(self, message):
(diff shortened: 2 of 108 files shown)


parlai/chat_service/core/agents.py
```diff
@@ -10,7 +10,7 @@
 from parlai.core.agents import Agent
 
 
-class ChatServiceAgent(ABC, Agent):
+class ChatServiceAgent(Agent, ABC):
```
klshuster: is this convention? or an actual bug?

parlai/chat_service/core/chat_service_manager.py
```diff
@@ -316,6 +319,12 @@ def _on_new_message(self, message):
             message to put on queue
         """
         agent_id = message['sender']['id']
+        if not self.world_runner.initialized:
```
klshuster: could we abstract this to a function `self.world_runner.is_initialized()`? so that we avoid accessing this prop directly

parlai/chat_service/services/websocket/agents.py
```diff
@@ -5,29 +5,85 @@
...
+                specifies whether or not the attachment is an image. If the
+                attachment is an image, either a `path` key must be specified
+                for the path to the image, or a `data` key holding a PIL Image.
+                A `quick_replies` key can be provided with a list of string quick
+                replies for any message
+        """
         logging.info(f"Sending new message: {act}")
-        if 'payload' in act:
-            raise ValueError("Payload not supported yet by websockets")
+        quick_replies = act.get('quick_replies', None)
```
klshuster: do quick replies make sense in the context of websockets?

parlai/chat_service/core/chat_service_manager.py
ideanl: Done

parlai/chat_service/core/agents.py
ideanl: I did this as convention since that order determines the priority of inherited methods (not that it'll actually make a difference right now)

parlai/chat_service/services/websocket/agents.py
ideanl: I included quick_replies in the message structure primarily since some UI's may want to implement a form of quick replies. It can be ignored by the WebSocket client or handled in any way so it's just an extra feature of the messages. It also allows for a more uniform protocol between various chat services so tasks can be agnostic of whether the service supports quick_replies.

parlai/chat_service/services/websocket/agents.py
```diff
@@ -5,29 +5,85 @@
...
+            attachment = act['attachment']
+            if attachment['type'] == 'image':
+                if 'path' in attachment:
+                    image = Image.open(attachment['path'])
+                    msg = self._get_message_from_image(image)
+                elif 'data' in attachment:
+                    msg = self._get_message_from_image(attachment['data'])
+                else:
+                    raise ValueError("Invalid attachment format for type 'image'")
+            else:
```
emilydinan: instead of putting everything in an if/else block here (since we have several nested "ifs") can we just assert that `attachment['type'] == 'image'` above?

parlai/chat_service/services/websocket/agents.py
```diff
@@ -5,29 +5,85 @@
...
+                for the path to the image, or a `data` key holding a PIL Image.
+                A `quick_replies` key can be provided with a list of string quick
+                replies for any message
+        """
         logging.info(f"Sending new message: {act}")
-        if 'payload' in act:
-            raise ValueError("Payload not supported yet by websockets")
+        quick_replies = act.get('quick_replies', None)
+        if act.get('attachment', None) is not None:
+            attachment = act['attachment']
```
emilydinan: nit: might be cleaner if we put this in some sort of helper function like `_get_attachment(act)`

parlai/chat_service/services/websocket/websocket_manager.py
```diff
@@ -107,6 +108,16 @@ def _done_callback(fut):
                 agent_state = self.get_agent_state(agent.id)
                 agent_state.set_active_agent(agent_state.get_overworld_agent())
 
+        # Monitor for crashed overworld futures
+        for agent_id, overworld_fut in self.agent_id_to_overworld_future.items():
+            if overworld_fut.done():
+                try:
```
emilydinan: nit: this logic is a bit vague here. can we just check that the future didn't finish with an exception by checking `overworld_fut.exception()` directly?

parlai/chat_service/core/chat_service_manager.py
```diff
@@ -159,8 +160,10 @@ def _parse_config(self, opt):
         self.task_configs = self.config['configs']
         self.max_workers = self.config['max_workers']
         self.opt['task'] = self.config['task_name']
+        # Deepcopy the opts so the manager opts aren't changed by the world runner
```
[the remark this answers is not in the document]
ideanl: When the world runner calls `module_initialize` on a worlds file, it may add objects into the manager opts which get deep copied whenever a new agent is created. This causes infinite recursion if one of the objects added to the opts stores a reference to the same opts (ran into this issue and found this solution used internally)

parlai/chat_service/services/websocket/websocket_manager.py
ideanl: I added a done callback to the overworld futures that checks for exceptions and handles completed overworlds, so this block of code is no longer needed

PR #2183, opened 2019-11-19
Fix a few lint errors

stephenroller:
**Patch description**
Small stuff that made it through at some point.

### DIFF
--- a/parlai/agents/bert_classifier/bert_classifier.py
+++ b/parlai/agents/bert_classifier/bert_classifier.py
@@ -12,7 +12,6 @@ from parlai.agents.bert_ranker.helpers import (
     get_bert_optimizer,
     MODEL_PATH,
 )
-from parlai.utils.misc import load_opt_file
 from parlai.core.torch_agent import History
 from parlai.core.torch_classifier_agent import TorchClassifierAgent
 from parlai.utils.misc import warn_once
--- a/parlai/agents/bert_ranker/cross_encoder_ranker.py
+++ b/parlai/agents/bert_ranker/cross_encoder_ranker.py
@@ -109,11 +109,9 @@ class CrossEncoderRankerAgent(TorchRankerAgent):
             .contiguous()
             .view(nb_cands * size_batch, -1)
         )
-        segments_context = tokens_context * 0
 
         # remove the start token ["CLS"] from candidates
         tokens_cands = cand_vecs.view(nb_cands * size_batch, -1)
-        segments_cands = tokens_cands * 0 + 1
         all_tokens, all_segments = concat_without_padding(
             tokens_context, tokens_cands, self.use_cuda, self.NULL_IDX
         )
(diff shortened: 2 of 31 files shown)


stephenroller:
```
There are 166787 missing_labels in wizard_of_wikipedia:WizardOfWikipediaTeacher.
F
FAIL: test_verify_data (__main__.TestNewTasks)
  File "tests/datatests/test_new_tasks.py", line 70, in test_verify_data
    self.assertFalse(found_errors, "Errors were found.")
AssertionError: True is not false : Errors were found.

Ran 1 test in 756.354s

FAILED (failures=1)
```

Looks like wiz fails but it seems intentional? I'm thinking we might want to make teachers that start with a `_` immune to the test and then rename this teacher. Also file an issue.

EricMichaelSmith:
> ```
> There are 166787 missing_labels in wizard_of_wikipedia:WizardOfWikipediaTeacher.
> F
> ======================================================================
> FAIL: test_verify_data (__main__.TestNewTasks)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "tests/datatests/test_new_tasks.py", line 70, in test_verify_data
>     self.assertFalse(found_errors, "Errors were found.")
> AssertionError: True is not false : Errors were found.
> 
> ----------------------------------------------------------------------
> Ran 1 test in 756.354s
> 
> FAILED (failures=1)
> python tests/datatests/test_new_tasks.py
> ```
> 
> Looks like wiz fails but it seems intentional? I'm thinking we might want to make teachers that start with a `_` immune to the test and then rename this teacher. Also file an issue.

The failure here is expected because the base `WizardOfWikipediaTeacher` doesn't return a label field, right? Yeah, I'm definitely in favor of some kind of signifier to prevent this kind of teacher from being checked by this unit test. A `_` might work, but I'm not sure a priori if it's always the case that an incomplete teacher == a private teacher

stephenroller:
> The failure here is expected because the base `WizardOfWikipediaTeacher` doesn't return a label field, right? Yeah, I'm definitely in favor of some kind of signifier to prevent this kind of teacher from being checked by this unit test. A `_` might work, but I'm not sure a priori if it's always the case that an incomplete teacher == a private teacher

Maybe we should discuss? It seems that an incomplete teacher also shouldn't be used by models etc, so marking it private would be trying to avoid that.

EricMichaelSmith:
> > The failure here is expected because the base `WizardOfWikipediaTeacher` doesn't return a label field, right? Yeah, I'm definitely in favor of some kind of signifier to prevent this kind of teacher from being checked by this unit test. A `_` might work, but I'm not sure a priori if it's always the case that an incomplete teacher == a private teacher
> 
> Maybe we should discuss? It seems that an incomplete teacher also shouldn't be used by models etc, so marking it private would be trying to avoid that.

Yeah, I can't immediately think of any cases in which we'd *need* to use an incomplete teacher, so I'm fine making the implicit equivalence between "private teacher" and "teacher we don't run this test for"

PR #2190, opened 2019-11-19
wizard: basic dialog can have topic

jaseweston:


parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -382,19 +388,62 @@ def get(self, episode_idx, entry_idx=0):
...
+            default=True,
+            help='prepend chosen topic to first turn',
+        )
+
+
+class BasicApprenticeDialogTeacher(BasicdialogTeacher):
+    @staticmethod
+    def add_cmdline_args(argparser):
+        agent = argparser.add_argument_group('Basic Dialog Arguments')
+        agent.add_argument(
```
emilydinan: hmm, instead of changing the argument defaults (which is a bit vague/hidden), can we just override the init and hard code `self.speaker_label = 'apprentice'`?

jaseweston: ok, done

jaseweston:
> are you going to add the fix for "doubling" the data by adding the `__SILENCE__` token as a dialogue turn here?

no i'm not.  feel free to do it, but it's not a priority, as this isnt the real teacher we want to use anyway -- it has no knowledge and no label_candidates.
we really want to put the apprentice possibility into the normal teachers (for generation and retrieval.. actually why do we have different teachers for generation and retrieval anyway?)

parlai/tasks/wizard_of_wikipedia/agents.py
jaseweston: ping

PR #2149, opened 2019-11-11
Refactors messenger to use base classes

jp848:
**Patch description**
Without changing the functionality, changes the messenger in `chat_services` to use the implemented base classes. 

**Testing steps**
`python run.py --config-path ../../tasks/chatbot/config.yml`

**Logs**
```
[ optional arguments: ] 
[  datapath: /Users/juuhnpark/ParlAI/data ]
[ Facebook Messenger: ] 
[  bypass_server_setup: False ]
[  config_path: ../../tasks/chatbot/config.yml ]
[  force_page_token: False ]
[  is_debug: False ]
[  local: False ]
[  log_level: 20 ]
[  password: None ]
[  verbose: False ]
[ Current ParlAI commit: 5b89068dea1c367ecba196d3218b81ada3e1ec86 ]

You are going to allow people on Facebook to be agents in ParlAI.
During this process, Internet connection is required, and you should turn off your computer's auto-sleep feature.

Please press Enter to continue... 

Setting up Messenger webhook...
Heroku: Collecting files...
Heroku: Starting server...
 ›   Warning: token will expire 11/09/2020
 ›   Use heroku authorizations:create to generate a long-term token
Enabling http-session-affinity for juuhnpark-parlai-messenger-cha... done
Scaling dynos... done, now running web at 1:Free
Webhook address: https://juuhnpark-parlai-messenger-cha.herokuapp.com/webhook
Local: Setting up WebSocket...
done with websocket
[ warning: overriding opt['model'] to legacy:seq2seq:0 (previously: seq2seq )]
[ Loading existing model params from /Users/juuhnpark/ParlAI/data/models/convai2/seq2seq/convai2_self_seq2seq_model ]
Dictionary: loading dictionary from /Users/juuhnpark/ParlAI/data/models/convai2/seq2seq/convai2_self_seq2seq_model.dict
[ num words =  18803 ]
/anaconda3/lib/python3.7/site-packages/torch/nn/_reduction.py:46: UserWarning: size_average and reduce args will be deprecated, please use reduction='sum' instead.
  warnings.warn(warning.format(ret))

```

PR #2201, opened 2019-11-21
Fix running Messenger chat service locally

ideanl:
**Patch description**
This PR fixes the ability to run the FB Messenger chat service locally using `--local`. This functionality was previously not fully working because the port could not be customized and SSL certificates could not be provided (which Messenger requires).

If the `--local` flag is provided and the environment variable `LOCAL_HTTPS=true`, this change will accept SSL certificates in the `~/.ssl` directory under the names `privkey.pem`, `fullchain.pem`, and `cert.pem`. The `https://localhost` hostname is also no longer hardcoded, as this is customizable through the prompts anyways so removing it only improves flexibility. This is useful since trusted certs will not be for the `localhost` domain.

In addition, there are some minor changes to the WS message structure to improve its uniformity with the Messenger message structure.

The changes to the MessengerManager are primarily copied from the ChatServiceManager, while #2149 refactors the manager to use the base class.

**Testing steps**
After getting trusted SSL certificates, run `LOCAL_HTTPS=true python parlai/chat_service/services/messenger/run.py --local --config-path parlai/chat_service/tasks/chatbot/config.yml` and ensure the local server successfully works.

**Screenshots**
Screen Shot 2019-11-21 at 10 54 55 AM
Screen Shot 2019-11-21 at 10 55 26 AM

### DIFF
--- a/parlai/chat_service/services/messenger/message_socket.py
+++ b/parlai/chat_service/services/messenger/message_socket.py
@@ -134,7 +134,7 @@ class MessageSocket:
             url_base_name = self.server_url.split('https://')[1]
             while self.keep_running:
                 try:
-                    sock_addr = "ws://{}/".format(url_base_name)
+                    sock_addr = "wss://{}/".format(url_base_name)
                     self.ws = websocket.WebSocketApp(
                         sock_addr,
                         on_message=on_message,
--- a/parlai/chat_service/services/websocket/agents.py
+++ b/parlai/chat_service/services/websocket/agents.py
@@ -39,37 +39,10 @@ class WebsocketAgent(ChatServiceAgent):
         """
         logging.info(f"Sending new message: {act}")
         quick_replies = act.get('quick_replies', None)
-        attachment_msg = self._get_attachment_msg(act)
-        if attachment_msg is not None:
-            self.manager.observe_payload(self.id, attachment_msg, quick_replies)
+        if act.get('payload', None):
+            self.manager.observe_payload(self.id, act['payload'], quick_replies)
         else:
             self.manager.observe_message(self.id, act['text'], quick_replies)
-
-    def _get_attachment_msg(self, act):
-        """
-        Gets the message for an attachment given an act
-
-        Args:
-            act: dict. See `observe` for the structure of this dict
-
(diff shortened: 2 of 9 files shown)


parlai/chat_service/services/messenger/message_socket.py
```diff
@@ -134,7 +134,7 @@ def run_socket(*args):
             url_base_name = self.server_url.split('https://')[1]
             while self.keep_running:
                 try:
-                    sock_addr = "ws://{}/".format(url_base_name)
+                    sock_addr = "wss://{}/".format(url_base_name)
```
klshuster: what is this change exactly?

parlai/chat_service/services/websocket/agents.py
```diff
@@ -39,38 +39,11 @@ def observe(self, act):
...
         quick_replies = act.get('quick_replies', None)
-        attachment_msg = self._get_attachment_msg(act)
-        if attachment_msg is not None:
-            self.manager.observe_payload(self.id, attachment_msg, quick_replies)
+        if act.get('payload', None):
+            self.manager.observe_payload(self.id, act['payload'], quick_replies)
         else:
             self.manager.observe_message(self.id, act['text'], quick_replies)
 
-    def _get_attachment_msg(self, act):
```
klshuster: what's the rationale for removing this?

parlai/chat_service/services/messenger/message_socket.py
ideanl: This change is to use WebSockets with SSL. Makes sense for all scenarios since Messenger requires SSL already for the webhooks to work.

parlai/chat_service/services/websocket/agents.py
ideanl: Removed this because rather than using the `attachment` key with either a `path` or `data` argument, we now accept a `payload` key with `data` and expect that the caller has loaded a base 64 encoded image in `data`.
Basically, shifting the responsibility so that the caller has to perform this logic which makes the payload more consistent with Messenger.
I just updated the documentation for `observe` to reflect the change.

PR #2210, opened 2019-11-22
Move copyright check away from unit tests.

stephenroller:
**Patch description**
Move the copyright check out of the unit tests, so that we don't see some otherwise strange failures.

### DIFF
--- a/tests/check_copyright.py
+++ b/tests/check_copyright.py
@@ -9,10 +9,7 @@ import re
 import parlai.utils.testing as testing_utils
 
 FILENAME_EXTENSIONS = r'.*\.(rst|py|sh|js|css)$'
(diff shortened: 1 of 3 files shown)


tests/check_copyright.py
```diff
@@ -0,0 +1,55 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import sys
+import re
+import parlai.utils.testing as testing_utils
+
+FILENAME_EXTENSIONS = r'.*\.(rst|py|sh|js|css)$'
+WHITELIST_PHRASES = [
+    'Moscow Institute of Physics and Technology.',
+    'https://github.com/fartashf/vsepp',
```
klshuster: might not need this anymore since we deleted vsepp agent

PR #2211, opened 2019-11-22
Self chat general script + Wizard self chat example world V2

jaseweston:


parlai/core/torch_ranker_agent.py
```diff
@@ -924,8 +944,8 @@ def _make_candidate_encs(self, vecs):
         self.model.eval()
         with torch.no_grad():
             for vec_batch in tqdm(vec_batches):
-                cand_encs.append(self.encode_candidates(vec_batch))
-        return torch.cat(cand_encs, 0)
+                cand_encs.append(self.encode_candidates(vec_batch).cpu())
```
stephenroller: I'm not sure we want this. What led you to adding it?

Thoughts from @emilydinan, @dexterju?

parlai/scripts/build_candidates.py
```diff
@@ -43,7 +43,7 @@ def build_cands(opt):
 
     print('[ starting to build candidates from task.. (ex:' + str(num_examples) + ')]')
     print('[ saving output to {} ]'.format(outfile))
-    cands = []
+    cands = set()
```
stephenroller: This is fine, but why was it made a set?

parlai/tasks/self_chat/agents.py
```diff
@@ -0,0 +1,29 @@
...
+from .build import build
+
+import json
+import os
+import random
+
+
+class DefaultTeacher(Teacher):
+    # Dummy class.
+    pass
```
stephenroller: Instead can you do:

```python
class DefaultTeacher(Teacher):
        raise RuntimeError('-t self_chat is a dummy helper, and not meant to be used directly.')
```

parlai/tasks/self_chat/build.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+def build(opt):
```
stephenroller: I don't think you need this file at all.

parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -907,6 +918,14 @@ class DefaultTeacher(WizardDialogKnowledgeTeacher):
     pass
 
 
+class SelfchatTeacher(WizardDialogKnowledgeTeacher):
```
stephenroller: SelfChatTeacher?

parlai/core/torch_ranker_agent.py
jaseweston: it was suggested fix from @klshuster  -- currently it oom without this. it seems superior to me

parlai/scripts/build_candidates.py
jaseweston: so we don't have repeat utterances in fixed set which is both slower for no reason, and also produces some problems when the topk is all the same

parlai/tasks/self_chat/agents.py
jaseweston: sure

parlai/tasks/self_chat/build.py
jaseweston: ah yea you prolly right

parlai/tasks/wizard_of_wikipedia/agents.py
jaseweston: i think the intention originally of alex and i in parlai was that you could enter lowercase in commandline and it would work because it would uppercase the first letter:
https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/worlds.py#L1147
but now we seem to have a smattering of both types (only first letter uppercase,  every word uppercase) because people didnt follow this..    note all the tasks are in lowercase as well

jaseweston:
> You need to edit the task_list.py to add self_chat to the list of tasks.

I think the tasks in the list should be actual tasks, like datasets, not helper tasks  like this?  we don't have interactive or fromfile there either for example

parlai/tasks/wizard_of_wikipedia/agents.py
jaseweston: so should we abandon this thing and try and make everything cased in future?  it's half and half right now..  or go back to the old way

stephenroller: I think if it has an underscore the next letter becomes caps. I agree we should standardize

stephenroller: You should do what you like here

parlai/core/torch_ranker_agent.py
```diff
@@ -444,7 +457,14 @@ def eval_step(self, batch):
...
             cand_preds = self.block_repeats(cand_preds)
 
-        preds = [cand_preds[i][0] for i in range(batchsize)]
+        if self.opt.get('inference', 'max') == 'max':
+            preds = [cand_preds[i][0] for i in range(batchsize)]
+        else:
+            # Top-k inference.
+            preds = []
+            for i in range(batchsize):
+                preds.append(random.choice(cand_preds[i][0 : self.opt['topk']]))
```
stephenroller: Topk in then generator is stochastic to the logits. Should we here too?

parlai/core/torch_ranker_agent.py
stephenroller: Hmm this will be slower due to transfer but it does have merit.

parlai/core/torch_ranker_agent.py
```diff
@@ -924,8 +944,8 @@ def _make_candidate_encs(self, vecs):
         self.model.eval()
         with torch.no_grad():
             for vec_batch in tqdm(vec_batches):
-                cand_encs.append(self.encode_candidates(vec_batch))
-        return torch.cat(cand_encs, 0)
+                cand_encs.append(self.encode_candidates(vec_batch).cpu())
+        return torch.cat(cand_encs, 0).cuda()
```
stephenroller: Make this `.to(vec_batch.device)` and it should fix the tests

parlai/core/torch_ranker_agent.py
klshuster: yeah there's a speed tradeoff but if we're encoding too many cands i think it might be necessary

PR #2214, opened 2019-11-23
Trying flake8 with github actions

stephenroller:
**Patch description**
Namely [this thing](https://github.com/marketplace/actions/flake8-action). It should actually annotate the commit with the errors from flake8.


stephenroller:
We don't have access to github actions yet.

PR #2216, opened 2019-11-23
Add type checking

stephenroller:
**Patch description**
Add the basics of type checking to ParlAI. There are lots and lots of errors that are not fixed (~126), and hopefully we'll add more annotations so more can be caught.

Currently, this CircleCI should always pass (because it's difficult to fix many of these errors), but the LOGS will be made available.

**Testing steps**
CircleCI.


parlai/core/teachers.py
```diff
@@ -525,15 +518,17 @@ def num_episodes(self):
         except AttributeError:
             return super().num_episodes()
 
-    @lru_cache(maxsize=1)
-    def num_examples(self):
+    def num_examples(self) -> int:
```
stephenroller: lrucache isn't type safe, so i'm manually doing a cache here.

parlai/agents/starspace/starspace.py
```diff
@@ -31,14 +31,14 @@ class StarspaceAgent(Agent):
     """
 
     OPTIM_OPTS = {
-        'adadelta': optim.Adadelta,
-        'adagrad': optim.Adagrad,
+        'adadelta': optim.Adadelta,  # type: ignore
```
stephenroller: pytorch doesn't have these types annotated yet

parlai/agents/bert_ranker/helpers.py
```diff
@@ -261,14 +260,14 @@ class AdamWithDecay(Optimizer):
     def __init__(
         self,
         params,
-        lr=required,
+        lr,
```
stephenroller: this seems just not really necessary. we should delete the whole class.

parlai/core/teachers.py
```diff
@@ -324,23 +324,16 @@ def next_batch(self):
...
         """
         Get the total number of examples in this dataset.
         """
-        if self.use_batch_act:
-            # when using batch_act, this is length of sorted data
-            return len(self.sorted_data)
```
stephenroller: note that `sorted_data` was never called anywhere in the codebase. this was an example of a static failure being caught by the type checker. Really not sure where it came from, but better to just require the user implement the method.

parlai/mturk/core/dev/test/test_full_system.py
```diff
@@ -343,6 +342,9 @@ class TestMTurkManagerWorkflows(unittest.TestCase):
     """
 
     def setUp(self):
+        patcher = mock.patch('builtins.input', return_value='y')
```
stephenroller: @JackUrb  this party you review.

Note that input was being done a nasty way before: we would set the context here, and that would cause calls to input later to resolve to a nonlocal context, instead of the builtins context. Very gross.

Now we'd like to just use the supported monkeypatching. If this weren't all being done in setUp, we could just use a decorator, but that doesn't affect setUp. So we need to manually patch.

parlai/mturk/webapp/server.py
```diff
@@ -36,12 +36,16 @@
 from parlai.mturk.core.mturk_data_handler import MTurkDataHandler
 from parlai.mturk.core.mturk_manager import MTurkManager
 from parlai.mturk.webapp.run_mocks.mock_turk_manager import MockTurkManager
-from parlai import __path__ as parlai_path
+from typing import Dict, Any
+
+# https://github.com/python/mypy/issues/1422
+from parlai import __path__ as parlai_path  # type: ignore
```
stephenroller: be real nice if we didn't even need this but ya know

parlai/tasks/dailydialog/parse.py
```diff
@@ -14,6 +14,7 @@
 import os
```
stephenroller: This is reasonably dead code that is left there as reference. We could easily delete the whole script with a comment left about the last commit it lived in.

tests/suites.py
```diff
@@ -101,7 +101,3 @@ def internal_tests():
     test_loader = unittest.TestLoader()
     test_suite = test_loader.discover("parlai_internal/tests")
     return test_suite
-
-
-if __name__ == '__main__':
```
stephenroller: this was just straight up broken.

parlai/agents/bert_ranker/helpers.py
klshuster: does torch now provide adam with decay? was HF's different?

parlai/agents/legacy_agents/seq2seq/seq2seq_v0.py
```diff
@@ -112,14 +112,14 @@ class Seq2seqAgent(Agent):
     """
 
     OPTIM_OPTS = {
-        'adadelta': optim.Adadelta,
-        'adagrad': optim.Adagrad,
+        'adadelta': optim.Adadelta,  # type: ignore
+        'adagrad': optim.Adagrad,  # type: ignore
         'adam': optim.Adam,
```
klshuster: is there a reason we don't ignore Adam's type?

parlai/agents/starspace/starspace.py
klshuster: should we just have this be in a utils file somewhere? looks like seq2seq uses it as does starspace

parlai/core/teachers.py
klshuster: it's an artifact of the old batchsort, thanks for removing

parlai/utils/misc.py
```diff
@@ -479,6 +433,32 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+def round_sigfigs(x: Union[float, 'torch.Tensor'], sigfigs=4) -> float:
+    """
+    Round value to specified significant figures.
+
+    :param x: input number
+    :param sigfigs: number of significant figures to return
+
+    :returns: float number rounded to specified sigfigs
+    """
+    x_: float
```
klshuster: this is a thing now??

parlai/utils/testing.py
```diff
@@ -371,7 +371,7 @@ def display_data(opt):
     return (train_output.getvalue(), valid_output.getvalue(), test_output.getvalue())
 
 
-def display_model(opt) -> Tuple[str, str]:
+def display_model(opt) -> Tuple[str, str, str]:
     """
     Run display_model.py.
```
klshuster: nit: in line below, return value is incorrect

parlai/utils/torch.py
```diff
@@ -0,0 +1,223 @@
...
+    :param bool fp16friendly: if True, pads the time dimension to be a multiple of 8.
+
+    :returns: (padded, lengths) tuple
+    :rtype: (Tensor[int64], list[int])
+    """
+
+    # number of items
+    n = len(items)
+    # length of each item
+    lens: List[int] = [len(item) for item in items]  # type: ignore
```
klshuster: why ignore this type?

projects/taskntalk/coopgame_agent.py
```diff
@@ -25,14 +25,14 @@ class CooperativeGameAgent(Agent):
     """
 
     OPTIM_OPTS = {
-        'adadelta': optim.Adadelta,
-        'adagrad': optim.Adagrad,
+        'adadelta': optim.Adadelta,  # type: ignore
```
klshuster: this is everywhere lol

parlai/agents/legacy_agents/seq2seq/seq2seq_v0.py
stephenroller: Adam has types!

parlai/utils/misc.py
stephenroller: yeah useful if you want to declare a variable that's going to be set in two different code paths. it's not entirely necessary though.

parlai/agents/starspace/starspace.py
stephenroller: I'm not sure if I care to refactor. I'm going to retire that seq2seq soon enough, and torch agent is the canonical way to handle this.

parlai/utils/torch.py
stephenroller: torch.Tensor is not of type Sizeable, so this throws an error. (Generally len(tensor) does). I'd like to make a patch for that.

projects/taskntalk/coopgame_agent.py
stephenroller: yes recall when we used to copy pasta all the agents?

parlai/agents/bert_ranker/helpers.py
stephenroller: Yes since torch 1.2 I believe

parlai/mturk/core/dev/test/test_full_system.py
JackUrb: This is much cleaner than the silly solution I stumbled on. Looks good to me!

projects/taskntalk/coopgame_agent.py
klshuster: 🤦‍♂

PR #2217, opened 2019-11-23
Try again with the build website.

stephenroller:
**Patch description**
Still trying to get all of the kinks worked out of CircleCI. Changing to `>=` for some key dependencies managed to fix things.

Thought there were some problems with the GPU setup, so I ended up also upgrading the distribution of linux we're running GPU jobs on, and adding caching to the nvidia downloads.

Changes:
- Add a `pip install -r requirements.txt`
- Change some key dependencies to `>=` allowing pip to sort things out a bit better
- Upgrade linux distribution in GPU tests so we can use python 3.7. This mandated upgrading the NVIDIA/CUDA installation as well.
- Add caching to nvidia downloads
- Bumped the cache on all the dependencies for safety.

**Testing steps**
CircleCI long finally passes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,15 +108,18 @@ setupcuda: &setupcuda
     command: |
       # download and install nvidia drivers, cuda, etc
       wget --no-clobber -P ~/nvidia-downloads 'https://s3.amazonaws.com/ossci-linux/nvidia_driver/NVIDIA-Linux-x86_64-430.40.run'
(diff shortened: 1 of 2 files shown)


vedantpuri:
@stephenroller this [issue](https://github.com/cvxgrp/scs/issues/85) seems exactly what we are having, how is the version of setuptools dictated ? Didn't find it in the requirements, maybe changing the version of setup tools could be helpful

PR #2218, opened 2019-11-24
Constrain quickstart.sh to only one node when running nightly

### LINKED ISSUE #2193: GPU nightly tests have many workers, but gpu quickstart is running on all of them
The GPU tests are actually split across many hosts to speed them up:

https://github.com/facebookresearch/ParlAI/blob/250891883a628b2d793ad0469ecb8561bfea3b29/.circleci/config.yml#L292-L295

https://github.com/facebookresearch/ParlAI/blob/250891883a628b2d793ad0469ecb8561bfea3b29/tests/suites.py#L15-L31

But the quickstart runs on all of them:

https://github.com/facebookresearch/ParlAI/blob/250891883a628b2d793ad0469ecb8561bfea3b29/.circleci/config.yml#L313-L315

We should just shortcut the quickstart tests so they only on rank 0 or similar.

vedantpuri:
**Patch description**
Attempts to fix #2193. 

**Testing steps**
Run [long] tests. Quickstart should appear only in node 0.

**Logs**
NA

**Other information**
NA

NA


vedantpuri:
@stephenroller What do you think of the current way to go about this?

Also, I have no clue why there is a dependency install error in the [mturk test](https://circleci.com/gh/facebookresearch/ParlAI/20056), [data test](https://circleci.com/gh/facebookresearch/ParlAI/20055), [copyright test](https://circleci.com/gh/facebookresearch/ParlAI/20048). They all fail at the same point, while fetching scipy (it fails due to an import error related to numpy), I am not even sure why it goes there since it should ideally just use the cache like it has been (maybe a cache miss ?). Need a little help regarding that, since it might affect future commits as well.

stephenroller:
I think because I updated the Debian version and python version. Somewhere it says “pyenv” and that should be updated, and all the dependency caches should be reset.

vedantpuri:
I don't think that's the issue since before checking out this branch things seem to be working just fine. Also, even now if you see, things are being installed in the path where I have dependencies cached, so I am not sure what's going on.
Also its only these cache dependencies being affected, not others why is that ?
@stephenroller

stephenroller:
It's happening on master too. I'm trying to fix it in #2217.

vedantpuri:
I see, commented on your PR.
Also I am unsure why the quickstart wasn't there on any of the parallel instances in the latest run, any thoughts on that ? @stephenroller

stephenroller:
I'm wary of the way you did it with the test calling a subprocess. I'd rather just keep it the way it is, and add a conditional fast exit at the top of the shell script if the parallelism isn't 0 or undefined.

vedantpuri:
@stephenroller I think the current solution works, let me know what you think!

tests/test_quickstart.sh
```diff
@@ -3,6 +3,8 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+([ -z "$CIRCLE_NODE_INDEX" ] || [ $CIRCLE_NODE_INDEX != 0 ])  && exit
```
stephenroller: Lgtm. Can you also echo that you’re skipping the test?

PR #2220, opened 2019-11-24
ConvAI2 self-chats

jaseweston:
Run with:

python parlai/scripts/self_chat.py -mf zoo:pretrained_transformers/model_poly/model   -t convai2:selfchat --inference topk  -ne 10 --display-examples True -dt valid


stephenroller:
tests are because broken CI. See #2217

PR #2234, opened 2019-12-02
self-chat logs:  label->labels bug

jaseweston:
Currently saves out the data, but the label is in the wrongly named field.


stephenroller:
We should talk about a way to just sort this out in general. We all make this bug frequently.

PR #2237, opened 2019-12-03
Small modernizations to controllable dialogue mturk task.

stephenroller:
**Patch description**
Small modernizations to an older mturk task that we used in [controllable dialogue](https://parl.ai/projects/controllable_dialogue/)

**Testing steps**
Sandbox several runs, will launch a real collection shortly.


stephenroller:
yeah I launched and it's pretty reliable.

PR #2244, opened 2019-12-04
[TGA] Fix uncaught exception.

stephenroller:
**Patch description**
An error message to prevent unexpected behavior wasn't being tripped due to a hardcoded value. Add a test that it's being checked, and fix the issue.

The exact behavior was increasing --beam-size in torch generator agent without also setting --inference to be not greedy.

**Testing**
New unit test.

### DIFF
--- a/tests/test_tga.py
+++ b/tests/test_tga.py
@@ -36,7 +36,7 @@ class TestUpgradeOpt(unittest.TestCase):
         """
         testing_utils.download_unittest_models()
 
-        # we should have an exception if we mix beam size > 1 with greedy
+        # we should have an exception if we mix beam size > 1 with inference greedy
         with self.assertRaises(ValueError):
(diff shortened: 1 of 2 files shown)


tests/test_tga.py
```diff
@@ -30,6 +30,33 @@ def test_inference(self):
...
+            testing_utils.display_model(
+                dict(
+                    task='integration_tests:multipass',
+                    model_file='zoo:unittest/transformer_generator2/model',
+                    beam_size=5,
+                    inference='greedy',
+                )
+            )
+
+        # and we shouldn't if we have bug
```
klshuster: if we have bug?

emilydinan: i don't understand this comment, "we shouldn't if we have bug". do you mean "we shouldn't if we have beam"?

PR #2247, opened 2019-12-05
[WebSockets Service] Cleanup WebSockets

ideanl:
**Patch description**
This PR cleans up parts of websockets by removing an unused function `_get_message_from_image` (references to it were removed in #2201). Also, fixes documentation for `observe_payload` to be updated to the correct specification, which is generalized for all payloads and more consistent with the messenger service.

**Testing steps**
Ran the websockets server in the following way using a UI that can send image payloads: `python ~/ParlAI/parlai/chat_service/services/websocket/run.py --config-path <path to config>`


parlai/chat_service/services/websocket/websocket_manager.py
```diff
@@ -269,14 +269,14 @@ def observe_payload(self, socket_id, payload, quick_replies=None):
         :param socket_id:
             int identifier for agent socket to send message to
         :param payload:
-            (dict) payload to send through the socket. The keys should be:
-                'image': (True/False) whether the message is an image
-                'mime_type': str. Mime type of the message
-                'data': str. base64 encoded content
+            (dict) payload to send through the socket. The mandatory keys are:
+                    'type': (True/False) whether the message is an image
```
klshuster: nit: it appears the values for `type` are no longer `(True/False)`, could you update that as well?

PR #2259, opened 2019-12-10
Length penalty as flag (e.g., so you can turn it off)

jaseweston:


jaseweston:
abandoned, put in another branch

stephenroller:
Changed to #2372.

PR #2261, opened 2019-12-10
eval_model script saves log in json format

jaseweston:
Previously, it had no log, which sucked

### DIFF
--- a/parlai/scripts/eval_model.py
+++ b/parlai/scripts/eval_model.py
@@ -37,9 +37,9 @@ def setup_args(parser=None):
         '-rf',
         '--report-filename',
         type=str,
-        default='.eval_model',
+        default='',
         help='Saves a json file of the evaluation report either as an '
-        'extension to the model name (if begins with a ".") or a whole '
+        'extension to the model-file (if begins with a ".") or a whole '
         'file path. Set to the empty string to not save at all.',
     )
     parser.add_argument('-ne', '--num-examples', type=int, default=-1)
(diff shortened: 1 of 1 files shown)


parlai/scripts/eval_model.py
```diff
@@ -32,6 +33,15 @@ def setup_args(parser=None):
         parser = ParlaiParser(True, True, 'Evaluate a model')
     parser.add_pytorch_datateacher_args()
     # Get command line arguments
+    parser.add_argument(
+        '-rf',
+        '--report-filename',
+        type=str,
+        default='.eval_model',
+        help='Saves a json file of the evaluation report either as an '
+        'extension to the model name (if begins with a ".") or a whole '
```
stephenroller: nit: model_file or model_filename

parlai/scripts/eval_model.py
```diff
@@ -32,6 +33,15 @@ def setup_args(parser=None):
         parser = ParlaiParser(True, True, 'Evaluate a model')
     parser.add_pytorch_datateacher_args()
     # Get command line arguments
+    parser.add_argument(
+        '-rf',
```
stephenroller: What do you think about having this parameter empty by default?

Consider if runs an eval_model on a zoo file. We're going to be touching that directory, even in unit tests. I'd rather not have it smash the directories *by default*.

klshuster: i think i agree with this, though one way to fix unit tests is to add this arg to `testing_utils.eval_model`

parlai/scripts/eval_model.py
jaseweston: ok

parlai/scripts/eval_model.py
jaseweston: ok, i'm not sure, i kind of liked how it saved if you forget, but i see your point.
i'm wondering if there's a file there if it could append to the report (like the training stats log), possibly with a timestamp
but for now, i've made it empty as requested

klshuster: perhaps if we made the default a `/tmp/` location we would not run into this issue

PR #2263, opened 2019-12-10
Image Seq2Seq Agent

klshuster:
**Patch description**
Addition of the `Image+Seq2Seq` agent, an agent that combines a transformer generator with an image encoder.

**Testing steps**
Included `tests/test_image_seq2seq.py`, which tests that the model can train in the following setups:

1. With a text-based task
2. With an image-based task
3. With a text-based task AND an image-based task

**Logs**
```
$ python tests/test_image_seq2seq.py
....
Ran 4 tests in 102.956s

OK
```

**Other Information**
I am seeking opinions on the `load_state_dict` function; I made it very flexible to account for various types of init_model schemes (encoder only, old dictionary without image tokens, whole model without image encoder, etc) which made it a little bit bloated.


parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+        """
+        Override to build appropriate model.
+        """
+        self.model = ImageSeq2seqModel(self.opt, self.dict)
+        if self.opt['embedding_type'] != 'random':
+            self._copy_embeddings(
+                self.model.embeddings.weight, self.opt['embedding_type']
+            )
+        if self.use_cuda:
+            self.model = self.model.cuda()
```
emilydinan: just fyi, i think this is done on init of TGA. did you find that you needed to do it again here?

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+        shared['nltk_bleu'] = self.nltk_bleu
+        shared['nltk_bleu_cnts'] = self.nltk_bleu_cnts
+        return shared
+
+    def build_model(self) -> ImageSeq2seqModel:
+        """
+        Override to build appropriate model.
+        """
+        self.model = ImageSeq2seqModel(self.opt, self.dict)
+        if self.opt['embedding_type'] != 'random':
```
emilydinan: this should be taken care of by `copy_embeddings` in torch_agent

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def _set_text_vec(self, *args, **kwargs) -> dict:
+        """
+        Override to include image token.
+        """
+        obs = super()._set_text_vec(*args, **kwargs)
+        if 'text' not in obs or 'text_vec' not in obs:
+            return obs
+        if self.opt.get('include_image_token', False):
+            truncate = args[2] - 1 if args[2] is not None else None
```
emilydinan: can you add a comment here? what is `args[2]`? a bit  opaque

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+        """
+        Override for custom loading.
+
+        Three reasons:
+            1. When using an init model without an image encoder
+            2. When using an init model with only an encoder provided
+                In this case, we may need to add the START token to the state_dict
+            3. When using an init model without image tokens in the embeddings
+        """
+        # Case 1 -> No Image Encoder
```
emilydinan: great commentary here 😄

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+                state_dict['START'] = self.model.START
+
+        cnt = 0
+        if self.opt['init_model'] is not None:
+            try:
+                self.model.load_state_dict(state_dict)
+                return
+            except Exception as e:
+                # Case 3 --> Check for Embedding Diffs
+                print('exception in loading state dict: {}'.format(e))
```
emilydinan: interesting... do we always want to do this? or do we want to surface the error?

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+        metrics.update({'fairseq_bleu': 'N/A', 'nltk_bleu_unnormalized': 'N/A'})
+        if not self.skip_generation and self.compute_tokenized_bleu:
+            if fairseq_bleu is not None:
+                try:
+                    fairseq_bleu_scores = {
+                        k: self.fairseq_bleu_scorer.result_string(order=k)
+                        for k in range(1, 5)
+                    }
+                except ZeroDivisionError:
+                    # some preds are REAL bad
```
emilydinan: LOL

parlai/core/torch_generator_agent.py
```diff
@@ -735,12 +741,22 @@ def _generate(self, batch, beam_size, max_ts):
         if isinstance(model, torch.nn.parallel.DistributedDataParallel):
             model = self.model.module
         encoder_states = model.encoder(*self._model_input(batch))
-        dev = batch.text_vec.device
+        if batch.text_vec is not None:
+            dev = batch.text_vec.device
+        else:
```
emilydinan: can you explain this?

klshuster:
just FYI that the gpu unittests are failing due to what I _think_ is an issue with `typing` with python3.7. I am investigating...

parlai/agents/image_seq2seq/image_seq2seq.py
klshuster: ahh yes you're right, I can remove this

parlai/agents/image_seq2seq/image_seq2seq.py
klshuster: i think i actually do need this here, as it's taken care of in `TransformerGeneratorAgent.build_model` (which I subclass); I wasn't able to find `copy_embeddings` in regular TA

parlai/agents/image_seq2seq/image_seq2seq.py
klshuster: for now I assume that the error is embedding based and print the exception; if there is some other error that happens, it'll raise later as I still call `self.model.load_state_dict` after this try/catch

parlai/core/torch_generator_agent.py
klshuster: this is related to the long-standing issue that we still do not specify `batchsize` in a batch and instead have code that calls `batch.text_vec`; this breaks for any task that is not exclusively text based unfortunately

klshuster:
@ Everyone --> The tests _finally_ pass (not sure what was going on...), so this is **ready for review**

parlai/agents/image_seq2seq/image_seq2seq.py
stephenroller: So @EricMichaelSmith's tutorial on TGA actually covers the recommended usage:

https://github.com/facebookresearch/ParlAI/blob/45e7516ef4ed54d8a70a39cf32d10b529a8250bb/parlai/agents/examples/seq2seq.py#L192-L201

parlai/agents/image_seq2seq/image_seq2seq.py
stephenroller: Can you leave a comment explaining this? My instinct was definitely like "wow, are we sure we want to catch all exceptions here?" It might be worth trying to narrow down beyond just "Exception" to what kind of exception we're actually expecting (KeyError?)

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+                enc_embs = state_dict['encoder.embeddings.weight']
+                dec_embs = state_dict['decoder.embeddings.weight']
+                num_embs = min(self.model.embeddings.weight.shape[0], embs.shape[0])
+                num_enc_embs = min(
+                    self.model.encoder.embeddings.weight.shape[0], enc_embs.shape[0]
+                )
+                num_dec_embs = min(
+                    self.model.decoder.embeddings.weight.shape[0], dec_embs.shape[0]
+                )
+                for i in range(num_embs):
```
stephenroller: i'm sorry, what is going on in all this? I see a lot of very opaque editing of embeddings and I'm unsure why we're going through these motions

stephenroller: you know what, checking the cases above again, it looks like you're trying to initialize a decoder. so like, what's with the loops?

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+                state_dict[
+                    'encoder.embeddings.weight'
+                ] = self.model.encoder.embeddings.weight
+                state_dict[
+                    'decoder.embeddings.weight'
+                ] = self.model.decoder.embeddings.weight
+
+                print(
+                    'Initialized embeddings for {} tokens ({}%).'
+                    ''.format(cnt, round(cnt * 100 / len(self.dict), 1))
```
stephenroller: Consider `f'Initialized embeddings for {cnt} tokens ({cnt/len(self.dict).1%})'`

stephenroller: (I personally don't like expressions in f-strings, I prefer to rip out the expression into a temp variable, but the `{var:.1%}` is what I want to highlight)

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -0,0 +1,349 @@
...
+                    for k, v in fairseq_bleu_scores.items()
+                }
+            if nltkbleu is not None:
+                metrics['nltk_bleu_unnormalized'] = {
+                    k: round_sigfigs(v / self.nltk_bleu_cnts[k], 4)
+                    for k, v in self.nltk_bleu.items()
+                }
+        return metrics
+
+    def _compute_fairseq_bleu(self, batch: Batch, texts: List[str]):
```
stephenroller: worth bubbling up to TGA?

parlai/agents/image_seq2seq/image_seq2seq.py
stephenroller: solid

parlai/agents/image_seq2seq/modules.py
```diff
@@ -0,0 +1,201 @@
...
+        input
+        """
+        self.n_img_layers = kwargs.pop('image_encoder_num_layers')
+        self.img_dim = kwargs.pop('image_features_dim')
+        self.use_cuda = kwargs.pop('use_cuda')
+        super().__init__(*args, **kwargs)
+        self._build_image_encoder()
+        self.dummy_image_enc = torch.zeros((self.embedding_size))
+        self.ones_mask = torch.ones(1).bool()
+        if self.use_cuda:
```
stephenroller: so it's better to avoid passing around a `use_cuda` or `device` flag in model code. Instead, create the tensor and use `tensor.to(known_tensor.device)`. For example here, you might use self.somelayer.device

parlai/agents/image_seq2seq/modules.py
```diff
@@ -0,0 +1,201 @@
...
+    def __init__(self, *args, **kwargs):
+        """
+        Override TransformerEncoder __init__.
+
+        Setup the image encoder; create some dummy tensors for inserting image into
+        input
+        """
+        self.n_img_layers = kwargs.pop('image_encoder_num_layers')
+        self.img_dim = kwargs.pop('image_features_dim')
+        self.use_cuda = kwargs.pop('use_cuda')
```
stephenroller: :/ I don't looove these kwargs pops but like, idk, I get it. One might consider the (tedious) but exhaustive enumeration of args, but that isn't without flaws.

parlai/agents/image_seq2seq/modules.py
```diff
@@ -0,0 +1,201 @@
...
+                    image_encoded.append(self.dummy_image_enc)
+
+            image_masks = torch.stack(image_masks)
+            image_encoded = torch.stack(image_encoded).unsqueeze(1)
+
+        return image_encoded, image_masks
+
+    def forward(
+        self, src_tokens: torch.Tensor, image_features: List[object]
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
```
stephenroller: dope, dope dope dope

parlai/core/torch_generator_agent.py
```diff
@@ -603,9 +606,12 @@ def eval_step(self, batch):
         """
         Evaluate a single batch of examples.
         """
-        if batch.text_vec is None:
+        if batch.text_vec is None and batch.image is None:
```
stephenroller: So we've got this in a bunch of places now. I'm wondering if life would be easier if batch world didn't give the agents padding examples, so that `is_valid` and checks like this could be completely removed. Opinions from the team?

stephenroller:
Also nit, but we use `-m seq2seq` to mean an RNN, and "Image Seq2Seq" to mean a transformer. Maybe that's misleading? Technically all the encoder-decoders (RNNs and Transformers alike) are all seq2seq models. (i.e. we should've really renamed `-m seq2seq` to `-m rnn_seq2seq`

parlai/agents/image_seq2seq/image_seq2seq.py
klshuster: i will change the exception here to make it more clear what we're handling

parlai/agents/image_seq2seq/image_seq2seq.py
klshuster: this whole function was the one i was least comfortable with, so i tried documenting the 3 cases in the docstring. This is a case where we are initializing an **Image** TGA with an init model of a **Standard** TGA. The issue that comes up is that the new **image** TGA has two additional tokens in its dictionary - `__image__` and `__no_image__`.  As a result, the embeddings are not the right dimension. (e.g. 50,000 vs. 50,002). So, as the image tokens are presumably at the end of the dictionary, I copy over the first N-2 embedding weights from the `init_model` to the new model.

perhaps a more clear way of handling this is to check that, indeed, the embedding dimensions only differ by 2. Also, I can probably remove most of these for loops and just have one.

parlai/agents/image_seq2seq/image_seq2seq.py
klshuster: I suppose I could include there, initially i did not want to clutter TGA with this but if it's useful...

parlai/agents/image_seq2seq/modules.py
klshuster: could you elaborate what this solution would look like?

parlai/core/torch_generator_agent.py
klshuster: at the very least, can we please put a `batchsize` attr in the batch? or maybe just a `batch.valid` attr that marks whether there are any valid examples in the batch? since that is a way more clear check than just checking if text or image exists

klshuster:
To respond to your top-level nit - i would think that `-m seq2seq` is the misnomer here. we described this model as "Image+Seq2Seq" in dodeca paper

parlai/agents/image_seq2seq/modules.py
klshuster: nvm i know what you mean

klshuster:
i addressed most of your comments @stephenroller if you'd like to take another look

Issue #2165, opened 2019-11-18
Immutable opt


stephenroller:
Opt should be made immutable.

- Remove its `dict` superclass
- Similar to message, throw a hissy fit on `__setitem__`
- Add a new fork option, which allows for values to be overridden. The fork will essentially do the current copy we have now.
- Should also do a debug log that a value was updated
- The point is to make it so that you can't mutate opts without being intentional, and it won't update anything anywhere else
- Ideally, forked opts should keep their parental links, that'd be super dope

PR #2266, opened 2019-12-11
Move opt; annotate some parameters with Opt

stephenroller:
**Patch description**
This is mostly an annotation patch, where I add a some typing to more places. This will eventually feed into #2165.

**Testing steps**
mypy, ci.

### DIFF
--- a/parlai/core/opt.py
+++ b/parlai/core/opt.py
@@ -18,8 +18,8 @@ class Opt(dict):
     """
     Class for tracking options.
 
-    Functions like a dict, but allows us to track the history of arguments
-    as they are set.
+    Functions like a dict, but allows us to track the history of arguments as they are
+    set.
     """
 
     def __init__(self, *args, **kwargs):
@@ -43,7 +43,9 @@ class Opt(dict):
         return (Opt, (), self.__getstate__())
 
     def __deepcopy__(self, memo):
-        """Override deepcopy so that history is copied over to new object."""
+        """
+        Override deepcopy so that history is copied over to new object.
+        """
         # track location of deepcopy
         loc = traceback.format_stack()[-3]
         self.deepcopies.append(loc)
@@ -58,7 +60,9 @@ class Opt(dict):
         return memo
 
     def display_deepcopies(self):
-        """Display all deepcopies."""
+        """
+        Display all deepcopies.
+        """
         if len(self.deepcopies) == 0:
             print('No deepcopies performed on this opt.')
             return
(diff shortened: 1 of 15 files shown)


parlai/core/opt.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.history = {}
+        self.deepcopies = []
+
+    def __setitem__(self, key, val):
+        loc = traceback.format_stack()[-2]
+        self.history.setdefault(key, []).append((loc, val))
```
EricMichaelSmith: This is the first time I've seen the Opt implementation - it looks super useful! I assume the second-to-last line of the stack is the call just before `__setitem__()`? (I'm trying to think if there'd be a need to have different opts for different calls of the same function, which would cause a collision here, but that's probably an edge case)

parlai/core/opt.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __reduce__(self):
+        return (Opt, (), self.__getstate__())
+
+    def __deepcopy__(self, memo):
+        """
+        Override deepcopy so that history is copied over to new object.
+        """
+        # track location of deepcopy
+        loc = traceback.format_stack()[-3]
```
EricMichaelSmith: Hmm - why is it using the third-to-last call of the stack here?

parlai/core/torch_agent.py
```diff
@@ -701,7 +703,7 @@ def __init__(self, opt, shared=None):
         self.id = type(self).__name__.replace("Agent", "")
 
         # now set up any fields that all instances may need
-        self.EMPTY = torch.LongTensor([])
+        self.EMPTY = torch.zeros(0, dtype=torch.long)
```
EricMichaelSmith: Interesting - is this a subtly different empty tensor than before?

stephenroller: The difference is the type checker doesn't complain :D

parlai/core/opt.py
stephenroller: Three levels:
- `copy.deepcopy(opt)` (-3)
- inside the source of deepcopy it calls `opt__deepcopy__` (-2)
- here (-1)

We want to record the line that called deepcopy.

parlai/core/opt.py
stephenroller: -1 is the current line (opt.py:30), -2 is whatever called `opt[key] = value`.

PR #2272, opened 2019-12-15
Dynamic batching

### LINKED ISSUE #2192: Dynamic Batching
We'd like to support dynamic batching, which is an advancement over batchsort and similar. The idea is that if we pack shorter conversations together, we can fit MORE of them in.

Dependencies:
#2043, #2191 

The basic idea is that we'll have something like 1024 or so copies of our batch worlds, and we'll be dynamically deciding which to advance at a time. This is why the self observe stuff is so important: we won't have any promise about what "index" the subworld is in at any given time.

A few requirements
- [ ] Should be an option (`-dynbs`?).
- [ ] Should essentially do batchsort "for free"
- [ ] Good testing is important
- [ ] Prototype can just hardcode some stuff or use a `--words-per-batch` option or similar.

Optional features:
- [ ] Doing batchify off the main thread would be incredible.
- [ ] Should use `--truncate` and/or a magic iteration to try to identify the max batch size for free, by catching OOMs. OOM handling is extremely finicky, so the truncate magic might be the way to go.

stephenroller:
**Patch description**
Dynamic batching is a method for increasing throughput by ensuring that every batch maximizes the amount of GPU memory. The key idea is that we can pack MANY small examples into memory or we can pack FEW large examples. (Closes #2192).

This is a large and sweeping change that requires extensive testing and careful implementations. It likely will have edge cases and bugs.

Key steps and requirements:
- [x] Initial implementation, which exploits self_observe.
- [x] Ensure all metrics are correct
- [x] Measure actual throughput improvements
- [x] Add CI testing
- [x] Support MultiWorld

**Testing steps**
- Manual testing.
- (there will be) new CIs.


stephenroller:
Looks like it does about the same as batchsort tends to get:

Without dynamic batching
```
$ python examples/train_model.py -t convai2 -m transformer/generator --text-truncate 1024 --label-truncate 128 --dict-tokenizer gpt2 -mf /tmp/testdyn -esz 128 --ffn-size 128 --skip-generation true  -bs 12 -eps 1 -dynb false
[ time:327.0s total_exs:131448 epochs:1.0 time_left:0s ] {'exs': 456, 'lr': 1, 'total_train_updates': 10954, 'gnorm': 2.475, 'clip': 1.0, 'gpu_mem_percent': 0.146, 'loss': 167.4, 'token_acc': 0.2867, 'nll_loss': 4.421, 'ppl': 83.15}
[ num_epochs completed:1.0 time elapsed:327.50010323524475s ]
```
With dynamic batching
```
$ python examples/train_model.py -t convai2 -m transformer/generator --text-truncate 1024 --label-truncate 128 --dict-tokenizer gpt2 -mf /tmp/testdyn -esz 128 --ffn-size 128 --skip-generation true  -bs 12 -eps 1 -dynb true
[ time:126.0s total_exs:131510 epochs:1.0 time_left:0s ] {'exs': 1232, 'lr': 1, 'total_train_updates': 1690, 'gnorm': 1.357, 'clip': 1.0, 'gpu_mem_percent': 0.351, 'loss': 83.72, 'token_acc': 0.2414, 'nll_loss': 5.149, 'ppl': 172.3}
[ num_epochs completed:1.0 time elapsed:126.10264015197754s ]
```
About a 2.6x speedup.

stephenroller:
Interestingly, if I fix the batchsize here (so that this becomes effectively batchsort), I don't see any speedup compared to fixed batches (in fact, slightly slower).

With cheaper tokenization (`--dict-tokenizer space`):
Without normal batching: 226.7s (5477 SGD steps)
With dynamic batching: 70.8s (746 SGD steps)
With batch sorting: 233.6s (5477 SGD steps)

so 3.22x speedup over normal batching., even though we're seeing a 7.3x reduction in training steps, and things are are even slightly slower with batch sorting. There is clearly a high overhead of this algorithm in terms of doing the preprocessing and selecting the batch. I suspect getting preprocessing off the main thread will be the key to the next step function in performance.

stephenroller:
Another data point:

Overhead only (randomly select fixed batches):
[1x sgd updates, 1x wall]
```
[ time:233.0s total_exs:131448 epochs:1.0 time_left:0s ] {'exs': 48, 'lr': 0.0001, 'total_train_updates': 5477, 'gnorm': 2.081, 'clip': 1.0, 'gpu_mem_percent': 0.0994, 'loss': 8.746, 'token_acc': 0.2864, 'nll_loss': 4.372, 'ppl': 79.19}
[ num_epochs completed:1.0 time elapsed:233.35133862495422s ]
```

[1x SGD updates, 0.92x wall reduction] Batch sorting:
```
[ time:253.0s total_exs:131448 epochs:1.0 time_left:0s ] {'exs': 672, 'lr': 0.0001, 'total_train_updates': 5477, 'gnorm': 2.107, 'clip': 1.0, 'gpu_mem_percent': 0.104, 'loss': 120.2, 'token_acc': 0.281, 'nll_loss': 4.324, 'ppl': 75.51}
[ num_epochs completed:1.0 time elapsed:253.21780014038086s ]
```

The patch as is (full dynamic batching with greedy packing):
[7.3x step reduction, 3.34x wall speedup]
```
[ time:69.0s total_exs:131561 epochs:1.0 time_left:0s ] {'exs': 3584, 'lr': 0.0001, 'total_train_updates': 744, 'gnorm': 0.9361, 'clip': 1.0, 'gpu_mem_percent': 0.313, 'loss': 103.7, 'token_acc': 0.2349, 'nll_loss': 4.893, 'ppl': 133.3}
[ num_epochs completed:1.0 time elapsed:69.85396957397461s ]
```

@aszlam's suggestion of doing greedy packing without sorting:
[3.7x step reduction, 2.2x wall speedup]
```
[ time:106.0s total_exs:131469 epochs:1.0 time_left:0s ] {'exs': 882, 'lr': 0.0001, 'total_train_updates': 1465, 'gnorm': 1.32, 'clip': 1.0, 'gpu_mem_percent': 0.163, 'loss': 46.9, 'token_acc': 0.2556, 'nll_loss': 4.681, 'ppl': 107.9}
[ num_epochs completed:1.0 time elapsed:106.76953077316284s ]
```

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+        # check some assumptions
+        if isinstance(
+            world, (ExecutableWorld, BatchWorld, HogwildWorld, MultiAgentDialogWorld)
+        ):
+            raise TypeError(
+                'World must be a DialogPartnerWorld or a '
+                'MultiWorld of DialogPartnerWorld'
+            )
+
+        if not hasattr(world, 'agents'):
```
emilydinan: hmm.. when would this happen?

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+            )
+
+        # TODO: check to ensure the agent has self_observe
+        shared = world.share()
+        self.world = world
+        # TODO: maybe generalize this
+        self.max_size = self.truncate * opt['batchsize']
+
+        # buffer worlds
+        self.worlds = [
```
emilydinan: that's a lotta worlds 😮

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+
+    def num_examples(self):
+        return self.world.num_examples()
+
+    def num_episodes(self):
+        return self.world.num_episodes()
+
+    def _ceil(self, n, r=4):
+        """
+        Round to the nearest multiple of 4.
```
emilydinan: why make `r` an argument if we only ever use it for 4?

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+
+        TensorCores only work when a tensor is a multiple of 4 in all dimensions. This
+        means all examples cost is related to their nearest multiple of 4.
+        """
+        # round up to 4, all things are equal
+        return ((n + r - 1) // r) * r
+
+    def _score(self, obs):
+        if 'text_vec' in obs:
+            # note that all examples have a cost that is based on their
```
emilydinan: nice

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+        assert not self.epoch_done()
+        indices = []
+        for i in range(self._BUFFER_SIZE):
+            if self._scores[i] is not None:
+                indices.append(i)
+                continue
+            if self.worlds[i].epoch_done():
+                continue
+
+            act = self.worlds[i].agents[0].act()
```
emilydinan: why do we do this parley call manually? can we just call `self.worlds[i].parley()`?

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+                indices.append(i)
+
+        assert len(indices) != 0, "DynamicBatchWorld ran out of data!"
+
+        for i in indices:
+            assert self._scores is not None
+
+        # sort all the indices by their score, so that we can find similarly lengthed
+        # items in O(1)
+        indices = sorted(indices, key=lambda i: (self._scores[i], random.random()))
```
emilydinan: what is the random for? just for shuffling in the case of a tie?

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+        # indices = sorted(indices, key=lambda i: (random.random()))
+
+        # now let's build the batch
+        batch = []
+
+        # start with a random item. index_index is the lookup into indices, but
+        # index is the actual world.
+        width = 0
+        index_index = random.randint(0, len(indices) - 1)
```
emilydinan: this is a confusing name

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+
+        # we picked a random spot, but we can get better packing if we start at the
+        # last example with the same score, since we always pick move down to smaller examples.
+        while index_index < len(indices) - 1 and (
+            self._scores[indices[index_index]] == self._scores[indices[index_index + 1]]
+        ):
+            index_index += 1
```
emilydinan: dumb q: why aren't we doing index_index -= 1?

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+                width = new_width
+                batch.append(index)
+                indices.pop(index_index)
+                index_index = max(index_index - 1, 0)
+            else:
+                # we'd overfill our buffer, give up
+                break
+
+        # double check our assumed invariant
+        assert self._ceil(width) * self._ceil(len(batch)) <= self.max_size
```
emilydinan: is it even possible that this wouldn't be true?

parlai/core/worlds.py
```diff
@@ -975,6 +975,202 @@ def shutdown(self):
...
+        ):
+            index_index += 1
+
+        # quit early if we eat our full buffer
+        while indices:
+            index = indices[index_index]
+            this_width = self._ceil(self._scores[index])
+            new_width = max(width, this_width)
+            # compute the cost of the new batch
+            new_size = new_width * self._ceil(len(batch) + 1)
```
emilydinan: why the +1 here?  arent' you already rounding up?

parlai/core/worlds.py
stephenroller: Because I used think it was 8. :D

parlai/core/worlds.py
stephenroller: Because that's a full round trip, and we only want HALF of a parley here.

parlai/core/worlds.py
stephenroller: Yes, maximize stochasticity for SGD reasons

parlai/core/worlds.py
stephenroller: sorted for shortest to longest. shortest examples are at index 0, longest at the end. We pick a point and then count down. Here though, we find the last spot with the equivalent cost so that we can maximize packing. I actually experimented with this and in my experiments it took it from like 5.5x to 6.5x.

parlai/core/worlds.py
stephenroller: no, it's more there for my sanity to annotate the invariant condition

parlai/core/worlds.py
stephenroller: we haven't added this example yet, and this example MIGHT bump us up to the next multiple of 4. (Consider when len(batch) is 4 and we want to add the 5th example)

parlai/core/dict.py
```diff
@@ -292,6 +292,9 @@ def __init__(self, opt: Opt, shared=None):
                 # don't check isfile first, should fail if file not found
                 self.load(opt['dict_initpath'])
 
+        # cache unk token for later
```
stephenroller: All the changes in this file caused small speed ups in tokenization. They're more pronounced in dynamic batching, since we may have to tokenize more examples between each parley, causing more overhead per GPU-hit.

parlai/core/teachers.py
```diff
@@ -865,7 +872,6 @@ def _data_generator(self, data_loader, datafile):
         while True:
             for episode in self._read_episode(data_loader(datafile)):
                 yield episode
-            yield -1
```
stephenroller: (This was a doozie, took me like 3 days)

This was previously used to signal the end of streaming data, even when cycling (looping forever). This signal would then get used to indicate epoch_done, which was used by build_dict to stop.

In normal batching, that's okay. We'll have one item in the batch be invalid, and move on with our lives. However, in dynamic batching, we have a large buffer of in-progress worlds. Reaching the end of the data marks that world clone as done, but it becomes difficult to signal to all the others to stop. 

If you just get rid of this, then build dict runs for ever. If you try to cleverly keep track of all the unseen examples, then after your first epoch you can end up with dynamic batches of 1 forever, as no world ever moves forward (they're all stuck at epoch done). If you drop the check in dynamic batching for epoch_done (thus pushing them all forward regardless), then you end up double-observing the filler examples, causing the self-observe invariants to fail.

So my solution is just to actually carve out build_dict's exception properly. It's confusing that we ever had this sort of dual loop-forever-but-also-skip-once-to-mark-done.

stephenroller:
Still has a bad combination with --validation-max-exs., happy to take suggestions for that.

parlai/core/teachers.py
```diff
@@ -927,6 +933,8 @@ def get(self):
                 if self.entry_idx == 0:
                     self.cur_episode = self.next_episode
                     self.next_episode = next(self.data)
+                if self.cur_episode == -1:
```
stephenroller: all this code can be simplified to this:

```python
if self.cur_episode is None:
    # first go around, always read off the episode
    self.cur_episode = next(self.data)
if self.cur_episode == -1:
    # we're done here, but try to prep for next time
    return {'episode_done': True}, True
entry = self.cur_episode[self.entry_idx]
table = self.build_table(entry)
episode_done = self.entry_idx == len(self.cur_episode) - 1
table['episode_done'] = episode_done
if episode_done:
    self.cur_episode = next(self.data)
    self.entry_idx = 0
else:
    self.entry_idx += 1
return table, self.cur_episode == -1
raise ValueError
```

If you think this is better than the current next_episode madness, then we should update this change.

parlai/core/dict.py
```diff
@@ -378,6 +381,14 @@ def __contains__(self, key):
         elif type(key) == str:
             return key in self.tok2ind
 
+    def _word_lookup(self, key):
+        # return index from token, or unk_token's index, or None
+        return self.tok2ind.get(key, self._unk_token_idx)
+
+    def _int_lookup(self, key):
```
klshuster: perhaps `_ind_lookup`?

parlai/core/teachers.py
klshuster: this whole `-1` thing is so bizarre, could we have this like defined as a constant or a type? like `self.cur_episode == END_OF_EPOCH` or something?

klshuster: assuming the `raise ValueError` here is a typo?

klshuster: i think this solution is better than the `next_episode` madness btw

parlai/core/torch_generator_agent.py
```diff
@@ -571,7 +572,7 @@ def train_step(self, batch):
                 print(
                     '| WARNING: ran out of memory, skipping batch. '
                     'if this happens frequently, decrease batchsize or '
-                    'truncate the inputs to the model.'
+                    'truncate the inputs to the model.: ' + str(e)
```
klshuster: nit: remove period if putting a colon here

parlai/core/worlds.py
```diff
@@ -975,6 +975,209 @@ def shutdown(self):
...
+                'World must be a DialogPartnerWorld or a '
+                'MultiWorld of DialogPartnerWorld'
+            )
+
+        if len(world.get_agents()) != 2:
+            raise AssertionError(
+                "Dynamic batch only works in a fixed dialog world with two agents."
+            )
+
+        if not hasattr(world.get_agents()[1], 'batch_act'):
```
klshuster: we have a function on worlds `get_task_agent`, should we have one for `get_model_agent` as well? rather than accessing `get_agents()[1]` directly

parlai/core/worlds.py
```diff
@@ -975,6 +975,209 @@ def shutdown(self):
...
+
+        # size of the buffer we will use to find worlds
+        self._BUFFER_SIZE = 1021  # chosen as a prime number
+
+        # TODO: check to ensure the agent has self_observe
+        shared = world.share()
+        self.world = world
+        # TODO: maybe generalize this
+        self.max_size = (self.l_truncate + self.truncate) * opt['batchsize']
+        self.rng = random.Random(4)
```
klshuster: i think this declaration is unnecessary as it is instantiated in `.reset()` which is called a few lines down

parlai/core/worlds.py
```diff
@@ -975,6 +975,209 @@ def shutdown(self):
...
+            if self._scores[i] is not None:
+                indices.append(i)
+                continue
+            if self.worlds[i].epoch_done():
+                continue
+
+            if hasattr(self.world, 'parley_init'):
+                self.worlds[i].parley_init()
+
+            act = self.worlds[i].get_agents()[0].act()
```
klshuster: u can use `get_task_agent` here no?

parlai/core/worlds.py
```diff
@@ -975,6 +975,209 @@ def shutdown(self):
...
+                batch.append(index)
+                indices.pop(index_index)
+                index_index = max(index_index - 1, 0)
+            else:
+                # we'd overfill our buffer, give up
+                break
+
+        # Always have a batch size that's a multiple of 4, for fp16's sake.
+        while len(batch) > 4 and len(batch) % 4 != 0:
+            batch.pop(0)
```
klshuster: why pop the beginning? is this getting rid of the largest examples?

parlai/core/worlds.py
```diff
@@ -975,6 +975,209 @@ def shutdown(self):
...
+        # update metrics
+        self.total_parleys += 1
+        self.total_exs += len(batch)
+
+    def get_total_epochs(self):
+        return self.total_exs / self.num_examples()
+
+    def report(self):
+        report = self.world.report()
+        # override that world's view of 'exs'
```
klshuster: is this a `TODO`?

tests/test_dynamicbatching.py
```diff
@@ -0,0 +1,88 @@
...
+    'embedding_size': 32,
+    'truncate': 8,
+    'model': 'transformer/generator',
+    'task': _TASK,
+    'no_cuda': True,
+}
+
+
+class TestDynamicBatching(unittest.TestCase):
+    def _test_correct_processed(self, num_goal, **kwargs: Dict[str, Any]):
```
klshuster: is `num_goal` just an `int` here?

parlai/core/teachers.py
stephenroller: Okay, I'll propose the new change.

parlai/core/worlds.py
stephenroller: Okay, I like that better.

parlai/core/worlds.py
stephenroller: Yeah, I don't think it matters what you pop really. Just trying to round it down to a multiple of 4. We could remove the shortest (they are easier to pack later...)

parlai/core/teachers.py
emilydinan: agree with @klshuster on the -1 and that this is better. i also find `build_table` to be a confusing name for just creating the next observation/action dict.

emilydinan: also do we need that `_lock()` still

parlai/core/worlds.py
emilydinan: yes, i find this confusing in worlds everywhere where we have `agents[0]` and `agents[1]` and we are just supposed to know which is which

parlai/core/worlds.py
klshuster: That was my thought too, that it'd be easier to pack those in later (i.e. let's get rid of the longer episodes first)

parlai/core/worlds.py
```diff
@@ -975,6 +975,209 @@ def shutdown(self):
...
+        # round up to 4, all things are equal
+        return ((n + r - 1) // r) * r
+
+    def _score(self, obs):
+        if 'text_vec' in obs:
+            # note that all examples have a cost that is based on their
+            # nearest multiple of 4. We can therefore mix-and-match
+            # anything with the same cost for increased stochasticity,
+            # while not really wasting much padding.
+            return tuple(
```
emilydinan: how does this play with label candidates for ranking models?

emilydinan: i.e. even if we train with batch candidates, we typically evaluate with inline label candidates, so the memory usage woudl go way up

stephenroller: Probably very poorly. Worth testing

parlai/core/worlds.py
stephenroller: Went with indices_idx.

parlai/core/teachers.py
stephenroller: Ya I think we probably do.

parlai/core/worlds.py
stephenroller: This has been tested.

tests/test_pytorch_data_teacher.py
```diff
@@ -309,10 +309,6 @@ def test_pyt_train(self):
     def test_pyt_train_stream(self):
         self._pyt_train('train:stream')
 
-    @testing_utils.retry()
```
stephenroller: this was hanging but we're deleting pytd ASAP so I think just whatever.

parlai/core/torch_generator_agent.py
```diff
@@ -635,12 +635,10 @@ def train_step(self, batch):
...
         """
-        if batch.text_vec is not None:
-            batchsize = batch.text_vec.size(0)
-        elif batch.image is not None:
-            batchsize = len(batch.image)
         # helps with memory usage
-        self._init_cuda_buffer(batchsize, self.truncate or 256)
+        # note we want to use the opt's batchsize instead of the observed batch size
+        # in case dynamic batching is in use
+        self._init_cuda_buffer(self.opt['batchsize'], self.label_truncate or 256)
```
klshuster: why did this change to label_truncate?

parlai/core/worlds.py
```diff
@@ -317,6 +323,12 @@ def get_task_agent(self):
         """
         return self.get_agents()[0]
 
+    def get_model_agent(self):
```
klshuster: this is great, I actually implemented this exact thing in my preprocess branch, will just pull in these changes once merged

parlai/core/worlds.py
```diff
@@ -974,6 +1004,216 @@ def shutdown(self):
...
+
+        # TODO: check to ensure the agent has self_observe
+        shared = world.share()
+        self.world = world
+        # TODO: maybe generalize this
+        self.max_words = (self.l_truncate + self.truncate) * opt['batchsize']
+
+        # buffer worlds
+        self.worlds = [
+            shared['world_class'](opt, None, shared) for _ in range(self._BUFFER_SIZE)
```
klshuster: you mentioned this in my pr - do we want to maintain this `(opt, None, shared)` initialization?

parlai/core/worlds.py
```diff
@@ -974,6 +1004,216 @@ def shutdown(self):
...
+
+    def _ceil(self, n, r=4):
+        """
+        Round to the nearest multiple of 4.
+
+        TensorCores only work when a tensor is a multiple of 4 in all dimensions. This
+        means all examples cost is related to their nearest multiple of 4.
+        """
+        # round up to 4, all things are equal
+        return ((n + r - 1) // r) * r
```
klshuster: do we anticipate using this function for `r !=4`?

parlai/core/worlds.py
```diff
@@ -974,6 +1004,216 @@ def shutdown(self):
...
+        # now let's build the batch
+        batch = []
+
+        # start with a random item. indices_idx is the lookup into indices, but
+        # index is the actual world.
+        width = 0
+        indices_idx = random.randint(0, len(indices) - 1)
+
+        # we picked a random spot, but we can get better packing if we start at
+        # the last example with the same score, since we always pick move down
```
klshuster: nit: some grammar issues with the second sentence here

tests/test_dynamicbatching.py
```diff
@@ -0,0 +1,136 @@
...
+    'n_heads': 1,
+    'ffn_size': 32,
+    'embedding_size': 32,
+    'truncate': 8,
+    'model': 'transformer/generator',
+    'task': _TASK,
+}
+
+
+# TODO tests to write:
```
klshuster: nit: looks like you can get rid of this TODO?

parlai/core/torch_generator_agent.py
stephenroller: It should have always been label truncate.

parlai/core/worlds.py
stephenroller: agreed, will change to match yours

parlai/core/worlds.py
stephenroller: I'm still not sure if it's 4 or 8 :)

stephenroller: Keeping it as 8, pulling it out into a constant in the utilities.

tests/test_dynamicbatching.py
stephenroller: i think it's not complete, i'd rather keep it.

parlai/core/torch_generator_agent.py
stephenroller: Actually, I'm wrong. Hm.

PR #2278, opened 2019-12-17
Loaders for tasks, teachers, agents, and world

emilydinan:
**Patch description**
Add a loader file into core which contains functions for loading modules for tasks, teachers, agents, and worlds.

This simplifies and removes some of this logic from other places like params, agents, and worlds.

It also adds functionality for world to have command line args, with a proof of concept added for the interactive world for ConvAI2.

Left to do:
- [x] Fill out docstrings and add documentation
- [x] Add tests
- [ ] Fix imports in internal.


parlai/core/agents.py
```diff
@@ -519,7 +499,7 @@ def compare_init_model_opts(opt: Opt, curr_opt: Opt):
         print('*' * 75)
 
 
-def load_agent_module(opt: Opt):
+def create_agent_from_file_opt(opt: Opt):
```
klshuster: super nit: for some reason i feel like `create_agent_from_opt_file` sounds better here, unless there's something i'm missing

parlai/core/params.py
```diff
@@ -809,14 +809,26 @@ def add_task_args(self, task):
...
+    def add_world_args(self, task, interactive_task):
+        """
+        Add arguments specific to the world.
+        """
+        world_class = load_world_module(task, interactive_task)
+        if world_class is not None and hasattr(world_class, 'add_cmdline_args'):
+            try:
+                world_class.add_cmdline_args(self)
+            except argparse.ArgumentError:
+                # already added
```
klshuster: does this mean that the world is trying to add arguments that have already been added to the argparser?

parlai/core/loader.py
```diff
@@ -0,0 +1,325 @@
...
+        world_class = getattr(core_worlds, world_name)
+    else:
+        return None
+
+    return world_class
+
+
+def load_world_module(
+    taskname: str,
+    interactive_task: bool,
```
klshuster: another nit: maybe we could `interactive_task` be an optional argument, as in the test below it is a little confusing that this is just `True` or `False`

parlai/core/agents.py
emilydinan: yes it definitely does lol...idk what i was thinking

parlai/core/params.py
emilydinan: yes. i'm not sure when this happens, but it was in the previous code

parlai/core/loader.py
emilydinan: i think it should just default to `False`, yes

PR #2286, opened 2019-12-18
[DocDay] Agent README Updates

klshuster:
**Patch description**
Lots of fun Agent Docs Updates!!

1. Fill in READMEs for missing agents
2. Autogenerate Docs with Agent READMEs
3. Move `TransformerClassifierAgent` from `transformer_classifier` to `transformer/classifier`

**Testing steps**
CircleCI (?)

**Logs**
```
docs$ make html
. . .
build succeeded.
```

**Screenshots**
Screenshot 2019-12-18 13 53 47


klshuster:
Update: per @stephenroller suggestions, each agent is now it's own html page. This does indeed force a TOC entry for each agent, which I am not sure is too bad of a thing...


Screenshot 2019-12-18 15 41 14

docs/source/generate_agent_list.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Load agent readme, add title if necessary.
+
+    :param agent:
+        string indicating agent module
+
+    :return:
+        agent's readme
+    """
+    readme_path = f'../../parlai/agents/{agent}/README.md'
```
stephenroller: wouldn't mind if this imported parlai to find the actual path, but I don't think this is the only spot that assumes you're running the doc builder in this particular folder.

docs/source/generate_agent_list.py
```diff
@@ -0,0 +1,102 @@
...
+    os.makedirs('agent_refs', exist_ok=True)
+    agents = [
+        name
+        for _, name, _ in pkgutil.iter_modules(
+            [os.path.dirname(parlai.agents.__file__)]
+        )
+    ]
+    for agent in agents:
+        with open(f'agent_refs/{agent}.rst', 'w') as fout:
+            write_agent(fout, prepare_agent_readme(agent))
```
stephenroller: do you want to try with .md to see if it looks substantially different? There is some probability that some MD will be lost in your simple translation.

stephenroller:
@jaseweston I wonder if it would make sense to do this for projects too, unifying our view.

docs/source/generate_agent_list.py
klshuster: indeed it is not the only place, hence why i wrote it this way. i'll look into whether i can import that path

docs/source/generate_agent_list.py
klshuster: sure i'll give it a try

docs/source/generate_agent_list.py
klshuster: indeed it is not the only place, hence why i wrote it this way. i'll look into whether i can import that path

docs/source/generate_agent_list.py
klshuster: Screenshot 2019-12-19 11 44 16
 it turns out pretty poorly from the screenshot. I'll leave as is for now. There is definitely a greater than 0% chance I am losing some MD but I am willing to take that risk

docs/source/generate_agent_list.py
klshuster: yeah i found a way to just import that path so i changed this

PR #2289, opened 2019-12-18
Interactive  + Self-chat + Pretty display docs

jaseweston:
Some more docs!! (see title)


docs/source/tutorial_tipsntricks.rst
```diff
@@ -44,6 +44,39 @@ That is, by adding a colon ":" followed by the flag name, an equals sign, and th
 You can add multiple flags, all separated by ":".
 
 
+Self-Chats
+##########
+
+Sometimes it is useful to generate models talking to themselves. You can do this with:
```
klshuster: could you maybe elaborate a bit on why you would specify something like `convai2:selfchat`? and perhaps include other tasks for which this is a thing?

docs/source/tutorial_tipsntricks.rst
```diff
@@ -44,6 +44,39 @@ That is, by adding a colon ":" followed by the flag name, an equals sign, and th
...
+.. code-block:: python
+
+   # Self-chatting Poly-Encoder model on a generic task (so e.g., no ConvAI2 personas are input)
+   python parlai/scripts/self_chat.py -mf zoo:pretrained_transformers/model_poly/model -t self_chat --inference topk -ne 10 --display-examples True -dt valid
+
+
+Prettifying Display of Chats
+############################
+
+This handy script can prettify the display of json file of chats (sequences of parlai messages):
```
klshuster: does the data need to be in a certain format to be prettified?

jaseweston: it does but i think we need to fix that to be more permissive of any json of parlai messages, so i didn't really want to document the current given that it should be fixed. what i do give are example files so they can be looked at..

docs/source/tutorial_tipsntricks.rst
jaseweston: I have written that below:

The task set by '-t' (in the above case "convai2:selfchat") links to a parlAI world that handles the particular nature of interactions, see e.g. `here <https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/convai2/worlds.py#L98>`_ 
or `here <https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/wizard_of_wikipedia/worlds.py#L106>`_.

klshuster: not sure how i missed that, sorry

docs/source/tutorial_tipsntricks.rst
klshuster: works for me

docs/source/tutorial_tipsntricks.rst
jaseweston: well it was kind of there, but i updated it also!

PR #2303, opened 2019-12-20
Update EmpatheticDialogues teachers

EricMichaelSmith:
**Patch description**
- Updating EmpatheticDialogues teacher with the current parlai-internal version, which utilizes inline cands and allows for the option to train on either both sides of the conversation or on only the side where the Speaker (i.e. Experiencer) is speaking and the Listener (i.e. Responder) is responding
- Adds an `EmotionClassificationSituationTeacher` to classify emotion labels of conversations given situation strings. Removes a similar but unused teacher
- Adds unit tests for both teachers

**Testing steps**
- Displaying data from teachers:
```
python examples/display_data.py -t empathetic_dialogues
python examples/display_data.py -t empathetic_dialogues:EmotionClassificationSituationTeacher
```
- Reproducing ParlALL result (11.18 ppl and 19.34 f1):
```
python -u examples/eval_model.py -t empathetic_dialogues -bs 16 --dict-file /checkpoint/kshuster/projects/parlall/dict/parlall_agent.dict --model-file /checkpoint/kshuster/projects/parlall/parlall_multi_12/parlall_multi_12_sweep10_Fri_Oct_25/5eae/model --skip-generation false --metrics all --inference beam --beam-size 5 --beam-min-length 10 --beam-block-ngram 3
```
- Reproducing all-in-one-dialogue result (51.97% hits@1):
```
python -u examples/eval_model.py \
-dt test \
-t parlai.tasks.empathetic_dialogues.agents:EmpatheticDialoguesTeacher \
--batchsize 16 \
--model-file /checkpoint/ems/all_in_one_dialogue/scripts/s2019_11_19__aio_context_change/01_re_fine_tune/005/34a05_jobid=24/model
```
Result of running `python tests/datatests/test_new_tasks.py`:
```
(py37) ems@devfair0229:~/GitHub/facebookresearch/ParlAI$ python tests/datatests/test_new_tasks.py
.
Ran 1 test in 61.368s

OK
```
Result of running `python setup.py test -s tests.suites.datatests`:
```
test_verify_data (test_new_tasks.TestNewTasks) ... ok

Ran 1 test in 59.750s

OK
```


EricMichaelSmith:
If I'm understanding correctly, it looks like the "datatests" CI check above *won't* pass, and that (given what the error log says) I should instead run `python tests/datatests/test_new_tasks.py`: I've done so and added the results to the PR message above

EricMichaelSmith:
> Love the teacher test. How can we generalize it?

Thanks @stephenroller ! Good question about generalization. I have two tests, one that checks the number of episodes/examples and one that checks the value of `.get()` for a specific example. I think one way to generalize this would be to create two scripts:
1. A setup script that someone can run to print the result of `Teacher.get()` for a specific example, as well as to print the actual number of episodes and examples for all combinations of values of the specified flags
1. A generalized teacher test script that will loop through all supported teachers and, for each teacher, check its episode/example counts and sample `.get()` outputs against the results of the first script
I've added an issue for this at https://github.com/facebookresearch/ParlAI/issues/2304, since I'm definitely onboard with us streamlining this process.

PR #2312, opened 2020-01-03
GitHub Actions flake8

stephenroller:
**Patch description**
As of January 1, 2020, we now have access to [GitHub Actions](https://github.com/features/actions)! This is GitHub's new CI/CD system, and a potential replacement to CircleCI. It has some special features available, like it can annotate PRs inline with the error messages, which has a nice application for linting and type-checking.

This PR:
- adds support for doing flake8/type-checking in ParlAI using @suo's awesome annotation system for adding the annotations [pytorch/add-annotations-github-action](https://github.com/pytorch/add-annotations-github-action).
- deletes relevant parts of the CircleCI config that are no longer needed.
- adds the `flake8-black` plugin so that "not formatted with black" is a lint error and shows the problematic areas
- And just for fun, I learned how to write custom flake8 plugins, so now copyright header checks and shebang checks happen in our little flake8 plugin! They have the prefix `PAI`.
- Also checks the docstring formatting as well.
- New mypy logic so that only NEW mypy errors are shown (comparing to master/HEAD). These are presented as errors, but can be subsequently ignored later if not addressed immediately.

The type errors present in this are fixed in #2318.

**Testing steps**
85 attempts at writing this PR. :)
Submitting this PR for review with intentional lint and typing issues. The reviewer should verify it looks cool and then I'll revert it before landing.

Here's a screenshot of what the UI looks like when there's an error:
image


stephenroller:
image

omg so dope

.github/workflows/lint.yml
```diff
@@ -0,0 +1,123 @@
+name: Lint
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+
+jobs:
+  lint:
```
EricMichaelSmith: Ha, it looks like the new YML settings are a lot more verbose than before. But it'd be good to have inline linting and type-checking messages, yes!

parlai/utils/flake8.py
```diff
@@ -0,0 +1,89 @@
...
+- Checks for python3 shebang
+- Check for copyright message
+- Docformatter issues (TODO)
+"""
+
+import docformatter
+import difflib
+
+PYTHON_SHEBANG = '#!/usr/bin/env python3'
+WHITELIST_PHRASES = ['Moscow Institute of Physics and Technology.']
```
EricMichaelSmith: Ha - what is this whitelist phrase used for? Was there a case in which we had to make an exception for this?

parlai/utils/flake8.py
```diff
@@ -0,0 +1,89 @@
...
+                f'before/{self.filename}',
+                f'after/{self.filename}',
+                n=0,
+                lineterm='',
+            )
+            for line in diff:
+                if line.startswith('@@'):
+                    fields = line.split()
+                    line_no, _ = fields[1].split(',')
+                    line_no = -int(line_no)
```
EricMichaelSmith: ha, I suppose "-" can mean both negation and how GitHub uses it to notate diffs. Maybe it's worth adding a comment here to explain this?

parlai/utils/flake8.py
stephenroller: yeah there's some code owned by moscow in parlai

PR #2317, opened 2020-01-03
[Loader] test internal agent load

emilydinan:
**Patch description**
Test loading an internal agent. Copies from example_parlai_internal and then cleans up.

### DIFF
--- a/tests/test_loader.py
+++ b/tests/test_loader.py
@@ -73,9 +73,9 @@ class TestLoader(unittest.TestCase):
         shutil.copytree(
             'example_parlai_internal/agents/parrot', 'parlai_internal/agents/parrot'
         )
-        agent_module = load_agent_module('internal:parrot')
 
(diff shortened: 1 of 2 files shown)


tests/test_loader.py
```diff
@@ -53,6 +53,36 @@ def test_load_dialog_partner_world(self):
         )
         self.assertEqual(world_module, DialogPartnerWorld)
 
+    def test_load_internal_agent(self):
+        def cleanup(parlai_internal_exists, agent_folder_exists):
```
stephenroller: the right way to do this i think is:

```
class TestParlaiInternal(unittest.TestCase):
    def setUp(self):
        # do the copying here
    def tearDown(self):
        # do the rm'ing here
    def test_load_internal_agent(self):
        # setup will be run before everything named test_, and teardown will be run after
```

PR #2319, opened 2020-01-05
Delay import of tensorboardX.

stephenroller:
**Patch description**
TensorboardX is a rather heavy package, that imports things like botocore etc. We're currently importing this even on basic tests, due to it being imported by train loop. We can slightly improve startup time by delaying this import until it's actually used.

**Testing steps**
CI.


dexterju:
LGTM, Thanks for doing this!

Issue #2162, opened 2019-11-18
Metrics revisiting


stephenroller:
# Overview

This is a grand issue, that will probably need to be fixed with multiple PRs. Self-reported metrics are unreliable. They have a number of issues:

- We cannot map back to individual examples. As such, at training time, we don't have any way of knowing things like `ppl` on a given task
- At validation time, we solve this by running each task in serial.
- We aren't sure how to combine them in a distributed setting. As a result, we opt to run validation only on the primary worker, while all the secondary workers just sit idle. This is extremely wasteful of expensive GPU time. If we could fix the metrics averaging, then we could easily shard the validation set across all the workers

# Part 1: Metric object

We need a generic Metric object, as well as a few variants. It should define a very simple interface:

```python
class Metric(object):
    @abstractmethod
    def report(self) -> Number:
         pass

    def __str__(self) -> str:
        return '{:.4g}'.format(self.report())

    def __repr__(self) -> str:
        return '{}({:.4g})'.format(str(self.__class__), self.report())

class SumMetric(Metric):
    # keeps a running sum
    def __init__(self, value: Number)
    def __add__(self, other: SumMetric) -> SumMetric
    def report(self) -> Number

class AverageMetric(Metric):
    def __init__(self, value: Number, denom: Integer)
    def __add__(self, other: AverageMetric) -> AverageMetric
         # Combines two metrics. This should create a new metric, which has the
         # numerator and denominator combined
    def report(self) -> Number
        # should "render" the object, as in return 
 ```

Most things will just use AverageMetric. Something like PPLMetric can be implemented via a call to `exp()` in the `report()` call.

The point of having clear semantics for *combining* metrics

## Requirements

- [x] I believe Metric should be immutable. We should encourage *combining* metrics rather than editing them.
- [x] Metric should be happy taking item tensors and converting them. It should throw an error if they aren't scalars though.
- [x] Metric should live in metrics.py
- [x] Metric should be strongly typed.
- [x] Metric should be well-documented
- [x] Must have unit tests

Maybe:
- [ ] Define comparators like > or < for Metrics, it could be super convenient in train_model.

# Part 2: Cleanup of metrics.py

With the AverageMetric well defined, we can now use it in metrics.py to clean up a lot of the tracking behavior. All of the `_cnt` madness in the current stuff should be removed and added for Metrics.

# Part 3: Handling metrics

[...]

PR #2329, opened 2020-01-09
Creating Metric object

EricMichaelSmith:
**Patch description**
Create a Metric object for the project to overhaul metrics (https://github.com/facebookresearch/ParlAI/issues/2162). Metrics are immutable, are combined using `Metric.__add__()`, and return floats with the `Metric.report()` method. I include some base Metric subclasses, SumMetric and AverageMetric, which are useful for aggregating sums and averages across multiple GPUs, hosts, worlds, etc.

Also:
- Moving `Teacher`, `MultiTaskTeacher`, and `AssignState` in order to avoid having MTurk unit test import torch
- Fix bAbI teacher to specify default task num

**Testing steps**
`python tests/test_metrics.py`
- Added PyTorch to MTurk CI check because `metrics.py` now imports it


parlai/core/metrics.py
```diff
@@ -42,6 +46,86 @@
...
+        if isinstance(sum_, torch.Tensor):
+            sum_ = sum_.item()
+        assert isinstance(sum_, Number)
+        sum_ = float(sum_)
+
+        self.sum = sum_
+
+    def __add__(self, other: 'SumMetric') -> 'SumMetric':
+        # NOTE: hinting can be cleaned up with "from __future__ import annotations" when
+        # we drop Python 3.6
```
stephenroller: oh? i'm real curious about this, what's this?

EricMichaelSmith: That's from https://stackoverflow.com/questions/41135033/type-hinting-within-a-class (see the second answer). In Python <= 3.6, you have to pass the class name as a string, but Python >= 3.7 allows for using the class name itself as the hinted type if you add `from __future__ import annotations`

parlai/core/metrics.py
```diff
@@ -42,6 +47,94 @@
...
+    """
+    Class that keeps a running average of some metric.
+    """
+
+    def __init__(
+        self, numer: TScalar, denom: TScalar,
+    ):
+
+        # Convert numer to float
+        if isinstance(numer, torch.Tensor):
```
emilydinan: looks like we do this "to number" operation many times throughout. can this just be utility function that we call? (perhps with options for int/float)

parlai/core/metrics.py
```diff
@@ -42,6 +47,94 @@
...
 
+class Metric(ABC):
+    """
+    Base class for storing metrics.
+
+    Subclasses should define .report().
+    """
+
+    @abstractmethod
+    def report(self) -> float:
```
klshuster: nit: could you please add a docstring here indicating what `report` does?

parlai/core/metrics.py
klshuster: could also be a function of the base `Metric` class

parlai/utils/typing.py
```diff
@@ -22,3 +24,8 @@ class _Shared(Dict[str, Any]):
 
 
 TShared = TypeVar('TShared', bound=_Shared)
+
+# ParlAI type to represent an object that is theoretically expressible as a scalar
```
klshuster: nit: could you perhaps follow the convention above of defining a class `_Scalar`? The reason for this is to have cleaner docstrings for these types (i.e. instead of `#` comments). Then you could have `TScalar = TypeVar('Tscalar', bound=_Scalar)`

parlai/core/metrics.py
```diff
@@ -10,13 +10,18 @@
...
-from parlai.utils.misc import warn_once
 from numbers import Number
+from typing import Union
 
-import re
+import torch
+
+from parlai.utils.thread import SharedTable
+from parlai.utils.misc import round_sigfigs, no_lock
+from parlai.utils.misc import warn_once
```
klshuster: nit: you can combine this import with the one above

.circleci/config.yml
```diff
@@ -318,6 +318,7 @@ jobs:
           key: deps-mturk-v4{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/lib"
+      - <<: *installtorchcpu37
```
stephenroller: This seems like a leak that we can avoid.

parlai/utils/typing.py
stephenroller: Yeah it's interesting. The typevar approach given here means that it can be used to indicate it's one of these types, but it's FIXED, and that cannot change. E.g. two references to TScalar in the same context have to either both be floats or both be ints or both be tensors.

I think for metrics we especially want to be able to mix/match so the pseudoclass which is a Union might be best. Also this excellent comment can then become a docstring.

emilydinan:
> What's everyone's opinion of the + syntax? There's also options for using like bitwise `|` but it's less obvious to me this is better.

I think I like the `+` syntax.

parlai/core/metrics.py
```diff
@@ -42,6 +47,94 @@
...
+
+        # Convert sum to float
+        if isinstance(sum_, torch.Tensor):
+            sum_as_number: Union[int, float, bool] = sum_.item()
+        else:
+            sum_as_number = sum_
+        assert isinstance(sum_as_number, Number)
+        sum_as_float = float(sum_as_number)
+
+        self.sum: float = sum_as_float
```
stephenroller: Can we make these private?

parlai/core/metrics.py
```diff
@@ -42,6 +47,95 @@
...
+        pass
+
+    def __str__(self) -> str:
+        return f'{self.value():.4g}'
+
+    def __repr__(self) -> str:
+        return f'{self.__class__} ({self.value():.4g})'
+
+    @staticmethod
+    def as_number(obj: Any) -> Number:
```
stephenroller: maybe we want as_float in addition to as_number?

Also obj should probably be explicitly a Tensor, a float, or an int.

It's shitty that mypy doesn't properly recognize float is a number. We might just want to use Union[float, int]. I do that elsewhere for the same reason (and we don't necessarily guarantee we can support complex numbers, for example)

parlai/mturk/core/agents.py
```diff
@@ -22,154 +24,6 @@
 # relative to heartbeats. This will come with more thorough testing.
 
 
-class AssignState:
```
stephenroller: lol please explain wtf is going on

EricMichaelSmith: It seems like this has to happen in order to get MTurk tests to not import torch =/ Honestly, I'm not sure if it's worth it

EricMichaelSmith: There are two ways that MTurk tests import torch, and moving AssignState to shared_utils.py fixes one of them

EricMichaelSmith: The other way can be fixed by moving the two `metrics` imports in `parlai.mturk.code.agents` into the specific methods that use them, because a lot of MTurk code relies on importing `Agent`. This is a lot of import/function shuffling, so lemme know if you still think it's worth it to avoid this torch import

parlai/core/metrics.py
EricMichaelSmith: Yes, agreed - added as_float() and as_int() methods, and I'm now more explicit with obj's type

parlai/utils/typing.py
EricMichaelSmith: Lemme know what you think about my current solution, which is to just create an alias of `Union[]` rather than a subclass, because you can't subclass `Union`. See https://docs.python.org/3/library/typing.html#user-defined-generic-types

parlai/tasks/self_feeding/agents.py
```diff
@@ -10,11 +10,9 @@
 
 import numpy as np
 
-from .build import build
-
-import parlai.core.agents as core_agents
-from parlai.core.teachers import ParlAIDialogTeacher
+from parlai.core.teachers import ParlAIDialogTeacher, MultiTaskTeacher
 from projects.self_feeding.utils import add_person_tokens
+from .build import build
```
stephenroller: huh, I'm curious about this one.

stephenroller:
don't bother running datatests, it will take 5 years to complete.

EricMichaelSmith:
> don't bother running datatests, it will take 5 years to complete.

Hmm, I've already started, but I won't make merging this PR dependent on that unless I find something obviously horrendous. I've already tweaked some small thing in babi because it was failing datatests

parlai/tasks/self_feeding/agents.py
EricMichaelSmith: Do you think it's worth changing? A lot of agents do relative imports like that

stephenroller: I mean i'm just a little surprised something was broken, i thought this was tested.

EricMichaelSmith: Hmm, nothing in this agent was broken - the same things are imported before and after this PR, but I think my IDE shuffled around the order of the import lines when I moved `MultiTaskTeacher` into a different module

PR #2327, opened 2020-01-08
Pytest + Speed improvements in unit tests.

stephenroller:
**Patch description**
Attempting to speed up the unit tests to improve developer feedback.

Changes:
- Switch to using pytests and having the output collected by CircleCI. We get lots of convenient meta data.
- Fix a bug in the TGA tutorial that accidentally ignored parameters; reduced some others.
- Delete a teacher that was 100% redundant with another. Deleted the one that was less used. Update the corresponding tests.
- A few things can be switched to evaluating pretrained models instead of training new ones and be safe imho.
- Catch a bunch of tests that were calling parse_args without explicit parameters so that they can be safe.
- Moved a few things around to be better compatible with pytests
- Added explicit empty command line args to some scripts, so the test versions do not actually use sys.argv.
- Update relevant READMEs.

**Testing steps**
CI timing.

PR #2333, opened 2020-01-10
[TRA tests] Speed up eval fixed cands test

emilydinan:
**Patch description**
Decrease number of epochs that the eval_fixed_cands test runs for. As far as I can tell this doesn't affect whether or not the test passes usually. For the three models that run this test, the speedups are as follows:

**Before**
```
[ Test test_eval_fixed completed in 10.13738s ]
.
[ Test test_eval_fixed completed in 18.37439s ]
..
[ Test test_eval_fixed completed in 14.78047s ]
```

**After**
```
[ Test test_eval_fixed completed in 2.79047s ]
........
[ Test test_eval_fixed completed in 6.67095s ]
.........
[ Test test_eval_fixed completed in 5.17016s ]
```

I also added a decorator to be able to easily track this when you are debugging. Idk if generally useful, up to @stephenroller ...

### DIFF
--- a/parlai/utils/testing.py
+++ b/parlai/utils/testing.py
@@ -16,6 +16,7 @@ import tempfile
 import shutil
 import io
 import signal
+import time
 from typing import Tuple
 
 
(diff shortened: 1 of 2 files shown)


parlai/utils/testing.py
```diff
@@ -133,6 +134,16 @@ def _wrapper(testself, *args, **kwargs):
         return _wrapper
 
 
+def print_test_time(func):
```
stephenroller: delete this, I have something more general in #2327.

PR #2341, opened 2020-01-16
Fix pretrained embeddings initialization

stephenroller:
**Patch description**
Last second of #2327, I made it so that modelzoo_path fails loudly if it couldn't find the zoo file. This is a welcome usability improvement (rather than silently picking an typoed path, it makes sure if you use the `zoo:` prefix, it really must be a pretrained model).

However, it turns out we were subtly depending on that behavior in the pretrained embeddings. Removing it caused an infinite loop when using some of the pretrained embeddings, and crashed when using the glove embeddings.

This broke master. This PR should fix it.

For simplicity of the fix, this also removes support for glove-twitter embeddings. From what I can tell, these were not generally used. If we want to bring them back, they should become a proper zoo file themselves.

**Testing steps**
CI, especially nightly.

```
$ python examples/train_model.py -m legacy:seq2seq:0 -t convai2 -emb glove -mf /tmp/gtest
$ python examples/train_model.py -m seq2seq -t convai2 -emb glove -mf /tmp/gtest2
$ python examples/train_model.py -m seq2seq -t convai2 -emb fasttext_cc -mf /tmp/ftest2
$ python examples/train_model.py -m seq2seq -t convai2 -emb fasttext -mf /tmp/ftest3
```

### DIFF
--- a/parlai/agents/legacy_agents/seq2seq/seq2seq_v0.py
+++ b/parlai/agents/legacy_agents/seq2seq/seq2seq_v0.py
@@ -282,7 +282,7 @@ class Seq2seqAgent(Agent):
             '-emb',
             '--embedding-type',
             default='random',
-            choices=['random', 'glove', 'glove-fixed', 'fasttext', 'fasttext-fixed',],
+            choices=['random', 'glove', 'glove-fixed', 'fasttext', 'fasttext-fixed'],
             help='Choose between different strategies '
             'for word embeddings. Default is random, '
             'but can also preinitialize from Glove or '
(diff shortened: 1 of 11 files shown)


stephenroller:
We can access them that way, but that would be recursive. modelzoo_path looks to call download. if download calls modelzoo_path, then it's a loop!

stephenroller:
Alright I think I got it all.

stephenroller:
(To clarify @emilydinan, you can refer to them this way, just not in the download function :D)

emilydinan:
> (To clarify @emilydinan, you can refer to them this way, just not in the download function :D)

Ah I see... So will you change the other paths then (that are outside of the `build` files)?

PR #2342, opened 2020-01-16
[Wiki task] add test and valid split

emilydinan:
**Patch description**
The `wikipedia:full` task was not split into train/test/valid. I added this split by declaring the last subdirectory as half test and half valid.

### DIFF
--- a/parlai/tasks/wikipedia/agents.py
+++ b/parlai/tasks/wikipedia/agents.py
@@ -34,15 +34,31 @@ class FullTeacher(DialogTeacher):
             opt['datapath'], 'wikipedia/full/wiki_full_extracted'
         )
         self.id = 'wikipedia'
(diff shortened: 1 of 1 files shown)


stephenroller:
What trains on this? I'm assuming we've only used this on things like wizard and drqa, where we don't want to split them.

stephenroller:
also TIL you can use `@` in your branch names... what's next, emojis?

stephenroller:
image

omg @hadasah you have to see this

hadasah:
@stephenroller start practicing your emoji interpretation, I'm never naming a branch with words again

emilydinan:
> What trains on this? I'm assuming we've only used this on things like wizard and drqa, where we don't want to split them.

I'm training on it.

PR #2344, opened 2020-01-17
Mark some mturk tests as flaky.

stephenroller:
**Patch description**
Since the upgrade to pytest, a few mturk tests have been flaky. They are always tests that depend on some condition being met within a certain time frame. I'm not sure why pytests changes it: maybe it has some background threads that are slowing it down.

I've opted to mark the tests that use time intervals with `retry` to improve their signal.

**Testing steps**
CI run 5 times.


stephenroller:
Testing results:
- Run 1 successful.
- Run 2 successful.
- ...

Issue #2185, opened 2019-11-19
Easier multi-gpu usage


stephenroller:
Right now multi-gpu stuff is a bit of a mess:
- To use multigpu with a generator, you should use distributed_train.py
- To use multigpu with a retrieval model, you should use regular train and use --data-parallel true
- To use multigpu in validation time, you're just out of luck (See #2162)

We should strive for easy multigpu training by default. There's some mess with multiprocessing, of course (See #1969 and related).

A vague attack plan would be to finish #1596, then try to merge `multiprocessing_train` and `train_model` with some refactoring, so we use DDP everywhere. We might as well wait on some of the other tasks.

Hat tip @uralik for the idea.

Tests will be important, but we need to figure out #1969. Intro docs should explain what to do, and probably the `--gpu` arg that a lot of these things support should be upgraded to support a list?

stephenroller:
#1596 will need to be revived for this. It's worth noting that general setup works well in [pytorch-lightning](https://github.com/williamFalcon/pytorch-lightning)

PR #2345, opened 2020-01-17
[Teachers] Chunk teacher

emilydinan:
**Patch description**
Teacher for loading large datasets in chunks, and an example implementation for a Wikipedia teacher. Based largely on @stephenroller 's internal implementation for [redacted]

**Testing steps**
```
python examples/display_data.py -t wikipedia:fullSplit
```


parlai/core/teachers.py
```diff
@@ -1944,6 +1945,223 @@ def shutdown(self):
             t.shutdown()
 
 
+class ChunkTeacher(FixedDialogTeacher, ABC):
```
stephenroller: We could follow Mary and call it AbstractChunkTeacher. I'm ambivalent.

parlai/core/teachers.py
```diff
@@ -1944,6 +1945,223 @@ def shutdown(self):
...
+        chunks = self.fold_chunks[:]
+
+        if (
+            shared is None
+            and self.is_train
+            and self.opt.get('distributed_world_size') is not None
+        ):
+            dws = int(self.opt['distributed_world_size'])
+            rank = int(self.opt['rank'])
+            self.fold_chunks = [c for c in self.fold_chunks if c % dws == rank]
```
stephenroller: Nothing you need to change here, but this will need to be carefully revisited in some of the aspects of #2185.

parlai/core/teachers.py
```diff
@@ -1944,6 +1945,223 @@ def shutdown(self):
...
+    Useful for loading large amounts of data.
+
+    Data is separated into chunks and loaded one chunk at a
+    time. Loads the data off of the main thread.
+    """
+    def __init__(self, opt, shared=None):
+        super().__init__(opt, shared)
+
+        # make logging print to stdout
+        logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
```
stephenroller: let's strip the logging from this please. This was just for me when I was debugging this originally.

parlai/core/teachers.py
```diff
@@ -1944,6 +1945,223 @@ def shutdown(self):
...
+            self.chunks = shared['chunks']
+            self.samples = shared['samples']
+            self.rng = shared['rng']
+            self.is_root_teacher = False
+        else:
+            self.is_root_teacher = True
+            self.samples = queue.Queue(maxsize=self.buffersize)
+            self.chunks = queue.Queue()
+            if self.is_train:
+                self.rng = random.Random()
```
stephenroller: We should leave a TODO here marking it as possibly needing a fixed seed in future works.

parlai/core/teachers.py
```diff
@@ -1944,6 +1945,223 @@ def shutdown(self):
...
+        if not self.opt.get('datafile'):
+            raise RuntimeError(
+                'Must specify datafile or override this function '
+                'to return the data folder.'
+            )
+
+        return self.opt['datafile']
+
+    @abstractmethod
+    def _get_num_samples(self, datatype) -> int:
```
stephenroller: abstract methods shouldn't be private. The things I need to implement should be visible to me.

parlai/core/teachers.py
```diff
@@ -1944,6 +1945,223 @@ def shutdown(self):
...
+        shared = super().share()
+        shared['samples'] = self.samples
+        shared['chunks'] = self.chunks
+        shared['rng'] = self.rng
+        return shared
+
+    def _setup_data(self, datatype):
+        pass
+
+    def num_episodes(self):
```
stephenroller: this isn't necessarily true. It was in my [redacted] teacher and might be in yours, but not in the general case.

parlai/core/teachers.py
```diff
@@ -1944,6 +1945,223 @@ def shutdown(self):
...
+            # buffer gets full
+            self.samples.put(data.pop(0))
+        # and start loading the next chunk
+        self._enqueue_request()
+
+    def _enqueue_chunks(self):
+        """
+        Shuffles and queues fold chunks for loading.
+        """
+        self.rng.shuffle(self.fold_chunks)
```
stephenroller: I think we have to protect against this if it's valid/test/train:evalmode. They should produce the same outputs every time.

parlai/core/teachers.py
```diff
@@ -1944,6 +1945,223 @@ def shutdown(self):
...
+        for c in self.fold_chunks:
+            self.chunks.put(c)
+
+    @abstractmethod
+    def _load_chunk_idx(self, chunk_idx: int) -> List[Tuple[str, ...]]:
```
stephenroller: We should use [generics](https://mypy.readthedocs.io/en/latest/generics.html) here. The chunkyboi doesn't care what this outputs, it just needs to be consistent as the input for `create_message`.

parlai/tasks/wikipedia/agents.py
```diff
@@ -76,6 +78,85 @@ def get_extraction_instructions(self):
...
+    def _create_message(self, queue_output: Tuple[str, ...]) -> 'Message':
+        """
+        [Abstract] Given the tuple output of the queue, return an act.
+        """
+        title, text = queue_output
+        return Message(
+            {
+                'title': title,
+                'text': text,
+                'labels': [''],
```
stephenroller: :-O

parlai/core/teachers.py
EricMichaelSmith: I'm on the fence about that too - after all, TorchRankerAgent has an abstract method and it's not AbstractTorchRankerAgent. I think I'm leaning against the "Abstract" for conciseness, since it's clear from the inheritance of ABC that it's abstract

parlai/core/teachers.py
EricMichaelSmith: But they can still be private and visible to the concrete subclass, which means they'd still warrant an underscore

parlai/core/teachers.py
```diff
@@ -1944,6 +1945,217 @@ def shutdown(self):
...
+        self.is_train = 'train' in datatype and 'evalmode' not in datatype
+
+    def share(self):
+        shared = super().share()
+        shared['samples'] = self.samples
+        shared['chunks'] = self.chunks
+        shared['rng'] = self.rng
+        return shared
+
+    def _setup_data(self, datatype):
```
EricMichaelSmith: Maybe a comment indicating when this wouldn't just be a passthrough?

parlai/tasks/wikipedia/agents.py
```diff
@@ -76,6 +78,73 @@ def get_extraction_instructions(self):
...
+
+    def _get_data_folder(self):
+        return os.path.join(self.opt['datapath'], 'wikipedia/full/wiki_full_extracted')
+
+    def _get_num_samples(self, datatype) -> int:
+        """
+        Return the number of samples given the datatype.
+        """
+        if 'train' in datatype:
+            return 5437097
```
EricMichaelSmith: Hmm, might be a good to have a unit test for this teacher as well (if it doesn't already exist) to check that this number of samples remains accurate

parlai/core/teachers.py
stephenroller: Sure they’re protected. But why are we going to hide the things the user needs to implement from the docs?

EricMichaelSmith: Hmm - will the docs (for instance, the docs you got to be auto-created Friday) only give documentation for public methods? In that case, yeah, maybe it makes sense to make this method public to increase its visibility

parlai/core/teachers.py
```diff
@@ -1944,6 +1947,209 @@ def shutdown(self):
...
+        """
+        Shuffles and queues fold chunks for loading.
+        """
+        if self.is_train:
+            self.rng.shuffle(self.fold_chunks)
+        for c in self.fold_chunks:
+            self.chunks.put(c)
+
+    @abstractmethod
+    def load_chunk_idx(self, chunk_idx: int) -> List[Tuple[T, ...]]:
```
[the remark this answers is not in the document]
klshuster: also this function name is a little confusing - it is not loading the `chunk_idx` but rather, loading examples _from_ the chunk idx `chunk_idx`. Perhaps something like `load_from_chunk` would be better?

parlai/tasks/wikipedia/agents.py
```diff
@@ -76,6 +78,92 @@ def get_extraction_instructions(self):
...
+            self.opt = opt
+            self._set_chunk_idx_to_file()
+        else:
+            self.chunk_idx_to_file = shared['chunk_idx_to_file']
+        super().__init__(opt, shared)
+
+    def _get_data_folder(self):
+        return os.path.join(self.opt['datapath'], 'wikipedia/full/wiki_full_extracted')
+
+    def get_num_samples(self, datatype) -> int:
```
klshuster: type warning here

parlai/tasks/wikipedia/agents.py
```diff
@@ -76,6 +78,92 @@ def get_extraction_instructions(self):
...
+        """
+        Return the number of samples given the datatype.
+        """
+        if 'train' in datatype:
+            return 5437097, 5437097
```
klshuster: could we make these numbers constants at the top?

parlai/tasks/wikipedia/agents.py
```diff
@@ -76,6 +78,92 @@ def get_extraction_instructions(self):
...
+        self.chunk_idx_to_file = {i: x for i, x in enumerate(all_subdirs)}
+
+    def get_fold_chunks(self, datatype) -> List[int]:  # type: ignore
+        """
+        Return a list of chunk IDs (integer).
+
+        Given the datatype (train/test/valid), return the list of chunk IDs that
+        correspond to that split.
+        """
+        all_subdirs = sorted([x for x in os.listdir(self.folder) if 'README' not in x])
```
klshuster: it looks like this func is called on every `__init__`, perhaps we could cache this `all_subdirs` somewhere so we don't have to sort it every time? not necessary if this is not a lot of subdirs, just didn't know the number

parlai/tasks/wikipedia/agents.py
```diff
@@ -76,6 +78,92 @@ def get_extraction_instructions(self):
...
+
+    def load_chunk_idx(self, chunk_idx: int) -> List[Tuple[str, str]]:
+        """
+        Given the chunk index, load examples from that chunk.
+
+        Return a list of tuples. The function `_create_message` will take these tuples
+        to form the Message object that is returned by the teacher.
+        """
+        output = []
+        chunk_path = os.path.join(self.folder, self.chunk_idx_to_file[chunk_idx],)
```
klshuster: nit: unnecessary trailing comma?

parlai/core/teachers.py
```diff
@@ -1944,6 +1947,209 @@ def shutdown(self):
...
+            self.rng = shared['rng']
+        else:
+            self.is_root_teacher = True
+            self.samples = queue.Queue(maxsize=self.buffersize)
+            self.chunks = queue.Queue()
+            if self.is_train:
+                # TODO: possible need a fixed seed here in the future
+                self.rng = random.Random()
+            else:
+                self.rng = random.Random(4)
```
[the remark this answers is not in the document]
stephenroller: haha no but we should def be consistent

parlai/core/teachers.py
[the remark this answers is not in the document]
stephenroller: It doesn't need to be a tuple. T is a placeholder, but the choice of going with a Generic is to ensure that it's the SAME type from load_chunk_idx going into the other method.

parlai/core/teachers.py
```diff
@@ -1944,6 +1947,209 @@ def shutdown(self):
...
+    """
+    Useful for loading large amounts of data.
+
+    Data is separated into chunks and loaded one chunk at a time. Loads the data off of
+    the main thread.
+    """
+
+    def __init__(self, opt, shared=None):
+        super().__init__(opt, shared)
+        self.buffersize = 100000
```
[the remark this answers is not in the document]
stephenroller: yeah, it probably needs to be configurable or like, (num_samples / num_chunks ) / 5. It should be smaller than the size of a chunk, but large enough that we have sufficient time to background load the chunk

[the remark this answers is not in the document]
emilydinan: i will update to make it dependent on the number of samples and number of chunks

parlai/tasks/wikipedia/agents.py
emilydinan: oh yes... i made a mistake here. i half-fixed this problem and did not push the other half lol

parlai/core/teachers.py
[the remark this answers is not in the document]
stephenroller: it'll try to load 2 chunks in a row then i think, but prolly nbd

PR #2348, opened 2020-01-19
Don't die if --fp16 true but no apex.

stephenroller:
**Patch description**
If the user specifies `--fp16 true`, we currently raise an error saying the user should install APEX. This has been annoying to some users (including internal ones, and on fbcode) who just wanted things to work without the failure.

**Testing steps**
Uninstalled apex and ran
```
$ python examples/interactive.py -m transformer/polyencoder  -mf zoo:pretrained_transformers/model_poly/model --fp16 true
[ Using CUDA ]
/private/home/roller/working/parlai/parlai/utils/torch.py:234: UserWarning: You set --fp16 true, but fp16 is unavailable. To use fp16, please install APEX from https://github.com/NVIDIA/apex.
  'You set --fp16 true, but fp16 is unavailable. To use fp16, please '
Dictionary: loading dictionary from /private/home/roller/working/parlai/data/models/pretrained_transformers/model_poly/model.dict
[ num words =  54944 ]
[Polyencoder: full interactive mode on.]
```

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -36,7 +36,12 @@ from parlai.core.dict import DictionaryAgent
 from parlai.nn.lr_scheduler import ParlAILRScheduler
 from parlai.core.message import Message
 from parlai.utils.misc import AttrDict, warn_once, round_sigfigs
-from parlai.utils.torch import argsort, fp16_optimizer_wrapper, padded_tensor
+from parlai.utils.torch import (
+    argsort,
+    fp16_optimizer_wrapper,
+    padded_tensor,
+    fp16_available,
+)
 
 
 class Batch(AttrDict):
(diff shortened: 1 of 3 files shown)


parlai/core/torch_agent.py
```diff
@@ -606,7 +611,7 @@ def __init__(self, opt: Opt, shared=None):
             if not shared and opt['gpu'] != -1:
                 torch.cuda.set_device(opt['gpu'])
         # indicate whether using fp16
-        self.fp16 = self.use_cuda and self.opt.get('fp16', False)
```
hadasah: This is where it was dying before?

hadasah:
Are the unit tests just flaking?

parlai/core/torch_agent.py
stephenroller: Well this would be set as true, and then eventually it would call `parlai.utils.torch.fp16_optimizer` and that will throw an import error saying the user needs to install APEX.

Issue #2357, opened 2020-01-23
Fb Messenger deployment tutorial


liloup1789:
Hello, 
is there any tutorial for implementing a bot on fb messenger ? To test and/or interact with it in a lifelong learning fashion ? 

When I click on the link `https://parl.ai/docs/tutorial_messenger.html` from the README.md page, I get a 404 error 
Thanks

liloup1789:
And when i run the run.py script, I have this error : 
`python parlai/chat_service/services/messenger/run.py --config-path parlai/chat_service/tasks/chatbot/config.yml 
[ optional arguments: ] 
[  datapath: /home/tf/Documents/Leon/parlai/ParlAI/data ]
[ Chat Services: ] 
[  config_path: parlai/chat_service/tasks/chatbot/config.yml ]
[  is_debug: False ]
[  password: None ]
[ Facebook Messenger: ] 
[  bypass_server_setup: False ]
[  force_page_token: False ]
[  local: False ]
[  log_level: 20 ]
[  verbose: False ]
[ Current ParlAI commit: 2927c796d1fdb0d5b396bd6bd514d3713fda4fd3 ]
/home/tf/Documents/Leon/parlai/ParlAI/parlai/chat_service/core/shared_utils.py:191: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  cfg = yaml.load(f.read())

You are going to allow people on Facebook to be agents in ParlAI.
During this process, Internet connection is required, and you should turn off your computer's auto-sleep feature.

Please press Enter to continue... 

Setting up Messenger webhook...
Heroku: Collecting files...
Traceback (most recent call last):
  File "parlai/chat_service/services/messenger/run.py", line 47, in <module>
    run(opt)
  File "parlai/chat_service/services/messenger/run.py", line 32, in run
    manager = MessengerManager(opt)
  File "/home/tf/Documents/Leon/parlai/ParlAI/parlai/chat_service/services/messenger/messenger_manager.py", line 43, in __init__
    self._complete_setup()
  File "/home/tf/Documents/Leon/parlai/ParlAI/parlai/chat_service/services/messenger/messenger_manager.py", line 61, in _complete_setup
    self.setup_server()
  File "/home/tf/Documents/Leon/parlai/ParlAI/parlai/chat_service/services/messenger/messenger_manager.py", line 199, in setup_server
[...]

PR #2123, opened 2019-11-01
Move shared_utils and server_utils to chat_service/core/

young-k:
**Patch description**

Neither `shared_utils` nor `server_utils` is messenger-implementation specific, so I moved them to `core/`. 

**Testing steps**

~CircleCI :pray:~ will look into better ways to test this

**Other information**

N/A

PR #2366, opened 2020-01-27
[Chat Service] Bug Fixes

klshuster:
**Patch description**
#2123 introduced an issue with setting up the heroku server for use with ParlAI messenger (specifically dealing with the path to the `server/` directory. As a temporary fix, I have moved the `server/` directory out of the `chat_service/services/messenger` directory into `chat_service/core/`.  

Fixes the issue described in #2357 

Finally, there was a slight issue with the way the args were being parsed (i.e. separate from what is described in the `README`). This PR also fixes those issues. 

**Testing steps**
Tested the chatbot task with page ID [REDACTED] (though we do need some good tests for this directory at some point).

**Logs**
```
$ python parlai/chat_service/services/messenger/run.py --config-path parlai/chat_service/tasks/chatbot/config.yml
.
.
.
```
Screenshot 2020-01-27 18 14 20


stephenroller:
Will the heroku stuff still make sense in the context of Jack's larger project?

klshuster:
This part specifically is separate as it has no ties to Mturk

PR #2367, opened 2020-01-28
[Not for merge][Feedback Welcome]All-in-one image modifications for ParlAI

dexterju:
**Patch description**
This PR is for collecting feedbacks on how I can get these All-in-One Image-Grounded Conversational Agents https://arxiv.org/abs/1912.12394 paper code merge back to ParlAI. A joint work with @klshuster @ylannb @jaseweston 
This includes several modifications:
1. parlai/core/torch_classifier_agent.py: [A nice plus]
I added codes to load a list of classes file from disk and use them as classification classes on top of the original binary classifier setting.
2. parlai/core/torch_ranker_agent.py [Necessary]
My changes let torch ranker agent take a binary matrix of size [btz * target_classes], this allows it handle multi-label classification. (In the VQA dataset setting, more than one label is considered as correct), so we don't penalize the rest of the correct labels.  
3.  parlai/core/worlds.py[Necessary]
My changes create a new world which flips the multi and batch settings as we had in master. 
4. parlai/utils/misc.py[For compatibility]
For a bug I observed when I tried to load the old model file back. Some previous versions of @klshuster's work packed the opt file with the model file together, and it failed without this line of code.

### DIFF
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -13,6 +13,7 @@ from parlai.core.opt import Opt
 from parlai.utils.distributed import is_distributed
 from parlai.core.torch_agent import TorchAgent, Output
 from parlai.utils.misc import round_sigfigs, warn_once
+import os
 from collections import defaultdict
 
 import torch
@@ -98,16 +99,18 @@ class TorchClassifierAgent(TorchAgent):
         super().__init__(opt, shared)
 
         # set up classes
-        if opt.get('classes') is None and opt.get('classes_from_file') is None:
-            raise RuntimeError(
-                'Must specify --classes or --classes-from-file argument.'
-            )
         if not shared:
-            if opt['classes_from_file'] is not None:
-                with open(opt['classes_from_file']) as f:
-                    self.class_list = f.read().splitlines()
+            if opt.get('classes') is None:
+                raise RuntimeError('Must specify --classes argument.')
+            if os.path.isfile(opt['classes'][0]):
+                # Load candidates
+                cand_path = opt['classes'][0]
+                print("[ Loading fixed classes set from {} ]".format(cand_path))
+                with open(cand_path, 'r') as f:
+                    cands = [line.strip() for line in f.readlines()]
             else:
-                self.class_list = opt['classes']
+                cands = opt['classes']
+            self.class_list = cands
             self.class_dict = {val: i for i, val in enumerate(self.class_list)}
             if opt.get('class_weights', None) is not None:
                 self.class_weights = opt['class_weights']
@@ -158,13 +161,14 @@ class TorchClassifierAgent(TorchAgent):
             if init_model:
                 print('Loading existing model parameters from ' + init_model)
                 self.load(init_model)
-            if self.use_cuda:
-                if self.opt['data_parallel']:
-                    if is_distributed():
-                        raise ValueError(
-                            'Cannot combine --data-parallel and distributed mode'
-                        )
-                    self.model = torch.nn.DataParallel(self.model)
+        if self.use_cuda:
+            if self.opt['data_parallel']:
+                if is_distributed():
+                    raise ValueError(
+                        'Cannot combine --data-parallel and distributed mode'
+                    )
+                self.model_data_parallel = torch.nn.DataParallel(self.model)
+                self.model = self.model_data_parallel.module
         if shared:
             # We don't use get here because hasattr is used on optimizer later.
             if 'optimizer' in shared:
@@ -245,7 +249,8 @@ class TorchClassifierAgent(TorchAgent):
 
         # calculate loss
         labels = self._get_labels(batch)
-        scores = self.score(batch)
+        scores = self.score(batch).float()
+        # import ipdb; ipdb.set_trace()
         loss = self.criterion(scores, labels)
         loss.backward()
         self.update_params()
@@ -269,7 +274,7 @@ class TorchClassifierAgent(TorchAgent):
             return
 
         self.model.eval()
-        scores = self.score(batch)
+        scores = self.score(batch).float()
         probs = F.softmax(scores, dim=1)
         if self.threshold is None:
             _, prediction_id = torch.max(probs.cpu(), 1)
@@ -346,6 +351,7 @@ class TorchClassifierAgent(TorchAgent):
         examples = self.metrics['examples']
         if examples > 0:
             m['examples'] = examples
+            m['loss'] = self.metrics['loss']
             m['mean_loss'] = self.metrics['loss'] / examples
 
             # get prec/recall metrics
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -251,7 +251,7 @@ class TorchClassifierAgent(TorchAgent):
         labels = self._get_labels(batch)
         scores = self.score(batch)
         loss = self.criterion(scores, labels)
-        loss = loss * scores.size(1)
+        # loss = loss * scores.size(1)
         loss.backward()
         self.update_params()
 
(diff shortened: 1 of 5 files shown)


parlai/core/torch_classifier_agent.py
```diff
@@ -98,16 +99,18 @@ def __init__(self, opt: Opt, shared=None):
...
-            raise RuntimeError(
-                'Must specify --classes or --classes-from-file argument.'
-            )
         if not shared:
-            if opt['classes_from_file'] is not None:
-                with open(opt['classes_from_file']) as f:
-                    self.class_list = f.read().splitlines()
+            if opt.get('classes') is None:
+                raise RuntimeError('Must specify --classes argument.')
+            if os.path.isfile(opt['classes'][0]):
```
klshuster: was this not supported before with the `opt['classes_from_file']`?

parlai/core/torch_classifier_agent.py
```diff
@@ -158,25 +161,26 @@ def __init__(self, opt: Opt, shared=None):
...
+                    )
+                self.model_data_parallel = torch.nn.DataParallel(self.model)
+                self.model = self.model_data_parallel.module
         if shared:
             # We don't use get here because hasattr is used on optimizer later.
             if 'optimizer' in shared:
                 self.optimizer = shared['optimizer']
         else:
-            optim_params = [p for p in self.model.parameters() if p.requires_grad]
+            optim_params = [p for p in self.model.parameters() if p.modus_grad]
```
klshuster: what is `modus_grad`?

parlai/core/torch_classifier_agent.py
```diff
@@ -158,25 +161,26 @@ def __init__(self, opt: Opt, shared=None):
...
-                        )
-                    self.model = torch.nn.DataParallel(self.model)
+        if self.use_cuda:
+            if self.opt['data_parallel']:
+                if is_distributed():
+                    raise ValueError(
+                        'Cannot combine --data-parallel and distributed mode'
+                    )
+                self.model_data_parallel = torch.nn.DataParallel(self.model)
+                self.model = self.model_data_parallel.module
```
klshuster: does this not mess up later computation? I think we should be calling `self.model()` rather than `self.model_data_parallel()`

parlai/core/torch_classifier_agent.py
```diff
@@ -247,6 +251,7 @@ def train_step(self, batch):
         labels = self._get_labels(batch)
         scores = self.score(batch)
         loss = self.criterion(scores, labels)
+        loss = loss * scores.size(1)
```
klshuster: what does this do?

parlai/core/torch_classifier_agent.py
```diff
@@ -269,7 +274,7 @@ def eval_step(self, batch):
             return
 
         self.model.eval()
-        scores = self.score(batch)
+        scores = self.score(batch).float()
```
klshuster: can you explain what this fixes too?

parlai/core/torch_ranker_agent.py
```diff
@@ -311,7 +311,7 @@ def _get_train_preds(self, scores, label_inds, cands, cand_vecs):
         else:
             _, ranks = scores.sort(1, descending=True)
         for b in range(batchsize):
-            rank = (ranks[b] == label_inds[b]).nonzero()
+            rank = (ranks[b] == label_inds[b].nonzero()[0]).nonzero()
```
klshuster: are these the changes for multi-class classification? does this still work with non-multiclass classification?

parlai/core/torch_ranker_agent.py
```diff
@@ -664,7 +663,9 @@ def _build_candidates(self, batch, source, mode):
                 use_cuda=self.use_cuda,
                 fp16friendly=self.fp16,
             )
-            if label_vecs is not None:
+            if label_vecs is not None and not batch.id.startswith(
+                'internal:vqa_v2:PersonalityTeacher'
```
klshuster: i'm not sure we should include this in the actual `torch_ranker_agent` code, perhaps this could be in the model?

parlai/core/torch_ranker_agent.py
```diff
@@ -702,7 +703,9 @@ def _build_candidates(self, batch, source, mode):
             cands = self.fixed_candidates
             cand_vecs = self.fixed_candidate_vecs
 
-            if label_vecs is not None:
+            if label_vecs is not None and not batch.id.startswith(
+                'internal:vqa_v2:PersonalityTeacher'
```
klshuster: ditto?

parlai/core/worlds.py
```diff
@@ -656,7 +656,7 @@ def parley_init(self):
         if self.new_world:
             self.new_world = False
             self.parleys = 0
-            if self.random:
+            if False:
```
klshuster: what is this change for?

parlai/core/worlds.py
```diff
@@ -839,6 +867,33 @@ def batch_act(self, agent_idx, batch_observation):
...
+    def parley_init(self):
+        """
+        Update the current subworld.
+
+        If we are in the middle of an episode, keep the same world and finish this
+        episode. If we have finished this episode, pick a new world (either in a
+        random or round-robin fashion).
+        """
+        self.parleys = self.parleys + 1
+        if self.world_idx >= 0 and self.worlds[self.world_idx].episode_done():
```
klshuster: when is `world_idx` not >= 0?

parlai/utils/misc.py
```diff
@@ -13,6 +13,7 @@
 import time
 from typing import Union, Optional, Set, Any, Dict, List
 import warnings
+from parlai.core.opt import Opt
```
klshuster: ahh yes I am aware this broke previous models, we should not add it back here however, those models should be re-saved if necessary (I have already done this with my old models)

parlai/core/torch_classifier_agent.py
emilydinan: this was indeed supported with the 'classes_from_file' argument

emilydinan: i suppose we could overload that argument, i'm not sure how i feel about it. if we decide to do that, then we should get rid of that argument

parlai/core/torch_classifier_agent.py
emilydinan: yeah i agree. what is going on here? i am pretty sure it worked fine before

parlai/core/torch_ranker_agent.py
emilydinan: I agree, I do'nt think this should be in TRA

parlai/core/worlds.py
```diff
@@ -747,6 +747,34 @@ def _override_opts_in_shared(table, overrides):
     return table
 
 
+class BatchMultiwWorld(MultiWorld):
```
emilydinan: What is this for? can you add a docstring?

parlai/core/worlds.py
```diff
@@ -747,6 +747,34 @@ def _override_opts_in_shared(table, overrides):
...
+            if len(weights) > i:
+                weight = weights[i]
+            else:
+                weight = 1
+            self.cum_task_weights[i] = weight + sum
+            sum += weight
+
+    def parley_init(self):
+        """
+        Flipping world id
```
emilydinan: Can you add a more descriptive docstring here? What does "flipping world id" mean?

parlai/utils/misc.py
stephenroller: Yes, this is resolved in master already

parlai/core/torch_classifier_agent.py
```diff
@@ -346,6 +351,7 @@ def report(self):
         examples = self.metrics['examples']
         if examples > 0:
             m['examples'] = examples
+            m['loss'] = self.metrics['loss']
```
EricMichaelSmith: Oh this wasn't used before?

dexterju:
Close, since the issues have been somehow solved.

PR #2371, opened 2020-01-30
[Scaling] Background preprocessing

klshuster:
# TODO

- [ ] Profile/benchmark single-gpu training
- [ ] Profile/benchmark distributed training

# Patch description
Offline Preprocessing.

Lots (LOTS) of moving parts here, so I will try to outline what the heck is going on.

**NOTE: I am accepting initial comments, however I am waiting on a sweep to outline speed improvements (and also to ensure the logic here is correct)**

## QueueWorld
A QueueWorld for offline preprocessing. Triggered with ``opt['numworkers']`` > 1

The QueueWorld spawns ``opt['numworkers']`` subprocesses, which handle
preprocessing of batches off the main thread.

The QueueWorld has a number of important structures:
    1. ``consume_queues`` - each subprocess has its own consume queue
        for model's actions.
    2. ``produce_queue`` - the queue from which QueueWorld pulls batches
    3. ``report_queue`` - the queue from which QueueWorld pulls reports
    4. ``buffers`` - tensor buffers for each attr in a Batch that is a tensor. This allows buffers to be reused amongst processes and signficantly reduces overhead. The reason for this is that copying tensors to/from queues is  _costly_.

## PWorld
A PWorld is a world that can be used in a subprocess of the QueueWorld.

A subclass must implement two methods:
1. ``produce_observation()``:
        returns an observation for an agent; this can be either a List[Message] or a Batch, depending on the agent
2. ``consume_act(act)``:
        Consumes a model's action, resulting from the observation in ``produce_observation``

The PWorld places produced observations on its ``produe_queue``;  it obtains actions from its ``consume_queue``. Finally, it places reports on its ``report_queue``.

## PBatchWorld & PDynamicBatchWorld
Subclasses of PWorld that correspond to their respective use cases.

## QueueSignal
An ``Enum`` used in the queues to signal various parts of the TrainLoop lifecycle

## TODO
- [x] Validation
- [x] Building Dictionaries
- [x] Multitasking?
- [x] Dynamic Batching Support
- [x] Tests
- [x] Model Perf. Match (sweep)

# Testing steps
tests/test_offline_preprocess.py

# Logs
```
$ python -m pytest -m unit tests/test_offline_preprocess.py
================================================================================ test session starts =================================================================================
rootdir: /private/home/kshuster/ParlAI, inifile: pytest.ini
plugins: requests-mock-1.7.0
collected 4 items

tests/test_offline_preprocess.py ....                                                                                                                                          [100%]

============================================================================= slowest 10 test durations ==============================================================================
631.46s call     tests/test_offline_preprocess.py::TestOfflinePreprocess::test_multitask
437.26s call     tests/test_offline_preprocess.py::TestOfflinePreprocess::test_multitask_dynamic_batching
230.92s call     tests/test_offline_preprocess.py::TestOfflinePreprocess::test_singletask_dynb
172.43s call     tests/test_offline_preprocess.py::TestOfflinePreprocess::test_singletask

=========================================================================== 4 passed in 1474.12s (0:24:34) ===========================================================================

```

Results from finetuning a large pre-trained model on convai2:

image


parlai/core/worlds.py
```diff
@@ -1245,6 +1252,127 @@ def shutdown(self):
...
+                    'produce_queue': produce_queue,
+                    'consume_queue': consume_queue,
+                    'produce_cond': produce_cond,
+                    'consume_cond': consume_cond,
+                }
+            )
+            self.processes.append(
+                Process(
+                    target=run_q_world,
+                    args=(opt, subworld, self.structs[-1]),
```
stephenroller: Nit: I think you can give kwargs

parlai/core/worlds.py
```diff
@@ -1245,6 +1252,127 @@ def shutdown(self):
...
+        self.processes: List[Process] = []
+        self.structs: List[Dict[str, Union[Queue, Condition]]] = []
+        self.worlds: List[World] = []
+        for _ in range(opt['numworkers']):
+            produce_cond = Condition()
+            consume_cond = Condition()
+            produce_queue = Queue(maxsize=1)
+            consume_queue = Queue(maxsize=1)
+            shared = world.share()
+            subworld = shared['world_class'](opt, None, shared)
```
stephenroller: is the none the agent?!

parlai/core/worlds.py
```diff
@@ -1245,6 +1252,127 @@ def shutdown(self):
...
+        # Get batch obs
+        while batch_obs is None:
+            chosen_p = random.choice(self.structs)
+            with chosen_p['produce_cond']:
+                if chosen_p['produce_queue'].full():
+                    batch_obs = chosen_p['produce_queue'].get()
+        # Batch Act on that obs
+        batch_acts = self.batch_act(1, batch_obs)
+        with chosen_p['consume_cond']:
+            chosen_p['consume_queue'].put_nowait(batch_acts)
```
stephenroller: I think we can just put. We don't need the cond, right?

parlai/core/worlds.py
```diff
@@ -1245,6 +1252,127 @@ def shutdown(self):
...
+            )
+        for p in self.processes:
+            p.start()
+
+    def parley(self):
+        batch_obs: List[Message] = None
+        chosen_p: Dict[str, Union[Queue, Condition]] = None
+        # Get batch obs
+        while batch_obs is None:
+            chosen_p = random.choice(self.structs)
```
stephenroller: I think this is why we aren't getting the speedup. We should have a SINGLE produce q, which all workers can put onto. This world should be reading *solely* off that queue.

parlai/core/worlds.py
klshuster: i think so; i took this logic directly from `BatchWorld`

stephenroller: Okay I need to understand this cuz it scares me

stephenroller: okay I see. It has two mutually exclusive arguments, shared and agents. Maybe we can drop the none and do `(opt, shared=shared)` to make it slightly more clear

parlai/core/teachers.py
```diff
@@ -256,7 +262,7 @@ def __init__(self, opt, shared=None):
         else:
             self.index = AttrDict(value=-1)
 
-        if not hasattr(self, 'data_loader'):
+        if not hasattr(self, 'data_loader') and opt['numworkers'] <= 1:
```
klshuster: we cannot mix QueueWorld with the dataloader in its current form, as it cannot be pickled.

klshuster:
Marking this ready for review. I have tentative results from performance sweeps that indicate this works well. Will post results in the description when they are available.

parlai/core/params.py
```diff
@@ -673,12 +673,20 @@ def add_parlai_args(self, args=None):
             type=int,
             help='batch size for minibatch training schemes',
         )
+        parlai.add_argument(
+            '-nw',
+            '--numworkers',
```
EricMichaelSmith: nit: hyphen between num and workers?

parlai/core/torch_agent.py
```diff
@@ -1364,6 +1361,11 @@ def batchify(self, obs_batch, sort=False):
         if any('image' in ex for ex in exs):
             imgs = [ex.get('image', None) for ex in exs]
 
+        kleenexs = [
```
EricMichaelSmith: kleenexs as a variable name would probably be very confusing =/ Why are we stripping out the Tensors here?

parlai/core/worlds.py
```diff
@@ -1123,7 +1132,15 @@ def _score(self, obs):
...
 
-    def parley(self):
+    def _build_batch(self) -> Tuple[List[int], List[Message]]:
+        """
+        Build the batch of observations for the model.
+
+        Return the indices into `self._obs` as well as the list of messages
+
+        :return: (batch, obs)
+        :rtype: (List[Int], List[Message])
```
EricMichaelSmith: Hmm, might help to say what self._obs indexes here for clarity

parlai/utils/misc.py
```diff
@@ -342,6 +344,22 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
         pass
 
 
+class ForkedPdb(pdb.Pdb):
```
EricMichaelSmith: Really cool!

parlai/core/params.py
stephenroller: Or just —workers?

parlai/core/params.py
```diff
@@ -673,12 +673,20 @@ def add_parlai_args(self, args=None):
...
+            help='how many workers the Pytorch dataloader should use',
+            hidden=True,
+        )
         parlai.add_argument(
             '-dynb',
             '--dynamic-batching',
             default=None,
             type='nonestr',
-            choices={'none', 'full', 'batchsort'},
+            choices={None, 'full', 'batchsort'},
```
stephenroller: This seems off. Are choices processed after parsing with the type?

parlai/core/teachers.py
```diff
@@ -244,7 +245,8 @@ def __init__(self, opt, shared=None):
         if not hasattr(self, 'datafile'):
             self.datafile = opt.get('datafile')
         # set up support for multithreaded data loading
-        self.data_queue = queue.Queue()
+        if opt['numworkers'] <= 1:
```
stephenroller: Yeah num_workers would be slightly nicer

parlai/core/teachers.py
stephenroller: Cute. 😡

stephenroller: Why can’t it be pickled?

parlai/core/torch_agent.py
stephenroller: Yeah sadly I don’t think this should stay lol

Def add the comment on why we strip

parlai/core/torch_agent.py
```diff
@@ -1708,19 +1710,26 @@ def batch_act(self, observations):
         ``eval_step`` methods instead. The former is called when labels are
         present in the observations batch; otherwise, the latter is called.
         """
-        batch_size = len(observations)
+        if not isinstance(observations, Batch):
+            batch_size = len(observations)
+        else:
+            batch_size = observations.text_vec.size(0)
```
stephenroller: This seems like a good opportunity to add that batchsize field you’ve always wanted

parlai/core/torch_agent.py
```diff
@@ -1708,19 +1710,26 @@ def batch_act(self, observations):
...
+        if not isinstance(observations, Batch):
+            batch = self.batchify(observations)
+        else:
+            batch = observations
+
         # check if there are any labels available, if so we will train on them
+        observations = batch.observations
         self.is_training = any('labels' in obs for obs in observations)
 
-        # create a batch from the vectors
```
stephenroller: Huh, why move this?

parlai/core/worlds.py
```diff
@@ -42,22 +42,29 @@
 """
 
 import copy
+import queue
 import random
 import time
 
+from abc import ABC, abstractmethod
+from enum import auto, Enum
```
stephenroller: Enum?!

parlai/core/worlds.py
```diff
@@ -42,22 +42,29 @@
...
+    from multiprocessing import Process, Value, Semaphore, Queue  # noqa: F401
 
-from parlai.core.agents import create_agents_from_shared
+from parlai.core.agents import create_agents_from_shared, Agent
 from parlai.core.loader import load_task_module, load_world_module
+from parlai.core.message import Message
 from parlai.core.metrics import aggregate_metrics
 from parlai.core.opt import Opt
 from parlai.core.teachers import create_task_agent_from_taskname
+from parlai.core.torch_agent import Batch
```
stephenroller: Hmm I think we need to yank out batch into a new file to prevent torch from being needed for mturk

parlai/core/worlds.py
```diff
@@ -814,7 +821,9 @@ def batch_observe(self, index, batch_actions, index_acting):
         for i, w in enumerate(self.worlds):
             agents = w.get_agents()
             observation = None
-            if batch_actions[i] is None:
+            if i >= len(batch_actions):
```
stephenroller: ??

parlai/core/worlds.py
stephenroller: Rtype is redundant with the annotation you gave

parlai/core/worlds.py
```diff
@@ -1499,6 +1533,622 @@ def shutdown(self):
...
 
+class QueueSignal(Enum):
+    """
+    Command signal used in the Queues for QueueWorld.
+
+    BATCH:                normal result; a batch is in the Queue
+    WORKER_FINISHED:      a worker is finished sending batch
+    RESET:                tell a worker to reset
+    RESET_METRICS:        tell a worker to reset_metrics
+    REPORT:               tell a worker to report
```
stephenroller: Lordy. Do we need this with agent level metrics? Let’s coordinate

parlai/core/worlds.py
```diff
@@ -1499,6 +1533,622 @@ def shutdown(self):
...
+        self._observe_batch_act(self.batch_indices, act)
+        self.batch_indices = None
+
+    def reset(self):
+        DynamicBatchWorld.reset(self)
+        PWorld.reset(self)
+        self.batch_indices = None
+
+
+def run_p_world(
```
stephenroller: Without committing to a choice, some functions as large as this work well as Classes that implement __call__.

Also, should we consider making this private? I’m hoping we don’t ever import this anywhere else

parlai/core/worlds.py
```diff
@@ -1499,6 +1533,622 @@ def shutdown(self):
...
+
+        :param pworld_class:
+            Which PWorld to use.
+        """
+        for worker_idx in range(self.numworkers):
+            consume_queue: Queue = Queue()
+            shared = world.share()
+            subworld = shared['world_class'](opt, None, shared)
+            self.processes.append(
+                Process(
```
stephenroller: Yes so you can inherent from Process and implement the run function. It tends to read cleaner since you introduce class variables etc

parlai/core/worlds.py
```diff
@@ -1499,6 +1533,622 @@ def shutdown(self):
...
+
+    def reset_metrics(self):
+        """
+        Reset metrics.
+
+        (TODO possibly reset process metrics)
+        """
+        super().reset_metrics()
+        self.world.reset_metrics()
+        # for q in self.consume_queues:
```
stephenroller: Heads up

parlai/core/worlds.py
```diff
@@ -1576,6 +2226,12 @@ def create_task(opt: Opt, user_agents, default_world=None):
         # use hogwild world if more than one thread requested
         # hogwild world will create sub batch worlds as well if bsz > 1
         world = HogwildWorld(opt, world)
+    elif opt.get('numworkers', 1) > 1:
```
stephenroller: Meh this is a bit cumbersome but the rest is reasonable so okay

tests/test_offline_preprocess.py
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env python3
```
stephenroller: Let’s discuss a better name? Maybe background? Offline implies we do it ahead of time instead of on the fly

parlai/core/teachers.py
klshuster: The Thread RLock's cannot be pickled specifically (i.e. the normal queue.Queue is not picklable)

parlai/core/params.py
klshuster: i can switch to `--num-workers`. This was originally a flag from the [deprecated] PytorchDataTeacher, as it's passed into the pytorch dataloader `numworkers` field. Given the deprecation, we can change

parlai/core/torch_agent.py
klshuster: I'll add a comment, but long story short we must avoid passing tensors along multiprocessing queues as it results in an expensive tensor copy operation

the name is staying 😄

parlai/core/params.py
klshuster: yes that is exactly what was going on - if i specified `'none'`, it would tell me that `None` was not a valid choice:

```
Parse Error: argument -dynb/--dynamic-batching: invalid choice: None (choose from 'full', 'batchsort', 'none')
```

parlai/core/torch_agent.py
klshuster: I generalized to just access `observations` from the batch object now, since we need to use that in line 1731 (`self.is_training`)

parlai/core/worlds.py
klshuster: ok, i am moving to `parlai/utils/batch.py`

parlai/core/worlds.py
klshuster: hmm i don't remember when i ran into this issue. i'll try removing this

parlai/core/worlds.py
klshuster: unclear. i'll pull in the agent level metrics when it's merged and see what's necessary going from there

parlai/core/worlds.py
klshuster: i actually need to uncomment this thanks

tests/test_offline_preprocess.py
klshuster: discussed offline, `background` is the way to go

klshuster:
Updates to address comments:

- Changed numworkers to num-workers & num_workers
- Add comment for why kleenexs
- Added size field to Batch
- Batch is in its own file now
- Run_p_world has been abstracted into its own Process object

stephenroller:
According to your pasted log, the unit tests take like 15 minutes. Any way we can bring that down a bit?

klshuster:
yeah i'll work on that

parlai/utils/misc.py
```diff
@@ -342,6 +344,22 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+    """
+    A Pdb subclass that may be used from a forked multiprocessing child.
+
+    See https://stackoverflow.com/questions/4716533/how-to-attach-debugger-to-a-python-subproccess
```
emilydinan: cool!

parlai/core/worlds.py
```diff
@@ -42,22 +42,29 @@
...
 from functools import lru_cache
-from typing import List, Dict, Any, Union
+from typing import List, Dict, Any, Union, Tuple, Optional
 
 try:
-    from torch.multiprocessing import Process, Value, Condition, Semaphore
+    from torch.multiprocessing import Process, Value, Semaphore, Queue
+    import torch
+    import torch.multiprocessing as mp
 except ImportError:
```
emilydinan: ????

parlai/core/worlds.py
```diff
@@ -1499,6 +1531,629 @@ def shutdown(self):
...
+        """
+        while True:
+            while not self.p_world.epoch_done():
+                self.p_world.parley()
+            # one more time to observe
+            self.produce_queue.put((QueueSignal.WORKER_FINISHED, None, self.worker_idx))
+            while True:
+                signal, batch = self.consume_queue.get()
+                if signal == QueueSignal.BATCH:
+                    # TODO: determine why this happens
```
emilydinan: what is this TODO?

parlai/core/worlds.py
```diff
@@ -1499,6 +1531,629 @@ def shutdown(self):
...
+
+        :return: report
+        :rtype: Dict
+        """
+        for q in self.consume_queues:
+            q.put((QueueSignal.REPORT, None))
+        reports: List[Dict] = []
+        while len(reports) < self.num_workers:
+            reports.append(self.report_queue.get())
+        # TODO: Aggregate correctly
```
emilydinan: why is this incorrect?

parlai/core/worlds.py
```diff
@@ -1499,6 +1531,629 @@ def shutdown(self):
...
+        :return: batch_actions
+        :rtype: List[Message]
+        """
+        a = self.world.get_model_agent()
+        if hasattr(a, 'batch_act'):
+            batch_actions = a.batch_act(batch_observation)
+        else:
+            batch_actions = []
+            for w in self.worlds:
+                agents = w.get_agents()
```
emilydinan: i'm a bit confused by this loop here. what is agent[1]?

parlai/core/worlds.py
klshuster: ahh yes i remember now. since we batchify offline, the agent does not see that we have a batchsize of e.g. 16 during the final step of validation if  `num_test_exs % 16 != 0`; it'll see something like 4. So it only responds with a list of 4 actions.

I'm adding some logic to the PBatchWorld to handle this instead of this opaque fix here

parlai/core/torch_agent.py
```diff
@@ -1364,6 +1286,17 @@ def batchify(self, obs_batch, sort=False):
         if any('image' in ex for ex in exs):
             imgs = [ex.get('image', None) for ex in exs]
 
+        # When passing Batch objects along multiprocessing queues, it is
+        # important that no tensors are unnecessarily placed in the batch,
+        # as this requires an expensive tensor copy. Thus, it is necessary
+        # to clean the exs of any tensors (giving us "clean" exs)
+        kleenexs = [
+            {k: v for k, v in ex.items() if not isinstance(v, torch.Tensor)}
+            for ex in exs
```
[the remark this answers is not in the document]
klshuster: i don't think so - this is purely removing the tensors from the `valid_obs`, which as far as I can tell (and confirmed this with @stephenroller) is not really used anywhere - or, at the very least, the tensors within these observations are not used

parlai/core/torch_agent.py
```diff
@@ -1708,19 +1642,26 @@ def batch_act(self, observations):
...
+        else:
+            batch_size = observations.batchsize
         # initialize a list of replies with this agent's id
         batch_reply = [
             Message({'id': self.getID(), 'episode_done': False})
             for _ in range(batch_size)
         ]
 
+        # create a batch from the vectors
+        if not isinstance(observations, Batch):
```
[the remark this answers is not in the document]
klshuster: i can add a comment, but basically this is when we call batchify in the background

parlai/core/worlds.py
klshuster: i need the `multiprocessing` module to set the start method, i can remove the `import torch` however

parlai/core/worlds.py
klshuster: still was not sure how I could reach this, will continue to test a little bit

parlai/core/worlds.py
klshuster: well so far the only thing I am aggregating is `exs`, which ad-hoc i think is correct but with the new metrics upgrade i'll just need to verify there is nothing else that needs to be aggregated

parlai/core/worlds.py
klshuster: ahh good catch, this should just be the `world.get_model_agent()`. 

To be honest, I have not even considered supporting agents who do not have batch act, and i'm not sure it makes sense to. So this branch of code is thus far not even touched

stephenroller:
I think all the child processes are building copies of the models and using up GPU memory :'(

parlai/core/worlds.py
emilydinan: the confusion was w.r.t. the except importerror. why do we have two different multiprocessing modules here?

klshuster: Ahh, well we want to use torch multiprocessing if we can but will resort to regular multiprocessing if it's not available. @stephenroller has pointed out to me that Mturk uses these worlds so we don't want people to need pytorch just to run mturk stuff

klshuster:
Test time reduced to ~520 seconds, and cuda is now solely on the main thread. Re-running performance sweep and will post numbers when they are available

stephenroller:
Can you add a multiprocessing test (tests/test_distributed) which uses multiprocessing + offline?

klshuster:
Added a multiprocessing train test that passes locally, currently disabled as we handle the issue with testing MP Train

klshuster:
A note that multiprocessing no longer works; will update when fixed

klshuster:
Distributed training now works, results from tests:

```
$ python -m pytest -m unit tests/test_background_preprocess.py
================================================================================ test session starts ================================================================================
rootdir: /private/home/kshuster/ParlAI, inifile: pytest.ini
plugins: requests-mock-1.7.0
collected 6 items

tests/test_background_preprocess.py ......                                                                                                                                    [100%]

============================================================================= slowest 10 test durations =============================================================================
147.86s call     tests/test_background_preprocess.py::TestBackgroundPreprocess::test_multitask_dynamic_batching
46.24s call     tests/test_background_preprocess.py::TestBackgroundPreprocess::test_singletask_dynb
42.59s call     tests/test_background_preprocess.py::TestBackgroundPreprocess::test_multitask
34.46s call     tests/test_background_preprocess.py::TestBackgroundPreprocess::test_singletask_distributed
10.79s call     tests/test_background_preprocess.py::TestBackgroundPreprocess::test_singletask
3.62s call     tests/test_background_preprocess.py::TestBackgroundPreprocess::test_stream

=================================================================== 6 passed, 3855 warnings in 303.62s (0:05:03) ====================================================================

```

I will run some more profiles/sweeps with both distributed training and single GPU training and post those results when I have them

parlai/scripts/multiprocessing_train.py
```diff
@@ -84,7 +84,11 @@ def multiprocess_train(
         distributed_utils.sync_object(None)
 
         # Run the actual training
-        return single_train.TrainLoop(opt).train()
+        retval = single_train.TrainLoop(opt).train()
+
+        dist.destroy_process_group()
```
stephenroller: huh, okay. That makes sense. Should you drop the one in the tearDown in the test?

parlai/scripts/multiprocessing_train.py
```diff
@@ -103,7 +107,11 @@ def launch_and_train(opt, port):
 
     try:
         retval = multiprocess_train(0, opt, port)
-        spawncontext.join()
+        if opt.get('num_workers', 1) > 1:
```
stephenroller: Do the subworkers cleanly join?

parlai/utils/testing.py
```diff
@@ -314,13 +314,15 @@ def distributed_train_model(opt: Opt) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         if 'dict_file' not in opt:
             opt['dict_file'] = os.path.join(tmpdir, 'model.dict')
 
-        parser = mp_train.setup_args()
-        popt = _forced_parse(parser, opt)
-
         # we need a prebuilt dictionary
         parser = build_dict.setup_args()
+        popt = _forced_parse(parser, opt)
```
stephenroller: ...what... what is going on over here?

tests/test_background_preprocess.py
```diff
@@ -134,12 +134,12 @@ def test_multitask(self):
                         rep['exs'], num_no_cand + num_multiturn, f'args: {args}'
                     )
                     self.assertEqual(
-                        rep['tasks']['integration_tests:nocandidate']['exs'],
+                        rep['integration_tests:nocandidate/exs'],
```
stephenroller: Thanks!

tests/test_background_preprocess.py
```diff
@@ -168,39 +168,36 @@ def test_multitask_dynamic_batching(self):
...
                 assertion(
-                    rep['tasks']['integration_tests:multiturn_nocandidate']['exs'],
+                    rep['integration_tests:multiturn_nocandidate/exs'],
                     NUM_TEST * 4,
                     f'args: {args}',
                 )
 
             self.assertEqual(valid['total_train_updates'], test['total_train_updates'])
 
-    @unittest.skipIf(True, 'Currently Skipping because of MP Train Testing Bug')
```
stephenroller: Where is this coming from, @mwilwork?

parlai/scripts/multiprocessing_train.py
klshuster: yes, this actually was causing an issue where any subsequent test called after `test_singletask_distributed` would have `is_distributed()` return True, since `dist.is_initialized()` would be True. What do you mean by dropping the tearDown in the test?

parlai/scripts/multiprocessing_train.py
klshuster: i dont think so

parlai/utils/testing.py
klshuster: forced parse probably isn't necessary but i just need to build the dict file before calling distributed train

parlai/core/torch_agent.py
```diff
@@ -35,16 +35,31 @@
 from parlai.core.dict import DictionaryAgent
 from parlai.nn.lr_scheduler import ParlAILRScheduler
 from parlai.core.message import Message
+<<<<<<< HEAD
```
stephenroller, line 38: !!!

stephenroller:
Abandoning for #3527

PR #2378, opened 2020-02-03
[FP16] FP16-safe cross entropy loss

emilydinan:
**Patch description**
Added an FP16-safe cross entropy loss, which computes the softmax in FP32 to avoid overflow.

Additionally, I added an edit to Torch Agent that allows us to override the way we pad tensors. The immediate use for this is in the GPT-2 agent implementation, to allow us to train in FP16.

EDIT: have confirmed that there is no performance difference between regular cross entropy and "safe" cross entropy in FP16.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -19,7 +19,7 @@ Contains the following main utilities:
 See below for documentation on each specific tool.
 """
 
-from typing import Dict, Any
+from typing import Dict, Any, Union, List, Tuple
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from collections import deque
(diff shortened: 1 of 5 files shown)


parlai/core/torch_agent.py
```diff
@@ -1323,12 +1344,8 @@ def batchify(self, obs_batch, sort=False):
             labels = [ex.get(field + '_choice') for ex in exs]
             y_lens = [y.shape[0] for y in label_vecs]
 
-            ys, y_lens = padded_tensor(
-                label_vecs,
-                self.NULL_IDX,
-                self.use_cuda,
-                fp16friendly=self.opt.get('fp16'),
-            )
+            ys, y_lens = self._pad_tensor(label_vecs)
```
stephenroller: ah this is gonna conflict hard with @klshuster

parlai/utils/torch.py
```diff
@@ -270,6 +271,27 @@ def fp16_available() -> bool:
...
+        self.NULL_IDX = ignore_index
+        self.reduction = reduction
+
+    def forward(self, scores, targets):
+        return F.nll_loss(
+            F.log_softmax(scores, 1, dtype=torch.float32),
+            targets,
+            ignore_index=self.NULL_IDX,
+            reduction=self.reduction,
+        )
```
stephenroller: i'm wondering if this needs an extra `.type_as(scores)` to bring it back to fp16 land. Can we check what myle does?

parlai/core/torch_agent.py
emilydinan: uh oh...really?

stephenroller:
Quick notes:
- This might remove instability we saw in fine-tuning some of our models in fp16.
- This does come with a performance hit.

I kinda want to test this more before we merge it but generally it's not a *bad* idea.

parlai/core/torch_agent.py
stephenroller: Not too bad, he's just gonna have a small merge conflict.

klshuster: yeah a little conflict but nothing major

PR #2374, opened 2020-01-31
Pass in non-string flags in task name

EricMichaelSmith:
**Patch description**
Pass in non-string flag values in task names using the ":" notation

**Testing steps**
```
>>> from parlai.core.teachers import _add_task_flags_to_agent_opt
>>> from parlai.core.opt import Opt
>>> opt = Opt({})
>>> flags='TestTeacher:flag1=string:flag2=true:flag3=false:flag4=4:flag5=5.0:flag6=foo'
>>> _add_task_flags_to_agent_opt(agent=None, opt=opt, flags=flags)
>>> opt
```
gives `{'flag1': 'string', 'flag2': True, 'flag3': False, 'flag4': 4, 'flag5': 5.0, 'flag6': 'foo', 'task': 'TestTeacher'}` as expected

PR #2379, opened 2020-02-03
[WIP] Fix WoW oddities

EricMichaelSmith:
1. Remove discrepancy in the number of examples in some of the WoW teachers
1. Don't add WoW topic by default for `BasicWizardDialogTeacher`, `BasicApprenticeDialogTeacher`, and `BasicBothDialogTeacher`, because it's not added by default to the other WoW teachers and it'd be good to be able to use these teachers without having WoW topic added. (This screwed me up for a bit when I was testing negative-history-candidate training.)

### DIFF
--- a/parlai/tasks/wizard_of_wikipedia/agents.py
+++ b/parlai/tasks/wizard_of_wikipedia/agents.py
@@ -413,14 +413,14 @@ class BasicdialogTeacher(WizardOfWikipediaTeacher):
 
 class BasicWizardDialogTeacher(BasicdialogTeacher):
     def __init__(self, opt, shared=None):
+        opt['speaker_label'] = "wizard"
         super().__init__(opt, shared)
-        self.speaker_label = "wizard"
 
 
 class BasicApprenticeDialogTeacher(BasicdialogTeacher):
     def __init__(self, opt, shared=None):
+        opt['speaker_label'] = 'apprentice'
         super().__init__(opt, shared)
-        self.speaker_label = 'apprentice'
 
 
 class BasicBothDialogTeacher(MultiTaskTeacher):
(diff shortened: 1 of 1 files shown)


parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -392,6 +392,7 @@ def get(self, episode_idx, entry_idx=0):
         text = dialog_entry_1['text']
         labels = [dialog_entry_2['text']]
 
+        assert isinstance(self.add_topic, bool)
```
klshuster: do we need this assert?

EricMichaelSmith: I think it'd be good to have it - I had a run train on the wrong data silently because self.add_topic was set to the string `"False"`, which the line below interpreted as `True` due to Python's sloppy polymorphic overloading of what `if` means

EricMichaelSmith:
(Will merge this in once I understand a problem preventing `python tests/datatests/test_new_tasks.py` from passing)

parlai/tasks/wizard_of_wikipedia/agents.py
klshuster: how did that happen?

EricMichaelSmith: Ha, it was a while ago, but I believe that that param was set to `"False"` as a result of the bug closed by https://github.com/facebookresearch/ParlAI/pull/2374 , where bool flags embedded within task names would be evaluated as strings. If it's really annoying to have this check we can remove it, but in general I like having checks like this that prevent silent errors, even if something else has to go wrong in order to trigger that error. Happy to discuss more :)

klshuster: I'm ok with leaving it, but yeah i suppose that's another reason that we should probably avoid flags in task names

EricMichaelSmith: Agreed - it's a useful hack to be able to put flags in task names but it's definitely problematic

PR #2383, opened 2020-02-04
Assertion to prevent double start in TGA.

stephenroller:
**Patch description**
Trying to fix some issues we've had before around double-start tokens being added to generation by just loudly failing.

**Testing steps**
The original failure of CI captured it a bit.


parlai/core/torch_generator_agent.py
```diff
@@ -564,7 +585,7 @@ def _init_cuda_buffer(self, batchsize, maxlen, force=False):
         if self.use_cuda and (force or not hasattr(self, 'buffer_initialized')):
             try:
                 self._control_local_metrics(disabled=True)
-                loss = self.compute_loss(self._dummy_batch(batchsize, maxlen))
+                loss = 0 * self.compute_loss(self._dummy_batch(batchsize, maxlen))
```
hadasah, line 588: was this causing a problem before, or is this just for extra clarity?

stephenroller: Extra clarity.

PR #2385, opened 2020-02-04
Acute-Eval helper code

EricMichaelSmith:
**Patch description**
- Allow passing in a file of workers to softban in Acute-Eval
- Move the controllable dialogue `.format_model_reply()` method into a new `parlai.utils.strings` module so that it can be used more broadly (for instance, to standardize the utterances of Acute-Eval dialogues)

**Testing steps**
Currently running command, which softblocks ~1100 workers:
```
python parlai/mturk/tasks/acute_eval/run.py \
--pairings-filepath /checkpoint/ems/parlai_scaling/scripts/s2020_01_31__acute_eval_meena/convo_sets/compiled_2020_02_03.jsonl \
-nc 250 \
--max-hits-per-worker 2 \
-r 0.5 \
--live \
--softblock-list-path /checkpoint/ems/parlai_scaling/scripts/s2020_01_31__acute_eval_meena/softblock_list.txt
```

### DIFF
--- a/parlai/utils/strings.py
+++ b/parlai/utils/strings.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Utility functions and classes for handling text strings.
+"""
+
+
+def format_model_reply(text: str) -> str:
+    """Standardize the capitalization and punctuation spacing of the input text."""
+
+    switch_list = [(' .', '.'), (' ,', ','), (' ?', '?'), (' !', '!'), (" ' ", "'")]
+
+    # add spaces so that words and punctuation can be seaprated
+    new_text = text.lower()
+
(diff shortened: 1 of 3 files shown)


EricMichaelSmith:
Got a CI error with `TestTransformerGenerator.test_beamsearch_contextblocking`, so that test may be flaky...

parlai/utils/strings.py
```diff
@@ -0,0 +1,55 @@
...
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Utility functions and classes for handling text strings.
+"""
+
+
+def standardize_string(text: str) -> str:
```
stephenroller: nit, let's find a more descriptive name. It makes larger assumptions than "is a string." `normalize_reply`? `normalize_english?` The assumptions are that this text is pretty much formatted by `re_tokenize`.

EricMichaelSmith: Yeah, definitely agree - changed to `normalize_reply`

PR #2389, opened 2020-02-06
Bring back code coverage.

stephenroller:
**Patch description**
I broke code coverage in #2363 when because the cache wasn't including coverage.py in the installation. As a fix, I switched to `python -m coverage`, but this conflicted with the `-m pytest`. The second was overriding, so we were running unit tests without coverage being generated. As such, it looked like our coverage metrics were plummeting.

**Testing steps**
We need to run CI and then rerun it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,17 +140,18 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: deps-v2-dt-{{ checksum "requirements.txt" }}
+          key: deps-v3-dt-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - save_cache:
-          key: deps-v2-dt-{{ checksum "requirements.txt" }}
+          key: deps-v3-dt-{{ checksum "requirements.txt" }}
           paths:
+            - "~/venv/bin"
             - "~/venv/lib"
       - <<: *installtorchcpu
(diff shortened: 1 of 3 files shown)


tests/test_eval_model.py
```diff
@@ -3,7 +3,7 @@
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-from examples.eval_model import setup_args
+from parlai.scripts.eval_model import setup_args
```
emilydinan: 😮

.circleci/config.yml
```diff
@@ -140,17 +140,18 @@ jobs:
...
       - restore_cache:
-          key: deps-v2-dt-{{ checksum "requirements.txt" }}
+          key: deps-v3-dt-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - save_cache:
-          key: deps-v2-dt-{{ checksum "requirements.txt" }}
+          key: deps-v3-dt-{{ checksum "requirements.txt" }}
           paths:
+            - "~/venv/bin"
```
emilydinan: why do you have to add this everywhere

stephenroller: So the `coverage` and `pytest` binaries are installed

PR #2391, opened 2020-02-06
Byte level bpe

dexterju:
**Patch description**
This PR adds a byte level bpe tokenizer into ParlAI, the original implementation is from hugging face.
We add this to provide more support to work closely with fairseq.
You can 
1. Load a pre-trained byte level bpe tokenizer 
2. train it from scratch by giving a test corpus. 

**Testing steps**
Test is added.
For saving:
```
python examples/train_model.py -t integration_tests  -mf /tmp/babi_new  -bs 1 -nt 4 -eps 5 -m memnn --no-cuda --dict-tokenizer bytelevelbpe --bpe-vocab ./data/test-byte-level-bpe-v1-vocab.json --bpe-merge ./data/test-byte-level-bpe-v1-merges.txt --validation-every-n-secs 5
```
For loading:
```
python examples/train_model.py -t integration_tests  -mf /tmp/babi_new  -bs 1 -nt 4 -eps 5 -m memnn --no-cuda --dict-tokenizer bytelevelbpe
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,7 @@ installdeps: &installdeps
       pip install -q -r requirements.txt
       python setup.py develop
       python -c "import nltk; nltk.download('punkt')"
+      pip3 install --progress-bar off tokenizers
 
 installtorchgpu14: &installtorchgpu14
   run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,8 @@ installtorchgpu14: &installtorchgpu14
       pip3 install --progress-bar off 'torchtext==0.5.0'
       pip3 install --progress-bar off 'torchvision==0.5.0'
       pip3 install --progress-bar off transformers
-      pip3 install --progress-bar off tokenizers
+      pip3 install --progress-bar off snips-nlu
+      pip3 install --progress-bar off --upgrade tokenizers
       python -c 'import torch; print("Torch version:", torch.__version__)'
       python -m torch.utils.collect_env
 
@@ -92,7 +93,8 @@ installtorchgpu13: &installtorchgpu13
       pip3 install --progress-bar off 'subword-nmt==0.3.7'
       pip3 install --progress-bar off pytorch-pretrained-bert
       pip3 install --progress-bar off transformers
-      pip3 install --progress-bar off tokenizers
+      pip3 install --progress-bar off snips-nlu
+      pip3 install --progress-bar off --upgrade tokenizers
       python -c 'import torch; print("Torch version:", torch.__version__)'
       python -m torch.utils.collect_env
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,11 +143,11 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: deps-v3-dt-{{ checksum "requirements.txt" }}
+          key: deps-v4-dt-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - save_cache:
-          key: deps-v3-dt-{{ checksum "requirements.txt" }}
+          key: deps-v4-dt-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
@@ -167,12 +167,12 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: deps-v3-osx-{{ checksum "requirements.txt" }}
+          key: deps-v4-osx-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchcpuosx
       - save_cache:
-          key: deps-v3-osx-{{ checksum "requirements.txt" }}
+          key: deps-v4-osx-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
@@ -194,12 +194,12 @@ jobs:
           command: circleci tests split --split-by=timings --timings-type=classname
       - <<: *fixgit
       - restore_cache:
-          key: deps-v3-ut36-{{ checksum "requirements.txt" }}
+          key: deps-v4-ut36-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchcpu
       - save_cache:
-          key: deps-v3-ut36-{{ checksum "requirements.txt" }}
+          key: deps-v4-ut36-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
@@ -218,12 +218,12 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: deps-v3-ut37-{{ checksum "requirements.txt" }}
+          key: deps-v4-ut37-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchcpu
       - save_cache:
-          key: deps-v3-ut37-{{ checksum "requirements.txt" }}
+          key: deps-v4-ut37-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
@@ -242,19 +242,19 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: nvidia-downloads-v3
+          key: nvidia-downloads-v4
       - <<: *setupcuda
       - save_cache:
-          key: nvidia-downloads-v3
+          key: nvidia-downloads-v4
           paths:
             - "~/nvidia-downloads/"
       - restore_cache:
-          key: deps-v3-gpu13-{{ checksum "requirements.txt" }}
+          key: deps-v4-gpu13-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchgpu13
       - save_cache:
-          key: deps-v3-gpu13-{{ checksum "requirements.txt" }}
+          key: deps-v4-gpu13-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
@@ -273,19 +273,19 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: nvidia-downloads-v3
+          key: nvidia-downloads-v4
       - <<: *setupcuda
       - save_cache:
-          key: nvidia-downloads-v3
+          key: nvidia-downloads-v4
           paths:
             - "~/nvidia-downloads/"
       - restore_cache:
-          key: deps-v3-gpu14-{{ checksum "requirements.txt" }}
+          key: deps-v4-gpu14-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchgpu14
       - save_cache:
-          key: deps-v3-gpu14-{{ checksum "requirements.txt" }}
+          key: deps-v4-gpu14-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
@@ -304,19 +304,19 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: nvidia-downloads-v3
+          key: nvidia-downloads-v4
       - <<: *setupcuda
       - save_cache:
-          key: nvidia-downloads-v3
+          key: nvidia-downloads-v4
           paths:
             - "~/nvidia-downloads/"
       - restore_cache:
-          key: deps-v3-nightly-{{ checksum "requirements.txt" }}
+          key: deps-v4-nightly-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchgpu14
       - save_cache:
-          key: deps-v3-nightly-{{ checksum "requirements.txt" }}
+          key: deps-v4-nightly-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
@@ -338,11 +338,11 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: deps-v3-mturk-{{ checksum "requirements.txt" }}
+          key: deps-v4-mturk-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - save_cache:
-          key: deps-v3-mturk-{{ checksum "requirements.txt" }}
+          key: deps-v4-mturk-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
@@ -360,7 +360,7 @@ jobs:
       - checkout
       - <<: *fixgit
       - restore_cache:
-          key: deps-v3-bw-{{ checksum "requirements.txt" }}
+          key: deps-v4-bw-{{ checksum "requirements.txt" }}
       - <<: *setup
       - <<: *installdeps
       - <<: *installtorchgpu14
@@ -371,7 +371,7 @@ jobs:
             pip install s3cmd
             sudo apt-get install linkchecker
       - save_cache:
-          key: deps-v3-bw-{{ checksum "requirements.txt" }}
+          key: deps-v4-bw-{{ checksum "requirements.txt" }}
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -78,6 +79,8 @@ installtorchgpu14: &installtorchgpu14
       pip3 install --progress-bar off 'torchtext==0.5.0'
       pip3 install --progress-bar off 'torchvision==0.5.0'
       pip3 install --progress-bar off transformers
+      pip3 install --progress-bar off snips-nlu
```
stephenroller: what is snips-nlu used for? Does tokenizers *really* require this?

parlai/core/dict.py
```diff
@@ -674,6 +695,8 @@ def save(self, filename=None, append=False, sort=True):
         elif self.tokenizer == 'gpt2':
             # never remove or sort tokens from gpt2
             pass
+        elif self.tokenizer == 'bytelevelbpe':
+            pass
```
stephenroller: leave the comment saying why

parlai/core/dict.py
```diff
@@ -690,6 +713,9 @@ def save(self, filename=None, append=False, sort=True):
         # save opt file
         with open(filename + '.opt', 'w', encoding='utf-8') as handle:
             json.dump(self.opt, handle, indent=4)
+        # save the byte level bpe model file as well
+        if self.tokenizer == 'bytelevelbpe':
+            self.byte_level_bpe.tokenizer(os.path.dirname(filename), filename)
```
stephenroller: is this right? This doesn't look right. Is the suffix implied? Is `tokenizer()` really the API call for the save? This just looks strange...

parlai/core/dict.py
```diff
@@ -818,6 +850,26 @@ def __str__(self):
         """
         return str(self.freq)
 
+    def sync_bytelevelbpe_dict(self):
```
stephenroller: make this private

parlai/core/dict.py
```diff
@@ -818,6 +850,26 @@ def __str__(self):
...
+        """
+        This will sync the dict agent and the hugging face bytelevel bpe dict.
```
stephenroller:
```python
"""
Sync the dictionary agent with Hugging Face tokenizer's BPE dict.

Called only once on initialization.
"""
```

parlai/core/dict.py
```diff
@@ -818,6 +850,26 @@ def __str__(self):
...
+            self.unk_token,
+            self.null_token,
+        ]
+        self.byte_level_bpe.tokenizer.add_special_tokens(special_tokens)
+        self.ind2tok = {}
+        self.tok2ind = {}
+        for i in range(self.byte_level_bpe.tokenizer.get_vocab_size()):
+            token = self.byte_level_bpe.tokenizer.id_to_token(i)
+            self.add_token(token)
+            self.freq[token] = 1
```
stephenroller: is there not a smarter value we can use?

parlai/core/dict.py
```diff
@@ -818,6 +850,26 @@ def __str__(self):
...
+            self.start_token,
+            self.end_token,
+            self.unk_token,
+            self.null_token,
+        ]
+        self.byte_level_bpe.tokenizer.add_special_tokens(special_tokens)
+        self.ind2tok = {}
+        self.tok2ind = {}
+        for i in range(self.byte_level_bpe.tokenizer.get_vocab_size()):
+            token = self.byte_level_bpe.tokenizer.id_to_token(i)
```
stephenroller: Is it worth doing the sort of "escaping" that we did like gpt-2?

tests/test_dict.py
```diff
@@ -18,6 +18,10 @@
 import os
 import shutil
 import unittest
+from parlai.core.build_data import download
+
+DEFAULT_BYTELEVEL_BPE_VOCAB = 'https://dl.fbaipublicfiles.com/parlai/test_bytelevel_bpe_v1/test-byte-level-bpe-v1-vocab.json'
```
stephenroller: These should be placed in a `parlai/_models/unittests/test_bytelevel_bpe_v1.tar.gz` and moved into the build statement in `parlai/zoo/unittests/build.py`. Then you can refer to them as `zoo:unittests/bytelevel_bpe_v1/convai2-vocab.json` and zoo:unittests/bytelevel_bpe_v1/convai2-merges.txt`.

tests/test_dict.py
```diff
@@ -60,6 +64,69 @@ def test_gpt2_bpe_tokenize(self):
             u'Hello, ParlAI! \U0001f600',
         )
 
+    def test_byte_level_bpe_tokenize(self):
+        # test loading
+        download(DEFAULT_BYTELEVEL_BPE_VOCAB, './', 'test-byte-level-bpe-vocab.json')
+        download(DEFAULT_BYTELEVEL_BPE_MERGE, './', 'test-byte-level-bpe-merges.txt')
```
stephenroller: See above.

parlai/core/hugging_face_bpe_helper.py
```diff
@@ -0,0 +1,43 @@
...
+from parlai.core.opt import Opt
+from typing import List
+
+
+class HuggingFaceBpeHelper(object):
+    @staticmethod
+    def add_cmdline_args(argparser):
+        parser = argparser.add_argument_group('HuggingFaceBpeHelper Arguments')
+        parser.add_argument(
+            '--bpe-vocab', type=str, help='path to pre-trained tokenizer vocab'
```
emilydinan: should these arguments have a default? (like the one you put in the zoo?)

tests/test_dict.py
```diff
@@ -60,6 +64,69 @@ def test_gpt2_bpe_tokenize(self):
...
+                'bpe_vocab': './test-byte-level-bpe-vocab.json',
+                'bpe_merge': './test-byte-level-bpe-merges.txt',
+            }
+        )
+        agent = DictionaryAgent(opt)
+        self.assertEqual(
+            # grinning face emoji
+            agent.bytelevelbpe_tokenize(u'Hello, ParlAI! \U0001f600'),
+            [
+                'H',
```
stephenroller: Worth pulling out into a constant at the top?

tests/test_dict.py
```diff
@@ -60,6 +64,69 @@ def test_gpt2_bpe_tokenize(self):
...
+                        'Ł',
+                        'ĺ',
+                        'Ģ',
+                    ]
+                ]
+            ),
+            # grinning face emoji
+            u'Hello, ParlAI! \U0001f600',
+        )
+        os.remove('./test-byte-level-bpe-vocab.json')
```
stephenroller: When you download into the correct folder, you won't need to clean up.

.circleci/config.yml
dexterju: This fixes a rust not found issue.

parlai/core/hugging_face_bpe_helper.py
```diff
@@ -0,0 +1,43 @@
...
+            raise ValueError('--bpe-merge is required for loading pretrained tokenizer')
+        self.vocab_path = opt.get('bpe_vocab')
+        self.merge_path = opt.get('bpe_merge')
+        self.tokenizer = ByteLevelBPETokenizer(self.vocab_path, self.merge_path)
+
+    def encode(self, text: str) -> List[str]:
+        return self.tokenizer.encode(text).tokens
+
+    def decode(self, x: List[str]) -> str:
+        return self.tokenizer.decode(self.tokenizer.token_to_id(c) for c in x)
```
emilydinan: are you sure that you need to call token_to_id individually here?

parlai/core/dict.py
dexterju: Thanks for flagging this, I will double check on this.

parlai/core/dict.py
dexterju: Could you elaborate on this?

tests/test_dict.py
```diff
@@ -60,6 +64,69 @@ def test_gpt2_bpe_tokenize(self):
...
+        )
+        self.assertEqual(
+            agent.vec2txt(
+                [
+                    agent.tok2ind[w]
+                    for w in [
+                        'H',
+                        'ello',
+                        ',',
+                        'Ġ',
```
EricMichaelSmith: Oy - how does the G with the dot equal a space between "Hello," and "ParlAI!" below?

parlai/core/hugging_face_bpe_helper.py
dexterju: Actually not, hmm, the purpose of having this is for loading fairseq / pretrained tokenizers, and they should be given each time. There is no universal one here.

parlai/core/hugging_face_bpe_helper.py
dexterju: It depends on how the decoding interface works, our dict expects a list of tokens, while the hugging face expects a list of token ids. 
What do you suggest here?

tests/test_dict.py
dexterju: It splits by byte level and then just cast it to Unicode I think.

parlai/core/dict.py
dexterju: What do you mean?

dexterju:
Thanks for reviewing!!! I will apply all the comments and leave the fight on whether we should merge it to you two lol @stephenroller @emilydinan haha

parlai/core/dict.py
stephenroller: Like using the “xff” type characters like we did with GPT2

stephenroller:
Yeah I’d like that too Emily. Just not today. DictAgent deserves a true refactor

tests/test_dict.py
dexterju: It should be unittest instead of unittests I think.

parlai/core/dict.py
stephenroller: resolved offline

parlai/core/hugging_face_bpe_helper.py
stephenroller: Actually... hold up. let's talk about this more, i just realized something

.circleci/config.yml
```diff
@@ -67,6 +67,7 @@ installdeps: &installdeps
       pip install -q -r requirements.txt
       python setup.py develop
       python -c "import nltk; nltk.download('punkt')"
+      pip3 install --progress-bar off tokenizers
```
stephenroller: I think we should just add tokenizers==0.4.2 to the requirements.txt.

.circleci/config.yml
stephenroller: I don't think we need it, can we try one more time to get rid of it with the requirements.txt change?

parlai/core/dict.py
```diff
@@ -336,6 +338,31 @@ def __init__(self, opt: Opt, shared=None):
...
+                )
+            if self.maxtokens > 0 or self.minfreq > 0:
+                raise ValueError(
+                    'You should not filter vocabulary with using --dict-tokenizer bytelevelbpe'
+                    ' (no --dict-minfreq or --dict-maxtokens).'
+                )
+            opt_for_byte_level_bpe = copy.copy(opt)
+            if loaded:
+                dict_basename = os.path.splitext(opt['dict_file'])[0]
+                if os.path.isfile(
```
stephenroller: should we just do them seperately?

parlai/core/dict.py
```diff
@@ -690,6 +726,11 @@ def save(self, filename=None, append=False, sort=True):
         # save opt file
         with open(filename + '.opt', 'w', encoding='utf-8') as handle:
             json.dump(self.opt, handle, indent=4)
+        # save the byte level bpe model file as well
+        if self.tokenizer == 'bytelevelbpe':
+            self.byte_level_bpe.tokenizer.save(
+                os.path.dirname(filename), os.path.splitext(filename)[0]
```
stephenroller: leave a comment saying "this puts it at filename.vocab and finlename.merges" or whatever

parlai/core/dict.py
stephenroller: We agreed not to do this.

parlai/core/hugging_face_bpe_helper.py
```diff
@@ -0,0 +1,43 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from parlai.core.opt import Opt
+from typing import List
+
+
+class HuggingFaceBpeHelper(object):
+    @staticmethod
+    def add_cmdline_args(argparser):
+        parser = argparser.add_argument_group('HuggingFaceBpeHelper Arguments')
```
stephenroller: Slightly better name here, this is user readable. Maybe ByteLevelBPE Arguments?

parlai/core/hugging_face_bpe_helper.py
stephenroller: Resolved offline.

parlai/core/hugging_face_bpe_helper.py
```diff
@@ -0,0 +1,43 @@
...
+            '--bpe-merge', type=str, help='path to pre-trained tokenizer merge'
+        )
+        return parser
+
+    def __init__(self, opt: Opt, shared=None):
+        try:
+            from tokenizers import ByteLevelBPETokenizer
+        except ImportError:
+            raise ImportError(
+                'Please install huggingface tokenizer with: pip install tokenizers'
```
stephenroller: Please install **H**ugging**F**ace tokenizer**s**

parlai/core/hugging_face_bpe_helper.py
```diff
@@ -0,0 +1,43 @@
...
+            )
+
+        if opt.get('bpe_vocab', None) is None:
```
stephenroller: prefer `'bpe_vocab' in opt`

parlai/core/hugging_face_bpe_helper.py
```diff
@@ -0,0 +1,43 @@
...
+            raise ValueError('--bpe-vocab is required for loading pretrained tokenizer')
+        if opt.get('bpe_merge', None) is None:
+            raise ValueError('--bpe-merge is required for loading pretrained tokenizer')
+        self.vocab_path = opt.get('bpe_vocab')
```
stephenroller: and now use `opt['bpe_vocab']`

parlai/core/hugging_face_bpe_helper.py
stephenroller: I think this is good, we can look for an improvement later if we think it's a bottleneck.

parlai/core/params.py
```diff
@@ -949,24 +947,17 @@ def _process_args_to_opts(self, args_that_override: Optional[List[str]] = None):
...
-        if self.opt.get('dict_file') is not None:
-            self.opt['dict_file'] = modelzoo_path(
-                self.opt.get('datapath'), self.opt['dict_file']
-            )
-        if self.opt['override'].get('dict_file') is not None:
-            # also check override
-            self.opt['override']['dict_file'] = modelzoo_path(
-                self.opt.get('datapath'), self.opt['override']['dict_file']
-            )
+        options_to_change = {'model_file', 'dict_file', 'bpe_vocab', 'bpe_merge'}
```
stephenroller: thanks for cleaning this up

tests/test_dict.py
```diff
@@ -60,6 +84,29 @@ def test_gpt2_bpe_tokenize(self):
             u'Hello, ParlAI! \U0001f600',
         )
 
+    def test_byte_level_bpe_tokenize(self):
+        # test loading
+        import parlai.scripts.train_model as tms
+
+        parser = tms.setup_args()
```
stephenroller: instead of `tms.setup_args` do `parser = ParlaiParser()`

tests/test_dict.py
```diff
@@ -60,6 +84,29 @@ def test_gpt2_bpe_tokenize(self):
             u'Hello, ParlAI! \U0001f600',
         )
 
+    def test_byte_level_bpe_tokenize(self):
```
stephenroller: leave a docstring describing what this is testing.

parlai/core/dict.py
```diff
@@ -336,6 +338,31 @@ def __init__(self, opt: Opt, shared=None):
...
+            if self.lower:
+                raise ValueError(
+                    'Only use --dict-lower false with --dict-tokenizer bytelevelbpe'
+                )
+            if self.maxtokens > 0 or self.minfreq > 0:
+                raise ValueError(
+                    'You should not filter vocabulary with using --dict-tokenizer bytelevelbpe'
+                    ' (no --dict-minfreq or --dict-maxtokens).'
+                )
+            opt_for_byte_level_bpe = copy.copy(opt)
```
stephenroller: why not just change the opt?

dexterju: Thought we didn't want to change opt?

dexterju:
I think that fix for most of the comments, will just wait the CI finishes and then double check the dependency.

parlai/core/dict.py
stephenroller: Yes, please change opt here.

parlai/core/dict.py
```diff
@@ -764,6 +809,11 @@ def vec2txt(self, vector, delimiter=' '):
         """
         if self.tokenizer == 'gpt2' and not self.opt.get('bpe_debug', False):
             return self.gpt2_bpe.decode(self[int(idx)] for idx in vector)
+        if self.tokenizer == 'bytelevelbpe' and not self.opt.get('bpe_debug', False):
+            vector = [
+                idx + len(self.tok2ind) - 4 if idx < 4 else idx - 4 for idx in vector
```
stephenroller: comment explaining why we do this

dexterju:
All comments addressed.
Add prefix space added.
Tests have been added accordingly. 
Let us wait to hear what the CI says before we merge them lol!

PR #2392, opened 2020-02-06
[FP16] Memory efficient optimizer

emilydinan:
**Patch description**
Add a wrapper for a memory efficient FP16 optimizer. Based heavily on the fairseq implementation, which can be found here: https://github.com/pytorch/fairseq/blob/master/fairseq/optim/fp16_optimizer.py. Preliminary tests seems to show this saves about ~6.5 GB when finetuning GPT2-large.

Additionally, added a compatible Adam and Adafactor optimizer (also based on fairseq implementation). 

**Testing steps**
`python -m pytest /private/home/edinan/ParlAI/tests/test_mem_efficient_fp16.py`

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -38,7 +38,7 @@ from parlai.core.message import Message
 from parlai.utils.fp16 import (
     fp16_apex_available,
     fp16_optimizer_wrapper,
-    ParlAIFP16MemoryEfficientOptimizer
+    ParlAIFP16MemoryEfficientOptimizer,
 )
 from parlai.utils.misc import AttrDict, warn_once, round_sigfigs
 from parlai.utils.torch import argsort, padded_tensor
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -40,6 +40,7 @@ from parlai.utils.fp16 import (
     fp16_optimizer_wrapper,
     MemoryEfficientFP16Optimizer,
     MemoryEfficientFP16Adam,
+    Adafactor,
 )
 from parlai.utils.misc import AttrDict, warn_once, round_sigfigs
 from parlai.utils.torch import argsort, padded_tensor
@@ -366,6 +367,7 @@ class TorchAgent(ABC, Agent):
 
         # now pull in memory efficient implementations
         optims['mem_eff_adam'] = MemoryEfficientFP16Adam
+        optims['adafactor'] = Adafactor
 
         return optims
 
@@ -478,6 +480,13 @@ class TorchAgent(ABC, Agent):
             hidden=True,
             help='Epsilon value for Adam optimizers. Set to 1e-6 if your '
             'large model has stability issues, but prefer the default.',
+        )
+        optim_group.add_argument(
+            '--adafactor-eps',
+            default='1e-30,1e-3',
+            type='floats',
+            help='Epsilon values for adafactor optimizer: regularization'
+            'constants for square gradienta nd parameter scale respectively',
         )
         optim_group.add_argument(
             '-mom',
@@ -811,7 +820,19 @@ class TorchAgent(ABC, Agent):
         elif opt['optimizer'] == 'qhadam':
             # set nus for qhadam
             kwargs['nus'] = opt.get('nus', (0.7, 1.0))
-        if opt['optimizer'] in ['adam', 'sparseadam', 'fused_adam', 'adamax', 'qhadam']:
+        elif opt['optimizer'] == 'adafactor':
+            # adafactor params
+            kwargs['beta1'] = opt.get('betas', (0.9, 0.999))[0]
+            kwargs['eps'] = opt['adafactor_eps']
+            kwargs['warmup_init'] = opt.get('warmup_updates', -1) > 0
+
+        if opt['optimizer'] in [
+            'adam',
+            'sparseadam',
+            'fused_adam',
+            'adamax',
+            'qhadam',
+        ]:
             # set betas for optims that use it
             kwargs['betas'] = opt.get('betas', (0.9, 0.999))
             # set adam optimizer, but only if user specified it
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -854,7 +854,6 @@ class TorchAgent(ABC, Agent):
                 'switch to --optimizer adam.'
             )
 
-        # TODO: if self.fp16 and self.fp16_impl == 'mem_efficient'... check that class is ok
         optim_class = self.optim_opts()[opt['optimizer']]
         self.optimizer = optim_class(params, **kwargs)
         if self.fp16:
@@ -862,6 +861,15 @@ class TorchAgent(ABC, Agent):
                 self.optimizer = fp16_optimizer_wrapper(self.optimizer)
             else:
                 # Using memory efficient optimizer
+                opt_name = opt['optimizer']
+                compatible_list = MemoryEfficientFP16Optimizer.compatible_optimizers()
+                is_compat = opt_name in compatible_list
+                if not is_compat:
+                    raise RuntimeError(
+                        f'The optimizer you selected {opt_name} is not compatible '
+                        'with Memory Efficient FP16. Please select from among this '
+                        f'list:\n{compatible_list}'
+                    )
                 self.optimizer = MemoryEfficientFP16Optimizer(self.optimizer)
         # TODO: we might want to hard reset optimizers here in the
         # case of fine tuning. Some rudimentary experiments seemed to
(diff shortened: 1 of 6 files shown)


parlai/core/torch_agent.py
```diff
@@ -433,6 +437,13 @@ def add_cmdline_args(cls, argparser):
         agent.add_argument(
             '--fp16', type='bool', default=False, help='Use fp16 computations.'
         )
+        agent.add_argument(
+            '--fp16-impl',
+            type=str,
+            default='apex',
+            choices=['apex', 'mem_efficient'],
```
EricMichaelSmith: Under what conditions would we still want to use apex? Maybe for the optimizers that this won't work with?

parlai/core/torch_agent.py
```diff
@@ -618,7 +629,12 @@ def __init__(self, opt: Opt, shared=None):
             if not shared and opt['gpu'] != -1:
                 torch.cuda.set_device(opt['gpu'])
         # indicate whether using fp16
-        self.fp16 = self.use_cuda and self.opt.get('fp16', False) and fp16_available()
+        self.fp16 = self.use_cuda and self.opt.get('fp16', False)
+        if self.fp16:
+            # check that the implementation requested is available
+            self.fp16_impl = self.opt.get('fp16_impl', 'apex')
+            if self.fp16_impl == 'apex' and not fp16_apex_available():
+                self.fp16 = False
```
EricMichaelSmith: Should we print a warning in this case?

parlai/core/torch_agent.py
```diff
@@ -814,11 +833,15 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
...
 
+        # TODO: if self.fp16 and self.fp16_impl == 'mem_efficient'... check that class is ok
         optim_class = self.optim_opts()[opt['optimizer']]
         self.optimizer = optim_class(params, **kwargs)
         if self.fp16:
-            self.optimizer = fp16_optimizer_wrapper(self.optimizer)
-
+            if self.fp16_impl == 'apex':
+                self.optimizer = fp16_optimizer_wrapper(self.optimizer)
+            else:
```
EricMichaelSmith: nit: `elif self.fp16_impl == 'mem_efficient'` here in case we want to add further choices down the line?

parlai/utils/fp16.py
```diff
@@ -0,0 +1,466 @@
...
+    params = list(params)
+    if len(params) == 1:
+        p = params[0]
+        grad_norm = torch.norm(p)
+        if grad_norm > max_norm > 0:
+            clip_coef = max_norm / (grad_norm + 1e-6)
+            p.mul_(clip_coef)
+        return grad_norm
+    elif max_norm > 0:
+        return torch.nn.utils.clip_grad_norm_(params, max_norm)
```
EricMichaelSmith: hmm - why not use `torch.nn.utils.clip_grad_norm_()` if len(params) == 1?

parlai/utils/fp16.py
```diff
@@ -0,0 +1,466 @@
...
+        self,
+        init_optimizer: torch.optim.Optimizer,  # type: ignore
+        loss_initial_scale: float = 2.0 ** 17,
+        min_loss_scale: float = 1e-4,
+    ):
+        self.optimizer = init_optimizer
+        # TODO: we should probably set a bunch of these args with opt?
+        self.min_loss_scale = min_loss_scale
+        self.scaler = DynamicLossScaler(init_scale=loss_initial_scale)
+        # TODO: add warning here about which optimizers this works with
```
EricMichaelSmith: 👍

parlai/core/torch_agent.py
emilydinan: Yeah. So I should write more about the tradeoffs inside the class docstring, but basically (1) memory efficinet is not compatible with everything and (2) you want to use apex if memory constraints are not an issue, as there are stability issues with the memory efficient optimizer

parlai/core/torch_agent.py
emilydinan: oh yeah. we didn't before but we probably should

emilydinan: actually the warning occurs in the function

emilydinan:
ready for review

parlai/core/torch_agent.py
```diff
@@ -433,6 +439,13 @@ def add_cmdline_args(cls, argparser):
         agent.add_argument(
             '--fp16', type='bool', default=False, help='Use fp16 computations.'
         )
+        agent.add_argument(
+            '--fp16-impl',
+            type=str,
+            default='apex',
+            choices=['apex', 'mem_efficient'],
```
klshuster: so we technically don't even need apex anymore to run fp16?

parlai/core/torch_agent.py
```diff
@@ -468,6 +481,13 @@ def add_cmdline_args(cls, argparser):
             help='Epsilon value for Adam optimizers. Set to 1e-6 if your '
             'large model has stability issues, but prefer the default.',
         )
+        optim_group.add_argument(
+            '--adafactor-eps',
+            default='1e-30,1e-3',
+            type='floats',
+            help='Epsilon values for adafactor optimizer: regularization'
```
klshuster: nit: need a space here at the end of the string

parlai/core/torch_agent.py
```diff
@@ -468,6 +481,13 @@ def add_cmdline_args(cls, argparser):
...
+            'constants for square gradienta nd parameter scale respectively',
```
klshuster: double nit: `for square gradients and`

parlai/core/torch_agent.py
```diff
@@ -468,6 +481,13 @@ def add_cmdline_args(cls, argparser):
             help='Epsilon value for Adam optimizers. Set to 1e-6 if your '
             'large model has stability issues, but prefer the default.',
         )
+        optim_group.add_argument(
+            '--adafactor-eps',
+            default='1e-30,1e-3',
```
klshuster: are these recommended values? i.e. might be good to have them marked as `recommended=` in the `add_argument` function

parlai/core/torch_agent.py
```diff
@@ -789,10 +814,25 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
...
+            kwargs['beta1'] = opt.get('betas', (0.9, 0.999))[0]
+            kwargs['eps'] = opt['adafactor_eps']
+            kwargs['warmup_init'] = opt.get('warmup_updates', -1) > 0
+
+        if opt['optimizer'] in [
+            'adam',
+            'sparseadam',
+            'fused_adam',
+            'adamax',
+            'qhadam',
```
klshuster: we don't need to set `betas` with the `mem_eff_adam`?

parlai/utils/fp16.py
```diff
@@ -0,0 +1,677 @@
...
+        return torch.sqrt(
+            sum(p.grad.data.norm() ** 2 for p in params if p.grad is not None)
+        )
+
+
+def has_overflow(grad_norm):
+    """
+    Detect inf and NaN in grad_norm.
+    """
+    if grad_norm == float('inf') or grad_norm != grad_norm:
```
klshuster: woah `grad_norm != grad_norm` works for NaN?

parlai/utils/fp16.py
```diff
@@ -0,0 +1,677 @@
...
+
+
+def fp16_apex_available() -> bool:
+    try:
+        import apex.fp16_utils  # noqa: F401
+
+        return True
+    except ImportError:
+        warn_once(
+            'You set --fp16 true, but fp16 is unavailable. To use fp16, please '
```
klshuster: should we modify this warning to indicate that one could also set `--fp16-impl` to `mem_efficient`?

parlai/utils/fp16.py
```diff
@@ -0,0 +1,677 @@
...
+
+
+###################################################
+## Memory Efficient Wrappers
+###################################################
+
+
+class DynamicLossScaler(object):
+    """
+    Shamelessly stolen from Fairseq.
```
klshuster: 😆

klshuster: We might want to change this wording before merging haha

parlai/core/torch_agent.py
emilydinan: technically. but you probably only want to use mem_efficient in the case where memory is a concern. the memory efficient version can be a bit unstable, and currently only supports a couple of optimizers (adam, adafactor)

parlai/core/torch_agent.py
emilydinan: no, instead we take the first beta only

parlai/utils/fp16.py
emilydinan: apparently! this was taken directly from fairseq!

parlai/utils/fp16.py
stephenroller:
```
$ ag stolen
docs/source/conf.py
186:    # Stolen shamelessly from Lasagne! Thanks Lasagne!

parlai/utils/distributed.py
146:    # stolen shamelessly from fairseq
```

PR #2393, opened 2020-02-07
Asymmetrical transformers

dexterju:
**Patch description**
This PR allows asymmetrical transformers. The transformers who have different number of layers in encoder and decoder.
We add two options, num-encoder-layer, num-decoder-layer, they will override the n-layers option if they are given. 

**Testing steps**
For testing the original transformers for backward compatible. 
```
python examples/train_model.py -t twitter -mf /tmp/tr_twitter -m transformer/ranker -bs 10 -vtim 3600 -cands batch -ecands batch --data-parallel True
```
```
 python examples/train_model.py -t twitter -mf /tmp/tr_twitter -m transformer/generator -bs 10 -vtim 3600   --num-encoder-layers 1 --num-decoder-layers 4
```
**Logs**
```
[ num words =  366522 ]
/private/home/daju/ParlAI/parlai/agents/transformer/modules.py:73: UserWarning: Building a Asymmetrical Transformer with encoder size of 1.
  warn_once('Building a Asymmetrical Transformer with encoder size of {}.'.format(opt['num_encoder_layers']))
/private/home/daju/ParlAI/parlai/agents/transformer/modules.py:104: UserWarning: Building a Asymmetrical Transformer with decoder size of 4.
  warn_once('Building a Asymmetrical Transformer with decoder size of {}.'.format(opt['num_decoder_layers']))
```

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -99,7 +99,7 @@ def _build_decoder(
     return TransformerDecoder(
         n_heads=opt['n_heads'],
         n_layers=(
-            opt['n_encoder_layers'] if opt['n_encoder_layers'] > 0 else opt['n_layers']
+            opt['n_decoder_layers'] if opt['n_decoder_layers'] > 0 else opt['n_layers']
         ),
         embedding_size=opt['embedding_size'],
         ffn_size=opt['ffn_size'],
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -72,9 +72,7 @@ def _build_encoder(
     return TransformerEncoder(
         n_heads=opt['n_heads'],
         n_layers=(
-            opt['n_encoder_layers']
-            if opt['n_encoder_layers'] > 0
-            else opt['n_layers']
+            opt['n_encoder_layers'] if opt['n_encoder_layers'] > 0 else opt['n_layers']
         ),
         embedding_size=opt['embedding_size'],
         ffn_size=opt['ffn_size'],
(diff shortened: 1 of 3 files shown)


parlai/agents/transformer/modules.py
```diff
@@ -69,9 +69,19 @@ def _build_encoder(
     n_positions=1024,
     n_segments=0,
 ):
+    if opt['num_encoder_layers'] > 0:
+        warn_once(
+            'Building a Asymmetrical Transformer with encoder size of {}.'.format(
```
stephenroller: we don't need the warning.

parlai/agents/transformer/transformer.py
```diff
@@ -102,6 +102,18 @@ def add_common_cmdline_args(argparser):
...
         'in the memory network',
     )
+    argparser.add_argument(
+        '--num-encoder-layers',
+        type=int,
+        default=-1,
+        help='This will overide the n-layers for asymmetrical transformers',
+    )
+    argparser.add_argument(
+        '--num-decoder-layers',
```
stephenroller: i guess let's use --n-decoder-layers for consistency with --n-layers

parlai/agents/transformer/transformer.py
```diff
@@ -102,6 +102,18 @@ def add_common_cmdline_args(argparser):
         help='Share word embeddings table for candidate and context'
         'in the memory network',
     )
+    argparser.add_argument(
+        '--num-encoder-layers',
```
stephenroller: add a short option `-nel` and `-ndl` for convenience i guess.

tests/test_transformers.py
```diff
@@ -516,6 +518,43 @@ def test_xlm(self):
         self.assertLessEqual(test['ppl'], 1.30)
         self.assertGreaterEqual(test['bleu-4'], 0.90)
 
+    def test_asymmetry(self):
```
dianaglzrico: Hi @dexterju :) this PR looks very useful, I really like it! Just a tiny question (reply to me when you have some free time, no worries) why is this function necessary? Is it just a sanity check to verify the parameters are being overridden?

dexterju: Hey, This unit test function will just check these parameters are set correctly. 
We want the logic to be if n-encoder-layers are explicitly given, it should use it, otherwise it should use n-layers instead to ensure backward compatibility (same case for the decoders). It is good to have this test ensuring this logic is respected.

dianaglzrico: Ahhh I see, thanks!

PR #2398, opened 2020-02-11
[TGA] Flexible context for n-gram blocking

emilydinan:
**Patch description**
Added a `_get_context` function so that more complex models can override the context that is used for n-gram blocking.


parlai/core/torch_generator_agent.py
```diff
@@ -911,8 +919,10 @@ def _generate(self, batch, beam_size, max_ts):
             else len(batch.image)
         )
         if batch.text_vec is not None:
+            batchsize = batch.text_vec.size(0)
```
stephenroller: we gotta kill this

emilydinan: why

PR #2405, opened 2020-02-14
Conversations

emilydinan:
**Patch description**
First pass at a Conversations class for ease of saving and viewing conversation data (and keeping around relevant metadata). 

The first use case I implemented is for saving self chats. This also comes with a pretty big self chat overhaul:
- Self Chats now function much the same way as Interactive Worlds, instead of before where they were like quasi-interactive worlds that involved a lot of opaque logic and required you to add `:selfchat` to the end of your task flag. Now, if you run the self chat script with `-t convai2` for instance, it automatically loads the appropriate self chat world.
- Got rid of metrics logging in self chat worlds (all it printed was time, and was distracting)
- Changed the logic around interactive worlds and self chat worlds so that they do not require adding a `create_agents` function to the associated task agents file that returned a `[]` (I always found this pretty confusing)
- Changed some arguments in self chat, to allow you to run multiple self chats in a row for example (`--num-self-chats 10`)
- Automatically save self chats in Conversations format.

Running for example:
`python parlai/scripts/self_chat.py -t convai2 -mf zoo:pretrained_transformers/model_bi/model --num-self-chats 10`

We can then load the conversations via
`convo = Conversations('/tmp/Biencoder_selfchat.jsonl')`
`convo.read_rand_conv()` yields:

```
==================================================
context: [{'text': "your persona: i eat mainly chinese food.\nyour persona: i'm single and live alone.\nyour persona: i make robots for a living.\nyour persona: i'm a night owl.\nyour persona: i do not have a good social life.\n", 'id': 'Biencoder_1'}, {'text': "your persona: my parents immigrated from europe before i was born.\nyour persona: while both my parents have thick european accents , i don't.\nyour persona: i drive a motorcycle instead of a car.\nyour persona: i am enrolled in several college credit courses.\nyour persona: i go to a private high school.\n", 'id': 'Biencoder_2'}]
--------------------------------------------------
Biencoder_1: hello , how are you ? i am in college and ride my bike .
Biencoder_2: i too live alone . my job is making robots . what about you ?
Biencoder_1: i am currently taking some college level classes at my private high school .
Biencoder_2: teaching must be hard because of getting up early , i am a night person .
Biencoder_1: my parents migrated here from germany . very little english .
Biencoder_2: nice there is a great chinese restraint i love to go to when i am there .
Biencoder_1: i think japanese is one of the harder languages to learn .
Biencoder_2: it is interesting . yet i do not have much of a social life .
Biencoder_1: maybe i could help you learn japanese .
Biencoder_2: yeah that would be cool , perhaps i can teach you some japanese
==================================================
```

`convo.read_metadata()` yields:
```
Saved at: 2020-02-26 14:18:13.988496
Self chat: True
Speaker 1: Biencoder_1
Speaker 2: Biencoder_2
Opt:
	init_opt: None
	show_advanced_args: False
	task: convai2
	download_path: /private/home/edinan/ParlAI/downloads
	...
	starttime: Feb26_14-17
```


parlai/utils/conversations.py
```diff
@@ -0,0 +1,188 @@
...
+        print(f'Speaker 2: {self.speaker_2}')
+        print('Opt:')
+        for k, v in self.opt.items():
+            print(f'\t{k}: {v}')
+        for k, v in self.extra_data.items():
+            print(f'{k}: {v}')
+
+    @staticmethod
+    def _get_path(datapath):
+        fle, _ = os.path.splittext(datapath)
```
klshuster: what does `splittext` do?

parlai/utils/conversations.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class Conversations:
+    """
+    Utility class for reading and writing from ParlAI Conversations format.
+
+    Conversations should be saved in JSONL format, where each line is
+    a JSON of the following form:
+    {
+        'possible_conversation_level_info': True,
```
klshuster: so what would this be? or is this just for the draft

parlai/utils/conversations.py
```diff
@@ -0,0 +1,188 @@
...
+            'speaker_1': <identity of speaker 1>,
+            'speaker_2': <identify of speaker 2>,
+            ...
+            Other arguments.
+        }
+        """
+        try:
+            metadata = Metadata(datapath)
+            return metadata
+        except RuntimeError:
```
klshuster: should we also `except` other errors like `KeyError` if we're missing some fields in the MetaData?

jsedoc:
Have you considered the forever chat format? https://github.com/jkeen/forever-chat-format

parlai/utils/conversations.py
emilydinan: it splits the path from the file extension

emilydinan:
> could we also include functions within the `Conversations` object for iterating through the conversations?

great idea

parlai/utils/conversations.py
emilydinan: this is just for the draft. i left it open such that you can add whatever other information you want in there.

parlai/utils/conversations.py
```diff
@@ -59,8 +59,8 @@ def read(self):
 
     @staticmethod
     def _get_path(datapath):
-        fle, _ = os.path.splittext(datapath)
-        return os.path.join(fle, '.metadata')
+        fle, _ = os.path.splitext(datapath)
```
stephenroller: nice, i did not know this!

jaseweston:
aren't we somehow duplicating with the WorldLogger class? why not update that?

parlai/utils/conversations.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python3
```
stephenroller: let's put it more central! parlai/core/conversations.py!

parlai/utils/conversations.py
```diff
@@ -0,0 +1,188 @@
...
+        self_chat=False,
+        speaker_1='human',
+        speaker_2=None,
+        **kwargs,
+    ):
+        """
+        Dump conversation metadata to file.
+        """
+        metadata = {}
+        metadata['date'] = str(datetime.datetime.now())
```
stephenroller: Other than opt, should we consider moving this meta data into the conversation itself? e.g. datetime might legitimately want to be controlled independently (say, as we output from a log).

parlai/utils/conversations.py
```diff
@@ -0,0 +1,188 @@
...
+        for k, v in kwargs.items():
+            metadata[k] = metadata[v]
+
+        metadata_path = cls._get_path(datapath)
+        print(f'[ Writing metadata to file {metadata_path} ]')
+        with open(metadata_path, 'w') as f:
+            f.write(json.dumps(metadata))
+
+
+class Conversations:
```
stephenroller: I'm wondering if the base level shouldn't be a `Conversation`, and conversations would fundamentally be a collection of objects (therefore implmenting `__len__`, `__iter__`, ...). Metadata would (as you have here) also be a part of Conversation.

The future I'm imagining is that the teacher code is also Conversation oriented. So when I need to create a new dataset from some arbitrary data, I need to parse out a single Conversation and emit it. Higher level logic could then turn a Conversation into a teacher, building the labels, handling the symmetry or not, even building label candidates automatically...

Building a conversation should be similarly easy, and grouping waves of them into a Conversation.

Probably we want to think this through more.

parlai/utils/conversations.py
```diff
@@ -0,0 +1,188 @@
...
+        }
+        """
+        try:
+            metadata = Metadata(datapath)
+            return metadata
+        except RuntimeError:
+            print('Metadata does not exist. Please double check ' 'your datapath.')
+            return None
+
+    def read_metadata(self):
```
stephenroller: why would this be called separately?

jaseweston:
> aren't we somehow duplicating with the WorldLogger class? why not update that?

ping. well guess you will have to update that as well to use this when it's finished?

emilydinan:
yes @stephenroller and @jaseweston, i planned to add all of the things you mentioned, this is a WIP draft right now. will ping when ready for final review. mostly wanted feedback on the format

jaseweston:
cool

parlai/utils/conversations.py
emilydinan: so having another class called `Conversation` to replace the idea of episodes?

parlai/utils/conversations.py
emilydinan: Metadata is supposed to be conversation**S** level and not conversation level. Datetime is written when the conversations logs are written. 

There can be other information included in each individual conversation, as keys in the dict other than 'dialog' (like 'persona' for instance).

emilydinan:
pinging, ready for review

parlai/scripts/self_chat.py
```diff
@@ -30,92 +28,97 @@ def setup_args(parser=None):
     )
     parser.add_argument(
         '-it',
```
klshuster: nit: perhaps change to `-st` as well

parlai/scripts/self_chat.py
```diff
@@ -30,92 +28,97 @@ def setup_args(parser=None):
...
+    for _ in range(opt['num_self_chats']):
+        _run_self_chat_episode(opt, world, logger)
+
+    # Save chats
+    if opt['outfile'] is None:
+        outfile = '/tmp/{}_selfchat'.format(model_id)
+    else:
+        outfile = opt['outfile']
+
+    if opt['save_format'] == 'conversations' and hasattr(world, 'write'):
```
klshuster: so we only use `world.write` if the `save_format` is `conversations`? should this maybe be a separate option from `conversations`? or are we only supporting `world.write` with `conversations`, assuming that `world.write` will write in the conversation format?

parlai/tasks/self_chat/worlds.py
```diff
@@ -124,16 +123,11 @@ def parley(self):
 
         if self.turn_cnt == 0:
             self.acts = [None, None]
-            # choose speaking order:
-            if random.choice([0, 1]):
```
klshuster: so is the order always the same now?

parlai/utils/conversations.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def _load(self, datapath):
+        metadata_path = self._get_path(datapath)
+        if not os.path.isfile(metadata_path):
+            raise RuntimeError(
+                f'Metadata at path {metadata_path} not found. '
+                'Double check your path.'
+            )
+
+        with open(metadata_path, 'rb') as f:
```
klshuster: why `rb`?

parlai/utils/conversations.py
```diff
@@ -0,0 +1,279 @@
...
+        Read the relevant metadata.
+        """
+        print(f'Saved at: {self.datetime}')
+        print(f'Self chat: {self.self_chat}')
+        print(f'Speaker 1: {self.speaker_1}')
+        print(f'Speaker 2: {self.speaker_2}')
+        print('Opt:')
+        for k, v in self.opt.items():
+            print(f'\t{k}: {v}')
+        for k, v in self.extra_data.items():
```
klshuster: perhaps `print('Extra Data: ')` before this as well?

parlai/utils/conversations.py
klshuster: The name `Conversation` makes more intuitive sense than `episode` anyway so @stephenroller might be onto something.

I kind of agree that we might want to go the `Conversation` route and have `Conversations` be a collection of `Conversation` objects. At the very least this would perhaps ease the logic around what conversation-level information is available, e.g. including the `personas` or other context (rather than just reading through the arbitrary keys in the dialogue).

Idk i do think this is good for now but could definitely warrant a more in-depth discussion

parlai/utils/conversations.py
```diff
@@ -0,0 +1,279 @@
...
+            self.metadata.read()
+
+    def next(self):
+        """
+        Return the next conversation.
+        """
+        if self.iterator_idx >= self.num_conversations:
+            print('You reached the end of the conversations.')
+            self.reset()  # return the iterator idx to 0
+            return None
```
klshuster: would we want to return a new convo anyway?

parlai/utils/conversations.py
```diff
@@ -0,0 +1,279 @@
...
+                        ex_id = ex.get('id')
+                        # possibly set speaker vars
+                        if i == 0 and speaker_1 is None and ex_id is not None:
+                            speaker_1 = ex_id
+                        elif i == 1 and speaker_2 is None and ex_id is not None:
+                            speaker_2 = ex_id
+
+                        # check if act is from speaker 1 or speaker 2
+                        context = False
+                        if (i % 2 == 0 and ex_id != speaker_1) or (
```
klshuster: is this always the case?

parlai/scripts/self_chat.py
```diff
@@ -30,92 +28,97 @@ def setup_args(parser=None):
...
+    parser.add_argument(
+        '--outfile', type=str, default=None, help='File to save self chat logs'
     )
-    parser.add_argument('--outfile', type=str, default='/tmp/selfchat.json')
     parser.add_argument(
-        '--format', type=str, default='json', choices={'parlai', 'json'}
+        '--save-format',
+        type=str,
+        default='conversations',
+        choices=['conversations', 'parlai', 'json'],
```
klshuster: this is actually a bug that i found in my experience with self-chat (it's in one of my other PRs) but i think you need to change `json` to `jsonl` here

jaseweston:
"Changed the logic around interactive worlds and self chat worlds so that they do not require adding a create_agents function to the associated task agents file that returned a [] (I always found this pretty confusing)" 
-- i think this also allowed to add flags.  we really need for it to also grab flags from the world, as well as from agents..

"Got rid of metrics logging in self chat worlds (all it printed was time, and was distracting)"
-- isnt it possible we might want metrics from self chats?

- does the conversations format only support 2 speakers?  it's better if it uses the ids to support > 2 (we're actually planning to do that scenario in light..)

parlai/tasks/self_chat/worlds.py
emilydinan: yeah. in self chat they are the exact same agent. so it shouldn't matter at all. the personas are chosen randomly.

emilydinan:
> "Changed the logic around interactive worlds and self chat worlds so that they do not require adding a create_agents function to the associated task agents file that returned a [] (I always found this pretty confusing)"
> -- i think this also allowed to add flags. we really need for it to also grab flags from the world, as well as from agents..
> 

I'm not sure how this allowed us to add flags -- all it did was return a `[]`right now. You can add flags to self chat worlds.


> "Got rid of metrics logging in self chat worlds (all it printed was time, and was distracting)"
> -- isnt it possible we might want metrics from self chats?

The only things that were logged was the time. So I'm not sure which metrics were logged. I would rather get rid of it until we have a clear use case.

> * does the conversations format only support 2 speakers?  it's better if it uses the ids to support > 2 (we're actually planning to do that scenario in light..)

Talked about this in ParlAI. Still no clear solution.

tests/test_conversations.py
```diff
@@ -0,0 +1,131 @@
...
+        act_list = [
+            [
+                [
+                    {
+                        'id': 'Emily',
+                        'text': 'Hello, do you like this test?',
+                    },
+                    {
+                        'id': 'Stephen',
+                        'text': 'Why yes! I love this test!',
```
stephenroller: I love all your tests

parlai/utils/conversations.py
```diff
@@ -64,6 +66,10 @@ def _get_path(datapath):
         fle, _ = os.path.splitext(datapath)
         return fle + '.metadata'
```
stephenroller: nit: i recently found a bug where some code that was using splitext (bytelevelbpe) wasn't working properly when I failed to give a file extension (which happened a lot there). It's just a potential red flag

PR #2407, opened 2020-02-18
Log loss scalar. Adjust when logs print/flush. Change token counting.

stephenroller:
**Patch description**
Small improvements to logging in train_model and torch agent:
- Only count label words as words-per-batch, following fairseq
- Add a tracker of the fp16 loss scalar
- Sync actual number of epochs across workers, rather than assuming it's linear
- Always log right before validation, and reset the log timer on validation so that we don't have partial metrics thanks to validation.

**Testing steps**
Several sweeps ran, CI.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1835,8 +1835,7 @@ class TorchAgent(ABC, Agent):
             # we divide by the binary is_primary_worker() so that the numerator is
             # num_tokens in all workers, and the denominator is 1.
             tbp = AverageMetric(
-                (batch.label_vec != self.NULL_IDX).sum().item()
-                + (batch.text_vec != self.NULL_IDX).sum().item(),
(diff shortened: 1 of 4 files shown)


parlai/scripts/train_model.py
```diff
@@ -678,17 +677,23 @@ def train(self):
                     >= self.val_every_n_epochs
                 ):
                     try:
+                        # log before we validate
+                        self.log()
```
emilydinan: nice

parlai/core/torch_agent.py
```diff
@@ -1942,6 +1941,10 @@ def update_params(self):
             self.global_metrics.add(
                 'clip', AverageMetric(float(grad_norm > self.opt['gradient_clip']))
             )
+        if self.fp16:
+            self.global_metrics.add(
+                'fp16_loss_scalar', AverageMetric(self.optimizer.loss_scale)
```
klshuster: this is causing unit test failures

PR #2412, opened 2020-02-18
Image features in polyencoder

EricMichaelSmith:
**Patch description**
Enable training the polyencoder on datasets that have image features (in `batch.image`). These image features, once encoded, can be combined with the context encoding by addition, prepending, or postpending. If image features are not included with any element of a batch, just use a tensor of zeros in its place.

Removing unused code:
- The `'none_with_pos_embs'` reduction type from the Transformer code
- The `--polyencoder-attention-keys` flag from the polyencoder. This means that we don't have to pass the positional embedding around the polyencoder anymore

Also, add two teachers used in the BlendedSkillTalk paper: one used to add WoW topics to ConvAI2 data and one used to add personas to WoW data. I use these when evaluating my polyencoder models trained on ImageChat. (I can move these into a separate PR if it's annoying to include them here :) )

**Testing steps**
Example of a command to train the polyencoder on ImageChat and BlendedSkillTalk:
```
python -u examples/train_model.py -veps 0.5 -vme 8000 --activation gelu --attention-dropout 0.3 --batchsize 16 --eval-batchsize 16 --candidates batch --data-parallel True --dict-endtoken __start__ --dict-file ./data/models/pretrained_transformers/model_bi.dict --dict-lower True --dict-tokenizer bpe --dropout 0.05 --embedding-size 768 --embeddings-scale False --ffn-size 3072 --fp16 True --history-size 20 --init-model /checkpoint/marywilliamson/all_in_one_dialogue/scripts/multitask_ft_aio_20191205/001/63a_jobid=15/model --label-truncate 72 --learn-embeddings True --learn-positional-embeddings True --log_every_n_secs 20 -lr 3e-06 --lr-scheduler reduceonplateau --lr-scheduler-decay 0.3 --lr-scheduler-patience 1 --max_train_time 200000 --model transformer/image_polyencoder --multitask-weights 1,0.33 --n-heads 12 --n-layers 12 --n-positions 1024 --n-segments 2 --num-epochs 100.0 --optimizer adamax --output-scaling 0.04 --poly-attention-type basic --poly-n-codes 300 --polyencoder-type n_first --reduction-type mean --relu-dropout 0.1 --save-after-valid True --share-encoders False --text-truncate 360 --validation-metric accuracy --validation-metric-mode max --variant xlm --warmup_updates 100 -t internal:blended_skill_talk,internal:comment_battle:imageDialog --image-mode-internal uru --swap-in-ccby-images True --image-encoder-num-layers 1 --image-features-dim 2048 --image-combination-mode add --model-file /checkpoint/ems/all_in_one_dialogue/scripts/s2020_02_12__imagechat_with_polyencoder/02_multitask/000/5e1cd_jobid=1/model
```
Example of a command to evaluate a polyencoder model on ImageChat and BlendedSkillTask:
```
python examples/eval_model.py --activation gelu --attention-dropout 0.1 --batchsize 16 --candidates inline --data-parallel True --dict-endtoken __start__ --dict-file ./data/models/pretrained_transformers/model_bi.dict --dict-lower True --dict-tokenizer bpe --dropout 0.15 --embedding-size 768 --embeddings-scale False --ffn-size 3072 --fp16 True --history-size 20 --init-model zoo:pretrained_transformers/poly_model_huge_reddit/model --label-truncate 72 --learn-embeddings True --learn-positional-embeddings True --log_every_n_secs 20 -lr 1e-05 --lr-scheduler reduceonplateau --lr-scheduler-decay 0.2 --lr-scheduler-patience 1 --model transformer/image_polyencoder --multitask-weights 1,2,1,3,100 --n-heads 12 --n-layers 12 --n-positions 1024 --n-segments 2 --optimizer adamax --output-scaling 0.045 --poly-attention-type basic --poly-n-codes 300 --polyencoder-type n_first --reduction-type max --relu-dropout 0.0 --share-encoders False --text-truncate 360 --variant xlm --warmup_updates 100 -t internal:blended_skill_talk,internal:comment_battle:imageDialog --image-mode-internal uru --swap-in-ccby-images True --image-encoder-num-layers 1 --image-features-dim 2048 --image-combination-mode prepend --model-file /checkpoint/ems/all_in_one_dialogue/scripts/s2020_02_12__imagechat_with_polyencoder/02_multitask/003/9df_jobid=22/model
```
Hits@1,100 results: BST: 76.20%, ImageChat: 46.75%


parlai/agents/transformer/polyencoder.py
```diff
@@ -368,19 +505,25 @@ def encode(self, ctxt_tokens, cand_tokens):
...
             # get context_representation. Now that depends on the cases.
-            ctxt_out, ctxt_mask, ctxt_pos = self.encoder_ctxt(ctxt_tokens)
-            att_keys = ctxt_out if self.attention_keys == 'context' else ctxt_pos
+            if self.use_image_features:
+                if not isinstance(ctxt_image, torch.Tensor) or ctxt_image.size() != (
+                    bsz,
+                    self.image_features_dim,
+                ):
+                    raise ValueError('Image feature tensor malformed!')
+                ctxt_out, ctxt_mask = self.encoder_ctxt(ctxt_tokens, ctxt_image)
```
stephenroller: This is starting to feel like we need `_model_input` or similar, like we have in TorchGeneratorAgent.

parlai/core/torch_agent.py
```diff
@@ -902,7 +902,12 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
             elif not optimstate_fp16 and self.fp16:
                 # old optimizer was fp32, but now we're doing fp16.
                 # this is a bit clunky, but alternatives are worse
-                self.optimizer.optimizer.load_state_dict(optim_states)
+                try:
+                    self.optimizer.optimizer.load_state_dict(optim_states)
+                except ValueError:
+                    print(
+                        'WARNING: not loading optim state since model params changed.'
```
stephenroller: warn_once instead maybe?

parlai/tasks/convai2/agents.py
```diff
@@ -105,3 +105,26 @@ def create_agents(opt):
...
         return []
+
+
+class PersonaTopicifierTeacher(SelfOriginalTeacher):
+    """
+    Adds WoW topics to ConvAI2 data.
+    """
+
+    def __init__(self, opt, shared=None):
+        from parlai_internal.projects.blended_skill_talk.add_personas_topics import (
```
jaseweston: you can't add internal stuff to public parlai!?

parlai/agents/transformer/polyencoder.py
```diff
@@ -41,6 +54,25 @@ def add_cmdline_args(cls, argparser):
             'simply take the first N vectors.',
             recommended='codes',
         )
+        agent.add_argument(
```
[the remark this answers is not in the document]
jaseweston: yea, can't we make a new class? ..idk..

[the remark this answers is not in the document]
klshuster: perhaps something like `image_polyencoder`

parlai/agents/transformer/polyencoder.py
```diff
@@ -137,6 +163,50 @@ def _set_text_vec(self, *args, **kwargs):
...
+            return features
+
+        if self.use_image_features:
+
+            # Checks/formatting of batch.image
+            bsz = batch.text_vec.size(0)
+            if batch.image is None or len(batch.image) == 0:
+                batch.image = [None] * bsz
+            else:
+                assert len(batch.image) == bsz
```
klshuster: hmm so this assumes that we only train with batches of _all_ images or _no_ images right? are we ruling out mixed batches?

parlai/agents/transformer/polyencoder.py
```diff
@@ -184,27 +256,35 @@ def score_candidates(self, batch, cand_vecs, cand_encs=None):
                 cand_rep = cand_encs.expand(bsz, cand_encs.size(1), -1)
         # bsz x num cands x seq len
         elif len(cand_vecs.shape) == 3:
-            _, _, _, cand_rep = self.model(cand_tokens=cand_vecs)
```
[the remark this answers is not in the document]
klshuster: i personally approve of this. i had changed the model to return the position embeddings as well at some point in the past, as I had been doing experiments with using the position embs as keys for the polyencoder attention; this never actually worked better than just using the context embs, so i don't think it's worth it to keep this around. I'm pretty sure the poly-encoder is the only model that uses this anyway

parlai/tasks/convai2/agents.py
klshuster: i'm not sure this should be in this PR either

parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -936,3 +936,26 @@ def create_agents(opt):
...
         return []
+
+
+class PersonaTopicifierTeacher(WizardDialogKnowledgeTeacher):
+    """
+    Adds personas to WoW data.
+    """
+
+    def __init__(self, opt, shared=None):
+        from parlai_internal.projects.blended_skill_talk.add_personas_topics import (
```
klshuster: ditto

parlai/tasks/convai2/agents.py
EricMichaelSmith: Ugh, good catch. Yes, agreed - removed

parlai/tasks/wizard_of_wikipedia/agents.py
EricMichaelSmith: Yes - removed

parlai/agents/image_seq2seq/modules.py
```diff
@@ -86,48 +88,51 @@ def __init__(
...
         embeddings_scale=False,
-        reduction_type='mean',
         n_positions=1024,
         activation='relu',
         variant='aiayn',
         n_segments=0,
         output_scaling=1.0,
         image_encoder_num_layers=1,
         image_features_dim=2048,
+        image_combination_mode='postpend',
```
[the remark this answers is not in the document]
EricMichaelSmith: Fair - changed

parlai/agents/image_seq2seq/modules.py
```diff
@@ -219,10 +231,36 @@ def forward(
...
+            full_mask = self.cat([context_mask, extra_masks])
+        elif self.image_combination_mode == 'prepend':
+            full_enc = self.cat([image_encoded, context_encoded])
+            full_mask = self.cat([extra_masks, context_mask])
+        else:
+            raise ValueError('Image combination mode not recognized!')
+
         return full_enc, full_mask
 
+    def add(self, tensors: List[torch.Tensor]) -> torch.Tensor:
```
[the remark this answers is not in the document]
EricMichaelSmith: Yes - changed

parlai/agents/transformer/modules.py
```diff
@@ -548,10 +548,7 @@ def forward(self, input, positions=None, segments=None):
             return output
         elif self.reduction_type is None or 'none' in self.reduction_type:
             output = tensor
-            ret = (output, mask)
-            if self.reduction_type == 'none_with_pos_embs':
-                ret = (output, mask, position_embs)
-            return ret
+            return output, mask
```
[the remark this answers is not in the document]
EricMichaelSmith: Yes - changed

parlai/agents/transformer/polyencoder.py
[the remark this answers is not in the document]
EricMichaelSmith: Yes, good point - I've split out new Agent and Model classes to prevent the original polyencoder from getting too heavy

parlai/agents/transformer/polyencoder.py
EricMichaelSmith: We can handle mixed batches: that's what the block with the `# Process all image feature vectors, or add in zero vectors if missing` is doing, filling in image features for specific examples that are missing them

parlai/agents/transformer/polyencoder.py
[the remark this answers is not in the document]
EricMichaelSmith: Yeah, given that this functionality isn't going to be used anymore, I'm in favor of ripping out dead code

parlai/agents/transformer/polyencoder.py
```diff
@@ -60,16 +92,6 @@ def add_cmdline_args(cls, argparser):
             'the key)',
             recommended='basic',
         )
-        agent.add_argument(
-            '--polyencoder-attention-keys',
```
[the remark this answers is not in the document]
EricMichaelSmith: Yes - added

parlai/core/torch_agent.py
EricMichaelSmith: Yes - changed

parlai/agents/transformer/polyencoder.py
EricMichaelSmith: Yes - I've re-written a few functions to reflect the fact that sometimes we'll be passing in ctxt image features and sometimes we won't

parlai/agents/transformer/polyencoder.py
```diff
@@ -118,15 +101,34 @@ def add_cmdline_args(cls, argparser):
         )
         return agent
 
+    @classmethod
+    def upgrade_opt(cls, opt_from_disk: Opt):
```
klshuster: great, thanks for this

parlai/agents/transformer/polyencoder.py
```diff
@@ -593,17 +496,180 @@ def forward(
...
+        """
+        return ImagePolyencoderModule(self.opt, self.dict, self.NULL_IDX)
+
+    def batchify(self, obs_batch: List[Message], sort: bool = False) -> Batch:
+        """
+        Override to handle image features.
+        """
+        batch = super().batchify(obs_batch, sort)
+
+        def _process_features(features: torch.Tensor) -> torch.Tensor:
```
klshuster: perhaps `_process_image_features` to make it more clear?

parlai/agents/transformer/polyencoder.py
```diff
@@ -593,17 +496,180 @@ def forward(
...
+
+        return batch
+
+    def _model_context_input(self, batch) -> Dict[str, Any]:
+        """Override PolyencoderAgent's context inputs into the model."""
+        return {'ctxt_tokens': batch.text_vec, 'ctxt_image': batch.image}
+
+    def load_state_dict(self, state_dict):
+        """
+        Override to account for codes and possibly weights used for image features.
```
klshuster: nit: this func only overridden for image weights

parlai/agents/transformer/polyencoder.py
EricMichaelSmith: Yes - changed

parlai/agents/transformer/polyencoder.py
EricMichaelSmith: Ugh, yes - didn't update this when I subclassed the polyencoder. Changed

parlai/agents/image_seq2seq/modules.py
```diff
@@ -137,8 +137,8 @@ def _build_image_encoder(self):
         for _ in range(self.n_img_layers - 1):
             image_layers += [
                 nn.ReLU(),
-                nn.Dropout(p=self.opt['dropout']),
-                nn.Linear(self.img_dim, self.embedding_size),
+                nn.Dropout(p=self.dropout_frac),
+                nn.Linear(self.embedding_size, self.embedding_size),
```
klshuster: thanks for these changes

parlai/agents/transformer/image_polyencoder.py
```diff
@@ -0,0 +1,211 @@
...
+        return ImagePolyencoderModule(self.opt, self.dict, self.NULL_IDX)
+
+    def batchify(self, obs_batch: List[Message], sort: bool = False) -> Batch:
+        """
+        Override to handle image features.
+        """
+        batch = super().batchify(obs_batch, sort)
+
+        def _process_image_features(features: torch.Tensor) -> torch.Tensor:
+            assert features.size() == (self.image_features_dim,)
```
klshuster: have you tried this with other image features? e.g. `resnet152` or `resnext101_32x48d_wsl`? these return 4d tensors with size `[1, IMAGE_FEATURES_DIM, 1, 1]`, so you'll need to extract the tensor appropriately.

parlai/agents/transformer/image_polyencoder.py
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env python3
```
stephenroller: I feel like this model deserves to go in image_seq2seq or another folder. It's deserves more than "minor transformer variant mixed in with polyencoder"

parlai/agents/transformer/image_polyencoder.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        Override PolyencoderAgent's context inputs into the model.
+        """
+        return {'ctxt_tokens': batch.text_vec, 'ctxt_image': batch.image}
+
+    def load_state_dict(self, state_dict):
+        """
+        Override to account for weights used for image features.
+        """
+        for tensor in ['dummy_image_enc', 'ones_mask']:
```
stephenroller: is this specifically for init model? I wonder if we need an argument for load_state_dict.

parlai/agents/transformer/polyencoder.py
```diff
@@ -96,6 +95,29 @@ def add_cmdline_args(cls, argparser):
...
+            # 2020-02-19 We are deprecating this flag because it was used for a one-time
+            # set of experiments and won't be used again. This flag was defaulted to
+            # 'context', so throw an exception otherwise.
+            if polyencoder_attention_keys_value == 'context':
+                warn_once(
+                    '--polyencoder-attention-keys has been deprecated and will be ignored.'
+                )
+            else:
+                raise NotImplementedError(
+                    'This --polyencoder-attention-keys mode is no longer supported!'
```
stephenroller: Suggest a commit where the user can find this option.

parlai/agents/transformer/polyencoder.py
```diff
@@ -96,6 +95,29 @@ def add_cmdline_args(cls, argparser):
...
+    @classmethod
+    def upgrade_opt(cls, opt_from_disk: Opt):
+        # call the parent upgrades
+        opt_from_disk = super(PolyencoderAgent, cls).upgrade_opt(opt_from_disk)
+
+        polyencoder_attention_keys_value = opt_from_disk.get(
+            'polyencoder_attention_keys'
+        )
+        if polyencoder_attention_keys_value is not None:
+            # 2020-02-19 We are deprecating this flag because it was used for a one-time
```
stephenroller: good

parlai/agents/transformer/polyencoder.py
```diff
@@ -96,6 +95,29 @@ def add_cmdline_args(cls, argparser):
...
+            # set of experiments and won't be used again. This flag was defaulted to
+            # 'context', so throw an exception otherwise.
+            if polyencoder_attention_keys_value == 'context':
+                warn_once(
```
stephenroller: This is the "I'm loading from an old model and I didn't specify this behavior so I used the default" case. The user doesn't need to be warned that nothing changed here. Just silently delete the key from the opt and go on.

parlai/agents/transformer/polyencoder.py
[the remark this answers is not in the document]
stephenroller: k, thanks y'all

parlai/agents/transformer/polyencoder.py
```diff
@@ -215,16 +247,27 @@ class PolyEncoderModule(torch.nn.Module):
     See https://arxiv.org/abs/1905.01969 for more details
     """
 
-    def __init__(self, opt, dict, null_idx):
+    def __init__(self, opt, dict_, null_idx):
```
stephenroller: nice

parlai/agents/transformer/polyencoder.py
```diff
@@ -265,33 +308,34 @@ def __init__(self, opt, dict, null_idx):
...
             )
 
-    def get_encoder(self, opt, dict, null_idx, reduction_type):
+    def get_encoder(self, opt, dict_, null_idx, reduction_type, for_context: bool):
         """
         Return encoder, given options.
 
         :param opt:
-            opt dict
+            option dictionary
```
stephenroller: nah, call it an Opt

parlai/agents/transformer/polyencoder.py
```diff
@@ -335,52 +386,59 @@ def attend(self, attention_layer, queries, keys, values, mask):
         else:
             raise Exception('Unrecognized type of attention')
 
-    def encode(self, ctxt_tokens, cand_tokens):
+    def encode(
+        self,
+        cand_tokens: Optional[torch.Tensor],
+        **ctxt_inputs: Dict[str, torch.Tensor],
+    ) -> Tuple[Optional[torch.tensor], Optional[torch.tensor], Optional[torch.tensor]]:
```
stephenroller: mypy found a bug

parlai/core/torch_image_agent.py
```diff
@@ -0,0 +1,88 @@
...
+from abc import abstractmethod
+from typing import List
+
+import torch
+
+from parlai.core.message import Message
+from parlai.core.torch_agent import Batch, TorchAgent
+
+
+class TorchImageAgent(TorchAgent):
```
stephenroller: :-O

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -6,23 +6,23 @@
...
+import torch
+
 from .modules import ImageSeq2seqModel
 from parlai.agents.transformer.transformer import TransformerGeneratorAgent
 from parlai.core.dict import DictionaryAgent
-from parlai.core.message import Message
 from parlai.core.torch_agent import Batch
-
-# from parlai.utils.typing import Dict, List
+from parlai.core.torch_image_agent import TorchImageAgent
```
stephenroller: love it

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -45,18 +45,9 @@ def add_cmdline_args(cls, argparser):
         """
         Override to add one arg.
         """
-        super(ImageSeq2seqAgent, cls).add_cmdline_args(argparser)
+        TransformerGeneratorAgent.add_cmdline_args(argparser)
+        TorchImageAgent.add_cmdline_args(argparser)
```
stephenroller: Can you try just super().add_cmdline_args(argparser)? You might have to do it in both places.

This is an experiment.

stephenroller:
(Also I wouldn't mind seeing this thing have some tests!)

parlai/agents/image_seq2seq/modules.py
```diff
@@ -96,6 +96,7 @@ def __init__(
         image_encoder_num_layers=1,
         image_features_dim=2048,
         image_combination_mode='append',
+        n_image_tokens=1,
```
stephenroller: What do you use in practice? This looks like it could cause that multiple-of-8 issue I was mentioning

EricMichaelSmith: So, my time steps definitely aren't multiples of 8. Given what you said, I'm sure that's causing a major slowdown - I first have to address some issues with the training *itself* not going as expected, but then I'll take a look at this :)

klshuster: if we don't end up seeing any gain from varying this I would say we can just keep it at 1 and reduce some unnecessary complexity later

parlai/agents/image_seq2seq/modules.py
```diff
@@ -219,11 +246,65 @@ def forward(
...
+            full_mask = self._cat([context_mask, extra_masks])
+        elif self.image_combination_mode == 'prepend':
+            full_enc = self._cat([image_encoded, context_encoded])
+            full_mask = self._cat([extra_masks, context_mask])
+        else:
+            raise ValueError('Image combination mode not recognized!')
+
+        # In fp16 mode, either remove extra tokens or add new ones on to get to a
+        # multiple of 8
+        if full_enc.dtype == torch.half and full_mask is not None:
```
klshuster: could we move this logic out of `forward` into a separate function?

parlai/agents/transformer/image_polyencoder.py
klshuster: hmm i think you're onto something, especially since they both use the `ContextWithImageEncoder`.

Not sure what the right move is here - perhaps combine in one folder `multimodal/image_seq2seq` and `multimodal/image_polyencoder`, though this is a bit opaque. 

I would be ok with leaving as is unless @stephenroller feels strongly

parlai/agents/transformer/image_polyencoder.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+
+    @classmethod
+    def add_cmdline_args(cls, argparser):
+        """
+        Add command-line arguments specifically for this agent.
+        """
+        PolyencoderAgent.add_cmdline_args(argparser)
+        TorchImageAgent.add_cmdline_args(argparser)
+        agent = argparser.add_argument_group('Image Encoder Args')
```
klshuster: nit: `Image PolyEncoder args`

parlai/agents/transformer/image_polyencoder.py
```diff
@@ -0,0 +1,215 @@
...
+        agent.add_argument(
+            '--image-combination-mode',
+            type=str,
+            default='prepend',
+            choices=['add', 'append', 'prepend'],
```
klshuster: once we have a solid grasp of which method is best, let's add a `recommended='<method>'` flag here (could just mark as `TODO` for now)

parlai/agents/transformer/image_polyencoder.py
```diff
@@ -0,0 +1,215 @@
...
+        # This agent doesn't support any encoder output reductions
+        return agent
+
+    def build_model(self, states=None):
+        """
+        Return built model.
+        """
+        return ImagePolyencoderModule(self.opt, self.dict, self.NULL_IDX)
+
+    def batchify_image_features(self, batch: Batch) -> Batch:
```
klshuster: could you maybe elaborate a bit about how this differs from the `batchify_image_features` in `ImageSeq2seq`? if it's not _too_ much different, it might be good to include in the `TorchImageAgent`

parlai/agents/transformer/image_polyencoder.py
```diff
@@ -0,0 +1,215 @@
...
+                )
+            else:
+                raise ValueError('Unsupported image feature format!')
+
+        # Turn into batchsize x image_features_dim for DataParallel
+        batch.image = torch.stack(processed_features_list)
+
+        return batch
+
+    def _get_batch_size(self, batch) -> int:
```
klshuster: yeah i have a fix for this in #2371 but fine for now

parlai/agents/transformer/image_polyencoder.py
klshuster: i believe it is

parlai/core/torch_image_agent.py
```diff
@@ -0,0 +1,88 @@
...
+
+    @abstractmethod
+    def batchify_image_features(self, batch: Batch) -> Batch:
+        """
+        Put this batch of images into the correct format for this agent.
+
+        self._process_image_features() will likely be useful for this.
+        """
+        raise NotImplementedError(
+            'Subclasses must implement method for batching images!'
```
klshuster: is there not any shared functionality that could go here?

tests/test_transformers.py
```diff
@@ -736,5 +734,59 @@ def test_invsqrt_learning_rate(self):
...
+    """
+    Unit tests for the ImagePolyencoderAgent.
+
+    Test that the model is able to handle simple train tasks.
+    """
+
+    @testing_utils.retry(ntries=3)
+    @testing_utils.skipUnlessTorch
+    @testing_utils.skipUnlessGPU
+    def test_image_task(self):
```
klshuster: ❤️

parlai/agents/image_seq2seq/modules.py
EricMichaelSmith: Hmm - so, preliminary sweeps show that it might be giving us a few-point boost, presumably due to the image features just taking up more space in the final encoding. I'd prefer to keep it in unless you all really hate it :)

parlai/agents/transformer/image_polyencoder.py
EricMichaelSmith: I've fixed this in torch_image_agent.py

parlai/agents/transformer/image_polyencoder.py
EricMichaelSmith: Yeah, this is for `TorchRankerAgent.load(init_model)`. @stephenroller  - what argument (other than `stated_dict`) did you have in mind for this?

parlai/agents/transformer/polyencoder.py
EricMichaelSmith: Yes - added

parlai/agents/transformer/polyencoder.py
EricMichaelSmith: Yes - changed

parlai/agents/transformer/polyencoder.py
EricMichaelSmith: Reverted

parlai/agents/transformer/polyencoder.py
EricMichaelSmith: subtle - fixed

stephenroller:
There's a large number of surprising mypy errors. If you really think they are noise, then GA, but some of them look legit (e.g. one type is `tensor` instead of `Tensor`). They are enough that they could easily be catching a bug.

parlai/agents/image_seq2seq/image_seq2seq.py
EricMichaelSmith: Hmm when I do this I get `Parse Error: unrecognized arguments: --image-encoder-num-layers 1 --image-features-dim 2048`, indicating that it couldn't find the `TorchImageAgent` args

EricMichaelSmith:
> (Also I wouldn't mind seeing this thing have some tests!)

Added 3 tests in `test_transformers.py` analogous to those in `test_image_seq2seq.py`

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -6,23 +6,23 @@
...
-# from parlai.utils.typing import Dict, List
+from parlai.core.torch_image_agent import TorchImageAgent
 
 
 TOKEN_IMAGE = '__image__'
 TOKEN_NO_IMAGE = '__no_image__'
 
 
-class ImageSeq2seqAgent(TransformerGeneratorAgent):
+class ImageSeq2seqAgent(TransformerGeneratorAgent, TorchImageAgent):
```
stephenroller: fyi, the ordering of the mixin determines the order of supers().

tests/test_transformers.py
```diff
@@ -793,5 +793,110 @@ def test_invsqrt_learning_rate(self):
...
+
+        Random chance is 10%, so this should be able to get much better than that very
+        quickly.
+        """
+        args = Opt({**self.base_args, **self.image_args})
+        valid, test = testing_utils.train_model(args)
+        self.assertGreater(
+            valid['accuracy'],
+            0.5,
+            f'ImagePolyencoderAgent val-set accuracy on a trivally simple task was {valid["accuracy"].value():0.2f}.',
```
stephenroller: You can do `assert valid['accuracy'] > 0.5` now, and it will print very cleanly with like, local variables and such.

stephenroller: (this is new since we switched to pytest)

parlai/agents/image_seq2seq/modules.py
EricMichaelSmith: Yes - moved

parlai/agents/transformer/image_polyencoder.py
EricMichaelSmith: Fix, fixed

parlai/agents/transformer/image_polyencoder.py
EricMichaelSmith: Added TODO for this

parlai/agents/transformer/image_polyencoder.py
EricMichaelSmith: Okay, I've added a note about this. Yes, I agree that it'd be good to test out whether `ImageSeq2seq` really requires a separate batchify method now. I've created an issue for this: https://github.com/facebookresearch/ParlAI/issues/2461

parlai/core/torch_image_agent.py
EricMichaelSmith: Perhaps - that's what I've created https://github.com/facebookresearch/ParlAI/issues/2461 to solve. I can try to look into unifying the `batchify_image_features()` methods now if it's really important, but given that there may be image_seq2seq edge cases that we have to account for, my preference would be to make this a separate PR

tests/test_transformers.py
EricMichaelSmith: Yes - changed

EricMichaelSmith:
> There's a large number of surprising mypy errors. If you really think they are noise, then GA, but some of them look legit (e.g. one type is `tensor` instead of `Tensor`). They are enough that they could easily be catching a bug.

Yes - went through these one by one and fixed all of the ones that I don't think are noise

PR #2413, opened 2020-02-18
[Self chat] Return logs

emilydinan:
**Patch description**
Edit the self chat function to return the logs.

### DIFF
--- a/parlai/scripts/self_chat.py
+++ b/parlai/scripts/self_chat.py
@@ -75,8 +75,8 @@ def self_chat(opt, print_parser=None):
     # Check for `selfchat` in the task name
     if 'selfchat' not in opt['task']:
         warn_once(
-            'You are using self chat with task {}.'.format(opt['task']) +
-            'If your task has an existing self chat world, then run with '
+            'You are using self chat with task {}. '.format(opt['task'])
+            + 'If your task has an existing self chat world, then run with '
             '-t {}:selfchat'.format(opt['task'])
         )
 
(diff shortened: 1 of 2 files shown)


parlai/scripts/self_chat.py
```diff
@@ -72,6 +72,14 @@ def self_chat(opt, print_parser=None):
     if hasattr(agent2, 'id'):
         agent2.id = agent2.id + "2"
 
+    # Check for `selfchat` in the task name
+    if 'selfchat' not in opt['task']:
+        warn_once(
+            'You are using self chat with task {}.'.format(opt['task']) +
+            'If your task has an existing self chat world, then run with '
```
stephenroller: You missed a space. Also you can use `f'strings'` if you want!

parlai/scripts/self_chat.py
```diff
@@ -103,7 +111,9 @@ def self_chat(opt, print_parser=None):
     if opt.get('display_examples'):
         print('-- end of episode --')
 
+    logger.reset_world()  # add to logs
```
stephenroller: this is so confusing, how does this add to logs?!

parlai/scripts/self_chat.py
stephenroller: I don't suppose we could check if such a self chat world exists but such is.

parlai/scripts/self_chat.py
jaseweston: hmm yea not sure what this is.. was it not working before?.. oh like the last episode maybe wasn't flushed, or..?

emilydinan: yeah it only flushes when the episode is done

PR #2414, opened 2020-02-18
[ACUTE] ACUTE Eval Cleanup

klshuster:
**Patch description**
I've re-written ACUTE-Eval to be more modular and easier to read. It now fits better with the FAST-Acute script I have in a separate internal PR.

The bulk of the re-write involves moving the functionality into an `AcuteEvaluator` class.

This also involves minor changes to the `self-chat.py` script.

**Testing steps**
Ran the example script, verified it works.


parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -3,28 +3,24 @@
...
 """
 Example script for running ACUTE-EVAL.
-The only arguments that *must* be modified for this to be run are:
-'dialogs_path':  Path to folder containing logs of all model conversations.
-    Each model should have one file named '<modelname>.jsonl'
-'model_comparisons': List of tuples indicating pairs of models to be compared.
-    The model names here should match the names of files in the folder
-'onboarding_tasks': List of tuples in format (id1, id2, name of matchup) where
-    id1, id2 are conversation ids where id2 is the correct conversation's id
+The only argument that *must* be modified for this to be run is:
```
stephenroller: I wonder if it's worth mentioning args that we MIGHT want to run.

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
-            opt['block_qualification'] = task_group_id
+    def _supplement_opt(self):
+        """
+        Add additional args to opt.
+
+        Useful to add relevant options after args are parsed.
+        """
+        self.opt.update(
+            {
+                'task': os.path.basename(os.path.dirname(os.path.abspath(__file__))),
```
stephenroller: is this right now? Hmm..

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+        pairs_path = self.opt.get('pairings_filepath')
+        if not os.path.exists(pairs_path):
+            raise RuntimeError('You MUST specify a valid pairings filepath')
+
+        with open(pairs_path) as pf:
+            for i, l in enumerate(pf.readlines()):
+                convo_pair = json.loads(l.strip())
+                task = {
+                    'task_specs': {
+                        'conversation_order': random.choice([[0, 1], [1, 0]]),
```
stephenroller: we can improve this. A better option is to have a key for model_left and a key for model_right, and then we should dump to disk the name of the winner, i.e. the value inside `model_left` or the value for `model_b`. This would greatly simplify the analysis logic.

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+                # count correct answers
+                num_correct += 1
+        if num_onboarding_tasks == 0:
+            # no onboarding tasks found
+            if worker_id in self.onboarding_failed_workers:
+                # worker already failed onboarding, add pairings back to queue
+                self.requeue_task_data(worker_id, all_task_data)
+            return
+        if (num_correct / num_onboarding_tasks) >= self.opt['onboarding_threshold']:
+            # worker did not fail onboarding
```
stephenroller: `# worker passed onboarding`

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
-
-        # Begin the task, allowing mturk_manager to start running the task
-        # world on any workers who connect
-        mturk_manager.start_task(
-            eligibility_function=check_worker_eligibility,
-            assign_role_function=assign_worker_roles,
-            task_function=run_conversation,
+    def run(self):
+        self.manager.setup_server(
+            task_directory_path=os.path.dirname(os.path.abspath(__file__))
```
stephenroller: ditto here probably

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+            print("This run id: {}".format(task_group_id))
+
+            # Begin the task, allowing mturk_manager to start running the task
+            # world on any workers who connect
+            self.manager.start_task(
+                eligibility_function=check_worker_eligibility,
+                assign_role_function=assign_worker_roles,
+                task_function=run_conversation,
+            )
+        except Exception as e:
```
stephenroller: useless catch/raise. Just drop these two lines.

parlai/mturk/tasks/acute_eval/README.md
```diff
@@ -38,10 +38,10 @@ This is a template of the expected format with the minimal expected fields:
     {
       "is_onboarding": false,
       "speakers_to_eval": ["first_modelname", "second_modelname"],
+      "dialogue_ids": [dialogue_1_id, dialogue_2_id],
```
EricMichaelSmith: Hmm - why the change? Now the dialogue IDs are a bit disassociated in placement from the dialogues themselves

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -3,28 +3,24 @@
...
+'pairings_filepath':  Path to pairings file in the format specified in the README.md
```
EricMichaelSmith: Great! Much cleaner to only have 1 required arg

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+        self.onboarding_failed_workers.add(worker_id)
+
+    def softblock_workers(self):
+        """
+        Softblock workers if necessary.
+        """
+        if not self.opt['is_sandbox'] and self.opt['softblock_list_path'] is not None:
             softblock_list = set()
-            with open(opt['softblock_list_path'], 'r') as f:
+            with open(self.opt['softblock_list_path']) as f:
```
EricMichaelSmith: This isn't still a read-only file?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
-                    save_data,
-                )
-            return save_data
+        ]
+        if tasks_still_needed < len(tasks_chosen):
+            tasks_chosen = random.sample(tasks_chosen, tasks_still_needed)
+        worker_data['tasks_completed'].extend(tasks_chosen)
+        for t in tasks_chosen:
+            worker_data['conversations_seen'].extend(
+                self._get_dialogue_ids(self.desired_tasks[t])
```
emilydinan: do we need to add teh `pair_id` to `tasks_completed` here??

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+        for subtask_data in task_data:
+            if subtask_data['task_specs'].get('is_onboarding', False):
+                worker_data['onboarding_todo'].append(subtask_data['pair_id'])
+            else:
+                self.task_queue.put(subtask_data)
+                try:
+                    worker_data['tasks_completed'].remove(subtask_data['pair_id'])
+                    for d_id in self._get_dialogue_ids(subtask_data):
+                        worker_data['conversations_seen'].remove(d_id)
+                except ValueError:
```
emilydinan: why does this happen?

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+        Soft block workers who fail onboarding tasks, keep track of their status.
+
+        :param worker_id:
+            worker id
+
+        :param save_data:
+            data from the worker's completed tasks
+        """
+        all_task_data = save_data['worker_data'][worker_id]['task_data']
+        response_data = save_data['worker_data'][worker_id]['response']['task_data']
```
emilydinan: what is `response_data` vs. `task_data`?

stephenroller: task data is what was served to the client, response data is what was sent back

parlai/mturk/tasks/acute_eval/README.md
klshuster: i mostly made this change to help with readability later on - yes, it does disassociate the `id`s from the dialogues themselves, but I figured the tradeoff was better later on. however, happy to change back if you feel strongly

parlai/mturk/tasks/acute_eval/example_script.py
klshuster: i'll add a comment regarding which other args might need tweaking

parlai/mturk/tasks/acute_eval/run.py
klshuster: alas, this is an old turk hack that we've been using forever

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+        Set block qualification if necessary.
+
+        :param task_group_id:
+            task id used to set block qualification, if necessary.
+        """
+        if (
+            self.opt['block_on_onboarding_fail']
+            and self.opt['block_qualification'] is None
+        ):
+            self.opt['block_qualification'] = task_group_id
```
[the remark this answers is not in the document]
klshuster: yes, we call this `set_block_qual` prior to creating HITs, allowing the manager to create the qual

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+                        's1_choice': self.opt['s1_choice'],
+                        's2_choice': self.opt['s2_choice'],
+                        'question': self.opt['question'],
+                        # TODO: make sure this doesn't break javascript
+                        'is_onboarding': convo_pair['is_onboarding'],
+                    },
+                    'pairing_dict': convo_pair,
+                    'pair_id': i,
+                }
+                if convo_pair.get('is_onboarding'):
```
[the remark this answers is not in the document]
klshuster: yes agreed, in my `fast_eval` script I actually am reading the onboarding tasks from file before constructing the dialogue-pairs file. I personally think this key in the json is OK, as you can imagine that each ACUTE run is associated with a _single_ conversations file

parlai/mturk/tasks/acute_eval/run.py
klshuster: agreed, i'll look into a better way of handling this

klshuster: this will affect the REACT component as well just FYI

klshuster: i'm just documenting my thought process at this point but this `conversation_order` is not actually used in analysis and is purely for the javascript rendering, so i'm going to leave as is

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,379 @@ def add_args(from_argv=False):
...
+                    return task_data
+            else:
+                self.task_queue.put(next_task)
+        # task queue containing num annotations requested of each pair is exhausted
+        # b/c we released enough hits to guarantee reaching the requested num on average
+        tasks_still_needed = tasks_per_hit - len(task_data)
+        tasks_remaining = [
+            t_id
+            for t_id in range(len(self.desired_tasks))
+            if t_id not in worker_data['tasks_completed']
```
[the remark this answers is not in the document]
klshuster: yeah this logic is a little tricky, i'm going to clean up this function. 

A "task" here is a pair of "dialogues", each with unique IDs. Essentially, we first iterate through the `task_queue` to see if the worker has some tasks they have not completed (checking `pair_id`); if we exhaust the `task_queue` before we get enough tasks for the worker to do, we go through all the tasks and find tasks that fit the following criteria:

1. The worker has not completed that task before; and 
2. The worker has not seen any dialogue in that task before

I believe this is to ensure that a worker does not have the same _dialogue_ show up twice in their task queue, even if it's being compared to a separate dialogue that the worker has not seen. This logic is from the old ACUTE stuff but that's how I understood it

parlai/mturk/tasks/acute_eval/run.py
klshuster: yes, as it is possible that the worker hasn't actually seen this task (given that the `task_queue` is shared amongst all workers)

parlai/mturk/tasks/acute_eval/run.py
klshuster: I am actually not sure why this would happen. presumably there is an issue where the worker's `tasks_completed` and `conversations_seen` are not in sync with the actual task data given to the worker. perhaps @hadasah might have more insight into why this error happened in the past?

parlai/mturk/tasks/acute_eval/run.py
klshuster: default option for `open` is `'r'`

klshuster:
I've pushed a change to reflect review comments. Specifically:

- Update `example_script` to include a comment regarding other arguments one may want to change depending on their needs.
- Add docstring to `AcuteEvaluator.__init__` indicating various object attribute details
- Clean up the `get_new_task_data` to simplify logic and highlight the steps in building the task data
- Fixed issue with worker data not being saved across HITs (something I caught)

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,459 @@ def add_args(from_argv=False):
...
+            {
+                'tasks_completed': [],
+                'conversations_seen': [],
+                'onboarding_todo': onboarding_todo,
+                'hits_completed': 0,
+            },
+        )
+        return self.worker_data[worker_id]
+
+    def _supplement_opt(self):
```
hadasah: 👍

parlai/mturk/tasks/acute_eval/run.py
hadasah: I'd already made the changes to analysis logic prior to this - I had it this way here to avoid dealing with names in this script, and I would rather have this than have rematch strings for display.

parlai/mturk/tasks/acute_eval/run.py
hadasah: This entire file is technically not thread-safe. It's theoretically possible (although incredibly unlikely) that you could have a race condition which gave the same worker the same conversation twice and then have this situation. I actually think we should get rid of this though, given that it's not a product. If that were to actually happen and a turker complains, we'll just pay them.

parlai/mturk/tasks/acute_eval/README.md
hadasah: I kind of prefer having the ids associated with the conversations, at the cost of splitting up a line of code over several visual lines later - but not a big deal

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -3,39 +3,44 @@
...
+Help strings for the other arguments can be found in run.py.
 """
 
 
 def set_args():
     args = add_args()
-    # task folder containing pairings file
-    args['task_folder'] = 'parlai/mturk/tasks/acute_eval/example/'
+    # pairings file
+    # args['pairings_filepath'] = 'parlai/mturk/tasks/acute_eval/example/pairings.jsonl'
```
hadasah: we can just delete this? If we intend to have the next line as the default example

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,459 @@ def add_args(from_argv=False):
...
+                ]
+                # make sure order is preserved
+                assert eval_speakers == convo_pair['speakers_to_eval']
+                task = {
+                    'task_specs': {
+                        'conversation_order': random.choice([[0, 1], [1, 0]]),
+                        's1_choice': self.opt['s1_choice'],
+                        's2_choice': self.opt['s2_choice'],
+                        'question': self.opt['question'],
+                        # TODO: make sure this doesn't break javascript
```
hadasah: @klshuster was this my comment or yours?

parlai/mturk/tasks/acute_eval/run.py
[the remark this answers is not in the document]
hadasah: @klshuster is right
The ideal case is that the worker only does comparisons that they've never seen which still haven't been completed as many times as we want. If we run out, we fill the rest of their task with comparisons consisting only of dialogues that they've never seen. We don't want the workers to see any conversation more than once, to minimize the noise of personal preference contributed by a single worker.

parlai/scripts/self_chat.py
```diff
@@ -48,7 +48,10 @@ def setup_args(parser=None):
     )
     parser.add_argument('--outfile', type=str, default='/tmp/selfchat.json')
     parser.add_argument(
-        '--format', type=str, default='json', choices={'parlai', 'json'}
+        '--format', type=str, default='jsonl', choices={'parlai', 'jsonl'}
+    )
+    parser.add_argument(
```
hadasah: Is there particularly a reason to need to vary the indent?

parlai/mturk/tasks/acute_eval/run.py
klshuster: Thanks for the insight @hadasah, for now I'll just add a comment indicating why this might cause an exception

parlai/mturk/tasks/acute_eval/example_script.py
klshuster: ahh oops i did not mean to push that change, thanks for catching

parlai/mturk/tasks/acute_eval/run.py
klshuster: mine, and it does not break javascript so we're good 😄

parlai/scripts/self_chat.py
klshuster: yes, the indent makes the file look pretty but absolutely kills processing via `json.loads`

parlai/mturk/tasks/acute_eval/example_script.py
```diff
@@ -44,9 +46,7 @@ def set_args():
     args['question'] = 'Who would you prefer to talk to for a long conversation?'
 
     args['num_conversations'] = int(
-        len(args['model_comparisons'])
-        * args['pairs_per_matchup']
-        / (args['task_description']['num_subtasks'] - 1)
+        args['pairs_per_matchup'] / max((args['subtasks_per_hit'] - 1), 1)
```
stephenroller: much better

parlai/mturk/tasks/acute_eval/run.py
stephenroller: Why can't we use model_left and model_right, and update the react component?

parlai/mturk/tasks/acute_eval/run.py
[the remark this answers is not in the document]
stephenroller: Thanks Kurt, that sounds much better

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,455 @@ def add_args(from_argv=False):
...
+            try:
+                next_task = self.task_queue.get()
+            except queue.Empty:
+                break
+            num_attempts += 1
+
+            pair_id = next_task['pair_id']
+            dialogue_ids = self._get_dialogue_ids(next_task)
+
+            if pair_id not in worker_data[  # make sure worker has not seen these conversations before
```
stephenroller: lol nit, that comment's spot is so awkward

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,455 @@ def add_args(from_argv=False):
...
+            else:
+                self.task_queue.put(next_task)
 
-        # Create the hits as specified by command line arguments
-        mturk_manager.create_hits()
+        return task_data
 
-        def check_worker_eligibility(worker):
-            return True
+    def _top_up_task_data(
```
stephenroller: Nice

parlai/mturk/tasks/acute_eval/run.py
```diff
@@ -112,341 +102,455 @@ def add_args(from_argv=False):
...
+            else:
+                self.task_queue.put(subtask_data)
+                try:
+                    worker_data['tasks_completed'].remove(subtask_data['pair_id'])
+                    for d_id in self._get_dialogue_ids(subtask_data):
+                        worker_data['conversations_seen'].remove(d_id)
+                except ValueError:
+                    # Task may have shown up in worker's task queue twice
+                    # due to some unfortunate race condition
+                    print("WARNING: couldn't remove task from worker's history")
```
stephenroller: Maybe `warning.warn`? Meh.

parlai/mturk/tasks/acute_eval/example_script.py
klshuster: i'm realizing now that this parameter, `pairs_per_matchup`, is a little ambiguous. Before, it indicated how many pairs _per model-matchup_ there were - that is, if there were 3 models (A,B,C) , we could have 100 pairs of comparisons for each combination (A,B), (B,C), (A, C), resulting in 300 pairs. 

Perhaps this argument should just be `num_matchup_pairs`.

klshuster:
Per your request @stephenroller, i've refactored `conversation_order`. Would appreciate your thoughts

klshuster:
I've confirmed via my fast eval script that this indeed works, will share results of that in the relevant PR. Will wait for tests to pass then merging

PR #2418, opened 2020-02-19
[Chat Services] Small Chat Service Updates

klshuster:
**Patch description**
Needed to fix a few things with the chat services implementations to get things working with messenger image uploads and loading models

1. the chat config now can specify that the messenger manager **not** load a model
2. More cases for image uploading are handled

**Testing steps**
Tested with my own demo tests

### DIFF
--- a/parlai/chat_service/services/messenger/agents.py
+++ b/parlai/chat_service/services/messenger/agents.py
@@ -56,23 +56,21 @@ class MessengerAgent(ChatServiceAgent):
         """
         self.manager.message_sender.typing_on(self.id, persona_id=persona_id)
 
-    def put_data(self, message):
+    def put_data(self, messenger_message):
         """
         Put data into the message queue if it hasn't already been seen.
         """
-        mid = message['message']['mid']
-        seq = message['message'].get('seq', None)
-        if 'text' not in message['message']:
+        mid = messenger_message['message']['mid']
+        seq = messenger_message['message'].get('seq', None)
+        message = messenger_message['message']
+        if 'text' not in message:
             print(
                 'Msg: {} could not be extracted to text format'.format(
-                    message['message']
+                    message
                 )
             )
-            return
-        text = message['message'].get('text')
-        if text is None:
-            text = message['message']['payload']
-        img_attempt = True if 'image' in message['message'] else False
+        text = message.get('text', None)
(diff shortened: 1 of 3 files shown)


parlai/chat_service/services/messenger/agents.py
```diff
@@ -56,23 +56,21 @@ def observe_typing_on(self, persona_id=None):
...
-                )
-            )
-            return
-        text = message['message'].get('text')
-        if text is None:
-            text = message['message']['payload']
-        img_attempt = True if 'image' in message['message'] else False
+        mid = messenger_message['message']['mid']
+        seq = messenger_message['message'].get('seq', None)
+        message = messenger_message['message']
```
emilydinan: message vs. messenger_message is slightly confusing, although i know what you mean by it. perhaps we could come up with a better name here, although i can't think of one off the top of my head

parlai/chat_service/services/messenger/agents.py
```diff
@@ -81,13 +79,17 @@ def put_data(self, message):
...
+                'report_sender': message.get('report_sender', None),
+                'sticker_sender': messenger_message.get('sticker_sender', None),
                 'img_attempt': img_attempt,
             }
             if img_attempt and self.data.get('allow_images', False):
-                action['image_url'] = message['message'].get('image_url')
-                action['attachment_url'] = message['message'].get('attachment_url')
+                action['image_url'] = message.get('image_url')
+                action['attachment_url'] = message.get('attachment_url')
+                if action['image_url'] is None and 'attachments' in message:
```
emilydinan: i believe you don't need to check for `'attachments' in message` here because `image_attempt` already checks for that

parlai/chat_service/services/messenger/agents.py
klshuster: it's honestly incredibly confusing, i figured if i included `messenger` in there it would be more clear that the message is coming from `messenger`; perhaps we can call it `messenger_data`?

klshuster:
i changed to `messenger_data`, if you're ok with that @emilydinan i'll go ahead and merge

parlai/chat_service/services/messenger/agents.py
emilydinan: yeah messenger data definitely sounds better

PR #2420, opened 2020-02-21
Remove original ConvAI2 models and related legacy code

stephenroller:
**Patch description**
ConvAI2 continues to be an important dataset, but we are currently maintaining models from the contest:
- seq2seq
- kvmemnn
- language_model

All three of these models have become cumbersome to manage. The nightly test for seq2seq, for example, fails with a CudNN error. The language_model isn't a torch agent, and I believe isn't used anywhere else. The seq2seq v0 is highly deprecated and shouldn't be used, and the few dependencies I can find (twitter).

Remaining pieces:
- [x] Chat service tutorial still refers to `seq2seq:0`
- [x] Quickstart still refers to `seq2seq:0`, should be updated.

**Testing steps**
CI should still pass.


stephenroller:
Kurt, this going to need you reviewing the chat services, we'll need to update those to remove the last bits.

klshuster:
yeah we'll need to update the following config: https://github.com/facebookresearch/ParlAI/blob/master/parlai/chat_service/tasks/chatbot/config.yml#L15

everything else should be fine. that config in the tutorial is actually out of date, as the model params should be under a `models` key (see linked config)

stephenroller:
This is ready.

docs/source/tutorial_chat_service.md
```diff
@@ -34,10 +34,8 @@ overworld: MessengerOverworld
 max_workers: 30
 opt:  # Additional model opts go here
   debug: True
-  model: legacy:seq2seq:0
-  model_file: models:convai2/seq2seq/convai2_self_seq2seq_model
-  override:
-    model: legacy:seq2seq:0
+  model: transformer/generator
```
klshuster, line 37: Totally ok if you don't want to do this, but for parity it'd be good to have the example config mirror the actual config you changed, i.e. to have a `models` key under `opt` with this config. e.g.:

```
opt:
  debug: True
  models:
    blender_90M:
      model: transformer/generator
      model_file: zoo:blender/blender_90M/model
```

PR #2421, opened 2020-02-21
Remove match_batch from TorchAgent

### LINKED ISSUE #2179: Investigate whether match_batch is still needed
In TorchAgent and seq2seq, we have some gross logic that tries to sort a batch so that the RNN is happy to accept it. Previously in older torch versions, something in the RNN required that the longest example be the very first example, and the match_batch madness was to deal with that.

I have an inkling that it may not be necessary anymore, but it would be great if someone checked.

Steps:
1. Create a subteacher in `integration_tests` that creates variable length outputs. This would be nice to have anyway.
2. Add a test with it for transformer/generator and seq2seq
3. Disable the flag that causes match_batch to sort code to run in the seq2seq.  See if this breaks your new tests.
4. If so, push the new tests and call it a day.
5. If not, delete the match_batch logic from everywhere. Make sure you run the long tests.

stephenroller:
**Patch description**
Fixes #2179.

**Testing steps**
Enter steps to test your pull request. Give a clear and concise description of
what you expected to happen during testing.

**Logs**
```
```

**Other information**

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1799,7 +1799,7 @@ class TorchAgent(ABC, Agent):
             for k, v in output.items():
                 if v is None:
                     continue
-                for i, sub_val in zip(batch.valid_inds, v):
+                for i, sub_val in zip(batch.valid_indices, v):
                     batch_reply[i][k] = sub_val
 
         # broadcast the metrics back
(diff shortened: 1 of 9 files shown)


parlai/core/torch_agent.py
```diff
@@ -1848,9 +1792,15 @@ def batch_act(self, observations):
                 # use `with torch.enable_grad()` to gain back gradients.
                 output = self.eval_step(batch)
 
+        # match up items of output to their corresponding Message
+        # e.g. we're transposing a {key -> [value1, value2, ...]} to
+        # [{key: value1}, {key: value2}, ...]
         if output is not None:
-            # local metrics are automatically matched up
-            self.match_batch(batch_reply, batch.valid_indices, output)
+            for k, v in output.items():
```
stephenroller: just folded in the logic from that method into here.

stephenroller:
Let's try again another time.

PR #2427, opened 2020-02-24
[TRA] Add option to return scores

emilydinan:
**Patch description**
Added option to return scores from eval_step. I've found this useful for debugging and thresholding. Default is False.


emilydinan:
> any reason we _wouldnt_ want to return the scores? we could just include it and forget the flag



> any reason we _wouldnt_ want to return the scores? we could just include it and forget the flag

I think this could be really expensive/ cause problems in mturk if we are trying to pass around these torch tensors. especially if you are ranking a large number of candidates. so i think the default should be False

PR #2419, opened 2020-02-20
Classifier probabilities & small fixes

jaseweston:

PR #2430, opened 2020-02-25
[Torch Classifier] print scores bug fix

emilydinan:
**Patch description**
Torch Classifier bug introduced by #2419. Scores should only be returned in interactive mode if `print_scores` = True. Otherwise we simply want the label to be printed.

### DIFF
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -287,7 +287,7 @@ class TorchClassifierAgent(TorchAgent):
 
         if batch.labels is None or self.opt['ignore_labels']:
             # interactive mode
-            if self.opt.get('interactive_mode', False):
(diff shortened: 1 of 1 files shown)


parlai/core/torch_classifier_agent.py
```diff
@@ -287,7 +287,7 @@ def eval_step(self, batch):
 
         if batch.labels is None or self.opt['ignore_labels']:
             # interactive mode
-            if self.opt.get('interactive_mode', False):
+            if self.opt.get('print_scores', False):
```
stephenroller: should print scores be set automatically true in interactive?

emilydinan: No I don't think so. We don't automatically want to return the scores.  Most often we just want the class prediction.

PR #2431, opened 2020-02-25
Tests for special tokens in bytelevel bpe

dexterju:
**Patch description**
In byte-level-bpe tokenizer, we add 4 special tokens to the Hugging Face tokenizer and to the ParlAI dict as well.
This test ensures that all these 4 tokens are added in the correct sequence in both cases, so the mapping between them is consistent. 

**Testing steps**
```
py tests/test_dict.py
```

### DIFF
--- a/tests/test_dict.py
+++ b/tests/test_dict.py
@@ -132,10 +132,20 @@ class TestDictionary(unittest.TestCase):
             agent.txt2vec(u'Hello, ParlAI! \U0001f600'),
             [agent.tok2ind[w] for w in BYTELEVEL_BPE_RESULT],
         )
-        # Test special token
-        self.assertEqual(
-            agent.txt2vec(agent.start_token), [agent.tok2ind[agent.start_token]]
-        )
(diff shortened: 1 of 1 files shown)


tests/test_dict.py
```diff
@@ -132,10 +132,20 @@ def test_byte_level_bpe_tokenize(self):
...
+        special_tokens = [
+            agent.null_token,
+            agent.start_token,
+            agent.end_token,
+            agent.unk_token,
+        ]
+        for each_token in special_tokens:
+            self.assertEqual(
+                agent.byte_level_bpe.tokenizer.token_to_id(each_token),
+                agent.tok2ind[each_token] - 4 + vocab_size,
```
stephenroller: fwiw this is a bit reductive and doesn't necessarily test what you want. it's testing the implementation, not the result. given that there are only 4, you probably just want to enumerate them:

assert txt2vec(null_token) == 0
assert txt2vec(start_token) == 1
assert txt2vec(end_token) == 2
...

Issue #2417, opened 2020-02-19
Issues with Wizard of Wikipedia


g-karthik:
Hello!

I'm trying to parse examples to train a knowledge selection model with the Wizard of Wikipedia dataset. I'm creating label candidates by following the code [here](https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/wizard_of_wikipedia/agents.py#L309L318).

I am finding that there are some cases where some candidates fetched using the code above have > 4000 tokens. Here's a couple sample candidates I printed from the `label_cands` list in the above linked code.

Search for `Index of philatelic articles` in `train.json` and you will find the below in the list of candidates.

```
[...]

PR #2437, opened 2020-02-26
[WoW] Wizard Teacher Fix

klshuster:
**Patch description**
Per #2417 we should be checking for `self.knowledge_separator` and adding if necessary to the `label`.

**Testing steps**
Ran `verify_data` before and after change.

**Logs**
Before:

```
$ python parlai/scripts/verify_data.py -t wizard_of_wikipedia:WizardDialogKnowledge --label-type chosen_sent --include-knowledge-separator true
.
.
.
28s elapsed: {"missing_text": 0, "missing_labels": 0, "missing_label_candidates": 0, "empty_string_label_candidates": 0, "label_candidates_with_missing_label": 69389, "did_not_return_message": 0, "exs": 74092, "%done": "100.00%", "time_left": "0s"}
```

After:

```
$ python parlai/scripts/verify_data.py -t wizard_of_wikipedia:WizardDialogKnowledge --label-type chosen_sent --include-knowledge-separator true
.
.
.
44s elapsed: {"missing_text": 0, "missing_labels": 0, "missing_label_candidates": 0, "empty_string_label_candidates": 0, "label_candidates_with_missing_label": 1, "did_not_return_message": 0, "exs": 74092, "%done": "100.00%", "time_left": "0s"}
```

### DIFF
--- a/parlai/tasks/wizard_of_wikipedia/agents.py
+++ b/parlai/tasks/wizard_of_wikipedia/agents.py
@@ -309,7 +309,7 @@ class WizardDialogKnowledgeTeacher(WizardOfWikipediaTeacher):
                 labels = ['{} {}'.format(title, sentence)]
 
         # finally, get label_candidates
-        label_cands = [TOKEN_NOCHOSEN + ' ' + TOKEN_NOCHOSEN]
+        label_cands = ['{} {}'.format(TOKEN_NOCHOSEN, TOKEN_NOCHOSEN)]
         knowledge_str = ''
         for title, passage in knowledge_dict.items():
             for p in passage:
--- a/parlai/tasks/wizard_of_wikipedia/agents.py
+++ b/parlai/tasks/wizard_of_wikipedia/agents.py
@@ -303,7 +303,10 @@ class WizardDialogKnowledgeTeacher(WizardOfWikipediaTeacher):
             labels = [wizard_entry['text']]
         else:
             title, sentence = _get_chosen_title_and_sent(wizard_entry, knowledge_dict)
-            labels = ['{} {}'.format(title, sentence)]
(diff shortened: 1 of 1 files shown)


parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -303,7 +303,10 @@ def get(self, episode_idx, entry_idx=0):
...
         else:
             title, sentence = _get_chosen_title_and_sent(wizard_entry, knowledge_dict)
-            labels = ['{} {}'.format(title, sentence)]
+            if self.knowledge_separator and title != TOKEN_NOCHOSEN:
+                labels = ['{} {} {}'.format(title, TOKEN_KNOWLEDGE, sentence)]
+            else:
+                labels = ['{} {}'.format(title, sentence)]
 
         # finally, get label_candidates
         label_cands = [TOKEN_NOCHOSEN + ' ' + TOKEN_NOCHOSEN]
```
g-karthik: Shouldn't you be checking for `self.knowledge_separator` here as well?

klshuster: i decided to not add the knowledge separator if the label is `TOKEN_NOCHOSEN`

g-karthik: Ah, I just noticed the additional `title != TOKEN_NOCHOSEN` condition you added in line 306. Your decision makes sense now.

Just a minor note: if you can change the new line 312 to `label_cands = ["{} {}".format(TOKEN_NOCHOSEN, TOKEN_NOCHOSEN)]`, that would make it consistent with the rest of the code.

stephenroller:
Thanks Kurt!

PR #2442, opened 2020-03-02
[WIP]Free memory before loading model back when overfitting

dexterju:
**Patch description**
The previous code caused CUDA OOM issues with large models, but those models were able to fit in CUDA mem when train script launched. 
It seems that python doesn't do garbage collection in time, so we would have 2 copies of the model at same time when we do self.agent = create_agent() when load best model back. 

**Testing steps**
Pending on testing jobs runs. Will update later.

### DIFF
--- a/parlai/scripts/train_model.py
+++ b/parlai/scripts/train_model.py
@@ -28,6 +28,7 @@ import json
 import numpy as np
 import os
 import signal
+import torch
 
 from parlai.core.metrics import Metric
 from parlai.core.agents import create_agent, create_agent_from_shared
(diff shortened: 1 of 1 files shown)


parlai/scripts/train_model.py
```diff
@@ -715,6 +716,8 @@ def train(self):
             self.save_model()
         elif opt.get('model_file'):
             # reload best validation model
+            del self.agent
```
stephenroller: I think you only need this, not the empty cache.  [from the docs](https://pytorch.org/docs/stable/cuda.html#torch.cuda.empty_cache)

image

dexterju:
This doesn't fix the OOM at loading time, Will look into it further.

stephenroller:
Supplanted by #2487

PR #2443, opened 2020-03-02
Improve adafactor optimizer

dexterju:
**Patch description**
The current implementation of ada factor is a bit confusing of the lr rate scheduler design. 
In my own case, it took me sometime to realize why my lr was not the same as I set and why it gets even larger than the lr given.

Changes I have done:
1.  Remove the relative step part, since that is the same as `inverse square root lr schduler`
2. Add options and warnings when we scale the lr by the model parameters. 

**Testing steps**
Tested with new fiarseq fine-tuning runs.

### DIFF
--- a/parlai/utils/fp16.py
+++ b/parlai/utils/fp16.py
@@ -523,9 +523,6 @@ class Adafactor(torch.optim.Optimizer):
     :param scale_parameter (bool):
         if true, learning rate is scaled by root mean square of parameter
         (default: True)
-    :param relative_step (bool):
-        if true, time-dependent learning rate is computed instead of external
-        learning rate (default: True)
     :param warmup_init (bool):
         time-dependent learning rate computation depends on whether warm-up
         initialization is being used (default: False)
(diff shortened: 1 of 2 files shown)


dexterju:
Leave the fight / argue to @stephenroller

parlai/utils/fp16.py
```diff
@@ -552,20 +548,15 @@ def __init__(
...
     def _get_lr(self, param_group, param_state):
         rel_step_sz = param_group['lr']
-        if param_group['relative_step']:
-            min_step = (
-                1e-6 * param_state['step'] if param_group['warmup_init'] else 1e-2
-            )
-            rel_step_sz = min(min_step, 1.0 / math.sqrt(param_state['step']))
         param_scale = 1.0
         if param_group['scale_parameter']:
+            warn_once('Lr is rescaled by Adafactor optimizer.')
```
stephenroller: I don't think we need the warning.

stephenroller:
Let's finish this up.

dexterju:
Let me finish this up.

dexterju:
Close due to #2547

PR #2447, opened 2020-03-03
Log metrics as a table

stephenroller:
**Patch description**
The introduction of agent level metrics has caused an explosion of the size of the metrics dictionary output. This is an attempt to clean things up a bit.

Changes:
- Mark some metrics as Global Metrics, which are not meant to be reported per-task.
- Have printing use pandas to output a tabular format

**Testing steps**
With a limited width terminal, multitasking
image

With a much longer terminal, multitasking:
image

With just a single task:
image

With a ranking model:
image


parlai/core/metrics.py
```diff
@@ -66,6 +66,12 @@ class Metric(ABC):
     Subclasses should define .value().
     """
 
+    def is_global(self) -> bool:
+        """
+        Indicates whether this metric should be reported globally or per-task.
+        """
+        return False
```
EricMichaelSmith: Hmm what makes this per-task by default? Just that it's not `GlobalMetric`?

parlai/core/metrics.py
```diff
@@ -391,12 +418,19 @@ def aggregate_named_reports(named_reports: Dict[str, Dict[str, Metric]]):
     :param reports: Dict of tasks -> metrics.
     """
     # reporters is a list of teachers or worlds
-    m: Dict[str, Metric] = {}
+    m: Dict[str, Union[Metric, Dict[str, Metric]]] = {}
```
EricMichaelSmith: So this allows nested dicts? I thought you didn't want that, right? And how below would nested dicts occur, since it looks like `task_report[each_metric]` needs to be a `Metric` in order to have an `.is_global()` method?

parlai/core/torch_agent.py
```diff
@@ -1835,11 +1843,11 @@ def batch_act(self, observations):
             # tokens per batch
             # we divide by the binary is_primary_worker() so that the numerator is
             # num_tokens in all workers, and the denominator is 1.
-            tbp = AverageMetric(
+            tbp = GlobalAverageMetric(
                 (batch.label_vec != self.NULL_IDX).sum().item(),
                 float(is_primary_worker()),
             )
-            self.global_metrics.add('tokens_per_batch', tbp)
+            self.global_metrics.add('tpb', tbp)
```
EricMichaelSmith: Nice but might obscure meaning...

parlai/scripts/train_model.py
```diff
@@ -74,7 +75,7 @@ def setup_args(parser=None) -> ParlaiParser:
     train.add_argument('--display-examples', type='bool', default=False, hidden=True)
     train.add_argument('-eps', '--num-epochs', type=float, default=-1)
     train.add_argument('-ttim', '--max-train-time', type=float, default=-1)
-    train.add_argument('-ltim', '--log-every-n-secs', type=float, default=2)
+    train.add_argument('-ltim', '--log-every-n-secs', type=float, default=10)
```
EricMichaelSmith: 👍 Didn't realize the default was so quick before!

parlai/scripts/train_model.py
```diff
@@ -444,10 +445,10 @@ def validate(self):
             or self.valid_optim * new_valid > self.valid_optim * self.best_valid
         ):
             print(
-                '[ new best {}: {}{} ]'.format(
+                '[ new best {}: {:.4g}{} ]'.format(
                     opt['validation_metric'],
                     new_valid,
-                    ' (previous best was {})'.format(self.best_valid)
+                    ' (previous best was {:.4g})'.format(self.best_valid)
```
EricMichaelSmith: Nit: hmm is this two nested `.format()`s? Might be worth cleaning up with f-strings

parlai/utils/misc.py
```diff
@@ -349,17 +350,74 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
         pass
 
 
-def nice_report(report):
+def _report_sort_key(report_key: str) -> Tuple[str, str]:
+    """
+    Sorting name for reports.
+
+    Sorts by main metric alphabetically, then by task.
```
EricMichaelSmith: Yay for clean sorting!

parlai/utils/misc.py
```diff
@@ -349,17 +350,74 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+    """
+    fields = report_key.split("/")
+    main_key = fields.pop(-1)
+    sub_key = '/'.join(fields)
+    return (sub_key or 'all', main_key)
+
+
+def _float_format(f):
+    if f != f:
```
EricMichaelSmith: hmm, a bit unclear - is this only to check for `None`s?

parlai/utils/misc.py
```diff
@@ -349,17 +350,74 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
     from parlai.core.metrics import Metric
 
-    output = {}
-    for k, v in report.items():
+    try:
+        import pandas as pd
+
+        use_pandas = True
+    except ImportError:
+        use_pandas = False
```
EricMichaelSmith: Maybe give a warning here?

parlai/core/torch_agent.py
stephenroller: I'm fine with that, it was too annoying

parlai/scripts/train_model.py
stephenroller: it's because the first go is different than others.

parlai/utils/misc.py
stephenroller: NaNs

parlai/core/metrics.py
stephenroller: ah actually no, no nested dicts.

parlai/utils/misc.py
```diff
@@ -384,7 +386,7 @@ def nice_report(report) -> str:
     try:
         import pandas as pd
 
-        use_pandas = True
+        use_pandas = False
```
EricMichaelSmith: Oof - why false now? Don't we want to use pandas if we can import it?

stephenroller: Debugging haha

stephenroller:
Okay this is finalized.

parlai/core/torch_agent.py
```diff
@@ -1849,11 +1857,11 @@ def batch_act(self, observations):
...
```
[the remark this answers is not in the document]
stephenroller: rofl

parlai/utils/misc.py
```diff
@@ -349,17 +351,81 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
 
-def nice_report(report):
+def _report_sort_key(report_key: str) -> Tuple[str, str]:
+    """
+    Sorting name for reports.
+
+    Sorts by main metric alphabetically, then by task.
+    """
+    fields = report_key.split("/")
+    main_key = fields.pop(-1)
```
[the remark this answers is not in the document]
stephenroller: Main key is the metric name, e.g fields can be like "empathetic_dialogues" and main_key will be "f1". the reason it could be multiple levels is sometimes names have /'s in them (like filenames)

parlai/utils/misc.py
```diff
@@ -349,17 +351,81 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+    sub_key = '/'.join(fields)
+    return (sub_key or 'all', main_key)
+
+
+def float_formatter(f):
+    if f != f:
+        return ""
```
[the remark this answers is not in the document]
stephenroller: then it renders as nan and it's real ugly.

PR #2448, opened 2020-03-04
Segments and softmaxes.

stephenroller:
**Patch description**
- Support `--n-segments` in polyencoder
- Use fp16 safe cross entropy loss for the ranking objective.

**Testing steps**
Internal sweeps.


stephenroller:
Ya I yolo'd the fp16 change, I'm still testing it internally.

stephenroller:
It worked well internally.

PR #2458, opened 2020-03-10
[metrics] Support macro averaging.

stephenroller:
**Patch description**
Bring back support for macro averaging, which was lost in the metrics.

After discussion during the team meeting, I continued to look at the code more and more and determined that I still did not like the proposed solution of mismatching metrics during train/eval time (which was one of the major motivations of user level metrics). As such, I've implemented proper macro averaging as a higher level metric.

The only part I'm really unhappy about is carving out the exception for SumMetric. It's a bit annoying because `exs` is a critical metric that must be reported, and we don't want that averaged per-task, we want it summed.

**Testing steps**
Manual testing. New CI. Long CI of older things. @klshuster  can we run a test with your dodeca models (just point me to the model files in chat and a task mixture)

~Running locally, I think there might be a hang around when using with multiprocessing train.~ Resolved.

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -461,7 +461,7 @@ def aggregate_named_reports(
 
     # reporters is a list of teachers or worlds
     m: Dict[str, Metric] = {}
-    macro_averages = {}
+    macro_averages: Dict[str, List[Metric]] = {}
     for task_id, task_report in named_reports.items():
         for each_metric, value in task_report.items():
             if value.is_global():
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -471,7 +471,7 @@ def aggregate_named_reports(
             else:
                 task_metric = f'{task_id}/{each_metric}'
                 m[task_metric] = m.get(task_metric) + value
-                if micro_average or isinstance(value, (SumMetric, FixedMetric)):
+                if micro_average or not isinstance(value, AverageMetric):
                     # none + a => a from implementation of Metric.__add__
                     m[each_metric] = m.get(each_metric) + value
                 else:
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -457,7 +457,7 @@ def aggregate_named_reports(
         raise ValueError("Cannot aggregate empty reports.")
     if len(named_reports) == 1:
         # no real aggregation to be done
-        return next(named_reports.values())
+        return next(iter(named_reports.values()))
 
     # reporters is a list of teachers or worlds
     m: Dict[str, Metric] = {}
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -263,10 +263,10 @@ class GlobalMetric:
     """
     A global metric is one that should not be aggregated across different tasks.
 
-    Key to it is the notion that any one worker or any one task already has a
-    global view of the value, and so no combinations should be done. Note this
-    is different then a FixedMetric, in that a GlobalMetric can be still
-    averaged across multiple parleys(), but a FixedMetric is always fixed.
+    Key to it is the notion that any one worker or any one task already has a global
+    view of the value, and so no combinations should be done. Note this is different
+    then a FixedMetric, in that a GlobalMetric can be still averaged across multiple
+    parleys(), but a FixedMetric is always fixed.
     """
 
     def is_global(self) -> bool:
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -231,9 +231,42 @@ class AverageMetric(Metric):
         return self._numer / self._denom
 
 
+class MacroAverageMetric(Metric):
+    """
+    Class that represents the macro average of several numbers.
+
+    Used for aggregating task level metrics.
+    """
+
+    __slots__ = ('_values',)
+
+    def __init__(self, metrics: List[Metric]) -> None:
+        self._values = metrics
+
+    def __add__(self, other: Optional['MacroAverageMetric']) -> 'MacroAverageMetric':
+        if other is None:
+            return self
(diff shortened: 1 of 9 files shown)


tests/test_metrics.py
```diff
@@ -89,10 +98,12 @@ def test_average_metric_additions(self):
             self.assertAlmostEqual(actual_output, output, places=6)
             self.assertIsInstance(actual_output, float)
 
-    def test_fixedmetric(self):
```
EricMichaelSmith: It's not still useful to test FixedMetric here?

tests/test_metrics.py
```diff
@@ -173,5 +184,82 @@ def test_largebuffer(self):
...
+            'fixed': FixedMetric(4),
+            'global_avg': GlobalAverageMetric(3, 4),
+        }
+        report2 = {
+            'avg': AverageMetric(1, 3),
+            'sum': SumMetric(4),
+            'fixed': FixedMetric(4),
+            'global_avg': GlobalAverageMetric(1, 3),
+        }
+        agg = aggregate_unnamed_reports([report1, report2])
```
EricMichaelSmith: Hmm I suppose we can safely assume that all unnamed reports will never need to be macroaveraged?

parlai/core/metrics.py
```diff
@@ -411,26 +444,43 @@ def normalize_answer(s):
...
                     m[each_metric] = value
             else:
-                # none + a => a from implementation of Metric.__add__
-                m[each_metric] = m.get(each_metric) + value
-                if len(named_reports) > 1:
-                    task_metric = f'{task_id}/{each_metric}'
-                    m[task_metric] = m.get(task_metric) + value
+                task_metric = f'{task_id}/{each_metric}'
+                m[task_metric] = m.get(task_metric) + value
+                if micro_average or isinstance(value, (SumMetric, FixedMetric)):
```
EricMichaelSmith: Ah, here's the `SumMetric` carve-out. Since we're implementing macro-*averaging* here, why don't we have the condition be that you only macroaverage if it's an `AverageMetric`?

tests/test_metrics.py
stephenroller: We can implement it later as we need. For now, unnamed reports is for two cases:
- Aggregating across workers
- Aggregating across batches

Both cases will always want micro averaging.

parlai/core/metrics.py
stephenroller: Yeah I wrestled with that. PPL stands out as a weird edge case. It is an average metric, so it would work. What does @klshuster think?

tests/test_metrics.py
stephenroller: You're right, I should leave it.

parlai/core/metrics.py
stephenroller: Updated per your suggestion.

parlai/core/metrics.py
```diff
@@ -231,9 +231,42 @@ def value(self) -> float:
...
+
+
 class GlobalMetric:
     """
     A global metric is one that should not be aggregated across different tasks.
+
+    Key to it is the notion that any one worker or any one task already has a global
+    view of the value, and so no combinations should be done. Note this is different
+    then a FixedMetric, in that a GlobalMetric can be still averaged across multiple
+    parleys(), but a FixedMetric is always fixed.
```
emilydinan: can you give examples of each of these metrics in the descriptiosn? I find it slightly confusing

parlai/core/metrics.py
```diff
@@ -411,26 +444,43 @@ def normalize_answer(s):
...
 
-def aggregate_named_reports(named_reports: Dict[str, Dict[str, Metric]]):
+def aggregate_named_reports(
+    named_reports: Dict[str, Dict[str, Metric]], micro_average: bool = False
+):
     """
     Aggregate metrics from multiple reports.
 
     :param reports: Dict of tasks -> metrics.
+    :param micro_average: If true, top level metrics will be the macro average.
```
klshuster: nit: `If true, top level metrics will be the _micro_ average`

parlai/nn/lr_scheduler.py
```diff
@@ -91,6 +92,7 @@ def load_state(self, states):
         if states.get('warmup_scheduler') and getattr(self, 'warmup_scheduler', None):
             self.warmup_scheduler.load_state_dict(states['warmup_scheduler'])
         self._number_training_updates = states.get('number_training_updates', 0)
+        self.step(self._number_training_updates)
```
klshuster: what is this change?

parlai/scripts/train_model.py
```diff
@@ -182,7 +182,16 @@ def setup_args(parser=None) -> ParlaiParser:
         'ppl,f1,accuracy,hits@1,rouge,bleu'
         'the rouge metrics will be computed as rouge-1, rouge-2 and rouge-l',
     )
+    train.add_argument(
```
klshuster: perhaps a totally different issue here but seeing this param added twice leads me to believe we should have some sort of `common_script_args` that we can add for all our scripts (would help standardize command line params e.g. how `-n` is num examples in some while `-ne` is num examples in others)

parlai/core/metrics.py
stephenroller: hardly a nit!

parlai/nn/lr_scheduler.py
stephenroller: oh boy, it's a bit of a pytorch hack.

Until we do a step(), the LR from a FRESHLY loaded model may report an incorrect value. The changes in train model that force a reload surfaced this.

parlai/core/metrics.py
stephenroller: Done for all the metrics classes.

PR #2464, opened 2020-03-11
[Conversations] Add Conversation and Turn classes

emilydinan:
**Patch description**
Add Conversation and Turn classes for easier iteration over conversations.

### DIFF
--- a/parlai/utils/conversations.py
+++ b/parlai/utils/conversations.py
@@ -11,6 +11,8 @@ import json
 import os
 import random
 
+from parlai.utils.misc import AttrDict
+
 
 BAR = '=' * 60
 SMALL_BAR = '-' * 60
@@ -93,6 +95,72 @@ class Metadata:
             f.write(json.dumps(metadata))
 
 
+class Turn(AttrDict):
+    """
+    Utility class for a dialog turn.
+    """
+    def __init__(self, id=None, text=None, **kwargs):
+        super().__init__(self, id=id, text=text, **kwargs)
+
+
+class Conversation:
+    """
+    Utility class for iterating through a single episode.
+
+    Used in the context of the Conversations class.
+    """
+    def __init__(self, episode):
+        self.episode = episode
+        self.context = episode.get('context')
+        self.metadata_path = episode.get('metadata_path')
+        self.turns = self._build_turns(episode)
+        self.iterator_idx = 0
+
+    def _build_turns(self, episode):
+        turns = []
+        for act_pair in episode['dialog']:
+            for act in act_pair:
+                turns.append(Turn(**act))
+        return turns
+
+    def read(self):
+        print(BAR)
+        high_level = [k for k in self.episode.keys() if k != 'dialog']
+        if high_level:
+            for key in high_level:
+                print(f'{key}: {self.episode[key]}')
+            print(SMALL_BAR)
+
+        for turn in self.turns:
+            print(f'{turn.id}: {turn.text}')
+
+        print(BAR)
+
+    @property
+    def num_turns(self):
+        return len(self.turns)
+
+    def __getitem__(self, index):
+        return self.turns[index]
+
+    def __next__(self):
+        """
+        Return the next conversation.
+        """
+        if self.iterator_idx >= len(self.turns):
+            print('You reached the end of the conversation.')
+            self.reset()  # return the iterator idx to 0
+            return None
+
+        conv = self.turns[self.iterator_idx]
+        self.iterator_idx += 1
+
+        return conv
+
+    def reset(self):
+        self.iterator_idx = 0
+
+
 class Conversations:
     """
     Utility class for reading and writing from ParlAI Conversations format.
@@ -139,7 +207,7 @@ class Conversations:
         with open(datapath, 'r') as f:
             lines = f.read().splitlines()
             for line in lines:
-                conversations.append(json.loads(line))
+                conversations.append(Conversation(json.loads(line)))
 
         return conversations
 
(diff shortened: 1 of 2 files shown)


parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+
+
+class Conversation:
+    """
+    Utility class for iterating through a single episode.
+
+    Used in the context of the Conversations class.
+    """
+
+    def __init__(self, episode):
```
klshuster: so the `Conversation` assumes that an episode is formatted a certain way right? would it be good to document here what fields the `Conversation` class is expecting (since you can't just load with any arbitrarily built episode)?

parlai/utils/conversations.py
```diff
@@ -187,21 +257,7 @@ def __next__(self):
...
-            print(SMALL_BAR)
-
-        for act_pair in convo['dialog']:
-            for turn in act_pair:
-                turn_id = turn['id']
-                text = turn['text']
-                print(f'{turn_id}: {text}')
-
-        print(BAR)
+        convo.read()
```
klshuster: i like this a lot. if we were feeling fancy we could also define the `__str__(self)` method for a `Conversation` that would return a string to print when we do `print(convo)`, but I am definitely happy with this

emilydinan: That's a good idea... will change it

parlai/utils/conversations.py
emilydinan: Yeah it assumes the episode is formatted in the way returned by the World Logger. Aka it is a list of act pairs, which are in turn a list of acts (aka Messages/dicts). Will document

parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+                turns.append(Turn(**act))
+        return turns
+
+    def read(self):
+        print(BAR)
+        high_level = [k for k in self.episode.keys() if k != 'dialog']
+        if high_level:
+            for key in high_level:
+                print(f'{key}: {self.episode[key]}')
+            print(SMALL_BAR)
```
stephenroller: lol

parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+        self.iterator_idx = 0
+
+    def _build_turns(self, episode):
+        turns = []
+        for act_pair in episode['dialog']:
+            for act in act_pair:
+                turns.append(Turn(**act))
+        return turns
+
+    def read(self):
```
stephenroller: I agree with Kurt, It would be nice to have a `__str__` that one would call print.

a `__repr__` would also be nice, but it could be more high level like `Conversation(turns=5, first="Hi how are, ...")` or something like that?

parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+        print(BAR)
+
+    @property
+    def num_turns(self):
+        return len(self.turns)
+
+    def __getitem__(self, index):
+        return self.turns[index]
+
+    def __next__(self):
```
stephenroller: does this work without an __iter__? I'm a bit surprised.

parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+        """
+        Return the next conversation.
+        """
+        if self.iterator_idx >= len(self.turns):
+            print('You reached the end of the conversation.')
```
stephenroller: Drop the print. Iterating through the conversation shouldn't have such a side effect...

parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+            self.reset()  # return the iterator idx to 0
+            return None
```
stephenroller: Normally you'd do `raise StopIteration` to indicate we're at the end. Python uses this for for loops...

parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+        for turn in self.turns:
+            print(f'{turn.id}: {turn.text}')
+
+        print(BAR)
+
+    @property
+    def num_turns(self):
+        return len(self.turns)
+
+    def __getitem__(self, index):
```
stephenroller: awesome

parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+                print(f'{key}: {self.episode[key]}')
+            print(SMALL_BAR)
+
+        for turn in self.turns:
+            print(f'{turn.id}: {turn.text}')
+
+        print(BAR)
+
+    @property
+    def num_turns(self):
```
stephenroller: maybe `__len__` instead?

parlai/utils/conversations.py
```diff
@@ -93,6 +95,74 @@ def save_metadata(
...
+            print('You reached the end of the conversation.')
+            self.reset()  # return the iterator idx to 0
+            return None
+
+        conv = self.turns[self.iterator_idx]
+        self.iterator_idx += 1
+
+        return conv
+
+    def reset(self):
```
stephenroller: so this is normally why you would do an `__iter__`, which removes the need for the reset logic

tests/test_conversations.py
```diff
@@ -97,6 +97,11 @@ def test_conversations(self):
         )
         self.assertIn(str_version, out.getvalue())
 
+        # test getting a specific turn
+        first = convos[0]  # Conversation
+        self.assertEqual(first[0].id, 'Emily')
+        self.assertEqual(first[3].text, 'Yes, I will do it right now!')
```
stephenroller: (y)

emilydinan:
> so nice! this looks very clean @emilydinan
> 
> context question, is there a specific task for which this is planned to be used in? is it multipurpose?

Good question @dianaglzrico --  no specific task yet. We are hoping to be able to utilize this for easy iteration through conversations. In the future we will have teachers that load from this conversations format. This is also the format that gets fed into the "acute eval" Mechanical turk task and which is saved by self chats.

stephenroller:
🥳

PR #2468, opened 2020-03-13
transformer classif default bug

jaseweston:
This was causing the reduction_type to be incorrect by default when e.g. using eval_model.
Found by the roller.


stephenroller:
Original source of the line:
https://github.com/facebookresearch/ParlAI/commit/a9f5928d7e19e5af1992defa4aca7b8ebe8f7e58

Talked to @emilydinan, doesn't remember why set_params was chosen over set_defaults.  Going to accept the fix, but we should maybe keep an eye out for if anything breaks.

PR #2469, opened 2020-03-13
[Self Chat] Reset bug fix

emilydinan:
**Patch description**
Self chat bug: world is not resetting after each self chat episode.


stephenroller:
This is only for metrics right? It appears like it could be more concerning in some other recent results we've run?

emilydinan:
> This is only for metrics right? It appears like it could be more concerning in some other recent results we've run?

It does not affect recent results AFAIK -- for self chats, I had been using my own script for generation up until now.

PR #2476, opened 2020-03-17
Normalize with trailing period

jaseweston:


parlai/utils/strings.py
```diff
@@ -42,6 +46,9 @@ def normalize_reply(text: str) -> str:
     new_text = new_text.strip()
     new_text = new_text.replace('  ', ' ')
 
+    if version > 1 and (new_text[-1] not in ['!', '.', '?', ')']):
```
stephenroller:
```suggestion
    if version > 1 and new_text and and new_text[-1] not in '!.?)"\''
```
Note the fix for empty string and the handling of quotations

parlai/utils/strings.py
```diff
@@ -8,9 +8,13 @@
 """
 
 
-def normalize_reply(text: str) -> str:
+def normalize_reply(text: str, version=1) -> str:
     """
     Standardize the capitalization and punctuation spacing of the input text.
+
+    Version 1: Fix sentence start casing, and punctuation.
+    Version 2: Add trailing period, if missing.
```
stephenroller:
```suggestion
    - Version 1: Fix sentence start casing, and punctuation.
    - Version 2: Add trailing period, if missing.
```

stephenroller: I think this will fix the autoformat complaint.

PR #2478, opened 2020-03-17
Change length penalty to match NMT paper

spencerp:
**Patch description**
I was reading the NMT paper to understand this length penalizing a little better, and realized we're not using the same constants they did. It kind of feels like we should use the constants they found to be optimal, or experiment with our own constants to find out what works best for dialog.

Screen Shot 2020-03-16 at 10 55 20 PM

The basic behavior of this function is the same either way, though, so there's a good case to be made for favoring reproducibility over using the optimal constants by keeping what we have.

**Testing steps**
Plotted the final score, given hypothesis lengths from 1 to 15, an imaginary hypothesis score of -2.0, and a length penalty of 0.65:
Screen Shot 2020-03-16 at 11 01 30 PM

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -1227,7 +1227,7 @@ class TreeSearch(object):
         for finished_item in self.finished:
             current_length = finished_item.timestep + 1
             # these weights are from Google NMT paper
-            length_penalty = math.pow((1 + current_length) / 6, self.length_penalty)
+            length_penalty = math.pow((5 + current_length) / 6, self.length_penalty)
             rescored_finished.append(
                 _HypothesisTail(
                     timestep=finished_item.timestep,


stephenroller:
So on one model, similar to our dodeca model, I find that the constant 1 is actually better:


With 1:
```
[ Finished evaluating tasks ['convai2'] using datatype valid ]
    accuracy  bleu-4  exs    f1  gpu_mem  loss        lr   ppl  token_acc  total_train_updates   tpb
     .002051  .01564 7801 .2035    .4580 2.433 2.344e-07 11.39      .4720               108862 429.7
```
With 5:
```
[ Finished evaluating tasks ['convai2'] using datatype valid ]
    accuracy  bleu-4  exs    f1  gpu_mem  loss        lr   ppl  token_acc  total_train_updates   tpb
     .002307   .0154 7801 .2023    .4580 2.433 2.344e-07 11.39      .4720               108862 429.7
```

We need to maintain backwards compatibility anyway. If you would like to change this, make it into a command line parameter, rather than changing the constant. Leave the default to the current value, but you can do `recommended=5` if you want.

spencerp:
Okay, great! Just wanted to bring this up in case it was unintentional.

PR #2481, opened 2020-03-18
Model Parallel Transformers

stephenroller:
**Patch description**
Initial patch supporting model parallel with transformers. Only supports memory efficient fp16 right now, and doesn't support fused_layer_norm in my version of Apex, but they recently patched it (NVIDIA/apex#371)  so I need to upgrade and test.

Edit: Layer placement is worked out. Distributed training needs to be tested, but I think it's going to happen.

Seeking reviews from anyone who can provide feedback.

A few things to work out:
- **Layer placement**. Since we only distribute the Layers, and not the optimizer, there's a weird imbalance of memory usage (kinda like DP does), since the optimizer lives on the 0th GPU. We can account for this and more heavily weight layers in later positions to balance it out, but ideally the encoder/decoder and the candidate/context encoders should balance it all out, and RnR should balance even better. I think a more dynamic algorithm, based on number of parameters assigned to each layer, would be valuable.
- **Distributed training** Mostly needs to be tested, Fairseq folk say they've not had a problem.
- Can/should I move the model_parallel call into TorchAgent. It's very helpful that it happens where it is, and the flag will still be only added by agents which support it.

**Testing steps**
- Fine tuning a model on multiple GPUs with Adam instead of Adafactor. Full training works great, as does eval model.
- Some new unit tests.


parlai/agents/transformer/polyencoder.py
```diff
@@ -129,11 +134,15 @@ def __init__(self, opt, shared=None):
...
 
+        if self.model_parallel and shared is None:
+            self.model.model_parallel()
+
     def build_model(self, states=None):
         """
         Return built model.
         """
-        return PolyEncoderModule(self.opt, self.dict, self.NULL_IDX)
+        model = PolyEncoderModule(self.opt, self.dict, self.NULL_IDX)
```
emilydinan: ?

parlai/agents/transformer/transformer.py
```diff
@@ -206,12 +212,17 @@ def add_cmdline_args(cls, argparser):
...
+        self.model_parallel = opt.get('model_parallel', False) and self.use_cuda
+        if self.data_parallel and self.model_parallel:
+            raise RuntimeError('Cannot combine --model-parallel and --data-parallel')
         if self.data_parallel:
             from parlai.utils.distributed import is_distributed
 
             if is_distributed():
                 raise ValueError('Cannot combine --data-parallel and distributed mode')
             self.model = torch.nn.DataParallel(self.model)
+        # elif self.model_parallel and shared is None:
```
emilydinan: did you mean to leave these commented out

parlai/utils/torch.py
```diff
@@ -235,3 +238,229 @@ def forward(self, xs):
...
+    def guess_split_size(item: Chunk, num_gpus: Optional[int] = None, dim=0) -> int:
+        """
+        Estimate the number of chunks we should split the batch into.
+
+        Uses some silly heuristics.
+        """
+        if num_gpus is None:
+            num_gpus = torch.cuda.device_count()  # type: ignore
+        if isinstance(item, torch.Tensor):
+            return max(1, item.size(dim) // int(num_gpus * 2))
```
emilydinan: why do we divide by `num_gpus * 2` ??

parlai/utils/torch.py
```diff
@@ -235,3 +238,229 @@ def forward(self, xs):
...
+        if isinstance(item, torch.Tensor):
+            # base case, just split the tensor
+            return list(torch.split(item, split_size, dim))
+        elif isinstance(item, tuple):
+            # We start with Tuple[Tensor] and we return List[Tuple[Tensor]]
+            return list(zip(*(PipelineHelper.split(i, split_size, dim) for i in item)))
+        elif isinstance(item, dict):
+            if item == {}:
+                # terrible edge case: the empty dict. return an infinite list
+                # of empty dicts and we'll figure out its correct size later
```
emilydinan: 😮

parlai/utils/torch.py
```diff
@@ -235,3 +238,229 @@ def forward(self, xs):
...
+        # device_no = layer_no // int(math.ceil(num_layers / num_gpus))
+        # return f'cuda:{device_no}'
+        if layer_no == 0:
+            return 'cuda:0'
+        elif num_layers < num_gpus:
+            return f'cuda:{layer_no}'
+        else:
+            assert isinstance(num_gpus, int)
+            device_no = 1 + (layer_no - 1) // int(
+                math.ceil((num_layers - 1) / (num_gpus - 1))
```
emilydinan: this is a bit confusing to me
say you have 12 layers and 2 gpus.
then layer 0 --> gpu 0 
layer n > 0 --> gpu 1

is this how it is supposed to work?

(also you don't need the cast to int if you are doing a math.ceil)

parlai/utils/torch.py
```diff
@@ -235,3 +238,229 @@ def forward(self, xs):
...
+            if not hasattr(l, '_mp_gpu'):
+                raise RuntimeError(
+                    'You must run PipelineHelper.make_parallel on the ModuleList '
+                    'before you can use iterate_layers_chunks.'
+                )
+
+        # devices maps device_idx -> (device, [(i, layers[i], ...])
+        # for example, if devices is 2 and there are 4 layers, we will have
+        # devices = {
+        #   0: ('cuda:0', [0, 1j]),
```
emilydinan: typo

tests/test_utils_torch.py
```diff
@@ -0,0 +1,188 @@
...
+
+class TestTorchUtils(unittest.TestCase):
+    def test_neginf(self):
+        assert neginf(torch.float32) < -1e15
+        assert neginf(torch.float16) > -1e15
+        assert neginf(torch.float16) < -1e4
+
+    def test_padded_tensor(self):
+        # list of lists
+        lol = [[1, 2], [3, 4, 5]]
```
klshuster: ha nice

tests/test_utils_torch.py
```diff
@@ -0,0 +1,188 @@
...
+        assert isinstance(left, torch.Tensor)
+        assert left.shape == (16, 3)
+        assert isinstance(right, dict)
+        assert len(right) == 2
+        assert 'x' in right
+        assert 'y' in right
+        assert right['x'].shape == (16, 5)
+        assert right['y'].shape == (16, 2)
+
+    def chunk_layer_iterator(self):
```
EricMichaelSmith: What's this for?

tests/test_utils_torch.py
stephenroller: lol that was just moving this function around.

parlai/agents/transformer/polyencoder.py
```diff
@@ -120,7 +120,12 @@ def __init__(self, opt, shared=None):
         self.rank_loss = torch.nn.CrossEntropyLoss(reduce=True, size_average=True)
         if self.use_cuda:
             self.rank_loss.cuda()
+
         self.data_parallel = opt.get('data_parallel') and self.use_cuda
+        self.model_parallel = opt.get('model_parallel') and self.use_cuda
+        if self.data_parallel and self.model_parallel:
+            raise RuntimeError('Cannot combine --model-parallel and --data-parallel')
```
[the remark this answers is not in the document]
stephenroller: Compared to distributed data parallel, it saves a lot. Compared to data parallel, it saves a little bit. Essentially, I don't have to full copies of the model on every GPU. I find that I can push batch size a good 25% when I'm very careful about layer placement.

tests/test_utils_torch.py
stephenroller: Unwritten unit test. It WILL become the workitem iterator.

parlai/utils/torch.py
```diff
@@ -235,3 +238,229 @@ def forward(self, xs):
...
+        # time.
+        num_chunks = len(chunks)
+        for l in layers:
+            if not hasattr(l, '_mp_gpu'):
+                raise RuntimeError(
+                    'You must run PipelineHelper.make_parallel on the ModuleList '
+                    'before you can use iterate_layers_chunks.'
+                )
+
+        # devices maps device_idx -> (device, [(i, layers[i], ...])
```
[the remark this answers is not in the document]
stephenroller: I've simplified this actually, need to update the comment.

parlai/utils/torch.py
```diff
@@ -235,3 +238,229 @@ def forward(self, xs):
...
+    For a tutorial on model parallelism, as it's implemented in parts of ParlAI,
+    see https://pytorch.org/tutorials/intermediate/model_parallel_tutorial.html.
+    """
+
+    @staticmethod
+    def guess_split_size(item: Chunk, num_gpus: Optional[int] = None, dim=0) -> int:
+        """
+        Estimate the number of chunks we should split the batch into.
+
+        Uses some silly heuristics.
```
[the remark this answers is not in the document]
stephenroller: This method assumes that ALL the tensors in the collection have a dimension 0 of batchsize. So if I have a dict of tensors, I want to return a list of dicts of smaller tensors. If I have a tuple of tensors, I want to return a list of tuples of smaller tensors. If I have just a tensor, return a list of tensors.

The main heuristic is to try to guess the number of chunks to split up into, to minimize the bubbling and idle GPU time. It's difficult to know, so I've made some (what I think) are reasonable guesses.

parlai/utils/torch.py
stephenroller: That's the heuristic. :) I've found it by just toying. Really it should be tuned per-model, but I'd rather have something that works roughly fine out of the box for all models.

image

The gist is that we want to minimize the size of this bubble, while not running out of memory. If you use the same number of chunks as GPUS, it's fairly straightforward to see that you will have a large bubble. If you chunk via 2x the number of GPUs, then the bubble becomes smaller, but I stretch the time dimension out a lot. So there's sorta a fine balance. In @mrshenli's excellent tutorial, he shows this graph, but it's specific to a single model:

image

dexterju:
Could you elaborate a bit on Layer placement? I am not sure I got your idea that  ```We can account for this and more heavily weight layers in later positions to balance it out```, so you mean fft size / number of parameters will be different in each layer?

parlai/agents/transformer/modules.py
```diff
@@ -21,14 +21,16 @@
...
 import torch.nn as nn
 import torch.nn.functional as F
 
 from parlai.core.torch_generator_agent import TorchGeneratorModel
 from parlai.utils.misc import warn_once
-from parlai.utils.torch import neginf
+from parlai.utils.torch import neginf, PipelineHelper
 
 try:
+    raise ImportError
```
stephenroller: drop this

stephenroller: Test with newer apex, or try to protect it.

parlai/agents/transformer/polyencoder.py
```diff
@@ -120,7 +120,12 @@ def __init__(self, opt, shared=None):
         self.rank_loss = torch.nn.CrossEntropyLoss(reduce=True, size_average=True)
         if self.use_cuda:
             self.rank_loss.cuda()
+
         self.data_parallel = opt.get('data_parallel') and self.use_cuda
+        self.model_parallel = opt.get('model_parallel') and self.use_cuda
```
stephenroller: move this to torchagent I think

parlai/agents/transformer/polyencoder.py
```diff
@@ -129,6 +134,9 @@ def __init__(self, opt, shared=None):
             if shared is None:
                 self.model = torch.nn.DataParallel(self.model)
 
+        if self.model_parallel and shared is None:
```
stephenroller: likewise

parlai/agents/transformer/transformer.py
```diff
@@ -206,12 +212,17 @@ def add_cmdline_args(cls, argparser):
     def __init__(self, opt, shared=None):
         super().__init__(opt, shared)
         self.data_parallel = opt.get('data_parallel') and self.use_cuda
+        self.model_parallel = opt.get('model_parallel', False) and self.use_cuda
+        if self.data_parallel and self.model_parallel:
```
stephenroller: likewise

parlai/utils/torch.py
emilydinan: nice, thanks for detailed explanation!

parlai/agents/transformer/modules.py
```diff
@@ -555,6 +553,22 @@ def forward(self, input, positions=None, segments=None):
...
             )
 
+    def _apply_model_parallel(self, tensor, mask):
+        """
+        Pipeline application of model parallelism.
+        """
+        chunks = PipelineHelper.split((tensor, mask))
+        work_items = PipelineHelper.schedule_work_items(self.layers, chunks)
+
+        for chunk_idx, layer_nos, next_device in work_items:
```
emilydinan: this is very clean

parlai/core/torch_ranker_agent.py
```diff
@@ -167,6 +172,12 @@ def __init__(self, opt: Opt, shared=None):
         opt['rank_candidates'] = True
         super().__init__(opt, shared)
 
+        self.model_parallel = opt.get('model_parallel', False) and self.use_cuda
+        self.data_parallel = opt.get('data_parallel', False) and self.use_cuda
```
emilydinan: can we just put this code in TorchAgent instead of in TorchRankerAgent and TorchGeneratorAgent?

parlai/utils/torch.py
```diff
@@ -235,3 +237,305 @@ def forward(self, xs):
...
+        assert isinstance(submodule, torch.nn.ModuleList)  # for typechecker
+        layers = submodule
+
+        # mark this section as MP
+        layers.is_model_parallel = True  # type: ignore
+
+        # next, let's figure out how many parameters we can assign to each GPU,
+        # but not make actual assignments yet. Assignments come later because we
+        # want consecutive layers to be collocated
+        keyfunc = self.__device_allocations.__getitem__
```
emilydinan: fancy

parlai/utils/torch.py
```diff
@@ -235,3 +237,305 @@ def forward(self, xs):
...
+
+        assert isinstance(submodule, torch.nn.ModuleList)  # for typechecker
+        layers = submodule
+
+        # mark this section as MP
+        layers.is_model_parallel = True  # type: ignore
+
+        # next, let's figure out how many parameters we can assign to each GPU,
+        # but not make actual assignments yet. Assignments come later because we
+        # want consecutive layers to be collocated
```
emilydinan: can you explain a bit more why we don't do the assignment at the same time?

stephenroller: Essentially first pass tries to pack all the groups, and second pass assigns the groups to the layers. I couldn’t figure out how to do it in one pass.

parlai/agents/transformer/modules.py
```diff
@@ -535,8 +540,15 @@ def forward(self, input, positions=None, segments=None):
         tensor = self.dropout(tensor)
 
         tensor *= mask.unsqueeze(-1).type_as(tensor)
-        for i in range(self.n_layers):
-            tensor = self.layers[i](tensor, mask)
+
+        if getattr(self.layers, 'is_model_parallel', False):
+            # factored out for readability. It is equivalent the other
```
klshuster: grammar nit: "equivalent to"

parlai/core/torch_classifier_agent.py
```diff
@@ -161,16 +161,16 @@ def __init__(self, opt: Opt, shared=None):
                 )
             if self.use_cuda:
                 self.model.cuda()
+                if self.data_parallel:
+                    self.model = torch.nn.DataParallel(self.model)
+                if self.model_parallel:
```
klshuster: just to make it abundantly clear, might it be good to have it be an `elif` here? since it is _never_ the case that both of these are applied?

parlai/scripts/distributed_train.py
```diff
@@ -61,7 +61,10 @@ def main():
         )
         main_host = hostnames.split()[0].decode('utf-8')
         distributed_rank = int(os.environ['SLURM_PROCID'])
-        device_id = int(os.environ['SLURM_LOCALID'])
+        if opt.get('model_parallel'):
+            device_id = -1
```
klshuster: what does setting this to `-1` do exactly?

parlai/utils/torch.py
```diff
@@ -235,3 +238,308 @@ def forward(self, xs):
...
+        keyfunc = self.__device_allocations.__getitem__
+        layer_assignments = {k: 0 for k in self.devices}
+        for layer_no, layer in enumerate(layers):
+            if layer_no == 0:
+                # hard code the first layer to be 0.
+                mostfree = 'cuda:0'
+            else:
+                # otherwise dynamic allocation
+                mostfree = min(self.devices, key=keyfunc)
+            self.__device_allocations[mostfree] += trainable_parameters(layer) * 32
```
klshuster: where does 32 come from?

stephenroller: totally made up arbitrary number

parlai/scripts/distributed_train.py
stephenroller: left a comment

parlai/utils/torch.py
stephenroller: left a comment

PR #2487, opened 2020-03-19
Fix a GPU leak

stephenroller:
**Patch description**
When working with models that barely fit on the GPU, we can easily OOM when we load up the final 
validation model. The memory of the agent was not clearing correctly.

This insidious bug was due to `lru_cache` keeping a copy of the the world in memory, and therefore the agent, and therefore the model.

Fix is to manually cache the thing that was bound by the LRU cache, and manually clear the memory before we load validation.

**Testing steps**
Loading a large model before and after this fix. CI.

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -379,7 +379,6 @@ class DialogPartnerWorld(World):
             self.total_exs += metrics['exs'].value()
         return metrics
 
-    @lru_cache(maxsize=1)
     def num_examples(self):
         """
         Return number of examples.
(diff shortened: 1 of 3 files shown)


parlai/core/worlds.py
```diff
@@ -1424,12 +1422,14 @@ def getID(self):
         """
         return self.inner_world.getID()
 
-    @lru_cache(maxsize=1)
     def num_examples(self):
         """
         Return the number of examples.
         """
-        return self.inner_world.num_examples()
+        if hasattr(self, '_num_examples'):
```
klshuster: i definitely prefer this logic anyway, the `lru_cache` seemed a little much for its purpose

PR #2492, opened 2020-03-20
Test Interactive and Self-Chat

stephenroller:
**Patch description**
We have a big hole in our testing around interactive and self chat talks. This creates some new tests for this.

A few changes:
- Support setting epoch_done in `local_human`, to signal to the world "hey, we're done here."
- Switch from `repeat_label` to `fixed_response` when collecting the personas. By using a `None` response, I skip the evaluation metrics, which gives a nice speed up
- Just use a fixed response for self chat. In the future, I think I would like to test the interactive mode of a few different agents (cc @jaseweston).

**Testing steps**
CI of new unit tests. Hopefully we see some good improvements in code coverage.

### DIFF
--- a/tests/test_interative.py
+++ b/tests/test_interative.py
@@ -1,11 +1,13 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright (c) Facebook, Inc. and its affiliates.
 # This source code is licensed under the MIT license found in the
 # http://google.com
 # LICENSE file in the root directory of this source tree.
 
-"""Tests that interactive.py is behaving well."""
+"""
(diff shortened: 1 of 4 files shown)


tests/test_interative.py
```diff
@@ -1,8 +1,59 @@
...
+        if self.turn <= self.max_turns:
+            return f'Turn {self.turn}'
+        else:
+            return '[EXIT]'
+
+
+class TestInteractive(unittest.TestCase):
+    def setUp(self):
+        # override input() with our fake, deterministic output
+        patcher = mock.patch('builtins.input', new=FakeInput())
```
emilydinan: this is pretty cool

PR #2493, opened 2020-03-20
[Chat Service] Consolidation of Utils

klshuster:
**Patch description**
The majority (95%) of this diff is the consolidation and reorganization of chat-service-related utils. Similarly to how we have `parlai/utils/...`, there is now `parlai/chat_service/utils/...` with respective utility functions. I pulled in a variety of utility classes, etc. from internal tasks [REDACTED] and [REDACTED] to clean up where these tasks are getting their utilities.

Changes in detail:
- Added type checking to `chat_service_manager.py`
- Created following utils files in `parlai/chat_service/utils/`:
  - `config.py`
    - Deals with parsing config files
  - `image.py`
    - Certain objects/classes for tracking image information, image licensing, image feature generation
  - `logging.py`
    - Deals with logging
  - `misc.py`
    - Various utils that do not fit anywhere else
  - `server.py`
    - Moved from `core`
  - `socket.py`
    - Has the `ChatServiceWebSocket` code
  - `timeout.py`
    - Utilities surrounding timeouts
- Deleted `parlai/chat_service/core/shared_utils.py`
- Moved `parlai/chat_service/core/server_utils.py` to `parlai/chat_service/utils/server.py`
- Deleted `parlai/chat_service/core/manager_utils.py`
  - Moved `ChatServiceWebSocket` to `utils/socket.py` above
  - Moved `ChatServiceWorldRunner` to `core/world_runner.py
- **The way models are specified in configs is now different; there is a top-level `models` parameter in `opt`, which is a dict mapping `model_name` to model args (e.g. model file, etc.). Then, a model is passed to a `ChatServiceWorld` via `opt[‘shared_bot_params’][<model_key>]`. This reflects internal changes.**
  - I updated the various task configs for the demo tasks to reflect this.
- Finally, I made the `OffensiveLanguageClassifier` shareable

**Testing steps**
Will mainly involve manually running the public tasks (chatbot, overworld_demo, qa_collection), as well as [REDACTED] and [REDACTED] to ensure that nothing is _broken_

One day I will incorporate a full testing suite. Today is not that day.

- [x] Overworld Demo
  - [x] messenger
  - [x] browser
  - [x] terminal chat
- [x] Chatbot
  - [x] messenger
  - [x] browser
  - [x] terminal chat
- [x] QA Collection
  - [x] messenger
  - [x] browser
  - [x] terminal chat
- [ ] [REDACTED TASK 1]
  - [ ] messenger
  - [ ] browser
  - [ ] terminal chat
- [x] [REDACTED REFACTORED TASK 1]
  - [x] messenger
  - [x] browser (ish)
  - [x] terminal chat (this actually works! just not when [REDACTED])

**Logs**
```
TODO
```

**Later TODO**
There is an issue with browser chat that it'll only send one message from the world at a time; i.e., if a world tries sending multiple messages to a person messaging the page, the person only sees the first. I cannot figure out what is going wrong here.


stephenroller:
redacted

parlai/chat_service/core/socket.py
```diff
@@ -0,0 +1,173 @@
...
+        self.keep_running = True
+        self._setup_socket()
+
+    def _safe_send(self, data, force=False):
+        if not self.alive and not force:
+            # Try to wait a second to send a packet
+            timeout = 1
+            while timeout > 0 and not self.alive:
+                time.sleep(0.1)
+                timeout -= 0.1
```
emilydinan: interesting

klshuster: hmm i didnt even write this lol

JackUrb: Yeaaah, I wrote it before learning about threading events. A better implementation would be to have `self.alive` be a `threading.Event` and call `self.alive.wait(1)` here, but there's no _need_ to update this as it would also require all `self.alive` checks to expect an `Event` rather than a bool.

parlai/chat_service/utils/misc.py
```diff
@@ -0,0 +1,240 @@
...
+import math
+import re
+from enum import Enum
+
+THREAD_MEDIUM_SLEEP = 0.3
+
+
+def set_is_debug(is_debug):
+    global debug
+    debug = is_debug
```
JackUrb: This really belongs to logging utils, it's a bit strange to set a global from a different file.

klshuster: ahh great point i'll move in a later PR

PR #2497, opened 2020-03-24
Interactive base world + some fixes

jaseweston:
- get_contexts seemed to have missing param sometimes, sometimes not, breaking stuff (change by spencer i believe)

- issue: seems a few of the contexts had an extra '\n' at the end. (This might confuse the model about which turn is next, blended_self_talk had this problem in interactive too, I pushed separately.)

### DIFF
--- a/parlai/tasks/convai2/worlds.py
+++ b/parlai/tasks/convai2/worlds.py
@@ -12,7 +12,6 @@ from parlai.agents.fixed_response.fixed_response import FixedResponseAgent
 from parlai.tasks.self_chat.worlds import SelfChatWorld as SelfChatBaseWorld
 
 import random
-from typing import List
 
 
 def load_personas(opt):
--- a/parlai/tasks/convai2/worlds.py
+++ b/parlai/tasks/convai2/worlds.py
@@ -44,8 +44,8 @@ def load_personas(opt):
                     )
                 if t.startswith('your persona:'):
                     a2_persona += t + '\n'
-                    personas.add(a1_persona.rstrip('\n'))
-                    personas.add(a2_persona.rstrip('\n'))
+            personas.add(a1_persona.rstrip('\n'))
+            personas.add(a2_persona.rstrip('\n'))
     print('[ loaded ' + str(len(personas)) + ' personas ]')
     return list(personas)
 
--- a/parlai/tasks/convai2/worlds.py
+++ b/parlai/tasks/convai2/worlds.py
@@ -38,7 +38,7 @@ def load_personas(opt):
             a2_persona = []
             for t in txt:
                 if t.startswith("partner's persona:"):
-                    a1_persona.appen(t.replace("partner's persona:", 'your persona:'))
+                    a1_persona.append(t.replace("partner's persona:", 'your persona:'))
                 if t.startswith('your persona:'):
                     a2_persona.append(t)
             personas.add('\n'.join(a1_persona))
--- a/parlai/tasks/convai2/worlds.py
+++ b/parlai/tasks/convai2/worlds.py
@@ -7,9 +7,10 @@
 from copy import deepcopy
 
 from parlai.core.worlds import create_task
-from parlai.core.worlds import DialogPartnerWorld, validate
+from parlai.core.worlds import validate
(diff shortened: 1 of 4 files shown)


parlai/tasks/convai2/worlds.py
```diff
@@ -44,8 +43,8 @@ def load_personas(opt):
                     )
                 if t.startswith('your persona:'):
                     a2_persona += t + '\n'
-            personas.add(a1_persona)
-            personas.add(a2_persona)
+            personas.add(a1_persona.rstrip('\n'))
```
emilydinan: why do we strip the new line here? is it becuase there is an extra newline at the end? maybe we shouldj ust edit the way we grab the personas. something like 
```
a1_persona = []
...
    a1_persona.append(t.replace("partner's persona:", "your persona:")
...
personas.add('\n'.join(a1_persona)

```

parlai/tasks/interactive/worlds.py
```diff
@@ -16,4 +18,57 @@ class InteractiveWorld(DialogPartnerWorld):
...
+        """
+        pass
+
+    def get_contexts(self):
+        """
+        Override to return a pair of contexts with which to seed the episode.
+
+        This function will be called before the first turn of every episode.
+        """
+        return ['__SILENCE__', '']
```
emilydinan: do we want this to be the default? not all datasets have a silence token?

parlai/tasks/interactive/worlds.py
```diff
@@ -16,4 +18,57 @@ class InteractiveWorld(DialogPartnerWorld):
...
+            self.p1, self.p2 = self.get_contexts()
+
+        acts = self.acts
+        agents = self.agents
+        if self.turn_cnt == 0:
+            # add the context on to the first message to agent 0
+            act = {}
+            act['text'] = self.p1
+            act['episode_done'] = False
+            act['id'] = 'context'
```
emilydinan: do we always want to add a contexT? i'm not sure about this 

cc @stephenroller

parlai/tasks/light_dialog/worlds.py
```diff
@@ -72,48 +66,11 @@ def get_new_personas(self):
...
-        self.cnt += 1
-
-        if act['episode_done']:
-            print("CHAT DONE ")
-            print("\n... preparing new chat... \n")
-            self.cnt = 0
-
 
-class InteractiveSelfchatWorld(SelfChatBaseWorld):
+class SelfChatWorld(SelfChatBaseWorld):
```
emilydinan: missed this one -- thanks!

parlai/tasks/convai2/worlds.py
jaseweston: sure, that would do the same thing

parlai/tasks/interactive/worlds.py
jaseweston: yes, i can remove this.

parlai/tasks/interactive/worlds.py
jaseweston: hmm, i guess i could just add if it isn't an empty string..

parlai/tasks/convai2/worlds.py
jaseweston: (but yes, the extra newline was a bug)

parlai/tasks/interactive/worlds.py
stephenroller: Does it make more sense to have get_context return something other than an empty string for that? E.g. it returns None instead?

This is fine i think, but maybe the semantics could be improved.

parlai/tasks/convai2/worlds.py
```diff
@@ -4,15 +4,12 @@
...
-from copy import deepcopy
-
 from parlai.core.worlds import create_task
-from parlai.core.worlds import DialogPartnerWorld, validate
 from parlai.agents.fixed_response.fixed_response import FixedResponseAgent
 from parlai.tasks.self_chat.worlds import SelfChatWorld as SelfChatBaseWorld
+from parlai.tasks.interactive.worlds import InteractiveWorld as InteractiveBaseWorld
 
 import random
-from typing import List
```
stephenroller: lol oh no, even less typing?!

PR #2504, opened 2020-03-25
Light task: small mods

jaseweston:
- make task name make it clear which label task, this makes eval able to differentiate the different tasks when multitasking
- in the new 4 way multitasking variant we don't give the task name in the input text to the model, so we now have a flag to turn that off


jaseweston:
> Just to be sure I see what's going on here - now the task labels are required in the LIGHT dialogue/action teachers, but the label is optional for the interactive world for the model?

-t light_dialog:simple_multi    doesn't use the task in the text (so we can easily multitask with 1 context)
-t light_dialog:simple (that was from before) still uses it

the interactive has a flag to switch

Issue #1977, opened 2019-08-28
[TA] report gnorm even when no clipping


stephenroller:
Presently, gradient norms are only reported in TA when `--gradient-clip` is > 0.

https://github.com/facebookresearch/ParlAI/blob/c1e35bac34e7336b794f3ac48b100f9b17d14e86/parlai/core/torch_agent.py#L1800-L1807

https://github.com/facebookresearch/ParlAI/blob/c1e35bac34e7336b794f3ac48b100f9b17d14e86/parlai/core/torch_agent.py#L1005-L1006

However, this information is always useful. In the case of gradient clip being <= 0, the norm should be computed anyway. See the `torch.nn.utils.clip_grad_norm_` code to see how this is done.

PR #2505, opened 2020-03-25
Add gnorm metric logging when gradient clipping < 0

hadasah:
**Patch description**
Patch for #1977
Show gnorm in metrics even when gradient clipping isn't set

**Testing steps**
Ran train_model with --gradient_clip -1, verified that gnorm is printed.
Command:
`python examples/train_model.py -m transformer/generator -t convai2 -mf /tmp/tg --gradient_clip -1 -lr 1e-10`
Output:
Screen Shot 2020-03-25 at 17 43 04

### DIFF
--- a/parlai/utils/torch.py
+++ b/parlai/utils/torch.py
@@ -229,6 +229,7 @@ def argsort(keys: List[Any], *lists: List[List[Any]], descending: bool = False):
 def compute_grad_norm(parameters, norm_type=2.0):
     """
     Compute norm over gradients of model parameters.
+
     Mostly taken from the body of torch.nn.utils.clip_grad_norm_
 
     :param parameters:
@@ -237,7 +238,6 @@ def compute_grad_norm(parameters, norm_type=2.0):
 
     :param norm_type
         type of p-norm to use
-
     """
     if isinstance(parameters, torch.Tensor):
         parameters = [parameters]
--- a/parlai/utils/torch.py
+++ b/parlai/utils/torch.py
@@ -241,7 +241,7 @@ def compute_grad_norm(parameters, norm_type=2.0):
     """
     if isinstance(parameters, torch.Tensor):
         parameters = [parameters]
-    parameters = list(filter(lambda p: p.grad is not None, parameters))
+    parameters = [p.grad for p in parameters if p is not None]
     total_norm = 0
     for p in parameters:
         param_norm = p.grad.data.norm(norm_type)
--- a/parlai/utils/torch.py
+++ b/parlai/utils/torch.py
@@ -230,12 +230,9 @@ def compute_grad_norm(parameters, norm_type=2.0):
     """
     Compute norm over gradients of model parameters.
 
-    Mostly taken from the body of torch.nn.utils.clip_grad_norm_
-
     :param parameters:
         the model parameters for gradient norm calculation. Iterable of
         Tensors or single Tensor
-
     :param norm_type:
         type of p-norm to use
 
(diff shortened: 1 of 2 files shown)


parlai/utils/torch.py
```diff
@@ -226,6 +226,29 @@ def argsort(keys: List[Any], *lists: List[List[Any]], descending: bool = False):
...
+        the model parameters for gradient norm calculation. Iterable of
+        Tensors or single Tensor
+
+    :param norm_type
+        type of p-norm to use
+
+    """
+    if isinstance(parameters, torch.Tensor):
+        parameters = [parameters]
+    parameters = list(filter(lambda p: p.grad is not None, parameters))
```
stephenroller:
```suggestion
    parameters = [p.grad for p in parameters if p is not None]
```

parlai/utils/torch.py
```diff
@@ -226,6 +226,29 @@ def argsort(keys: List[Any], *lists: List[List[Any]], descending: bool = False):
...
+
+    :param parameters:
+        the model parameters for gradient norm calculation. Iterable of
+        Tensors or single Tensor
+
+    :param norm_type
+        type of p-norm to use
+
+    """
+    if isinstance(parameters, torch.Tensor):
```
stephenroller: I don't understand this one

hadasah: torch.nn.utils.clip_grad_norm_ works in the case where it's passed a list of tensors, as well as a single tensor of parameters. Since I don't think we'll be needing that case (?) I can just get rid of that

parlai/utils/torch.py
hadasah: oops I forgot to change this, thanks!

parlai/utils/torch.py
stephenroller: Oh, that's cool. I like the mirroring, so you can leave it if you want.

Do you know what fixed the tests?

parlai/utils/torch.py
```diff
@@ -226,6 +226,29 @@ def argsort(keys: List[Any], *lists: List[List[Any]], descending: bool = False):
     return output
 
 
+def compute_grad_norm(parameters, norm_type=2.0):
+    """
+    Compute norm over gradients of model parameters.
```
stephenroller: It wants an extra newline between this and the next line, and it wants you to delete the extra empty line at the end.

parlai/utils/torch.py
```diff
@@ -476,9 +501,13 @@ def chunk_to(chunk: Chunk, device: str) -> Chunk:
         if isinstance(chunk, torch.Tensor):
             return chunk.to(device)  # type: ignore
         elif isinstance(chunk, tuple):
-            return tuple(PipelineHelper.chunk_to(c, device) for c in chunk)  # type: ignore
+            return tuple(
+                PipelineHelper.chunk_to(c, device) for c in chunk
+            )  # type: ignore
         elif isinstance(chunk, dict):
-            return {k: PipelineHelper.chunk_to(v, device) for k, v in chunk.items()}  # type: ignore
```
stephenroller: please don't change any of these?

hadasah:
> I wonder if there’s a test we could write....

I could write a function in test_utils_torch.py just for testing the compute_grad_norm() function but I don't think we currently test for which metrics get printed out during training (do we? I didn't find it but maybe I wasn't looking in the right place)

stephenroller:
Yeah that's correct. Go ahead and merge.

What was wrong with the docs?

hadasah:
it didn't like the reference to torch.nn.utils.clip_grad_norm_ - I think it tries to link to the functioin so you can only reference functions within ParlAI?

PR #2511, opened 2020-03-28
More efficient model parallel transfer to GPU.

stephenroller:
**Patch description**
1. Modify model parallel so that there is no intermediary step of putting all weights on one GPU. This is helpful when the model weights are too big for one GPU but fine when distributed. This is somewhat exacerbated because we don't half() the model until after initializing.
2. Add some new tests for DataParallel.

**Testing steps**
New CI tests. Manual tests with a large model.


parlai/core/torch_classifier_agent.py
```diff
@@ -163,11 +163,12 @@ def __init__(self, opt: Opt, shared=None):
                 print('Loading existing model parameters from ' + init_model)
                 self.load(init_model)
             if self.use_cuda:
-                self.model.cuda()
+                if self.model_parallel:
```
EricMichaelSmith, line 166: Since this code seems to now be repeated in TCA, TGA, and TRA, why not move to a new method in TA?

parlai/utils/torch.py
```diff
@@ -341,8 +341,15 @@ def make_parallel(self, model: torch.nn.Module) -> torch.nn.Module:
         self.__device_allocations['cuda:0'] += trainable_parameters(model) * 3
 
         model.apply(self._place_modulelist)
+        model._apply(self._move_rest_to_cuda0)
         return model
 
+    def _move_rest_to_cuda0(self, parameter):
+        if parameter.device.type == 'cpu':
+            return parameter.to('cuda:0')
```
EricMichaelSmith, line 349: Hm can you walk me through this? This looks like it is moving at least some weights just to GPU 0

parlai/core/torch_classifier_agent.py
stephenroller: Yeah, I do think there's a refactor incoming. I've wanted to do it for a while, but subtle differences kept me from being confident in it. I'd like a lot less custom init code in each of those classes...

parlai/utils/torch.py
stephenroller: Yes, it's moving everything that wasn't explicitly allocated to another GPU already to GPU0. The way I implemented model parallel was that I shard the layers (unevenly) across all the GPUs, but left things like the embedding weights on GPU0 (which is why I was uneven). Since the goal was to avoid initializing everything on one GPU, I move everything special around first, and then search for anything leftover.

PR #2512, opened 2020-03-28
Fix loading optimizer state from disk when using --fp16-impl mem_eff

stephenroller:
**Patch description**
Fixes a bug when loading memory efficient fp16 optimizers off of disk. The memory efficient fp16 optimizer keeps fp32 states for the weights, and hacks around a pytorch cast by keeping the original state dict. Since the original state dict was loaded on CPU, this would cause a mismatch of devices. Anything that changed the optimizer, switched to fp32, or switched to apex fp16, wasn't affected, since they didn't have this special code path resuming tensors.

This patch ensures that all tensors of the state are moved to the same device as the corresponding parameter weights.

Also prevents a crash on switching between memeff and apex fp16. This is a bit more finicky, but at least it won't crash now.

**Testing steps**
Manual testing with a command that @jaseweston gave. New integration tests that failed before this patch, and passes after.

### DIFF
--- a/parlai/utils/fp16.py
+++ b/parlai/utils/fp16.py
@@ -401,8 +401,14 @@ class MemoryEfficientFP16Optimizer(torch.optim.Optimizer):
         }
         for k, v in state_dict['state'].items():
             if k in id_map:
+                # make sure when we copy the original state back, we make sure
+                # that original state is on the correct device
                 param = id_map[k]
-                self.optimizer.state[param] = v
+                like_device_v = {
+                    j: w.to(param.device) if torch.is_tensor(w) else w
+                    for j, w in v.items()
+                }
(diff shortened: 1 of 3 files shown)


parlai/utils/fp16.py
```diff
@@ -401,8 +401,14 @@ def load_state_dict(self, state_dict):
         }
         for k, v in state_dict['state'].items():
             if k in id_map:
+                # make sure when we copy the original state back, we make sure
+                # that original state is on the correct device
                 param = id_map[k]
-                self.optimizer.state[param] = v
+                like_device_v = {
+                    j: w.to(param.device) if torch.is_tensor(w) else w
```
EricMichaelSmith: :+1:

PR #2518, opened 2020-03-30
Speed up some unit tests

stephenroller:
**Patch description**
- I noticed that deepcopying opts was taking as much as 50% of some of our tests, especially ones like dynamic batching, where we create so many world copies. This is because we deep copy the history and the list of deepcopies, and this just grows and grows. We can make this cheaper by not having opt.history be a dict of mutable lists.
- Various small hyperparameter changes in some tests, shortcuts, skip_generation, batch candidates, etc.

**Testing steps**
CI

### DIFF
--- a/parlai/core/opt.py
+++ b/parlai/core/opt.py
@@ -24,12 +24,12 @@ class Opt(dict):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.history = {}
+        self.history = []
         self.deepcopies = []
 
     def __setitem__(self, key, val):
         loc = traceback.format_stack()[-2]
-        self.history.setdefault(key, []).append((loc, val))
+        self.history.append((key, val, loc))
         super().__setitem__(key, val)
 
     def __getstate__(self):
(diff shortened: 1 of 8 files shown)


parlai/core/opt.py
```diff
@@ -49,41 +49,37 @@ def __deepcopy__(self, memo):
...
-        # deepcopy the dict
-        memo = copy.deepcopy(dict(self))
-        # make into Opt object
-        memo = Opt(memo)
-        # deepcopy the history
-        memo.history = copy.deepcopy(self.history)
-        # deepcopy the deepcopy history
-        memo.deepcopies = copy.deepcopy(self.deepcopies)
+        # copy all our children
+        memo = Opt({k: copy.deepcopy(v) for k, v in self.items()})
```
emilydinan: curious -- why is this faster than just deepcopying the dictionary?

stephenroller: Bc you were making 2 shallow copies before the deepcopy.

stephenroller: (The real savings is changing the history to be a shallow copy, the rest is fairly micro optimizations)

PR #2519, opened 2020-03-30
Only initialize optimizer if needed.

stephenroller:
**Patch description**
Only initialize the optimizer if we need it for training. Otherwise, we can save a lot of GPU memory and get larger batch sizes.

**Testing steps**
CI.


jaseweston:
yes!

PR #2520, opened 2020-03-30
Implement train:evalmode for ED teacher

EricMichaelSmith:
**Patch description**
`train:evalmode` is not currently compatible with the ED teacher because the teacher defines its own `self.datatype` attribute that strips away the `:evalmode` suffix instead of using the (correct) datatype string as defined by `FixedDialogTeacher`. This change removes this redefinition of `self.datatype` by the ED teacher and simply deletes the `:evalmode` suffix wherever is not desired.

**Testing steps**
- Checked that `python -m tests.tasks.test_empathetic_dialogues` passes
- Checked that `python tests/datatests/test_new_tasks.py` passes


EricMichaelSmith:
> I like the name base datatype. I wish we could get rid of these altogether.
> 
> I don't understand how this diff accomplished your goal tho. is self.datatype set in the teacher?

Thanks @stephenroller ! Yeah, self.datatype is set in the superclass at https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/teachers.py#L241

PR #2522, opened 2020-03-31
Safe interactive mode

jaseweston:
- depends on the color PR, it only actually changes local_human.py
- yes, I could make a new agent or new interactive script -- is it worth it?

right now you can just do:

py examples/interactive.py -m fixed_response --safety all

to make it work


jaseweston:
actually ... i might have changed my mind... i think safe_interactive might be better.. let me try dat

PR #2523, opened 2020-04-01
safe interactive mode

jaseweston:


parlai/agents/safe_local_human/safe_local_human.py
```diff
@@ -0,0 +1,94 @@
...
+        if opt['safety'] == 'string_matcher' or opt['safety'] == 'all':
+            self.offensive_string_matcher = OffensiveStringMatcher()
+        if opt['safety'] == 'classifier' or opt['safety'] == 'all':
+            self.offensive_classifier = OffensiveLanguageClassifier()
+        self.self_offensive = False
+
+    def offensive(self, text):
+        if hasattr(
+            self, 'offensive_string_matcher'
+        ) and self.offensive_string_matcher.__contains__(text):
```
klshuster, line 50: nit: note that you shouldn't  really call `__contains__` directly, you can actually just write:

```py
text in self.offensive_string_matcher
```

parlai/agents/safe_local_human/safe_local_human.py
```diff
@@ -0,0 +1,94 @@
...
+            return True
+        if hasattr(
+            self, 'offensive_classifier'
+        ) and self.offensive_classifier.__contains__(text):
```
klshuster, line 54: ditto here

PR #2527, opened 2020-04-01
[wip] Quantization and microspeedups

stephenroller:
**Patch description**
Work in Progress.

Attempts at doing quantization, combined with a few micro speedups to take any gain we can get.


stephenroller:
cry.

PR #2528, opened 2020-04-01
Raise errors when teachers have overlapped ids

dexterju:
**Patch description**
Two task teachers might have the same id. And this causes problems for metrics aggregation, since we use teacher.id as keys for that. 
Having overlapped teacher.id will lead metrics reports getting overwritten. 
**Testing steps**
 py examples/train_model.py -t integration_tests,integration_tests -mf /tmp/hkhk

**Logs**
```
  File "examples/train_model.py", line 17, in <module>
    TrainLoop(opt).train()
  File "/private/home/daju/ParlAI/parlai/scripts/train_model.py", line 277, in __init__
    build_dict(opt, skip_if_built=True)
  File "/private/home/daju/ParlAI/parlai/scripts/build_dict.py", line 113, in build_dict
    world_dict = create_task(ordered_opt, dictionary)
  File "/private/home/daju/ParlAI/parlai/core/worlds.py", line 1624, in create_task
    world = MultiWorld(opt, user_agents, default_world=default_world)
  File "/private/home/daju/ParlAI/parlai/core/worlds.py", line 624, in __init__
    world_id,
KeyError: "<parlai.tasks.integration_tests.agents.DefaultTeacher object at 0x7f66249b3fd0> and <class 'parlai.tasks.integration_tests.agents.DefaultTeacher'> teachers have overlap in id integration_tests."
```

### DIFF
--- a/tests/test_train_model.py
+++ b/tests/test_train_model.py
@@ -41,7 +41,7 @@ class TestTrainModel(unittest.TestCase):
         task2_acc = valid['integration_tests:multiturnCandidate/accuracy']
         total_acc = valid['accuracy']
         self.assertEqual(
-            total_acc, task1_acc + task2_acc, 'Task accuracy is averaged incorrectly',
+            total_acc, task1_acc + task2_acc, 'Task accuracy is averaged incorrectly'
         )
 
         valid, test = testing_utils.train_model(
@@ -58,7 +58,7 @@ class TestTrainModel(unittest.TestCase):
         total_acc = valid['accuracy']
         # metrics should be averaged equally across tasks
         self.assertEqual(
-            total_acc, task1_acc + task2_acc, 'Task accuracy is averaged incorrectly',
+            total_acc, task1_acc + task2_acc, 'Task accuracy is averaged incorrectly'
         )
 
     def test_multitasking_metrics_macro(self):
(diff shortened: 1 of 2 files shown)


jaseweston:
yes, this is going to crash a bunch of experiments in the short term lol, but probably worth it

stephenroller:
We could probably improve the error message if the error message included the `__class__` of the each of these overlapping teachers.

dexterju:
I can add this along with the test.

tests/test_train_model.py
```diff
@@ -100,6 +100,16 @@ def test_multitasking_metrics_macro(self):
             'Task accuracy is averaged incorrectly',
         )
 
+    def test_multitasking_id_overlap(self):
+        with self.assertRaises(KeyError) as context:
+            testing_utils.train_model(
```
stephenroller, line 105: maybe just create world instead of full blown train model?

tests/test_train_model.py
```diff
@@ -100,6 +100,16 @@ def test_multitasking_metrics_macro(self):
...
+                {'task': 'integration_tests,' 'integration_tests'}
+            )
+            self.assertTrue(
```
stephenroller, line 108: use `with self.assertRaises(AssertionError)`

PR #2534, opened 2020-04-03
Adding passing of Agent.data to ChatService for internal parity

JackUrb:
**Patch description**
Internally we set contents of an agent's data between worlds. Externally we only supported this via `onboard_data`, which lets an onboarding world return data that is put in the agent later. This proved too restrictive for some of our internal chats, and so we added passing of the `agent.data` field. 

This PR introduces the same mechanic to the open source version (essentially being straight better than `onboard_data`), but I've retained the old mechanic to make minimal changes.

**Testing steps**
Tested with the LIGHT worlds, which aren't yet public yet. Suffice to say, I can send data from one world to another world via the agent.data field.

### DIFF
--- a/parlai/chat_service/core/chat_service_manager.py
+++ b/parlai/chat_service/core/chat_service_manager.py
@@ -37,6 +37,7 @@ class AgentState:
         self.active_agent = overworld_agent
         self.task_id_to_agent: Dict[str, ChatServiceAgent] = {}
         self.onboard_data = None
+        self.data = {}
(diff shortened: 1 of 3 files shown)


klshuster:
i wonder if this is what `stored_data` was meant to be...

JackUrb:
Quite unsure?

At some point `chat_service` will definitely deserve a cleanup now that it's the same internally as externally, but I doubt we'll rush into doing that before we find ourselves annoyed with the upgrade so far.

PR #2536, opened 2020-04-06
Fix seq2seq with --fp16 true

stephenroller:
**Patch description**
Tried `--fp16 true` with `-m seq2seq` and noticed an error when making a recommendation for some UMass students. This fixes that bug.

**Testing steps**
Manual testing.

### DIFF
--- a/parlai/agents/seq2seq/modules.py
+++ b/parlai/agents/seq2seq/modules.py
@@ -15,7 +15,7 @@ from torch.nn.parameter import Parameter
 from torch.nn.utils.rnn import pad_packed_sequence, pack_padded_sequence
 import torch.nn.functional as F
 
-from parlai.utils.torch import NEAR_INF
+from parlai.utils.torch import neginf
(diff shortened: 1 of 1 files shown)


parlai/agents/seq2seq/modules.py
```diff
@@ -692,7 +692,7 @@ def forward(self, xes, hidden, attn_params):
             # calculate activation scores, apply mask if needed
             if attn_mask is not None:
                 # remove activation from NULL symbols
-                attn_w_premask.masked_fill_((~attn_mask), -NEAR_INF)
+                attn_w_premask.masked_fill_((~attn_mask), neginf(enc_out.dtype))
```
spencerp, line 695: nit; would it be slightly safer to use `neginf(attn_w_premask.dtype)`? Or is it necessarily the case that we want `enc_out` and `attn_w_premask` to have the same dtype?

stephenroller: Yeah you're right.

PR #2537, opened 2020-04-07
Don't copy pretrained embeddings when evaluating

stephenroller:
**Patch description**
We can speed up evaluation, especially in CI, if we don't perform the expensive download-and-copy pretrained embeddings step during evaluation time.

**Testing steps**
- Long tests should be noticeably faster.
- Code coverage may go down a little bit.


tests/nightly/gpu/test_controllable.py
```diff
@@ -116,7 +117,7 @@ def test_convai2_finetuned_norepetition(self):
                 'beam_min_n_best': 10,
                 'use_reply': 'model',
                 'batchsize': 64,
-                'num_examples': NUM_EXAMPLES,
+                'num_examples': NUM_EXAMPLES // 1,  # this test appears v. slow
```
emilydinan, line 120: what's this for

stephenroller: Leak from testing, thanks for flagging

stephenroller:
The test failure is due to the pytorch 1.5 release that broke master. This represents a significant speed up on most of the GPU tests.

PR #2542, opened 2020-04-08
Allow macroaverages with nonequal number of subreports.

stephenroller:
**Patch description**
I was working on a large multitask setting with a very small batch size in a distributed setting. Very easily, some workers would work on different subsets of tasks, which would mess up this assertion I had left in the macro-averaging code.  So we support the mixed case properly now, but it slightly changes the API.

**Testing steps**
New test. CI. Manual testing.

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -259,20 +259,23 @@ class MacroAverageMetric(Metric):
     AverageMetrics already.
     """
 
-    __slots__ = ('_values',)
+    __slots__ = ('_values', '_names')
 
-    def __init__(self, metrics: List[Metric]) -> None:
+    def __init__(self, metrics: List[Metric], names: List[str]) -> None:
         self._values = metrics
+        self._names = names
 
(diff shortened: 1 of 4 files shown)


parlai/core/metrics.py
```diff
@@ -259,20 +259,23 @@ class MacroAverageMetric(Metric):
     AverageMetrics already.
     """
 
-    __slots__ = ('_values',)
+    __slots__ = ('_values', '_names')
 
-    def __init__(self, metrics: List[Metric]) -> None:
+    def __init__(self, metrics: List[Metric], names: List[str]) -> None:
```
EricMichaelSmith, line 264: Is there any advantage to having two Lists as args instead of one arg that's Dict[str, Metric]? Could be a cleaner way to associate metrics with names

stephenroller: You're right. I was feeling worried about it yesterday for some reason I don't know, but this is much better.

stephenroller:
Fixed a few additional type annotation errors that are unreleated.

PR #2543, opened 2020-04-08
Higher default log time.

stephenroller:
**Patch description**
Just raise the default of the log time a bit.

**Testing steps**
N/A, CI.

### DIFF
--- a/parlai/scripts/build_dict.py
+++ b/parlai/scripts/build_dict.py
@@ -52,7 +52,7 @@ def setup_args(parser=None, hidden=True):
         hidden=hidden,
     )
     dict_loop.add_argument(
-        '-ltim', '--log-every-n-secs', type=float, default=2, hidden=hidden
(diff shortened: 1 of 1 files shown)


parlai/scripts/build_dict.py
```diff
@@ -52,7 +52,7 @@ def setup_args(parser=None, hidden=True):
         hidden=hidden,
     )
     dict_loop.add_argument(
-        '-ltim', '--log-every-n-secs', type=float, default=2, hidden=hidden
+        '-ltim', '--log-every-n-secs', type=float, default=10, hidden=hidden
```
jxmsML, line 55: is there any trick for setting those magic number: 2, 10, 20, etc?

klshuster: i think we usually pick these based on what feels right, no magic tricks 😄

stephenroller: Yes, this one is just a personal preference. 2 seconds feels like logging too often, so I raised it :D

PR #2545, opened 2020-04-09
[Acute Eval] Bug fix 

emilydinan:
**Patch description**
Bug fix for when the number of HITs exceeds the number of conversations to be completed. We `top_up_extra_data` so that the user still has a HIT to complete. However, there was an `extend` instead of an `append` here, causing bugs down the line.


dexterju:
Thanks!!!!!!!!!!

PR #2558, opened 2020-04-13
[TRA] Changes to interactive mode

emilydinan:
**Patch description**
Previous changes to interactive mode automatically set `eval_candidates` to `fixed`. While this is a useful default, it does not apply in all situations. For example, in the knowledge retriever, we want to use `inline` candidates in interactive mode (fix for this included in this diff). [I've also found this to be the case in other examples, particularly with classifiers.]

This also restructures the call to `set_interactive_mode` a bit; before we were setting a bunch of candidate-related variables inside this function, even though they had little to do with "interactive mode."

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -72,6 +72,15 @@ class TorchRankerAgent(TorchAgent):
             'value as --candidates if no flag is given)',
         )
         agent.add_argument(
+            '-icands',
+            '--interactive-candidates',
+            type=str,
+            default='fixed',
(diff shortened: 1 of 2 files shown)


emilydinan:
> The real issue seems that it was difficult to control how interactive mode candidates were being generated, since it would just forcefully override you?

yes. this allows you to change it. the default is still "fixed" but it's not a hard default

emilydinan:
> The real issue seems that it was difficult to control how interactive mode candidates were being generated, since it would just forcefully override you?

yes. this allows you to change it. the default is still "fixed" but it's not a hard default

PR #2561, opened 2020-04-13
[History] Add temporary string

emilydinan:
**Patch description**
Extend history to allow adding a string temporarily to the end of a history. It will be removed when `get_temp_history()` is called on the next turn.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -208,6 +208,7 @@ class History(object):
         self.history_strings = []
         self.history_raw_strings = []
         self.history_vecs = []
+        self.temp_history = None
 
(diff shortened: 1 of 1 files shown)


parlai/core/torch_agent.py
```diff
@@ -259,7 +260,7 @@ def add_reply(self, text):
         # update history vecs
         self._update_vecs(text)
 
-    def update_history(self, obs):
+    def update_history(self, obs, temp_history=None):
```
klshuster, line 263: so `temp_history` is persistent until the next call to `update_history`, right? maybe we can add a small comment to `update_history` to make that more explicit

PR #2562, opened 2020-04-13
Make ConvAI2 worlds faster to share

spencerp:
**Patch description**
Currently, if you duplicate a ConvAI2 world, it will go through the process of loading the personas from disk again. This change adds the personas to the `shared` dict so that they don't need to be reloaded.

**Testing steps/Logs**
Wrote a test to verify the sharing of personas works:
```
spoff:~/ParlAI(convai-share)
λ python -m unittest -v tests.tasks.convai2.test_worlds.TestConvAI2InteractiveWorld 
test_share (tests.tasks.convai2.test_worlds.TestConvAI2InteractiveWorld) ... [ optional arguments: ] 
[  display_ignore_fields: agent_reply ]
[  display_verbose: False ]
[  max_display_len: 1000 ]
[  num_examples: 10 ]
[ Main ParlAI Arguments: ] 
[  batchsize: 1 ]
[  datapath: /tmp/tmpqr8ygco1 ]
[  datatype: train:stream ]
[  download_path: /private/home/spoff/ParlAI/downloads ]
[  dynamic_batching: None ]
[  hide_labels: False ]
[  image_mode: raw ]
[  init_opt: None ]
[  multitask_weights: [1] ]
[  numthreads: 1 ]
[  show_advanced_args: False ]
[  task: convai2 ]
[ ParlAI Model Arguments: ] 
[  dict_class: None ]
[  init_model: None ]
[  model: None ]
[  model_file: None ]
[ ParlAI Image Preprocessing Arguments: ] 
[  image_cropsize: 224 ]
[  image_size: 256 ]
[ ConvAI2 Interactive World: ] 
[  display_partner_persona: True ]
[ Current ParlAI commit: da6a54a0d8e79694654066e3ffc550af6925203a ]
[ Current internal commit: aa65f52c2316d4610102be9213ab1c85611a986c ]
ok

Ran 1 test in 0.160s

OK
```


spencerp:
Seeing some unit test failures. Will address them before asking for a review.

parlai/tasks/convai2/worlds.py
```diff
@@ -76,10 +82,15 @@ def finalize_episode(self):
...
+    def share(self):
+        shared_data = super().share()
+        shared_data['personas_list'] = self.personas_list
+        return shared_data
+
 
 class SelfChatWorld(SelfChatBaseWorld):
     def init_contexts(self):
-        self.personas_list = load_personas(self.opt)
+        self.personas_list = get_personas(self.opt)
```
jaseweston, line 93: why not the same sharing in selfchat worlds?

parlai/core/worlds.py
```diff
@@ -1634,3 +1634,16 @@ def create_task(opt: Opt, user_agents, default_world=None):
         world = BatchWorld(opt, world)
 
     return world
+
+
+def create_task_from_shared(shared_world):
```
jaseweston, line 1639: we don't already have something like this?

spencerp: Semi-surprisingly, no. We just do something like `shared['world_class'](opt, shared=shared)` normally.

parlai/tasks/convai2/worlds.py
spencerp: I was just trying to limit the scope of this PR. But I'd be very happy to add it to SelfChat worlds as well!

parlai/core/worlds.py
stephenroller: I’m wondering world.clone() isn’t better. We have this for Agent but nothing else, and I kinda like clones better

parlai/core/worlds.py
```diff
@@ -150,6 +150,12 @@ def share(self):
         shared_data['agents'] = self._share_agents()
         return shared_data
 
+    def clone(self):
+        """
+        Create a duplicate of the world.
+        """
+        return type(self)(opt=copy.deepcopy(self.opt), agents=None, shared=self.share())
```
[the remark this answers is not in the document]
spencerp: It's the safest thing to do, so that if one world modifies anything in the opt it doesn't modify the content of the other world's opt. Though it costs memory to do so. I'll leave it up to you all who have more context on the tradeoffs that are best for this situation!

spencerp:
Re: context, this came up when working on an internal project.

parlai/core/worlds.py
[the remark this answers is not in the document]
stephenroller: `world.__init__` actually call deepcopy. But we do that for agents too. No guarantee a subchild behaves properly.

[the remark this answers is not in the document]
spencerp: Okay, in that case I'll keep the `deepcopy` here. If the `world.__init__` handles `deepcopy` that's great, and this intermediary copy will be garbage collected. If it doesn't, the `deepcopy` here will provide some extra safety.

PR #2563, opened 2020-04-14
add safety classifier to detect offensive script

jaseweston:
- options to use word list based safety, classifier or both
- prints out more stats

### DIFF
--- a/parlai/scripts/detect_offensive_language.py
+++ b/parlai/scripts/detect_offensive_language.py
@@ -17,7 +17,7 @@ Examples
 from parlai.core.params import ParlaiParser
 from parlai.core.agents import create_agent
 from parlai.core.worlds import create_task
-from parlai.utils.safety import OffensiveStringMatcher
+from parlai.utils.safety import OffensiveStringMatcher, OffensiveLanguageClassifier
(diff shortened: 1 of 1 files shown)


parlai/scripts/detect_offensive_language.py
```diff
@@ -59,39 +69,63 @@ def detect(opt, printargs=None, print_parser=None):
...
+    def report(world, stats):
+        report = world.report()
+        log = {
+            'word_offenses': stats['bad_words_cnt'],
+            'classifier_offenses%': 100
+            * (stats['classifier_offensive'] / stats['total']),
+            'string_offenses%': 100 * (stats['string_offensive'] / stats['total']),
+            'total_offenses%': 100 * (stats['total_offensive'] / stats['total']),
+        }
+        text, log = log_time.log(report['exs'], world.num_examples(), log)
```
emilydinan, line 90: q: what does this log_time.log do? does it make the `log` pretty?

jaseweston: hmm.. not sure.. i think this was there already.. i guess it prints it in some format, yes..!

PR #2571, opened 2020-04-16
bst: no start utterances

jaseweston:


parlai/tasks/blended_skill_talk/worlds.py
```diff
@@ -44,9 +44,10 @@ def load_personas(opt):
         if d['context_dataset'] == 'wizard_of_wikipedia':
             p1 += d['additional_context'] + '\n'
             p2 += d['additional_context'] + '\n'
-        ctxt = d['free_turker_utterance'] + '\n' + d['guided_turker_utterance']
-        p1 += ctxt
-        p2 += ctxt
+        if opt.get('include_initial_utterances', True):
```
EricMichaelSmith, line 47: Hmm - if `--include-initial-utterances` is False, though, `p1` and `p2` will end in trailing newlines now, in the cases where "context_dataset" is WoW or there are personas. Another way to do this could be to build up the context as a list of strings and just do `p1 = '\n'.join(p1_lines)` at the end

jaseweston: good point! fixing..

jaseweston:
better?

parlai/tasks/blended_skill_talk/worlds.py
EricMichaelSmith: Yeah, looks good to me :)

PR #2577, opened 2020-04-20
[BST] add multitask teacher

emilydinan:
**Patch description**
Add multi-task BST teacher for easier training on all "topicifier" teachers.

On a different note: this teacher takes an insane amount of time to build (several hours), given that it is a relatively small dataset. @EricMichaelSmith -- is there a reason we don't just downloaded the already-built data from AWS? Is this simply for the sake of reproducibility?


emilydinan:


.
Ran 1 test in 314.589s

OK

PR #2578, opened 2020-04-21
Catch some common errors in -t fromfile

stephenroller:
**Patch description**
Born from debugging with some novel users, this catches a few things that gave them problems, where we could have loudly failed.

**Testing steps**
New CI tests.

### DIFF
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -1391,18 +1391,18 @@ class ParlAIDialogTeacher(FixedDialogTeacher):
                         f"for you automatically. This is happening on Line {line_no} "
                         f"in {path}. The line is:\n\t{line}"
                     )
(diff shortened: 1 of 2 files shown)


parlai/core/teachers.py
```diff
@@ -1391,6 +1391,18 @@ def _setup_data(self, path):
                         f"for you automatically. This is happening on Line {line_no} "
                         f"in {path}. The line is:\n\t{line}"
                     )
+                if msg and 'text' not in msg:
```
dexterju, line 1394: A thought, image chat / captioning task does not have text, in this case, they should be None or ''?

stephenroller: They probably wouldn't use the fromfile teacher.

PR #2601, opened 2020-04-29
BST: add persona note

jaseweston:


parlai/tasks/blended_skill_talk/worlds.py
```diff
@@ -14,6 +14,10 @@
 
 def load_personas(opt):
     print('[ loading personas.. ]')
+    if opt.get('include_personas', True):
```
emilydinan, line 17: nit, i know this isn't your chagne but: if `not opt.get('include_personas', True)` we shouldn't even waste time loading the personas here.

jaseweston: hmm, yea tru i dont want to mess anything tho

PR #2611, opened 2020-04-30
[TCA] bug fix

emilydinan:
**Patch description**
Thresholding bug fix. I'm unsure why tests did not capture this before, or why this bug is newly introduced.

### DIFF
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -233,7 +233,7 @@ class TorchClassifierAgent(TorchAgent):
         Format interactive mode output with scores.
         """
         preds = []
-        for i, pred_id in enumerate(prediction_id.tolist()):
+        for i, pred_id in enumerate(prediction_id):
             prob = round_sigfigs(probs[i][pred_id], 4)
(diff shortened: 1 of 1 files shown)


parlai/core/torch_classifier_agent.py
```diff
@@ -279,10 +279,11 @@ def eval_step(self, batch):
         probs = F.softmax(scores, dim=1)
         if self.threshold is None:
             _, prediction_id = torch.max(probs.cpu(), 1)
+            prediction_id = prediction_id.tolist()
         else:
             ref_prob = probs.cpu()[:, 0]
             # choose ref class if Prob(ref class) > threshold
-            prediction_id = ref_prob <= self.threshold
+            prediction_id = [int(ref_prob <= self.threshold)]
```
dexterju, line 286: Maybe a dumb question, but will this work when ref_prob is a tensor with more than one element? In that case you cannot cast it to int with this right? You might want .to(torch.int64) instead and kept the rest as the same.

emilydinan: nice catch

PR #2612, opened 2020-04-30
[TGA] Beam blacklist

stephenroller:
**Patch description**
Add the ability to hard-blacklist ngrams in the beam search. Adds a new optional argument that points to a filename. If specified, the filename is expected to be a text file, with one phrase per line. The system tokenizes the strings, so no need to pre-convert to tokens.

**Testing steps**
Manual testing. New test in CI.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -18,7 +18,7 @@ Contains the following utilities:
 """
 
 from abc import ABC, abstractmethod
-from typing import TypeVar, List, Dict
+from typing import TypeVar, List, Dict, Optional, Tuple, Set, Iterable
 import math
 from operator import attrgetter
 
@@ -28,7 +28,7 @@ import torch.nn.functional as F
 
 from parlai.core.opt import Opt
 from parlai.utils.distributed import is_distributed, sync_parameters
-from parlai.core.torch_agent import TorchAgent, Batch, Output
+from parlai.core.torch_agent import TorchAgent, Batch, Output, DictionaryAgent
 from parlai.utils.misc import warn_once
 import parlai.utils.logging as logging
 from parlai.core.metrics import SumMetric, AverageMetric, BleuMetric, FairseqBleuMetric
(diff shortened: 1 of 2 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -954,6 +963,38 @@ def _generate(self, batch, beam_size, max_ts):
...
+    def _load_beam_blacklist(self) -> List[List[int]]:
+        """
+        Load the beam blacklist.
+
+        :return: a dict mapping ngram length to different ngrams
+        """
+        phrases = set()
+        for phrase in self._get_blocked_phrases():
+            phrases.add(phrase)
+            phrases.add(phrase.lower())
```
emilydinan, line 975: nice

parlai/core/torch_generator_agent.py
```diff
@@ -1052,6 +1094,14 @@ def set_context(self: TSType, context: torch.LongTensor) -> TSType:
         self.context = context.tolist()
         return self
 
+    def set_blacklist(self: TSType, blacklist: List[List[int]]) -> TSType:
+        self.blacklist: Dict[int, List[List[int]]] = {}
+        for ngram in blacklist:
```
emilydinan, line 1099: poss. dumb q... why do we do this `set_blacklist` every time the tree search factory is created? don't we only need to create this dictionary once?

parlai/core/torch_generator_agent.py
```diff
@@ -1115,6 +1165,15 @@ def _block_ngrams(
                     logprobs[beam_id][ngram[-1]] = neginf(logprobs.dtype)
         return logprobs
 
+    def _block_blacklist(self, logprobs: torch.Tensor):
+        for beam_id, hyp in enumerate(self.partial_hyps):
+            for ngram_size, bad_ngrams in self.blacklist.items():
+                prefix = hyp[-(ngram_size - 1) :]
+                for ngram in bad_ngrams:
+                    if ngram_size == 1 or prefix == list(ngram[:-1]):
+                        logprobs[beam_id][ngram[-1]] = neginf(logprobs.dtype)
```
emilydinan, line 1174: 👍

parlai/core/torch_generator_agent.py
stephenroller: hmm, ya you're right. i figured it was a more convenient form to keep it in the agent, in case you wanted to dynamically update it. I don't think the O(n) conversion is that bad tho.

emilydinan: i see .. still a bit weird. could create a blacklist object to allow you to dynamically update

klshuster: i agree that we'd want some way to dynamically update the blacklist, though perhaps Emily is right that we'd want to just pass along a shared object rather than rebuild. given that `n` is not gonna be too large here it's prob fine either way

stephenroller:
Updated to include a blacklist object you can dynamically add to.

PR #2613, opened 2020-05-01
Added --host and --port parameter to client.py and interactive_web.py

marcelodiaz558:
**Patch description**
The files client.py and interactive_web.py by default are just handling requests by localhost and port 8080, I added two new parameters: --host to both and --serving_port to client.py, if --host is 0.0.0.0 then the server will run allowing all incoming IPs, the defaults for both parameters are 'localhost' and 8080 respectively, if they're not specified then the behavior will be the same as before.

**Usage:**
python ~/ParlAI/parlai/chat_service/services/browser_chat/client.py --port 8080 --host 0.0.0.0 --serving_port 80

python ~/ParlAI/parlai/scripts/interactive_web.py -t skill -mf modelfile --host 0.0.0.0 --port 80

**Example log**
In this example the server will now allow any IP from the internet and will be listening in port 80
```
~/ParlAI/parlai/chat_service/services/browser_chat$ sudo python3 client.py --port 8081 --serving_port 80 --host 0.0.0.0
[ Browser Chat: ] 
[  host: 0.0.0.0 ]
[  port: 8081 ]
[  serving_port: 80 ]
[ Current ParlAI commit: 066f02d17dc71855733594b6643ea7a8a6b58efb ]
Connecting to port:  8081
Please connect to the link: http://0.0.0.0:80/
```

### DIFF
--- a/parlai/chat_service/services/browser_chat/client.py
+++ b/parlai/chat_service/services/browser_chat/client.py
@@ -132,7 +132,7 @@ def on_close(ws):
 def _run_browser():
     host = opt.get('host', 'localhost')
     serving_port = opt.get('serving_port', 8080)
-    
+
     httpd = HTTPServer((host, serving_port), BrowserHandler)
 
     print('Please connect to the link: http://{}:{}/'.format(host, serving_port))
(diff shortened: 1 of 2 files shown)


stephenroller:
The interactive_web change looks fine to me.

Please run `bash autoformat.sh` for us.

PR #2615, opened 2020-05-01
Use AWS versions of cached EDPersonaTopicifierTeacher data

EricMichaelSmith:
**Patch description**
Use versions of cached EDPersonaTopicifierTeacher datafiles from AWS instead of having to compute and cache the datafile the first time that EDPersonaTopicifierTeacher is used, which takes ~45 min for the train set. Still allow the option of re-compiling the datafile with `--recompile-persona-topic-data True`.

**Testing steps**
- `python tests/tasks/test_blended_skill_talk.py` (EDPersonaTopicifierTeacher test added because it can now run quickly due to these cached files)
- Testing that all files are loaded correctly with `display_data.py`
- Checking that `display_data.py` output is the same for the test set with and without `--recompile-persona-topic-data True`
```
(py37) ems@devfair0229:~/GitHub/facebookresearch/ParlAI$ python tests/datatests/test_new_tasks.py
.
Ran 1 test in 599.938s

OK
```

### DIFF
--- a/parlai/tasks/blended_skill_talk/agents.py
+++ b/parlai/tasks/blended_skill_talk/agents.py
@@ -150,7 +150,7 @@ class EDPersonaTopicifierTeacher(EmpatheticDialoguesTeacher):
             '--recompile-persona-topic-data',
             type='bool',
             default=cls.RECOMPILE_DEFAULT,
-            help='Re-compile data with ConvAI2 personas and WoW topics added',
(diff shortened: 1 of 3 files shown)


EricMichaelSmith:
Thanks @emilydinan ! I've clarified the remove_political_convos error message to say that it removes political conversations from the dataset, and I've specified that there's no need to recompile the data unless you want to see how it was compiled in the first place.

PR #2631, opened 2020-05-05
Remove value error in HuggingFace BPE Tokenizer to work with dict.lower

dexterju:
**Patch description**
The Hugging face bpe tokenizer works with dict.lower. The `ValueError` here should be removed. 
1. The ParlAI hugging face bpe tokenizer will just load and sync its dict with the saved hugging face bpe files. Dict.lower will not create any issue in this case.
2. Based on testing with some models we have already.


dexterju:
Sure, let us do this.

PR #2641, opened 2020-05-07
Use safe personas in interactive BST

EricMichaelSmith:
**Patch description**

By default, only use personas for the interactive BST world that appear on a previously-collected list of safe personas, downloaded from AWS. (The original behavior can be restored with the hidden flag `--safe-personas-only False`.)

**Testing steps**

Counts of numbers of persona pairs after safety filtering:
- Train: 929 removed and 3890 remaining
- Valid: 209 removed and 800 remaining
- Test: 193 removed and 787 remaining

Also see the new unit test for this added to `tests/tasks/test_blended_skill_talk.py`

```
$ python tests/datatests/test_new_tasks.py
.
Ran 1 test in 896.163s

OK
```

### DIFF
--- a/parlai/tasks/blended_skill_talk/agents.py
+++ b/parlai/tasks/blended_skill_talk/agents.py
@@ -56,6 +56,12 @@ def _cached_data_path(opt: Opt, experiencer_side_only: bool) -> str:
     )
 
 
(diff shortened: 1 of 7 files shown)


stephenroller:
Awesome

PR #2647, opened 2020-05-08
[Chat Service] API Updates, Thread Sleeping

klshuster:
**Patch description**
1. We were running on a **very** old version of the messenger send API (which they deprecated) so we need to update to the latest version.
2. Blocking in `TimeoutUtils.get_timeout_act()` steals the GIL from any threads executing async calls (e.g. `.cuda()` operations); adding a small sleep here mitigates this issue.
3. The change to the blender ascii is adding a newline (lint error)

Tagging @emilydinan just for the blender ascii change

**Testing steps**
Tested in chat services


parlai/chat_service/utils/timeout.py
```diff
@@ -53,6 +53,7 @@ def _is_valid(act):
                         "quick_replies": quick_replies,
                     }
                 )
+            time.sleep(1)
```
JackUrb, line 56: This is a fairly unresponsive sleep - definitely need to have one but `0.1` may be better for feeling snappy in response to user input (especially as this is shared between all kinds of chat services). Could also put it in a constant

parlai/chat_service/services/messenger/message_sender.py
```diff
@@ -13,6 +13,7 @@
 MAX_TEXT_CHARS = 640
 MAX_QUICK_REPLY_TITLE_CHARS = 20
 MAX_POSTBACK_CHARS = 1000
+API_VERSION = 'v6.0'
```
JackUrb, line 16: ❤️

parlai/chat_service/utils/timeout.py
hadasah: fwiw @JackUrb I also though 1 was too long but anecdotally I haven't felt the response time to be too long yet.

JackUrb: Fair enough - ultimately when it comes to using polling whatever feels like it's not a bad tradeoff between responsiveness and interruption is alright. We've solved this in Mephisto using a `threading.Event()` that is set whenever an action is available, this way we can wait on that even rather than polling.

klshuster: 1 was purely based off anecdotal evidence - indeed using a `threading.Event()` is the appropriate course of action here, but this was meant as a quick fix. I can reduce to `0.1` for now

PR #2649, opened 2020-05-09
Support add_argument(type=bool)

### LINKED ISSUE #2168: Throw error if bool is used in add_argument?
in our `ParlaiParser.add_argument`, if the type argument is a literal bool, we should either throw an error or maybe smartly switch to `'bool'` with a warning?

stephenroller:
**Patch description**
Fixes #2168. Let's  us do `add_argument(type=bool)`.

**Testing steps**
New CI test.


jxmsML:
Adding this test is awesome!

PR #2651, opened 2020-05-09
Fix off by one in build_data.

### LINKED ISSUE #2133: unimportant off-by-one in build_data.download
https://github.com/facebookresearch/ParlAI/blob/f261852f688dc1ae9e2382518c757da14a603cac/parlai/core/build_data.py#L75-L80

This reads like there are 5 retries, but there are really 6. Should change this code to be more readable.

stephenroller:
**Patch description**
Fixes #2133.

**Testing steps**
New test.


stephenroller:
Re-requesting review.

Issue #1918, opened 2019-08-13
Add additional tests for Torch Ranker Agents


emilydinan:
We currently have a test for TorchRankerAgents here: 
https://github.com/facebookresearch/ParlAI/blob/master/tests/test_tra.py

We should add additional models to test: transformer/crossencoder, as well as all of the bert_rankers. 

It would also be great to test other flags like `--train-predict True`, `--cap-num-predictions 5`, etc....

PR #2656, opened 2020-05-11
crossencoder and train-predict tra tests

domrigoglioso:
**Patch description**
Adds a test for transformer/crossencoder and --train-predict, as mentioned in #1918 

**Testing steps**
Ran `python tests/test_tra.py`, passed all tests
```
..s
Ran 29 tests in 80.242s
```


stephenroller:
Do you mind merging master into this PR so we can see if it passes after I fixed the unittests in master?

PR #2661, opened 2020-05-13
Enable pytorch 1.5.

stephenroller:
**Patch description**
pytorch 1.5 came out and broke our tests. This turns out to because our method for hogwild no longer works. Unfortunately, I think we're going to have to simply sunset hogwild, but this requires a lot of doc updates etc. The best way is to do this before 1.6, and for now just tell users that hogwild is 1.4 or earlier.

So in summary:
- Make the helper decorators for pytorch a little more usable.
- Upgrade the installation caches for CI
- Disable hogwild only when in 1.5.

**Testing steps**
Local testing, but CI requires a lot of updates.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,8 +121,8 @@ setupcuda: &setupcuda
     working_directory: ~/
     command: |
       # download and install nvidia drivers, cuda, etc
-      wget --quiet --no-clobber -P ~/nvidia-downloads 'https://s3.amazonaws.com/ossci-linux/nvidia_driver/NVIDIA-Linux-x86_64-430.40.run'
-      time sudo /bin/bash ~/nvidia-downloads/NVIDIA-Linux-x86_64-430.40.run --no-drm -q --ui=none
+      wget --quiet --no-clobber -P ~/nvidia-downloads 'https://s3.amazonaws.com/ossci-linux/nvidia_driver/NVIDIA-Linux-x86_64-440.64.run'
+      time sudo /bin/bash ~/nvidia-downloads/NVIDIA-Linux-x86_64-440.64.run --no-drm -q --ui=none
       echo "Done installing NVIDIA drivers."
       pyenv versions
       nvidia-smi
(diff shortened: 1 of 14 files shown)


parlai/utils/testing.py
```diff
@@ -89,15 +89,31 @@ def skipIfCircleCI(testfn, reason='Test disabled in CircleCI'):
...
         skip = True
     else:
-        from packaging import version
+        import packaging.version
 
-        skip = version.parse(torch.__version__) < version.parse('1.4.0')
-    return unittest.skipIf(skip, reason)(testfn)
+        target = packaging.version.parse(version)
+        skip = packaging.version.parse(torch.__version__) < target
+    reason = 'Test requires pytorch {version} or newer'
```
klshuster, line 101: nit: prob need an `f""` string here

parlai/utils/testing.py
```diff
@@ -89,15 +89,31 @@ def skipIfCircleCI(testfn, reason='Test disabled in CircleCI'):
...
+def skipIfTorchVersion(version):
+    skip = False
+    if not TORCH_AVAILABLE:
+        skip = True
+    else:
+        import packaging.version
+
+        target = packaging.version.parse(version)
+        skip = packaging.version.parse(torch.__version__) >= target
+    reason = 'Test requires pytorch OLDER than {version}'
```
klshuster, line 114: ditto

tests/test_memnn.py
```diff
@@ -73,6 +74,7 @@ def test_labelcands_multi(self):
         self.assertGreater(valid['hits@1'], 0.95)
         self.assertGreater(test['hits@1'], 0.95)
 
+    @unittest.skip
```
klshuster, line 77: should we just delete the test if we're skipping it?

.circleci/config.yml
```diff
@@ -147,12 +146,12 @@ jobs:
...
       - restore_cache:
-          key: deps-v5-osx-{{ checksum "requirements.txt" }}
+          key: deps-v5-osx-testing1-{{ checksum "requirements.txt" }}
       - <<: *setup
-      - <<: *installdeps
       - <<: *installtorchcpuosx
+      - <<: *installdeps
       - save_cache:
-          key: deps-v5-osx-{{ checksum "requirements.txt" }}
+          key: deps-v5-osx-testing1-{{ checksum "requirements.txt" }}
```
klshuster, line 154: just curious why you added the `testing1` for all the cache keys

stephenroller: Bc I was testing, of course

tests/test_memnn.py
stephenroller: whoops!

stephenroller:
Sigh, it looks like pytorch 1.5 hard breaks model parallelism (pytorch/pytorch#38605).

stephenroller:
We'll revive for 1.5.1.

PR #2663, opened 2020-05-14
AirDialogue dataset

domrigoglioso:
**Patch description**
Add AirDialogue dataset to tasks (https://github.com/google/airdialogue)

**Logs**

`python3 examples/display_data.py -t airdialogue`
```
[creating task(s): airdialogue]
- - - NEW EPISODE: airdialogue - - -
Hello
   Hello
How may I help you?
   Can you help me to change my recent reservation because my trip dates are got postponed?
I will help you with that please share your name to proceed further?
   Edward hall here.
Please wait for a while.
   Sure, take your own time.
There is no active reservation found under your name to amend it.
   That's ok, thank you for checking.
Thank you for choosing us.
   
- - - NEW EPISODE: airdialogue - - -
HI
   Hello
How may I be of your address?
   I want to book a flight ticket to Las Vegas with price under 1000. Can you please help me with it?
Sure, can I know your connection limit?
   I need a single connection.
Please let me know your boarding and landing points.
   Airport codes are HOU-LAS.
[ loaded 321459 episodes with a total of 2224199 examples ]
```

`python tests/datatests/test_new_tasks.py`
This was taking exceptionally long to run for me (wasn't finished after running overnight) so I'm not sure what to do here.


parlai/tasks/airdialogue/agents.py
```diff
@@ -0,0 +1,94 @@
...
+    return flights, price, connections, flight airlines, departure times, and other
+    flight information.
+    """
+
+    def __init__(self, opt, shared=None):
+        super().__init__(opt, shared)
+        jsons_path = _path(opt)
+        self.datatype = opt['datatype'].split(':')[0]
+        self.messages = []
+        self._setup_data(jsons_path)
```
emilydinan, line 38: when we do batching, we have batchsize_num teachers. to avoid loading this data batchsize_num times, how about we share `self.messages` across instances? for example
```
if shared is not None:
     self.messages = shared['messages']
else:
     self.messages = self._setup_data(jsons_path)
```
you will also have to override the `share` function to share messages

parlai/tasks/airdialogue/agents.py
```diff
@@ -0,0 +1,94 @@
...
+        self.id = 'airdialogue'
+        self.reset()
+
+    def _setup_data(self, jsons_path):
+        train_path = os.path.join(jsons_path, 'train_data.json')
+        test_valid_path = os.path.join(jsons_path, 'dev_data.json')
+        if self.datatype.startswith('test'):
+            with open(test_valid_path) as f:
```
emilydinan, line 46: looks like you have very similar blocks of code for test/valid/train. could we try making htis a bit more modular by calling a helper function?

parlai/tasks/airdialogue/agents.py
```diff
@@ -0,0 +1,94 @@
...
+        entry = self.messages[episode_idx][log_idx]
+        entry = entry.split(': ')[1]
+        last_backnforth_idx = len(self.messages[episode_idx]) - 2
+        episode_done = log_idx >= last_backnforth_idx
+        if log_idx < last_backnforth_idx:
+            label_text = self.messages[episode_idx][log_idx + 1]
+            label_text = label_text.split(': ')[1]
+            labels = [label_text]
+        else:
+            labels = ['']
```
emilydinan, line 83: when does this happen? i think we should avoid surfacing examples for which the label field is empty

domrigoglioso: This happens when the first speaker is also the very last speaker, so occasionally at the very end of episodes when they have an odd length.

domrigoglioso:
So it seems like the dataset was made private.
I'm getting a forbidden error from https://storage.googleapis.com/airdialogue/airdialogue_data.tar.gz

parlai/tasks/airdialogue/agents.py
emilydinan: in this case let's drop the last example

emilydinan:
> So it seems like the dataset was made private.
> I'm getting a forbidden error from https://storage.googleapis.com/airdialogue/airdialogue_data.tar.gz

hmmm, i wonder if this was on purpose. maybe we can file a bug on the repo

stephenroller:
Presumably, no one replied to my github issue. We should email them I suppose.

stephenroller:
I just emailed the first author, hopefully we get a response.

stephenroller:
Actually I just downloaded the dataset.

parlai/tasks/airdialogue/build.py
```diff
@@ -31,4 +32,14 @@ def build(opt):
         for downloadable_file in RESOURCES:
             downloadable_file.download_file(dpath)
 
+        # Re-organize the directory to be less redundant
+        print('reorganizing airdialogue directory')
```
stephenroller, line 36: would it make more sense to just download the file to `opt['datapath']` so it unzips properly?

domrigoglioso: Good point did not think of that! I'll fix that

parlai/tasks/airdialogue/agents.py
```diff
@@ -0,0 +1,94 @@
...
+import json
+
+
+def _path(opt):
+    # build the data if it does not exist
+    build(opt)
+
+    # set up path to data (specific to each dataset)
+    jsons_path = os.path.join(
+        opt['datapath'], 'airdialogue', 'airdialogue_data', 'airdialogue'
```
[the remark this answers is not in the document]
domrigoglioso: Would be `airdialogue_data/airdialogue` now and not `airdialogue/airdialogue`

parlai/tasks/airdialogue/agents.py
```diff
@@ -0,0 +1,99 @@
...
+        entry = entry.split(': ')[1]
+        last_backnforth_idx = len(self.messages[episode_idx]) - 2
+        # sometimes the first speaker is at the end with no reply
+        if len(self.messages[episode_idx]) % 2 == 1:
+            last_backnforth_idx -= 1
+        episode_done = log_idx >= last_backnforth_idx
+        label_text = self.messages[episode_idx][log_idx + 1]
+        label_text = label_text.split(': ')[1]
+        labels = [label_text]
+        action = {
```
emilydinan, line 89: last nit: can you wrap this in a Message object? See here for example: https://github.com/facebookresearch/ParlAI/blob/19b83d8aec0cf441c9a5a21cf08bc6f6a03f0a4c/parlai/tasks/light_genderation_bias/agents.py#L396

PR #2673, opened 2020-05-19
[Bug Fix]Display data should add :ordered properly

dexterju:
**Patch description**
The previous code forces display_data to have ordered in opt, this logic breaks for valid, test. Also breaks for train:stream
**Testing steps**
display_data with any chunk teacher.

### DIFF
--- a/parlai/scripts/display_data.py
+++ b/parlai/scripts/display_data.py
@@ -62,8 +62,8 @@ def simple_display(opt, world, turn):
 
 def display_data(opt):
     # force ordered data to prevent repeats
-    if 'ordered' not in opt['datatype']:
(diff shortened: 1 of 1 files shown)


stephenroller:
@emilydinan  is this what you were thinking?

dexterju:
Ping.

stephenroller:
Merge in master and rerun checks. This kind of change could be insidious and 4 checks flagging looks weird

dexterju:
Yes, I tested with valid:stream and test:stream

dexterju:
The unittests_38 is broken, right?

stephenroller:
Yeah that’s just a fluke

Issue #2177, opened 2019-11-19
ParlaiParser.add_argument should support no_save and no_override


stephenroller:
Depends on #2172.

- If no_save is True, then the option is not saved with the opt.
- If no_override is True, then the parser will throw a hard error if the value on disk is not what's loaded. This one might have to wait a bit longer.

PR #2676, opened 2020-05-20
Add Opt load/save utilities

### LINKED ISSUE #2172: Opt save/load
Build a save and load function into opt, and use them. Should help refactor out some nastiness win params.py.

stephenroller:
**Patch description**
Bundles these methods with Opt just for cleanliness. Corresponding internal changes upgrades old references.

Also adds a backstop for items that SHOULDN'T be saved or loaded from disk.

Although this may seem like a naming nit, centralizing the functionality will help keep attempts at #2177 clean.

Fixes #2172.

**Testing steps**
CI.

### DIFF
--- a/parlai/core/opt.py
+++ b/parlai/core/opt.py
@@ -13,6 +13,17 @@ import json
 import pickle
 import traceback
 
+# these keys are automatically removed upon save. This is a rather blunt hammer.
+# It's preferred you indicate this at option definiton time.
+__AUTOCLEAN_KEYS = [
+    "override",
+    "batchindex",
+    "download_path",
+    "datapath",
+    "batchindex",
+    "interactive_mode",
+]
+
 
 class Opt(dict):
     """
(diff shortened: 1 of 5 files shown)


parlai/core/opt.py
```diff
@@ -81,17 +94,38 @@ def display_history(self, key):
...
+        dct = dict(self)
+
+        # clean up some things we probably don't want to save
+        for key in __AUTOCLEAN_KEYS__:
+            if key in dct:
+                del dct[key]
+
+        with open(filename, 'w', encoding='utf-8') as f:
+            json.dump(dct, fp=f, indent=4)
+            f.write('\n')
```
EricMichaelSmith, line 113: Nit: maybe keep the comment about doing this for jq?

parlai/core/opt.py
```diff
@@ -13,6 +13,19 @@
...
+
+# these keys are automatically removed upon save. This is a rather blunt hammer.
+# It's preferred you indicate this at option definiton time.
+__AUTOCLEAN_KEYS__: List[str] = [
+    "override",
+    "batchindex",
+    "download_path",
+    "datapath",
+    "batchindex",
+    "interactive_mode",
```
EricMichaelSmith, line 26: Maybe keep your comment about why we're not saving interactive_mode?

PR #2692, opened 2020-05-26
Testing speedups/robustness

### LINKED ISSUE #2338: Remaining slow unit tests
Following up from #2327, here are a few possible improvements that could be done to unit tests to bring them to sub 5 minutes:

- there's a test_context_blocking and test_beam_blocking. Both of those train a model to get 100% performance on a custom task, and then use blocking to ensure performance degrades in an expected manner. These are also somewhat flaky. By moving them to pretrained models (distributed with the other unit test models), we could remove the flakiness and speed up their inference from ~50s to well under 10.
- Deleting the pytorch data teacher tests
- ?

stephenroller:
**Patch description**
- Use more machines & bigger machines lol
- Use slightly smarter parallelism, grouping tests by how much time they take
- Fixes #2338 via distributing a few more pretrained models.
- Shave a few unnecessary evaluations off of others
- Increase the timeout of the test failure. Unfortunately, sometimes we just get a very slow machine now, and some test will end up taking 15 minutes instead of 1... I'm not sure what's going on.

Todo: Distribute some pretrained models for test_tra.py (or probably just reuse the existing ones). We should see a substantial speedup on our slowest worker then.

**Testing steps**
It's all CI.


stephenroller:
Marking this as ready for review, I'll put off speeding up test_tra.py for later or someone else to do.

PR #2700, opened 2020-05-28
[Display Data] don't force ordered

emilydinan:
**Patch description**
`train:ordered` is the default for display data, but some teachers only work with streaming. We shouldn't FORCE `train:ordered` (it is set as default, but we should be able to override it).


emilydinan:
abandoning, as i see @dexterju already fixed this!

PR #2701, opened 2020-05-28
Convai2:normalized small fix

jaseweston:
The persona lines weren't normalized correctly.

We had  "Your persona: i love sweets."  rather than  "your persona: I love sweets."


jaseweston:
Yes, it's not quite how we trained some models then, but it's seems slightly wrong, so idk, just fix  I think..

stephenroller:
I mean blender shouldn't have much problem generalizing "Your" vs your""

PR #2704, opened 2020-05-29
Better catching of CUDA OOms.

stephenroller:
**Patch description**
A very old pytorch bug was resolved (pytorch/pytorch#18853) allowing us to more reliably recover from CUDA OOM issues. This uses this new trick.

**Testing steps**
LOL IM OOMING LIKE CRAZY IN MY SWEEP ITS SO GOOD.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -676,20 +676,29 @@ class TorchGeneratorAgent(TorchAgent, ABC):
             loss = self.compute_loss(batch)
             self.backward(loss)
             self.update_params()
+            oom_sync = False
         except RuntimeError as e:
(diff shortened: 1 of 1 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -676,21 +676,30 @@ def train_step(self, batch):
...
-                # gradients are synced on backward, now this model is going to be
-                # out of sync! catch up with the other workers
-                self._init_cuda_buffer(8, 8, True)
             else:
                 raise e
 
+        if oom_sync:
+            # moved outside of the try-except because the raised exception in scope
+            # actually prevents from the data being freed, which can sometimes cause
+            # us to OOM during our OOM handling.
```
EricMichaelSmith, line 696: 😢

PR #2717, opened 2020-06-02
Fromfile jsonl teacher

hadasah:
**Patch description**
Added a fromfile teacher which reads .jsonl files in the conversations format (see `parlai/utils/conversations.py` for more detail). Tl;dr, each conversation is on one line, formatted like this (line breaks for readability):
```
{
  'dialogue':[
    {'id':'model1', 'text': 'hi'},
    {'id':'model2', 'text': 'hi back'},
    ...
  ]
}
```

**Testing steps**
Ran `display_data.py` on a sample file, with `--label-turns` as default (secondspeaker), firstspeaker, and both.


parlai/tasks/fromfile/agents.py
```diff
@@ -106,6 +106,81 @@ def __init__(self, opt, shared=None):
...
+            datafile += "_" + self.opt['datatype'].split(':')[0] + '.jsonl'
+        if shared is None:
+            self._setup_data(datafile)
+        # Truncate datafile to just the immediate enclosing folder name and file name
+        dirname, basename = os.path.split(datafile)
+        self.id = os.path.join(os.path.split(dirname)[1], basename)
+        self.reset()
+
+
+class Json2Teacher(ConversationTeacher):
```
jaseweston, line 151: i think we can drop this

parlai/tasks/fromfile/agents.py
```diff
@@ -106,6 +106,81 @@ def __init__(self, opt, shared=None):
...
+            datafile += "_" + self.opt['datatype'].split(':')[0] + '.jsonl'
+        if shared is None:
+            self._setup_data(datafile)
+        # Truncate datafile to just the immediate enclosing folder name and file name
+        dirname, basename = os.path.split(datafile)
+        self.id = os.path.join(os.path.split(dirname)[1], basename)
+        self.reset()
+
+
 class DefaultTeacher(ParlaiformatTeacher):
```
jaseweston, line 184: i think it's going to be a bit annoying to access it as it won't be the default... could we maybe check the file extension to tell which type it is if you don't specify?
or else maybe just call this whole thing   -t jsonl  or -t convfile  instead of -t fromfile ... or ?

stephenroller:
Let's get some tests. I agree, let's maybe axe the json2 in place of just the 

@jaseweston you think `-t jsonfile` is better than `-t fromfile:json`? Wish we had asked this at the meeting today.

jaseweston:
> Let's get some tests. I agree, let's maybe axe the json2 in place of just the
> 
> @jaseweston you think `-t jsonfile` is better than `-t fromfile:json`? Wish we had asked this at the meeting today.

what about the file extension detection possibiity?

otherwise, if not the file extension detection, i think i  would like `-t jsonfile`  better, as it makes it more a first class citizen, and is something i and others are likely to be using all the time in the future

stephenroller:
I'm fine with jsonfile or detection

hadasah:
I'd thought fromfile:json felt more intuitive, but if the other ones make more sense, happy to switch it.
If there's not a big difference to jason, I'd prefer `-t jsonfile` to detection, I like the explicitness.

jaseweston:
one can do both "fromfile:json" and detection though i think...  one can
make the fromfile default teacher detect the extension and call the right
fromfile type.
If however "fromfile:json" is invoked, it would go directly to that teacher
.. ?
is that possible?

parlai/core/teachers.py
```diff
@@ -1399,6 +1403,142 @@ def _setup_data(self, path):
...
+        return len(self.episodes)
+
+    def get(self, episode_idx, entry_idx=None):
+        """
+        Get a specific example from the dataset.
+        """
+        return self.episodes[episode_idx][entry_idx]
+
+    def _setup_data(self, path):
+        print("[loading parlAI text data:" + path + "]")
```
emilydinan, line 1497: hmm, can we make this more descriptive?

parlai/core/teachers.py
```diff
@@ -1399,6 +1403,142 @@ def _setup_data(self, path):
...
+        self.episodes = []
+        self.num_exs = 0
+        eps = []
+        with open(path, newline='\n') as read:
```
emilydinan, line 1501: hmm... i think the idea here was to use the Conversations utilities to read and iterate through the data. can we do this instead of manually loading the json file? there are utilities for loading it up (`Conversations('<filename>')` and also for iterating through conversations (you should jsut be able to do something like `for convo in conversations ...`

parlai/core/teachers.py
```diff
@@ -1399,6 +1403,142 @@ def _setup_data(self, path):
...
+            self.episodes = []
+            self.num_exs = 0
+            self.label_turns = opt.get('label_turns')
+            if opt.get('conversationteacher_datafile') is not None:
+                self._setup_data(opt.get('conversationteacher_datafile'))
+        else:
+            self.episodes = shared['episodes']
+            self.num_exs = sum(len(e) for e in self.episodes)
+
+        self.id = opt['task']
```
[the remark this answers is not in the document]
hadasah: it would just be "jsonfile" if you use this ConversationTeacher directly, but when called with '-t jsonfile', the JsonTeacher overrides self.id with the name of the file

parlai/tasks/fromfile/agents.py
hadasah: I changed it to -t jsonfile because dynamically changing the class of DefaultTeacher is messy and I couldn't think of a clean way to do it. Usually for this sort of thing I use a factory but the way that defaultteachers are set up, that doesn't make sense. @stephenroller ?

parlai/core/teachers.py
```diff
@@ -1491,52 +1492,50 @@ def get(self, episode_idx, entry_idx=None):
         """
         Get a specific example from the dataset.
         """
-        return self.episodes[episode_idx][entry_idx]
+        return Message(self.episodes[episode_idx][entry_idx])
 
     def _setup_data(self, path):
         print("[loading parlAI text data:" + path + "]")
```
stephenroller, line 1498: also please use logging.info

parlai/core/teachers.py
```diff
@@ -1422,6 +1427,140 @@ def _setup_data(self, path):
...
+
+    def _setup_data(self, path):
+        logging.info("[loading data from json file into task:" + path + "]")
+        self.episodes = []
+        self.num_exs = 0
+        eps = []
+        conversations = Conversations(path)
+        self.num_exs = sum(len(c) for c in conversations)
+        for conv in conversations:
+            context = conv.context
```
hadasah, line 1528: I wasn't sure what to do with the context @emilydinan @stephenroller ? 
If you look at the sample json file I used for testing, the context got stored within the dialogue under the id 'context' so the dialogue doesn't really use it correctly.

parlai/tasks/jsonfile/agents.py
```diff
@@ -22,10 +22,10 @@ class JsonTeacher(ConversationTeacher):
...
     def add_cmdline_args(argparser):
-        agent = argparser.add_argument_group('FromFile Task Arguments')
+        agent = argparser.add_argument_group('JsonFile Task Arguments')
         agent.add_argument('-dp', '--fromfile-datapath', type=str, help="Data file")
         agent.add_argument(
-            '-ffdt',
+            '-jfdt',
             '--fromfile-datatype-extension',
             type='bool',
             default=False,
```
stephenroller, line 31: I wonder if we can make this true.

parlai/core/teachers.py
```diff
@@ -1422,6 +1427,147 @@ def _setup_data(self, path):
...
+        """
+        Return the number of episodes from the data.
+        """
+        return len(self.episodes)
+
+    def get(self, episode_idx, entry_idx=None):
+        """
+        Get a specific example from the dataset.
+        """
+        return Message(self.episodes[episode_idx][entry_idx])
```
stephenroller, line 1518: Why are we manually implementing get/num_episodes/etc. Shouldn't the `setup_data` basically handle all this, if we were yielding etc?

hadasah: is using yield better? None of the teachers in teachers.py uses yield except the (deprecated) fbdialogteacher

parlai/core/teachers.py
```diff
@@ -1422,6 +1427,147 @@ def _setup_data(self, path):
...
+                if eps:
+                    eps[-1]['episode_done'] = True
+                    self.episodes.append(eps)
+
+            # train on even turns as labels (turns w/ second speaker)
+            if self.label_turns in ['secondspeaker', 'both']:
+                eps = []
+                xturns = turns[1::2]
+                yturns = turns[2::2]
+                for xturn, yturn in zip(xturns, yturns):
```
emilydinan, line 1559: since this for loop is basically the same as repeated above, could we make a helper function here?

parlai/core/teachers.py
stephenroller: it's okay

PR #2718, opened 2020-06-02
Tweak the stale action.

stephenroller:
**Patch description**
Tweak the stale action a little bit.

**Testing steps**
No testing with github actions.


stephenroller:
Hm, I'm down, but I thought maybe we'd want them to be the same for the task synchronization tool but I can change it.

klshuster:
yeah i like `never-stale`, good to have both

Issue #1883, opened 2019-07-27
Parse verification for options passed in train_model, etc.


stephenroller:
Our current unit tests often use a case of passing opts directly into a model, and combined with some logic that uses set_default. This is also common for a variety of scripts which just set command line options.

However, set_defaults does not perform an error message if you try setting a default for an argument that hasn't been added to the parser. This has led to some cases where misspelled options were being set, and the true intended option was using its default.

We should switch to a version which, when given options, treats them as being set via command line parameters and does a proper `parse_args()` to identify they are indeed valid options. An argparser error should be raised otherwise.

stephenroller:
This is easily done now by upgrading the `parlai/utils/testing.py` to call `TrainModel.main_kwargs` instead of the funky ParlaiParser.set_defaults + loop that it does.

vedantpuri:
Is this only referring to the ```train_model``` function in testing.py? 
I don't completely understand how main_kwargs works, could you direct me to any sources where I could learn more ?

As far as I understand these lines are to be replaced (in the ```train_model``` function):
```python
        parser = tms.setup_args()
        # needed at the very least to set the overrides.
        parser.set_params(**opt)
        parser.set_params(log_every_n_secs=10)
        popt = parser.parse_args([], print_args=False)
        # in some rare cases, like for instance if the model class also
        # overrides its default params, the params override will not
        # be taken into account.
        for k, v in opt.items():
            popt[k] = v
        tl = tms.TrainLoop(popt)
        valid, test = tl.train()
```

with some call to main_kwargs which will setup everything as it does in ```scripts/train_model.py```.

Also, has something like this to be done for ```eval_model``` function as well ?

stephenroller:
That's correct, yes. We shouldn't need the set_params and the for loop anymore when you switch to tms.TrainModel.main(**kwargs)

PR #2719, opened 2020-06-03
Parse verification in utils/testing

vedantpuri:
**Patch description**
Fix for #1883 
Also fixes other typos that were revealed using parse verification

**Testing steps**
Run the CI tests and ensure it doesn't change from earlier

**Logs**
Stack Trace of Latest run:
[st.txt](https://github.com/facebookresearch/ParlAI/files/4725483/st.txt)

**Other information**
ParlaiScript wasn't returning in _run_args and _run_kwargs, and this PR fixes that.

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -159,8 +159,6 @@ def str2floats(s):
     """
     Look for single float or comma-separated floats.
     """
-    if "[" in s:
-        s = s[1:-1]
     return tuple(float(f) for f in s.split(','))
 
 
@@ -1089,10 +1087,10 @@ class ParlaiParser(argparse.ArgumentParser):
                 string_args.append(last_option_string)
             elif isinstance(action, argparse._StoreAction) and action.nargs is None:
                 string_args.append(last_option_string)
-                string_args.append(str(value))
+                string_args.append(self._value2argstr(value))
             elif isinstance(action, argparse._StoreAction) and action.nargs in '*+':
                 string_args.append(last_option_string)
-                string_args.extend([str(v) for v in value])
+                string_args.extend([self._value2argstr(value) for v in value])
             else:
                 raise TypeError(f"Don't know what to do with {action}")
 
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -1055,10 +1055,12 @@ class ParlaiParser(argparse.ArgumentParser):
                 string_args.append(last_option_string)
             elif isinstance(action, argparse._StoreAction) and action.nargs is None:
                 string_args.append(last_option_string)
-                string_args.append(self._value2argstr(value, space=True))
+                string_args.append(self._value2argstr(value))
             elif isinstance(action, argparse._StoreAction) and action.nargs in '*+':
                 string_args.append(last_option_string)
-                string_args.extend([self._value2argstr(value) for v in value])
+                string_args.extend(
+                    [self._value2argstr(value, space=True) for v in value]
+                )
             else:
                 raise TypeError(f"Don't know what to do with {action}")
 
@@ -1090,10 +1092,11 @@ class ParlaiParser(argparse.ArgumentParser):
                 string_args.append(last_option_string)
             elif isinstance(action, argparse._StoreAction) and action.nargs is None:
                 string_args.append(last_option_string)
-                string_args.append(self._value2argstr(value, space=True))
+                string_args.append(self._value2argstr(value))
             elif isinstance(action, argparse._StoreAction) and action.nargs in '*+':
                 string_args.append(last_option_string)
-                string_args.extend([self._value2argstr(value) for v in value])
+                # Special case: Labels
+                string_args.extend([str(v) for v in value])
             else:
                 raise TypeError(f"Don't know what to do with {action}")
 
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -1007,6 +1007,15 @@ class ParlaiParser(argparse.ArgumentParser):
 
         return self.opt
 
+    def _value2argstr(self, value: Any) -> str:
+        """
+        Reverse-parse an opt value into one interpretable by argparse.
+        """
+        if isinstance(value, [list, tuple]):
+            return ",".join(str(v) for v in value)
+        else:
+            return str(v)
+
     def _kwargs_to_str_args(self, **kwargs):
         """
         Attempt to map from python-code kwargs into CLI args.
@@ -1043,10 +1052,10 @@ class ParlaiParser(argparse.ArgumentParser):
                 string_args.append(last_option_string)
             elif isinstance(action, argparse._StoreAction) and action.nargs is None:
                 string_args.append(last_option_string)
-                string_args.append(str(value))
+                string_args.append(self._value2argstr(value))
             elif isinstance(action, argparse._StoreAction) and action.nargs in '*+':
                 string_args.append(last_option_string)
-                string_args.extend([str(v) for v in value])
+                string_args.extend([self._value2argstr(value) for v in value])
             else:
                 raise TypeError(f"Don't know what to do with {action}")
 
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -159,6 +159,8 @@ def str2floats(s):
     """
     Look for single float or comma-separated floats.
     """
+    if "[" in s:
+        s = s[1:-1]
     return tuple(float(f) for f in s.split(','))
 
 
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -1007,12 +1007,15 @@ class ParlaiParser(argparse.ArgumentParser):
 
         return self.opt
 
-    def _value2argstr(self, value) -> str:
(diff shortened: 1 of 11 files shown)


stephenroller:
Weird, why isn't CI running?

vedantpuri:
Yeah, I was also surprised why it didn't run. I initially thought maybe things changed since I last worked and only certain tests are triggered, but from your reply that seems False.
I am running the unit tests locally and will put the trace in the PR description in a bit.

stephenroller:
Yeah that'd be great, push with [long] and let's see why CI isn't being hit.

vedantpuri:
Yeah, I am unsure why CI isn't being triggered, I think its something in the CI config, any ideas @stephenroller ?

stephenroller:
I'm going to ask our internal support team to take a look

stephenroller:
dope

stephenroller:
- For Self-Feeding, it should be `learningrate` instead of `learning_rate`.
- For the others, we're rendering the option wrong. We're gonna need some smarter logic:

There's a few calls to str(v) in this method:
https://github.com/facebookresearch/ParlAI/blob/4e27a41285565085e76e6a6ce4b78055f9db483a/parlai/core/params.py#L1046

We're gonna have to refactor those out into some helper like `self._value2argstr` or similar which needs to look like:

```python
def _value2argstr(self, value: Any) -> str:
    """
    Reverse-parse an opt value into one interpretable by argparse.
    """
    if isinstance(value, [list, tuple]):
        return ",".join(str(v) for v in value)
    else:
        return str(v)
```

parlai/core/params.py
```diff
@@ -159,6 +159,8 @@ def str2floats(s):
     """
     Look for single float or comma-separated floats.
     """
+    if "[" in s:
```
stephenroller, line 162: is this really necessary?

stephenroller: at the very least, you need to lead a comment explaining.

parlai/scripts/script.py
```diff
@@ -63,7 +63,7 @@ def _run_args(cls, args: Optional[List[str]] = None):
         opt = parser.parse_args(args=args, print_args=False)
         script = cls(opt)
         script.parser = parser
-        script.run()
+        return script.run()
```
stephenroller, line 66: thanks!

parlai/utils/testing.py
```diff
@@ -274,18 +274,8 @@ def train_model(opt: Opt) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
-        popt = parser.parse_args([], print_args=False)
-        # in some rare cases, like for instance if the model class also
-        # overrides its default params, the params override will not
-        # be taken into account.
-        for k, v in opt.items():
-            popt[k] = v
-        tl = tms.TrainLoop(popt)
-        valid, test = tl.train()
+        # Parse verification
+        valid, test = tms.TrainModel.main(**opt)
```
stephenroller, line 278: it even reads cleaner...

parlai/core/params.py
vedantpuri: The issue is, sometimes (atleast I think its not always) a string is passed in that is of the form of a list (i.e. "[1, 2, 3]"). I am not super sure why that is the case but then the error thrown was cannot convert "[1" to float because it was splitting on comma.

I would be happy to change it if there is a better way/different place to deal with this. Would be grateful for ideas/suggestions.

stephenroller: Can you help me figure out WHERE the string is being passed in? I don't think that should be the case after you switched to `_value2str`

vedantpuri: See [here](https://circleci.com/gh/facebookresearch/ParlAI/42246?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). This stack trace prompted me to make the change

stephenroller: I see. Look back at the rest of _kwargs_to_str_args. There's another paragraph of code that needs the same str -> value2argstr change.

parlai/core/params.py
```diff
@@ -1043,10 +1054,10 @@ def _kwargs_to_str_args(self, **kwargs):
                 string_args.append(last_option_string)
```
vedantpuri, line 1054: @stephenroller are you referring to this also being changed to ```_value2argstr(last_option_string)``` ?

parlai/core/params.py
vedantpuri: I can add a comment explaining the precaution of that if statement.

parlai/core/params.py
stephenroller: No, I'm referring to the `string_args.append(str(value))` down on lines 1092 and 1095

parlai/core/params.py
stephenroller: I think after you fix the other thing, it won't be necessary to have these lines any more.

vedantpuri: Tried that in the latest push

parlai/core/params.py
vedantpuri: Unsure how I missed that, thanks for pointing out. Fixed them in latest push

parlai/core/params.py
```diff
@@ -1078,10 +1087,10 @@ def _kwargs_to_str_args(self, **kwargs):
...
             elif isinstance(action, argparse._StoreAction) and action.nargs is None:
                 string_args.append(last_option_string)
-                string_args.append(str(value))
+                string_args.append(self._value2argstr(value))
             elif isinstance(action, argparse._StoreAction) and action.nargs in '*+':
                 string_args.append(last_option_string)
-                string_args.extend([str(v) for v in value])
+                string_args.extend([self._value2argstr(value) for v in value])
```
stephenroller, line 1093: lol okay so THIS one is needs to be space separated, while the one  above needs to be comma separated.

sorry, nothing's easy.

parlai/core/params.py
```diff
@@ -1043,10 +1052,10 @@ def _kwargs_to_str_args(self, **kwargs):
...
```
stephenroller, line 1058: this one also space separated.

vedantpuri:
I don't understand in some cases its converting the number to letters, like 1 to one, 0 to zero and thats causing an issue in areas, any thoughts ?

stephenroller:
That’s coming from here:


https://github.com/facebookresearch/ParlAI/blob/11da27ba9858442281bd6c3ce93f9b844031eea1/parlai/tasks/integration_tests/agents.py#L404

parlai/core/params.py
```diff
@@ -1007,6 +1007,18 @@ def parse_args(self, args=None, namespace=None, print_args=True):
 
         return self.opt
 
+    def _value2argstr(self, value, space=False) -> str:
+        """
+        Reverse-parse an opt value into one interpretable by argparse.
+        """
+        if isinstance(value, (list, tuple)):
+            if not space:
```
stephenroller, line 1015: mind stripping out the space case since we don't need it anymore?

parlai/utils/testing.py
```diff
@@ -274,18 +274,8 @@ def train_model(opt: Opt) -> Tuple[Dict[str, Any], Dict[str, Any]]:
             opt['model_file'] = os.path.join(tmpdir, 'model')
         if 'dict_file' not in opt:
             opt['dict_file'] = os.path.join(tmpdir, 'model.dict')
-        parser = tms.setup_args()
```
stephenroller, line 277: can you also do eval_model and related now?

parlai/core/params.py
```diff
@@ -1043,10 +1055,12 @@ def _kwargs_to_str_args(self, **kwargs):
                 string_args.append(last_option_string)
             elif isinstance(action, argparse._StoreAction) and action.nargs is None:
                 string_args.append(last_option_string)
-                string_args.append(str(value))
+                string_args.append(self._value2argstr(value))
             elif isinstance(action, argparse._StoreAction) and action.nargs in '*+':
                 string_args.append(last_option_string)
-                string_args.extend([str(v) for v in value])
+                string_args.extend(
+                    [self._value2argstr(value, space=True) for v in value]
```
vedantpuri, line 1062: @stephenroller space is still in use here. I am not confident that it is actually being used anywhere though, would you know if this case is ever hit ?
If not I can get rid of it

stephenroller: The two sections should mirror each other

vedantpuri:
Seems like ```batch_size``` should be ```batchsize``` in a number of places, unsure about the ```num_epochs``` error.

parlai/core/params.py
```diff
@@ -1043,10 +1052,10 @@ def _kwargs_to_str_args(self, **kwargs):
                 string_args.append(last_option_string)
```
stephenroller, line 1052: This chunk (and its mirror below) needs to become:

```
if isinstance(action, ,argparse._StoreActionTrue):
    if bool(value):
        string_args.append(last_option_string)
```
That should fix the advanced args issue

vedantpuri:
Yielding this PR to @stephenroller. Hope to see it landed soon!

PR #2721, opened 2020-06-03
Change whitelist/blacklist terminology 

emilydinan:
**Patch description**
Get rid of all the whitelist/blacklist terminology in ParlAI and replace with allow list and block list. I also observed a bug with sharing the block list in torch_generator_agent that I resolved.

### DIFF
--- a/parlai/chat_service/utils/misc.py
+++ b/parlai/chat_service/utils/misc.py
@@ -124,12 +124,12 @@ class SafetyDetectionResult(Enum):
     Result of identfying offensive language in a response.
 
     SAFE:       the message is safe
-    BLACKLIST:  the message contains a word from the blacklist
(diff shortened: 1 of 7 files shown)


parlai/chat_service/utils/misc.py
```diff
@@ -124,12 +124,12 @@ class SafetyDetectionResult(Enum):
...
 
     SAFE:       the message is safe
-    BLACKLIST:  the message contains a word from the blacklist
+    BLOCKLIST:  the message contains a word from the blocklist
     UNSAFE:     the message is deemed unsafe by the safety classifier
     """
 
     SAFE = 0
-    BLACKLIST = 1
+    BLOCKLIST = 1
```
klshuster, line 132: Yeah this'll break stuff but i can make the change on my end once we sync - can you ping me when you merge?

PR #2724, opened 2020-06-03
[Beam block] upgrade opt bug fix

emilydinan:
**Patch description**
Don't override block list unless it is not None. I'm still not convinced this is the right solution, seems like an issue with opt overriding.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -334,9 +334,10 @@ class TorchGeneratorAgent(TorchAgent, ABC):
 
         # 2020-06-03: Changing "blacklist" --> "blocklist"
         if 'beam_blacklist_filename' in opt_from_disk:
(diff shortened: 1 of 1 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -334,9 +334,10 @@ def upgrade_opt(cls, opt_from_disk: Opt):
 
         # 2020-06-03: Changing "blacklist" --> "blocklist"
         if 'beam_blacklist_filename' in opt_from_disk:
-            opt_from_disk['beam_block_list_filename'] = opt_from_disk[
-                'beam_blacklist_filename'
-            ]
+            if opt_from_disk['beam_blacklist_filename'] is not None:
```
stephenroller, line 337: maybe it should be .get?

stephenroller: oh wait the first if protects it, duh

PR #2735, opened 2020-06-09
Update narrativeqa data checksum because of upstream change

sjmielke:
They fixed some URLs so now their "master" is a different zip 🤷‍♀️


stephenroller:
Wanna hit merge @sjmielke?

PR #2748, opened 2020-06-10
Don't require pre-built dictionaries.

stephenroller:
**Patch description**
For certain specialized models/tokenizers, the dictionary is fixed ahead of time, especially like GPT-2. In these instances, we don't necessarily need to mandate they be built ahead of time in distributed mode. This adds a new field to dictionaries indicating whether this is the case.

**Testing steps**
Manual testing. CI.

### DIFF
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -324,6 +324,12 @@ class DictionaryAgent(Agent):
             if opt.get('dict_file'):
                 self.save_path = opt['dict_file']
 
+    def is_prebuilt(self):
+        """
+        Indicates whether the dictionary is fixed, and does not require building.
+        """
+        return self.tokenizer in ['gpt2', 'bytelevelbpe']
+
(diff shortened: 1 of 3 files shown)


parlai/core/dict.py
```diff
@@ -324,6 +324,12 @@ def __init__(self, opt: Opt, shared=None):
             if opt.get('dict_file'):
                 self.save_path = opt['dict_file']
 
+    def is_prebuilt(self):
```
emilydinan, line 327: hmm so why not just return False if this is overridden by the Gpt2 dictionary?

also, there are other HF tokenizers out there in ParlAI (that Dexter wrote); these should probably return False too, no?

stephenroller: This is a misnomer. There's the HuggingFace GPT2 dictionary, and there's the gpt-2 tokenizer, and they're different.

PR #2752, opened 2020-06-11
[TriviaQA] no-evidence, union

sjmielke:
Adds NoEvidence teachers for Web and Wikipedia sections and adds a NoEvidenceUnion teacher that (thanks to evidence-dropping) joins the two sections. This is different from the previous "union" that did multitasking because most questions in TriviaQA are shared between Web and Wikipedia. New build()-ing merges the two into a single json with combined answer sets that can then be used by the teacher---marked as version "2.0" to make sure we don't have stale files :)

### DIFF
--- a/parlai/tasks/triviaqa/agents.py
+++ b/parlai/tasks/triviaqa/agents.py
@@ -24,11 +24,10 @@ class WebTeacher(DialogTeacher):
     def __init__(self, opt, shared=None):
         if not hasattr(self, 'prefix'):
             self.prefix = ''
-            if opt['datatype'].startswith('train'):
-                self.suffix = 'train'
-            else:
-                self.suffix = 'dev'
+            self.suffix = 'train' if opt['datatype'].startswith('train') else 'dev'
 
+        if "no_evidence" not in self.__dict__:
+            self.no_evidence = False
         qa_dir, self.evidence_dir = _path(opt)
         opt['datafile'] = os.path.join(
             qa_dir, self.prefix + 'web-' + self.suffix + '.json'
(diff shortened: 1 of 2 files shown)


parlai/tasks/triviaqa/agents.py
```diff
@@ -24,11 +24,10 @@ class WebTeacher(DialogTeacher):
     def __init__(self, opt, shared=None):
         if not hasattr(self, 'prefix'):
             self.prefix = ''
-            if opt['datatype'].startswith('train'):
-                self.suffix = 'train'
-            else:
-                self.suffix = 'dev'
+            self.suffix = 'train' if opt['datatype'].startswith('train') else 'dev'
 
+        if "no_evidence" not in self.__dict__:
```
emilydinan, line 29: hmm maybe it would be slightly better to do `if not hasattr(self, 'no_evidence')`. 

alternatively, you can set a function like `def no_evidence(self): return False` and override it in the subclasses to avoid checking if the attribute exits

parlai/tasks/triviaqa/agents.py
```diff
@@ -70,15 +72,20 @@ def __init__(self, opt, shared=None):
...
```
emilydinan, line 87: same here

parlai/tasks/triviaqa/agents.py
```diff
@@ -123,13 +133,42 @@ def __init__(self, opt, shared=None):
...
 class VerifiedTeacher(MultiTaskTeacher):
     def __init__(self, opt, shared=None):
         opt = copy.deepcopy(opt)
         opt['task'] = 'triviaqa:VerifiedWikipedia,triviaqa:VerifiedWeb'
         super().__init__(opt, shared)
 
 
+class NoEvidenceUnionTeacher(DialogTeacher):
+    def __init__(self, opt, shared=None):
+        if not hasattr(self, 'prefix'):
```
emilydinan, line 151: what does this prefix do here? where do we set prefix?

parlai/tasks/triviaqa/agents.py
```diff
@@ -123,13 +133,42 @@ def __init__(self, opt, shared=None):
...
+            self.prefix = ''
+            self.suffix = 'train' if opt['datatype'].startswith('train') else 'dev'
+
+        qa_dir, self.evidence_dir = _path(opt)
+        opt['datafile'] = os.path.join(
+            qa_dir, self.prefix + 'union-noevidence-' + self.suffix + '.json'
+        )
+        self.id = 'triviaqa'
```
emilydinan, line 159: can we make the id like `Trivia QA No Evidence Union`

parlai/tasks/triviaqa/agents.py
sjmielke: Iirc. `prefix` is there to get the `verified-` version of the dev data.

parlai/tasks/triviaqa/agents.py
sjmielke: adding triviaqa feels weird cause the other files don't have it, but yeah, should be consistent in order 🙃

PR #2755, opened 2020-06-12
Update tokens/batch metrics a bit.

stephenroller:
**Patch description**
Add some additional token-per-batch metrics: tpb_t and tpb_l which are text_vec tokens/batch and label_vec tokens/batch . The old tpb changes its semantics from label tokens/batch to tokens/batch including BOTH label and text.  We also add an "examples/batch" metric which is nice when using dynamic batching.

**Testing steps**
Manual testing. CI.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1912,9 +1912,18 @@ class TorchAgent(ABC, Agent):
             # tokens per batch
             # we divide by the binary is_primary_worker() so that the numerator is
             # num_tokens in all workers, and the denominator is 1.
-            tpb = GlobalAverageMetric(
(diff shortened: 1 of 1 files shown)


stephenroller:
Yes to confusing, but no to breaking anything. This is only reported. If y'all have suggestions for better name, i'm down to hear.

The reason was just because it was useful to me lol.

EricMichaelSmith:
Hmm, I suppose you could do 'total_tokens_per_batch', 'text_tokens_per_batch', and 'label_tokens_per_batch' - up to you, though

stephenroller:
My feeling was those were just tooooo long.

stephenroller:
Decided we don't really need this, I identified what I needed to and Eric's right, the confusion is unnecessary.

PR #2756, opened 2020-06-12
BST models in zoo

EricMichaelSmith:
**Patch description**
Code to add models used in BST paper (single-task and multi-task, base and fine-tuned on BST) to the zoo

**Testing steps**
Reproduced paper numbers on BST task with uploaded models (see `parlai/zoo/model_list.py`)


EricMichaelSmith:
> great, i know @stephenroller appreciates you splitting up these models into different tar files

Ha, yeah, the model files are pretty big, and downloading 6 models would take quite a while...

PR #2757, opened 2020-06-12
Turn global enc/dec build functions into class methods

EricMichaelSmith:
**Patch description**
Makes `_build_encoder()` and `_build_decoder()` in `parlai/agents/transformer/modules.py` class methods of `TransformerGeneratorModel`. (This means that `TransformerGeneratorModel.build_encoder()` is called when building the encoders of `TransformerMemNetModel`; if this calling of another class's class method is undesirable, the `.build_encoder()` method can be duplicated in `TransformerMemNetModel`.)

**Testing steps**
The new class methods are used by `TransformerGeneratorModel` and `TransformerMemNetModel`, the latter of which is instantiated by `TransformerClassifierAgent`, and so existing CI tests should cover both uses of these class methods.


emilydinan:
curious, can you explain what prompted this change?

EricMichaelSmith:
> curious, can you explain what prompted this change?

This would make it more straightforward to have subclasses of `TransformerGeneratorModel` override the default encoder and decoder. For instance, I override the `.build_decoder()` class method in my Blender-2.7B-based classifier for the style-gen paper (https://github.com/fairinternal/ParlAI-Internal/blob/master/agents/retnref_style/retnref_style.py#L477 )

parlai/agents/transformer/modules.py
```diff
@@ -192,7 +128,7 @@
         self.reduction_type = opt.get('reduction_type', 'mean')
         self.n_segments = opt.get('n_segments', 0)
 
-        self.context_encoder = _build_encoder(
+        self.context_encoder = TransformerGeneratorModel.build_encoder(
```
stephenroller, line 131: This should be self. Or maybe class(self).

EricMichaelSmith: Okay, yeah, I was on the fence about whether to add a separate `TransformerMemNetModel.build_encoder()` class method. I've just added one. This adds repetition of code, but it eliminates the strange calling of another class's class method

EricMichaelSmith:
> this is very convenient, i was going to do this next week.
> 
> can you run long tests just for paranoia? i think this could break wizard for example

Good point - running long tests now

PR #2760, opened 2020-06-13
Atomic saves

stephenroller:
**Patch description**
When working with these very large files, we have to worry about our cluster scheduler preempting our jobs in the middle of saving. This change makes the saving of model files atomic by saving to a temporary file, and then renaming the file to the goal.

Needs to have some spurious changes stripped from it (I was trying to make saving atomic).

**Testing steps**
Manual testing. CI


parlai/utils/torch.py
```diff
@@ -42,6 +45,37 @@ def neginf(dtype: torch.dtype) -> float:
...
+
+def atomic_save(state_dict: Any, path: str) -> None:
+    """
+    Atomic save is a wrapper around torch.save, but does so in an atomic manner.
+    """
+    # atomic save is used for async writes, and it helps to yield the thread
+    time.sleep(1.0)
+    timestamp = int(time.time())
+    randint = random.randint(0, 99)
+    tmppath = path + f'.tmp.{timestamp}.{randint}'
```
sjmielke, line 74: Why not use Python stdlib's `tempfile` (https://docs.python.org/3/library/tempfile.html)? :)

stephenroller: I want it to be inside the NFS directory so the move is free. If I used tempfile, the file would be saved on local disk, and the rename into NFS would involve the same slow data transfer that's frighteningly preëmptable.

That said, maybe the best pattern would be
1. dump to local disk synchronously
2. launch a background thread that copies to NFS temp file, and does the rename at the end.

sjmielke: Makes sense, but if we were to use a `TemporaryFile(dir=os.path.dirname(path))` we would be in the target filesystem still and avoid the teeny tiny probability of a conflict---not that that is likely at all, but just feels like the right™️ thing to do 😅

stephenroller: Docs say of TemporaryFile: "It will be destroyed as soon as it is closed (including an implicit close when the object is garbage collected)."

(I don't really want a temporary file, I want a permanent file. I just want a temporary name.)

stephenroller: NamedTemporaryFile looks like it has what we need tho, I'll switch to that

PR #2761, opened 2020-06-16
Always run long CI

stephenroller:
**Patch description**
Thanks to some speed optimizations, I think it's fairly safe to run the long tests on every branch now. Let's give it a shot.

This PR deletes a bunch of the code that was controlling when tests run.

**Testing steps**
New CI.


.circleci/config.yml
```diff
@@ -448,31 +448,6 @@ jobs:
             echo 'Testing docs css'
             curl -f -i 'https://parl.ai/docs/_static/css/parlai_theme.css'
 
-  check_extra_tests:
```
EricMichaelSmith, line 451: Oh hmm are these tests duplicates of others, since they're being deleted?

codecov.yml
```diff
@@ -1,5 +1,5 @@
 coverage:
   status:
     project: off
-    patch: off
+    patch: on
```
EricMichaelSmith, line 4: What does this control?

stephenroller: I think it controls whether codecov makes the posts on PRs or not.

.circleci/config.yml
stephenroller: This was the code that parsed the [long] bits.

PR #2768, opened 2020-06-17
More ChatService internal parity

JackUrb:
**Patch description**
In an effort to get more parity between internal and external uses of chat service, this PR includes the following changes:

- Overworlds are able to access and modify an `Agent`'s `data` attribute, and this `data` is then sent to onboarding and task worlds.
- Model options passed in `opt['models']` are allowed to have a `list` of overrides, and but the models loaded from this state should select the first override setting. The remaining overrides are exposed in a new `opt['model_info']` dict, which mirrors how we expose this internally.
- *Minor* Ensured we're waiting on some asyncio calls that were never awaited (which caused some log spam)
- *Minor* Set the tornado socket to have no sending delay, which slightly reduces the outgoing message write time. Turns out this is not the source of the longer delay that seems to only effect overworlds and onboarding worlds, though I haven't tracked that down yet.

**Testing steps**
Running some of the LIGHT project under the new architecture works fine.

### DIFF
--- a/parlai/chat_service/services/messenger/messenger_manager.py
+++ b/parlai/chat_service/services/messenger/messenger_manager.py
@@ -71,9 +71,16 @@ class MessengerManager(ChatServiceManager):
         """
         if 'models' in self.opt and self.should_load_model:
             model_params = {}
+            model_info = {}
             for model in self.opt['models']:
                 model_opt = self.opt['models'][model]
+                overrides = model_opt.get('overrides', {})
+                if type(overrides) is list:
+                    print("Got list overrides!")
+                    model_opt['overrides'] = overrides[0]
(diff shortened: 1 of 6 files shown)


parlai/chat_service/services/messenger/messenger_manager.py
```diff
@@ -71,9 +71,16 @@ def _load_model(self):
         """
         if 'models' in self.opt and self.should_load_model:
             model_params = {}
+            model_info = {}
             for model in self.opt['models']:
                 model_opt = self.opt['models'][model]
+                overrides = model_opt.get('overrides', {})
+                if type(overrides) is list:
+                    print("Got list overrides!")
```
klshuster, line 79: nit: you left a print in 😄

PR #2769, opened 2020-06-17
Open-sourcing BST context generator

EricMichaelSmith:
**Patch description**
- Open-sourcing `ContextGenerator`, used to generate the contexts used to start off BST conversations when crowdsourcing the dataset
- Cleaner specification of paths to two persona files from AWS
- Fix issue in ConvAI2PersonaTopicifierTeacher: base ConvAI2 teacher now returns an Opt, requiring a .force_set() when modifying the text

**Testing steps**
- Added `tests.tasks.test_blended_skill_talk.TestContextGenerator`
- Output of `test_new_tasks.py`:
```
Ran 1 test in 391.892s

OK
```

### DIFF
--- a/parlai/tasks/blended_skill_talk/agents.py
+++ b/parlai/tasks/blended_skill_talk/agents.py
@@ -454,12 +454,17 @@ class ContextGenerator:
     BST dataset.
     """
 
-    def __init__(self, opt, datatype: str = 'train'):
+    def __init__(self, opt, datatype: str = 'train', seed: Optional[int] = None):
         """
         Initalize the context generator.
 
         opt: only a 'datapath' key is required, to specify the ParlAI data folder
         """
+
+        if seed is not None:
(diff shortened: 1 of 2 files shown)


jxmsML:
LGTM, thanks so much for the fix! For the safe persona, I think it's better to put that later in a separate pr once the episode idx issue is fixed.

EricMichaelSmith:
> LGTM, thanks so much for the fix! For the safe persona, I think it's better to put that later in a separate pr once the episode idx issue is fixed.

Thanks! Will merge in once long GPU tests pass

EricMichaelSmith:
> > LGTM, thanks so much for the fix! For the safe persona, I think it's better to put that later in a separate pr once the episode idx issue is fixed.
> 
> Thanks! Will merge in once long GPU tests pass

And yes, agreed that the safe persona work can be separate

Issue #2713, opened 2020-06-01
 Improvements to CLI interface


stephenroller:
We recently introduced the [parlai script](https://github.com/facebookresearch/ParlAI/blob/master/bin/parlai) ([actual source](https://github.com/facebookresearch/ParlAI/blob/master/parlai/scripts/parlai_exe.py)). This allows one to do things like `parlai display_data` and `parlai train_model` as shorthand for the typical `parlai -m parlai.scripts.display_data`.

We would like the following improvements to it, each which can be completed as a separate PR.

## `python -m parlai` should be equivalent to running this command.

This is done by doing adding an `if __name__ == '__main__'` clause to the parlai `__init__.py`.

## Standardize the access to scripts.

We recently introduced a [`ParlaiScript` class](https://github.com/facebookresearch/ParlAI/blob/master/parlai/scripts/script.py), which standardizes the access to the script, and adds some ways to invoke a script programmatically with kwargs instead of manually set up args. You may [see examples of it](https://colab.research.google.com/drive/1bRMvN0lGXaTF5fuTidgvlAl-Lb41F7AD#scrollTo=KtVz5dCUmFkN) being used, and how it's set up:

https://github.com/facebookresearch/ParlAI/blob/dec6f2295eb8cb6f8589c4e760221ec736283d38/parlai/scripts/interactive.py#L86-L92

However, only a handful of the existing scripts do this. We should standardize this in all scripts.

## Add subcommands so `--help` works

The best way to do this is have the main parlai script set up [sub-parsers](https://docs.python.org/3/library/argparse.html#sub-commands) which would determine which sub-command arguments it is going.

I would suggest (but one doesn't have to do this), that we add a `register_script` decorator in `loader.py` (see `register_teacher` and `register_agent` for an example). This `register_script` decorator should take a "script name" and any additional aliases it might have, for example

```python
@register_script("display_data", aliases=["dd"])
class DisplayData(ParlaiScript):
    # ...
```

Then, when the super-script is created, it can enumerate through the script registry to set up all of the major commands and their sub-parsers. If we use this approach, we will need to use importlib to automatically import everything inside parlai.scripts so that all of the registry is hit.

This registry should probably be used to also [update the documentation automatically](https://github.com/facebookresearch/ParlAI/blob/master/docs/source/generate_cli.py)

## Update references to `python examples/...` to `parlai ...` in the docs.

Hopefully most of them can be fixed with a regular expression replace in the code base.

[...]

PR #2772, opened 2020-06-18
Script registry and arg completion

vedantpuri:
**Patch description**
Changes the syntax for scripts to use a new `@register_script` syntax. This can be used to set aliases for a model as well. Scripts that are registered will automatically be added to the `--help` of the `parlai` supercommand. Furthermore, we add support for bash completion (and documentation on how to use it).

See #2713.

**Testing steps**
Manual testing. New CI. Will upload a screenshot in a bit.


stephenroller:
Closing this to signal the fact that this is far from what Vedant originally submitted, and really needs to be redefined.

klshuster:
assuming you no longer need a review from me?

PR #2784, opened 2020-06-22
Retiring some old code

stephenroller:
**Patch description**

Delete the following old code.

- one data conversion script
- memnn_feedback, which is untested and untouched for 2 years now (other than lint/black/etc).

for memnn_feedback, I left a tag and a readme saying where to find the code.

**Testing steps**
CI


stephenroller:
I thought you ran `python -m cprofile`, not using these profile scripts... Anyway okay.

PR #2785, opened 2020-06-22
Fixed ParlAI entry point

stephenroller:
**Patch description**
Fixed the entry point into parlai so that:
- Things don't catastrophically fail when you upgrade one of parlai's dependencies
- You can also run with `python -m parlai` as an equivalent to running the command. This is good because it enables @emilydinan to do `python -m pdb -m parlai` and get her debug functionality.

**_This requires everyone rerun `python setup.py develop`_**. Furthermore, we should probably release a new pypi release with this.

**Testing steps**
Manual testing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ if __name__ == '__main__':
         package_data={'': ['*.txt', '*.md']},
         entry_points={
             "flake8.extension": ["PAI = parlai.utils.flake8:ParlAIChecker"],
-            "console_scripts": ["parlai = parlai:main"],
+            "console_scripts": ["parlai=parlai.__main__:main"],
         },
         classifiers=[
             "Programming Language :: Python :: 3",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -21,7 +21,15 @@
     readme = f.read().split('--------------------')[-1]
 
 with open('requirements.txt') as f:
-    reqs = f.read()
+    reqs = []
+    for line in f:
+        line = line.strip()
+        if '>' in line:
+            reqs.append(line.split('>')[0])
+        elif '=' in line:
```
jxmsML, line 29: we won't be expecting '~=' in , '<=' equirements.txt in the future?

jxmsML: only this nit, other LGTM.

stephenroller: Does ~= work?

stephenroller: Maybe I can just use line.split('=').rstrip('><=~')...

jxmsML: yeah that could work.

PR #2786, opened 2020-06-23
Support reversed history

stephenroller:
**Patch description**
Support reversed history in TorchAgent. With reversed history, utterances are stored in recent-to-oldest, but words remain in natural order. This provides a small inductive bias for some models by having the most recent utterance always appear in an early place.

While there, I also killed the usage of the deque in the History class. This could have an increase in memory usage on very long conversations, but I think it's worth keeping them around.

**Testing steps**
Manual testing, new CI.


jaseweston:
won't this be more confusing when you read the history_str ? isn't it better to just change the representation in the transformer module (or whatever) instead rather than here? or?

stephenroller:
That would be a lot of complicated engineering for what's really a preprocessing change...

PR #2788, opened 2020-06-24
Bug fix: Multi-task with datatype train:stream

emilydinan:
**Patch description**
@klshuster found a bug where if we multi-task with `-dt train:stream`, the shorter teacher will only complete 1 epoch. This is fixed by: (1) changing the condition we check for FixedDialogTeacher setting `epoch_done` and (2) changing `self.random` for multiworld. In the multiworld setting, we want to randomly choose between subworlds when training, even if the examples in the subworld may be ordered (as in `train:stream`).

I wrote a test that fails before these changes and now passes.


stephenroller:
Why aren't tests running tho

emilydinan:
> this `is_training` is much better

@klshuster 's idea

tests/test_multiworld.py
```diff
@@ -80,3 +80,30 @@ def test_stochastic(self):
...
+            fixed_response='Hello!',
+            datatype='train:stream',
+            batchsize=16,
+        )
+        agent = create_agent(opt)
+        world = create_task(opt, agent)
+
+        for i in range(100):
+            world.parley()
+            if i % 10 == 0 and i > 0:
```
klshuster, line 102: thank you for the test. could we also inspect the `world.acts` or something to see whether the examples are nontrivial (i.e. not just the default  `{'id':'teacher', 'episode_done': True}` act)

parlai/utils/data.py
```diff
@@ -0,0 +1,22 @@
...
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Utilities related to handling data.
+"""
+
+
+def is_training(datatype: str) -> bool:
```
klshuster, line 11: 😍

stephenroller:
tbh I'm surprised that chunky teacher doesn't hang.

klshuster:
also since tests aren't running here, maybe you can run tests locally? just to check

stephenroller:
Filed a bug with CircleCI. None of Emily's PRs will run tests until they fix something.

tests/test_multiworld.py
emilydinan: done

PR #2791, opened 2020-06-25
Dialogue Unlikelihood OS commit

hadasah:
**Patch description**
Open sourcing our dialogue unlikelihood code

**Testing steps**
`pytest -v -s tests/nightly/gpu/test_unlikelihood.py`

**Logs**
```

```
Currently the ELI5 tests are failing because the ELI5 task isn't built the way the rest of our tasks are.


parlai/core/build_data.py
```diff
@@ -383,7 +383,7 @@ def download_models(
                     url += model_type + '/'
                 url += fname
             else:
-                url = path + '/' + fname
+                url = path + '/' + model_folder + '/' + fname
```
emilydinan, line 386: hmmm... this worries me. will this break anything?

parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -55,6 +55,7 @@ def add_cmdline_args(cls, argparser):
             recommended=True,
             help='if true, include image token (or no image token) for each example',
         )
+        return group
```
emilydinan, line 58: does this change belong here?

projects/dialogue_unlikelihood/agents.py
```diff
@@ -0,0 +1,619 @@
...
+
+        maxlen = self.label_truncate or 256
+        beam_preds_scores, _ = self._generate(batch, self.beam_size, maxlen)
+        preds, scores = zip(*beam_preds_scores)
+
+        self._ngram_metrics(batch, preds)
+
+        cand_choices = None
+        # TODO: abstract out the scoring here?
+        if self.rank_candidates:
```
emilydinan, line 276: would be nice if this was abstracted out in TGA so that we avoid the copypasta here

projects/dialogue_unlikelihood/agents.py
```diff
@@ -0,0 +1,619 @@
...
+    """
+
+    pass
+
+
+class RepetitionUnlikelihoodAgent(RepetitionUnlikelihoodAgentTrait, ImageSeq2seqAgent):
+    """
+    Example usage:
+
+        -t internal:generation_safety -m parlai_internal.projects.dontsaythat.agents:RepetitionUnlikelihoodAgent
```
emilydinan, line 616: have to get rid of reference to parlai_internal here and elsewhere

parlai/core/build_data.py
hadasah: ah, I can get rid of this. I was using this during localhost testing (if you follow the quip we have, you don't end up wget-ing the right file)

parlai/agents/image_seq2seq/image_seq2seq.py
hadasah: yes, my repetitionunlikelihood agent inherits from image_seq2seq because we used an image_seq2seq model as our ELI5 baseline. when I call super().add_cmdline_args() from there, I get None without this line

stephenroller:
Hmm, looks like we have issues since we can't distribute the eli5 data. Let's think about alternatives to this.

projects/dialogue_unlikelihood/agents.py
```diff
@@ -0,0 +1,619 @@
...
+
+        if batch.label_vec is not None:
+            # calculate loss on targets with teacher forcing
+            _ = self.compute_loss(batch)  # noqa: F841  we need the side effects
+
+        maxlen = self.label_truncate or 256
+        beam_preds_scores, _ = self._generate(batch, self.beam_size, maxlen)
+        preds, scores = zip(*beam_preds_scores)
+
+        self._ngram_metrics(batch, preds)
```
stephenroller, line 272: all of this copy pasta is just so we can run this method, isn't it? Can we just carve out something in TGA that we can override?

stephenroller:
Margaret can you rebase this and fix the lint errors?

stephenroller:
So go ahead and throw a @unittest.skip on the ELI5 tests, and then fix the main unit tests breaks, and I think this is ready.

parlai/core/build_data.py
stephenroller: drop this still tho

projects/dialogue_unlikelihood/agents.py
stephenroller: it still would be kinda nice to see this not be the case. idk.

hadasah: I added a call to a new method _add_generation_metrics() in TGA's eval_step(), which I implement in the unlikelihood agent. Is this good or were you thinking of something different?

stephenroller: That's awesome.

stephenroller: i don't think the add to TGA's eval_step made it in to the PR, make sure you add that.

stephenroller:
🎉🎉🎉

PR #2797, opened 2020-06-26
[ImageS2S] Early fusion images

klshuster:
**Patch description**
Support early fusion in the `image_seq2seq` agent for 1D images.

Any image types beyond something that's `[image_features_dim X 1]` dimensionality will not work with early fusion.

**Testing steps**
Tested locally via training; also added two new tests to `test_image_seq2seq`

**Logs**
```
$ python -m pytest tests/test_image_seq2seq.py
.
.
.
python -m pytest tests/test_image_seq2seq.py
============================================================================ test session starts =============================================================================
rootdir: /private/home/kshuster/ParlAI, inifile: pytest.ini
plugins: requests-mock-1.7.0
collected 5 items

tests/test_image_seq2seq.py .....                                                                                                                                      [100%]

========================================================================= slowest 10 test durations ==========================================================================
26.91s call     tests/test_image_seq2seq.py::TestImageSeq2Seq::test_multitask_early_fusion
26.64s call     tests/test_image_seq2seq.py::TestImageSeq2Seq::test_multitask
10.88s call     tests/test_image_seq2seq.py::TestImageSeq2Seq::test_image_task
8.34s call     tests/test_image_seq2seq.py::TestImageSeq2Seq::test_text_task
3.85s call     tests/test_image_seq2seq.py::TestImageSeq2Seq::test_image_task_early_fusion

================================================================= 5 passed, 166 warnings in 78.61s (0:01:18) =================================================================```

### CODE REVIEW

File: parlai/agents/image_seq2seq/modules.py [hunk 1]
```diff
@@ -228,14 +265,80 @@ def forward(
             A (full_enc, full_mask) tuple, which represents the encoded context
             and the mask
         """
+        if self.fusion is FusionType.LATE:
+            return self._forward_late_fusion(src_tokens, image_features)
+        elif self.fusion is FusionType.EARLY:
+            return self._forward_early_fusion(src_tokens, image_features)
+
+        raise RuntimeError(f'Unsupported fusion type: {self.fusion}')
```
EricMichaelSmith, line 273: nit: I always think `else: raise RuntimeError` is cleaner, but this really doesn't matter :P

parlai/agents/image_seq2seq/modules.py
```diff
@@ -228,14 +265,80 @@ def forward(
...
+                segments=torch.ones(  # type: ignore
+                    len(image_features), dtype=torch.long, device=valid_img.device
+                ),
+            )
+
+        # perform early fusion
+        tensor = self._cat([context_tensor, image_tensor])
+        mask: torch.BoolTensor = self._cat([context_mask, image_mask])  # type: ignore
+
+        # Below copied from TransformerEncoder.forward
```
EricMichaelSmith, line 309: Hmm if it's copied should we extract this out into a method of `TransformerEncoder`, perhaps? Otherwise this could quickly diverge from `TransformerEncoder` (since, for instance, I imagine Stephen likely changed this code recently given his layer norm tweak)

parlai/agents/image_seq2seq/modules.py
```diff
@@ -154,8 +169,11 @@ def _build_image_encoder(self):
             ]
         self.image_encoder = nn.Sequential(*image_layers)
 
-    def encode_images(
-        self, images: Union[List[object], torch.Tensor]
+    def embed_images(
```
EricMichaelSmith, line 172: Hm why the name change here? It's not an embedding in the sense of a `torch.nn.Embedding`-style index lookup, for instance. And it'd be good to change the docstring if we rename the function

tests/test_image_seq2seq.py
```diff
@@ -103,7 +105,41 @@ def test_multitask(self):
 
         valid, test = testing_utils.train_model(args)
         self.assertLessEqual(
-            valid['ppl'], 5.0, 'failed to train image_seq2seq on image+text task',
+            valid['ppl'], 5.0, 'failed to train image_seq2seq on image+text task'
+        )
+
+    @testing_utils.retry(ntries=3)
```
EricMichaelSmith, line 111: Great to have a check here! Whenever I write a PR I need to have a little popup that tells me, "Write a dang check"...

parlai/agents/image_seq2seq/modules.py
klshuster: happy to change

parlai/agents/image_seq2seq/modules.py
klshuster: my initial thought was that it sounded better during early fusion, as we call `forward_embedding` on the tokens. but yeah you're right, it's technically not similar and makes more sense as `encode`, so i'll change back

parlai/agents/image_seq2seq/modules.py
klshuster: This is a great point, thank you for bringing it up. 

We've already made steps to factor out the `TransformerEncoder` forward to three "steps" --> embedding input, forwarding through layers, and reducing layer outputs. we *could* further separate into *just* two steps - 1) embedding input and 2) "everything else", but it might obfuscate things a bit more. 

As you mention, we run a risk that something is wrong with the layer norms, however, the specific change you mention does not affect this code directly. Still, I'll add a warning comment here about the layer norms - I would hope that any other fundamental changes to the encoder's forward functions will be implemented in `forward_layers` or `reduce_output` anyway, which is fine in our setup

PR #2805, opened 2020-07-01
[Chunk Teacher] Allow multi-turn

emilydinan:
**Patch description**
Some changes to Chunk Teacher:
-  allow for multi-turn conversation
-  allow for ambiguous output from load_chunk
-  allow the data split to depend on other arguments in opt instead of just datatype

### DIFF
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -29,7 +29,7 @@ This module also includes ``DataLoader``, a threadpool data loader for
 structures for accessing textual dialog data and utilized by ``DialogTeacher``
 """
 import copy
-from typing import List, Tuple, Optional
+from typing import List, Tuple, Optional, TypeVar
 
 from parlai.core.agents import Agent, create_agent_from_shared
 from parlai.core.image_featurizers import ImageLoader
@@ -56,6 +56,9 @@ import os
 import torch
 import json
 import argparse
+
+
+ChunkOutput = TypeVar('ChunkOutput')
 
 
 class DataLoader(Thread):
@@ -1989,7 +1992,7 @@ class ChunkTeacher(FixedDialogTeacher, ABC):
         if opt['numthreads'] > 1:
             raise ValueError('Chunk teacher is not compatible with Hogwild.')
 
-        self.set_datasettings(opt['datatype'])
+        self.set_datasettings(opt)
 
         self.dws = int(self.opt.get('distributed_world_size', 1))
         self.rank = int(self.opt.get('rank', 0))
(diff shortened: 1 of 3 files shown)


parlai/core/teachers.py
```diff
@@ -2021,6 +2024,7 @@
             self._enqueue_request()
 
         self.episode_done = True
+        self.last_queue_output = None
```
stephenroller, line 2028: Couldn’t you just use self.opt?

parlai/core/teachers.py
```diff
@@ -2164,12 +2170,21 @@
         return output
 
     def get(self, episode_idx, entry_idx=0):
-        queue_output = self.samples.get()
-        if queue_output is None:
-            return None
+        if self.episode_done:
```
stephenroller, line 2173: Would prefer this private I think

parlai/tasks/wikipedia/agents.py
```diff
@@ -151,7 +153,7 @@ def load_from_chunk(self, chunk_idx: int) -> List[Tuple[str, str]]:
 
         return output
 
-    def create_message(self, queue_output: Tuple[str, ...]) -> 'Message':
```
stephenroller, line 154: ChunkOutput?

parlai/core/teachers.py
emilydinan: What do you mean? I actually didn't add this varialbe, I believe it's something used by FixedDialogTeacher

klshuster: we could just make this `self._episode_done` since this teacher doesn't actually use the `self.episode_done` from `FixedDialogTeacher` right? or does it?

parlai/tasks/integration_tests/agents.py
```diff
@@ -575,7 +577,7 @@ def load_from_chunk(self, chunk_idx: int):
             output.append((text, resp))
         return output
 
-    def create_message(self, sample_item):
+    def create_message(self, sample_item, entry_idx=0):
```
klshuster, line 580: it'd probably be nice to have a `MultiturnChunkyTeacher` as well in the integration tests 😄

klshuster:
Or, perhaps @stephenroller knows better (I have personally never used `ChunkTeacher` in the past so perhaps do not appreciate if this indeed breaks things)

emilydinan:
I've tested all of the tasks publicly and internally with these changes... I have a PR in internal that's ready to land once this lands. I'll make episode_done private and land

PR #2819, opened 2020-07-07
[Params] Weird opt override bug

emilydinan:
**Patch description**
@jxmsML encountered a weird bug in which if you set an argument on the command line to a negative numerical value, it will not override the default values (or whatever value was loaded from opt). This is because we have a check in there for if the first character of a value is a `-` (which is the case for negative numbers). For now I'm just getting rid of this check, since it should be captured by the checks for store true and false above...and seeing what happens.


jxmsML:
Thanks for the fix!! 
💯

PR #2825, opened 2020-07-08
Releasing style-controlled generation data

EricMichaelSmith:
**Patch description**
Releasing teachers to loop over dialogue datasets with labeled with Image-Chat personalities. See README for details

**Testing steps**
```python -m unittest tests.tasks.test_style_gen_teachers```

**test_new_tasks.py**
```
Ran 2 tests in 79.740s

OK
```

### CODE REVIEW

File: tests/tasks/test_style_gen_teachers.py [hunk 1]
```diff
@@ -0,0 +1,149 @@
...
+
+import unittest
+
+import parlai.utils.testing as testing_utils
+from parlai.agents.repeat_label.repeat_label import RepeatLabelAgent
+from parlai.core.worlds import create_task
+from parlai.scripts.display_data import setup_args
+
+
+class TestStyleGenTeachers(unittest.TestCase):
```
klshuster, line 15: thanks for the test!

EricMichaelSmith:
> could you maybe update the `test_new_tasks` script to set that flag? seems like something we'd want to have anyway
> 
> everything looks good otherwise! would like to see test output prior to merging

Yes, makes sense - added this flag.

tests/datatests/test_new_tasks.py
```diff
@@ -59,6 +60,14 @@ def test_verify_data(self):
                 parser = setup_args()
                 opt = parser.parse_args(args=['--task', subt], print_args=False)
                 opt['task'] = subt
+                if subt.split(':')[-1] == 'ParlaiformatTeacher':
```
klshuster, line 63: wait actually now i am confused; why does your test fail if you do not include this block?

EricMichaelSmith: Because the `fromfile` agent requires `--fromfile-datapath` to be set: https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/fromfile/agents.py#L81

klshuster: but don't you set it before you call `super().__init__(opt, shared)`?

EricMichaelSmith: Hmm, this flag isn't set at all if you just call `fromfile` by itself. You can see this by calling `python examples/display_data.py -t fromfile`:
```
  File "examples/display_data.py", line 19, in <module>
    DisplayData.main()
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/scripts/script.py", line 79, in main
    return cls._run_args(None)
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/scripts/script.py", line 66, in _run_args
    return script.run()
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/scripts/display_data.py", line 108, in run
    return display_data(self.opt)
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/scripts/display_data.py", line 71, in display_data
    world = create_task(opt, agent)
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/core/worlds.py", line 1627, in create_task
    world = create_task_world(opt, user_agents, default_world=default_world)
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/core/worlds.py", line 1590, in create_task_world
    task_agents = _create_task_agents(opt)
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/core/worlds.py", line 1578, in _create_task_agents
    return create_task_agent_from_taskname(opt)
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/core/teachers.py", line 2401, in create_task_agent_from_taskname
    task_agents = teacher_class(opt)
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/tasks/fromfile/agents.py", line 114, in __init__
    super().__init__(opt, shared)
  File "/private/home/ems/GitHub/facebookresearch/ParlAI_dev/parlai/tasks/fromfile/agents.py", line 82, in __init__
    raise RuntimeError('fromfile_datapath not specified')
RuntimeError: fromfile_datapath not specified
```

The interesting thing about `test_new_tasks.py` is that it'll test any teachers imported *alongside* a module even if they're not defined in the module. This happens as a result of this line: https://github.com/facebookresearch/ParlAI/blob/b9e1027a72e7baad0d21fce00a0d58c66e3c3fef/tests/datatests/test_new_tasks.py#L45 This is why I have to get this test working for `-t fromfile` even though my PR doesn't change the `fromfile` teacher

klshuster: So given that your teachers are `ParlaiFormat`, would it be possible to just have them subclass `ParlAIDialogTeacher` instead?

EricMichaelSmith: Hmm, okay, I'll try that, but I feel like if we remove the code I just added to handle `fromfile`, we'll hit this issue again whenever as someone else subclasses fromfile

klshuster: Yeah i guess this begs a bigger question of whether we want people to subclass `fromfile` when you get equivalent functionality by subclassing `ParlAIDialogTeacher`. I think `fromfile` is meant for ease of use on the command-line / quick dataset viewing and manipulation. As a real teacher, fromfile as u see just wraps these actual teachers

PR #2803, opened 2020-06-29
Self chat with two different models

dianaglzrico:
**Patch description**
Allow self chat to be performed between two different models.

**Testing steps**
To self chat with two different models, add partner like so:
```
python parlai/scripts/self_chat.py -mf zoo:pretrained_transformers/model_poly/model -p zoo:blender/blender_90M/model
```

Self chat with same model, remains the same:
```
python parlai/scripts/self_chat.py -mf zoo:blender/blender_90M/model
```

**Logs**
```
$ py parlai/scripts/self_chat.py  -mf zoo:pretrained_transformers/model_poly/model -p zoo:blender/blender_90M/model
...
[context]: Hi!
14:09:55 WARN | [ Executing eval mode with a common set of fixed candidates (n = 131438). ]
[Polyencoder]: hey
   [TransformerGenerator]: hello , how are you today ? i just got back from a long day at work , so i ' m nervous .
[Polyencoder]: ahhh i remember those days , i am good though , stressing over work though
   [TransformerGenerator]: that ' s good to hear . do you have any other hobbies you like to do ?
[Polyencoder]: not really besides animals and exercising when i can ! you ?
   [TransformerGenerator]: i like to play video games and watch netflix . what kind of animals do you exercise ?
[Polyencoder]: all animals ! i love animals . do you have any hobbies ?
   [TransformerGenerator]: yes , i love playing video games . i also like to read . what ' s your favorite book ?
[Polyencoder]: that is a hard question . i would have to say the bible
   [TransformerGenerator]: i love the bible as well . it ' s a great book . what do you do for fun ?
-- end of episode --
```

```
$ py parlai/scripts/self_chat.py  -mf zoo:blender/blender_90M/model
...
[context]: Hi!
[TransformerGenerator_1]: hello , how are you today ? i just got back from a long day at work , so i ' m nervous .
   [TransformerGenerator_2]: i am nervous as well . what do you do for work ? i hope you have a good day !
[TransformerGenerator_1]: thank you ! i work at a bank . i hope your day goes as well as mine too !
   [TransformerGenerator_2]: thank you so much ! i hope it goes well for you too . i am sure you will have a great day !
[TransformerGenerator_1]: thank you , i ' ll try my best to make it go as smoothly as possible . do you have any hobbies ?
   [TransformerGenerator_2]: i like to play video games and watch netflix . i also like to read . what about you ?
[TransformerGenerator_1]: i love video games as well ! i ' ve been playing a lot of call of duty lately .
[...]

PR #2826, opened 2020-07-09
[Agents] Create agent from model file

emilydinan:
**Patch description**
Make it easier to create an agent from model file. Before, we did not add a `datapath` or `model_subargs` which could make loading or overriding opts difficult. Also cleaned up the `create_agent_from_opt_file` function (renaming a few things and moving some stuff around). It's still a bit of a mess though TBH.

Now you can do something like:
`agent = create_agent_from_model_file('zoo:blender/blender_90M/model', {'inference': 'topk'})` and it will automatically create the agent with all of the correct arguments added and override the inference to topk.

This will be useful for M-Turk and other tasks where we have to create "subagents" (like retrieve and refine). It is also incidentally useful for #2803 in which @dianaglzrico needs to create a `partner_agent` for two models to speak to each other.

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -328,7 +328,9 @@ def create_agent_from_opt_file(opt: Opt):
     # update dict file path
     if not opt_from_file.get('dict_file'):
         opt_from_file['dict_file'] = model_file + '.dict'
-    elif opt_from_file.get('dict_file') and not os.path.isfile(opt_from_file['dict_file']):
+    elif opt_from_file.get('dict_file') and not os.path.isfile(
+        opt_from_file['dict_file']
+    ):
         old_dict_file = opt_from_file['dict_file']
         opt_from_file['dict_file'] = model_file + '.dict'
     if not os.path.isfile(opt_from_file['dict_file']):
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -273,7 +273,7 @@ def create_agent_from_model_file(model_file, opt_overides=None):
     """
     opt = {}
     add_datapath_and_model_args(opt)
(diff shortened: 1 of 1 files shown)


emilydinan:
I do not know what this mypy error is

parlai/core/agents.py
```diff
@@ -256,7 +262,7 @@ def compare_init_model_opts(opt: Opt, curr_opt: Opt):
         )
 
 
-def create_agent_from_model_file(model_file, opt_overides=None):
+def create_agent_from_model_file(model_file, opt_overrides=None):
```
klshuster, line 265: nice

parlai/core/agents.py
```diff
@@ -286,54 +293,75 @@ def create_agent_from_opt_file(opt: Opt):
     """
     model_file = opt['model_file']
     optfile = model_file + '.opt'
-    if os.path.isfile(optfile):
```
klshuster, line 289: to clarify - the only functional change here was moving this block out of the conditional statement right?

parlai/core/agents.py
```diff
@@ -266,10 +272,11 @@ def create_agent_from_model_file(model_file, opt_overides=None):
         The agent
     """
     opt = {}
-    opt['model_file'] = model_file
-    if opt_overides is None:
-        opt_overides = {}
-    opt['override'] = opt_overides
+    add_datapath_and_model_args(opt)
```
klshuster, line 275: so this is really the fundamental change in this pr right?

parlai/core/agents.py
emilydinan: yes and some renaming. i changed "new_opt" to "opt_from_file" because I thought the naming was confusing

parlai/core/agents.py
emilydinan: Yep!

PR #2829, opened 2020-07-09
Add classifier for style-controlled generation

EricMichaelSmith:
**Patch description**
- Adds code for a classifier on top of a generator model, as used in the style-gen paper
- As part of this, `ClassificationMixin` is being added as well: this was previously used to add classification metrics to a BERT ranker model. A tweak was made to `ClassificationMixin._update_confusion_matrix()` to prevent adding empty observations in a batch.
- Uploads the main classifier used in the style-gen paper to the model zoo
- Adds a wrapper teacher to move the previous+current utterances into the context and the style into the label, which is useful for training/evaluating this style classifier
- Modularizes `FixedDialogTeacher.act()` so that the steps to get the raw action can be done independently of the processing/recording of the action; needed for proper tracking of the operations performed by the wrapper teacher
- Adds the style-gen paper to the projects page

Many thanks to @candacelax for testing this code and alerting me to issues with it!

**Testing steps**
```
pytest tests/nightly/gpu/test_style_gen.py::TestClassifierOnGenerator::test_accuracy
pytest tests/nightly/gpu/test_style_gen.py::TestClassifierOnGenerator::test_simple
```


parlai/agents/style_gen/classifier.py
```diff
@@ -0,0 +1,287 @@
...
+        else:
+
+            batch_reply = super(ClassificationMixin, self).batch_act(observations)
+
+            preds = self._get_preds(batch_reply)
+            if preds is None:
+                return batch_reply
+
+            labels = 'personality'
+            labels_lst = [[label] for label in self._get_labels(observations, labels)]
```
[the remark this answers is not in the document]
stephenroller: Eee, if it's `List[List[List[str]]]` maybe we should consider another abstraction. That's getting quite deep.

tests/nightly/gpu/test_style_gen.py
```diff
@@ -0,0 +1,46 @@
...
+Test agents used with style-controlled generation.
+"""
+
+import unittest
+
+import parlai.utils.testing as testing_utils
+from parlai.core.opt import Opt
+
+
+class TestClassifierOnGenerator(unittest.TestCase):
```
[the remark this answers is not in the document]
stephenroller: +1!!

stephenroller:
Yes, I think this probably is more suited for projects.

tests/nightly/gpu/test_style_gen.py
```diff
@@ -0,0 +1,46 @@
...
+                    batchsize=32,
+                    num_epochs=5,
+                    n_layers=1,
+                    n_heads=1,
+                    ffn_size=32,
+                    embedding_size=32,
+                )
+            )
+        )
+        assert valid['accuracy'] > 0.97
```
stephenroller, line 41: Don’t use >. Use isAlmostEqual.

We are concerned with ensuring performance doesn’t drift over time, either positive or negative

EricMichaelSmith:
> Overall this looks pretty good - agree with many of the choices made.
> 
> My one big question is whether this should live in `agents` or `projects`. If the former, it might indeed be good to actually act on those `TODOs` and refactor the appropriate parent classes (TransformerDecoder, TorchClassifierAgent) to avoid duplicate code. If the latter, then this is good as is.
> 
> Mostly nits and other questions as well

Yes, makes sense - just moved this folder to `projects`

parlai/agents/style_gen/classifier.py
```diff
@@ -0,0 +1,287 @@
...
+    def add_cmdline_args(cls, argparser):
+        """
+        Add CLI args.
+        """
+        TransformerClassifierAgent.add_cmdline_args(argparser)
+        agent = argparser.add_argument_group('ClassifierOnGenerator Arguments')
+        agent.add_argument(
+            '--freeze-enc-dec-weights',
+            type='bool',
+            default=False,
```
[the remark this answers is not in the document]
EricMichaelSmith: Hmm - sometimes I've frozen those weights and sometimes I haven't. I've changed the docstring to "The generator encoder/decoder weights can be frozen during classifier training."

parlai/core/torch_classifier_agent.py
```diff
@@ -506,3 +507,142 @@ def score(self, batch):
...
+            predicted by the classifier. A string will be None if the corresponding
+            observation is empty.
+        :param labels: List of label fields from the observations. Fields may be Nones
+            if the observations are empty.
+        """
+        f1_dict = {}
+        explode_labels = []
+        for x in labels:
+            if x is not None and len(x) > 0:
+                explode_labels.append(x[0])
```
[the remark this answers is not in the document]
EricMichaelSmith: Hmm, not currently - I've added a check for this.

parlai/agents/style_gen/classifier.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # Override the criterion
+        if not shared:
+            self.criterion = self.build_criterion()
+            if self.use_cuda:
+                self.criterion.cuda()
+
+        # Freeze generator encoder/decoder weights
+        if opt['freeze_enc_dec_weights']:
+            for param in chain(
```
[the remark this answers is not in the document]
EricMichaelSmith: Essentially, yeah - I've usually seen people perform operations on parameters by iterating over a generator, given how many parameters there could be, so I wanted to have a way to just chain iterators together

parlai/agents/style_gen/classifier.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # Get predictions
+        _, prediction_id = torch.max(scores.float().cpu(), 1)
+        preds = [self.class_list[idx] for idx in prediction_id]
+
+        return Output(preds)
+
+    def eval_step(self, batch):
+        """
+        Train on a single batch of examples.
```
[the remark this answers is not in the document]
EricMichaelSmith: oof, that was the docstring for this method in TCA as well - changed

parlai/agents/style_gen/modules.py
```diff
@@ -0,0 +1,266 @@
...
+              - text_vec: (LongTensor[bsz, text seqlen])
+              - label_vec: (LongTensor[bsz, label seqlen])
+                  (Only used if not self.personality_as_label)
+        :return:
+            - the model's predicted per-class scores.
+              (FloatTensor[bsz, len(class_list)])
+        """
+
+        if self.personality_as_label:
+            # All tokens go into the encoder and classification is learned from that.
```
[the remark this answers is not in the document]
EricMichaelSmith: Added more explanation for this

parlai/agents/style_gen/modules.py
```diff
@@ -0,0 +1,266 @@
...
+        #  TransformerDecoder methods.
+
+        encoder_output, encoder_mask = encoder_state
+
+        if input is not None:
+            seq_len = input.size(1)
+            positions = input.new(seq_len).long()
+        else:
+            seq_len = embedded_input.size(1)
+            positions = embedded_input.new(seq_len).long()
```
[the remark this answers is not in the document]
EricMichaelSmith: This is because, for the baseline in my style-gen paper, I sometimes needed to pass in "soft" token embeddings (i.e. embeddings weighted over the probability distribution of the next token) instead of the tokens themselves

parlai/agents/style_gen/modules.py
```diff
@@ -0,0 +1,266 @@
...
+                embedded_input = embedded_input[:, -1:, :]
+            if positions is not None:
+                positions = positions[:, -1:]
+        else:
+            incr_state = {}
+
+        if embedded_input is not None:
+            tensor = embedded_input  # No need to copy because we only reassign below
+        else:
+            tensor = self.embeddings(input)
```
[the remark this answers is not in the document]
EricMichaelSmith: Hmm, I can't replace this with a call to `TransformerEncoder.forward_embedding()` unless that function allows token embeddings to be passed in instead of token IDs - see my note on the following comment

tests/nightly/gpu/test_style_gen.py
EricMichaelSmith: Yes - changed

EricMichaelSmith:
Converting this to a draft for now - a lot of useful points were raised about de-duplicating code in the classifier agent, which will require a bit more digging into. I'll plan to submit a more slimmed-down PR soon to just open-source the style-gen agent now, with the classifier agent to follow later.

parlai/agents/style_gen/classifier.py
```diff
@@ -0,0 +1,287 @@
...
+
+class ClassifierAgent(ClassificationMixin, TransformerGeneratorAgent):
+    """
+    Agent that uses a generator model with a classifier head.
+
+    Useful for performing classification with a pretrained generator model. Note that
+    generator encoder/decoder weights will be frozen during classifier training.
+    """
+
+    # TODO: perhaps reduce the amount of code duplicated from TorchClassifierAgent. This
```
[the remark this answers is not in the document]
EricMichaelSmith: Hmm, I've just moved this code to `projects/style_gen/classifier.py` instead; I'm not sure that there's a huge demand for subclassing TCA at the moment, yeah

parlai/core/torch_classifier_agent.py
```diff
@@ -506,3 +507,142 @@ def score(self, batch):
             class.
         """
         raise NotImplementedError('Abstract class: user must implement score()')
+
+
+class ClassificationMixin(Agent):
```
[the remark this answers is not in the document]
EricMichaelSmith: Hmm I suppose it should we put this in `parlai/core/`, but I've moved this to `projects/style_gen/modules.py` instead

parlai/core/torch_classifier_agent.py
```diff
@@ -506,3 +507,142 @@ def score(self, batch):
...
+            return None
+
+        return preds
+
+    def _get_labels(self, observations, labels):
+        labels = [obs.get(labels) for obs in observations]
+        return labels
+
+    def batch_act(self, observations):
+        # clear local metrics before anything else
```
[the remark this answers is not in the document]
EricMichaelSmith: (moved this out of `parlai/core/torch_classifier_agent.py` as well)

EricMichaelSmith:
> Your PR contains a change to a task. Please paste the results of the following command into a comment:
> 
> ```shell
> python tests/datatests/test_new_tasks.py
> ```

```
(ParlAI_dev__2004b) ems@devfair0229:~/GitHub/facebookresearch/ParlAI_dev$ python tests/datatests/test_new_tasks.py

.
Ran 1 test in 0.299s

OK
```

parlai/agents/style_gen/classifier.py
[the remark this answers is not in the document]
EricMichaelSmith: `obs['personality']` is a string, so `self._get_labels()` returns `List[str]`; thus, `labels_lst` is `List[List[str]]`. I've added a note to clarify that

EricMichaelSmith:
@stephenroller @klshuster @dexterju @dianaglzrico @emilydinan Requesting re-review, after having moved the classifier code into `projects/`

parlai/core/teachers.py
```diff
@@ -480,6 +488,13 @@ def act(self):
         # TODO: all teachers should eventually create messages
         # while setting up the data, so this won't be necessary
         action = Message(action)
+
+        return action
+
+    def process_action(self, action: Message) -> Message:
```
klshuster, line 494: what is the purpose of adding these functions at such a high level? if it is important to keep, can we get a little more thorough documentation (why does this need to be split up into `get_orig_action` and `process_action`?

parlai/tasks/wrapper/agents.py
```diff
@@ -40,7 +41,13 @@ def add_cmdline_args(cls, parser):
             help='The task whose fields will be manipulated.',
         )
         known_args, _ = parser.parse_known_args(nohelp=True)
-        parser.add_task_args(known_args.wrapper_task)
+        try:
+            parser.add_task_args(known_args.wrapper_task)
+        except RuntimeError:
+            warn_once(
```
klshuster, line 47: why don't we want this to raise an error?

parlai/tasks/style_gen/agents.py
```diff
@@ -88,3 +89,45 @@ def __init__(self, opt, shared=None):
...
+
+    Because the dialogue history is effectively overwritten by this action, all episodes
+    will be flattened into one example each.
+    """
+
+    def __init__(self, opt: Opt, shared=None):
+        super().__init__(opt, shared)
+        assert isinstance(self.task, FixedDialogTeacher)
+
+    def act(self):
```
klshuster, line 112: this looks quite similar to the `LabelsToText` teacher in the `wrapper/agents.py` file; is the only difference that we set the label to `act['personality']`?

parlai/zoo/model_list.py
```diff
@@ -1461,14 +1461,27 @@
...
     },
+    {
+        "title": "Style-controlled generation: previous and current utterance classifier",
+        "id": "style_gen",
+        "path": "zoo:style_gen/prev_curr_classifier/model",
+        "agent": "projects.style_gen.classifier:ClassifierAgent",
+        "task": "style_gen:LabeledBlendedSkillTalk",
+        "project": 'https://github.com/facebookresearch/ParlAI/tree/master/projects/style_gen',
+        "description": "Classifier trained on Image-Chat turns 2 and 3 to classify the personality of an example given that utterance and the previous utterance.",
+        "example": "parlai eval_model --task style_gen:PrevCurrUttStyle --wrapper-task style_gen:LabeledBlendedSkillTalk --model-file zoo:style_gen/prev_curr_classifier/model --model projects.style_gen.classifier:ClassifierAgent --classes-from-file image_chat",
```
klshuster, line 1480: `--classes-from-file image_chat` seems a little off, maybe `image_chat_personalities_file`?

projects/style_gen/modules.py
```diff
@@ -57,3 +71,319 @@ def get_temp_history(self, observation: Message) -> Optional[str]:
...
+    """
+    Adds a 'personalities' field to the batch in the case where personality information
+    is not encoded in any other field.
+    """
+
+    def __init__(self, personalities=None, **kwargs):
+        super().__init__(personalities=personalities, **kwargs)
+
+
+class TransformerDecoderWithEmbeds(TransformerDecoder):
```
klshuster, line 176: This solution for passing in an `embedded_input` definitely works; another way you can get around this is by just subclassing `forward_embedding` to return the `input` directly (i.e. bypass the actual `embedding` phase); this would allow you to not have to repeat some of the decoder forward logic

projects/style_gen/modules.py
```diff
@@ -57,3 +71,319 @@ def get_temp_history(self, observation: Message) -> Optional[str]:
...
+                tensor, encoder_output, encoder_mask, incr_state
+            )
+        else:
+            for idx, layer in enumerate(self.layers):
+                tensor, new_incr_state[idx] = layer(
+                    x=tensor,
+                    encoder_output=encoder_output,
+                    encoder_mask=encoder_mask,
+                    incr_state=incr_state.get(idx),
+                )
```
klshuster, line 233: note that this is now `forward_layers` in the transformer decoder

projects/style_gen/modules.py
```diff
@@ -57,3 +71,319 @@ def get_temp_history(self, observation: Message) -> Optional[str]:
...
+            tensor = _normalize(tensor, self.norm_embeddings)
+        if positions.max().item() > self.n_positions:
+            warn_once(
+                'You are inputting a sequence of {x} length, but only have '
+                '--n-positions {y}. Set --truncate or increase --n-positions'.format(
+                    x=positions.max().item(), y=self.n_positions
+                )
+            )
+        tensor = tensor + self.position_embeddings(positions).expand_as(tensor)
+        tensor = self.dropout(tensor)  # --dropout
```
klshuster, line 219: yeah so this whole block is now in `forward_embedding` in `TransformerDecoder`

projects/style_gen/modules.py
klshuster: I wrote a few notes below about how some of the logic has been factored out but since this is a project agent it's probably good (if not better) to keep as is for maximum reproducibility (as those functions may change in the future...)

projects/style_gen/modules.py
```diff
@@ -57,3 +71,319 @@ def get_temp_history(self, observation: Message) -> Optional[str]:
...
+        if all(x is None for x in preds):
+            return None
+
+        return preds
+
+    def _get_labels(self, observations, labels_field: str):
+        labels = [obs.get(labels_field) for obs in observations]
+        return labels
+
+    def batch_act(self, observations):
```
klshuster, line 310: is there a difference between this and `TorchAgent.batch_act`? just curious

klshuster (commented):
sorry for the late review, all looks pretty good to me for merging into `projects`, not sure if u wanted to wait for other reviews as well

had a few questions, but again since this is going into `projects` i'm good with the current setup. my major question is regarding the change to `FixedDialogTeacher`

parlai/core/teachers.py
EricMichaelSmith: The purpose of adding these functions at such a high level is because any wrapper teacher that modifies the 'text' or 'label' fields will have to do so before self.last_act or self.lastY are set, or otherwise the action will become out of sync with those attributes. (The alternative is to simply duplicate the logic of FixedDialogTeacher.act() in any wrapper teacher, but this will silently break whenever this method is changed in the future.) I've added a comment to the `.get_orig_action()` method to indicate why these methods are split up the way that they are; happy to discuss this more if need be

parlai/tasks/wrapper/agents.py
EricMichaelSmith: Because there is still a use case for running this code without adding task args, specifically when using a wrapper teacher as part of a unit test in which args have to be passed in as an opt into `testing_utils.eval_model()` as opposed to using ParlaiParser

parlai/tasks/style_gen/agents.py
EricMichaelSmith: Yes - and also, here we truncate the context history to only the previous utterance, which is necessary for how we trained the style-gen classifier

parlai/zoo/model_list.py
EricMichaelSmith: Fair enough - changed

projects/style_gen/modules.py
EricMichaelSmith: > This solution for passing in an `embedded_input` definitely works; another way you can get around this is by just subclassing `forward_embedding` to return the `input` directly (i.e. bypass the actual `embedding` phase); this would allow you to not have to repeat some of the decoder forward logic

Yeah good point! I think this code pre-dates some of that more recent refactoring

projects/style_gen/modules.py
EricMichaelSmith: Hmm, there isn't anymore, due to a fix that Candace made a few days ago. I've deleted this `batch_act()` because it's no longer useful

EricMichaelSmith:
Thanks for the very useful comments @klshuster ! Addressed them and am requesting re-review

parlai/core/teachers.py
klshuster: so is it necessary for anyone who subclasses the `AbstractWrapperTeacher` to remember to call `get_orig_action` and then return `self.process_action`?

EricMichaelSmith: It won't be necessary to do that manually - my plan is to refactor `AbstractWrapperTeacher` after this PR is in in order to call `get_orig_action()` and return `self.process_action()` in that abstract teacher's `.act()` directly. I plan to do this because I expect that all subclassed wrapper teachers will likely need to add their act-modifying code in between those two methods

klshuster: Ahh yes great, that would make this less prone to user error

klshuster (approved):

PR #2840, opened 2020-07-14
[BART/TGA] BART and TGA API Updates

klshuster:
**Patch description**
BART API Updates:

1. Factor out functionality for getting conversion args, in case a subclassed model would like to add different conversion args.
2. Override `BartAgent.vectorize` to ensure that start token is explicitly in labels.
3. Override `BartModel.decode_forced` to bypass assertion about double BOS (since this is irrelevant for BART, as we actually start with EOS)

TGA API Updates:

1. Factor out initial forced decoder input, similarly to how `_initial_decoder_input` is factored out.
2. ~`TGA._get_initial_decoder_input` now returns the `inds_sz`, i.e. the number of batch indices to keep track of. This can vary with more complex decoding schemes.~ **EDIT: This change is no longer necessary**
3. In `TGA._generate`, there is now a call to `TGA._get_next_decoder_input` to determine the next output for the decoder.

The main goal of these changes is to allow flexibility of subclasses to modify decoder input/output accordingly

**Testing steps**
Relying on existing tests here.


parlai/agents/bart/modules.py
```diff
@@ -25,3 +26,27 @@ def output(self, tensor: torch.Tensor) -> torch.Tensor:
...
         output = F.linear(tensor, self.embeddings.weight)
         return output
+
+    def decode_forced(
+        self, encoder_states: Tuple[Any, ...], ys: torch.LongTensor
+    ) -> Tuple[torch.FloatTensor, torch.LongTensor]:
+        """
+        Decode with a fixed, true sequence, computing loss.
+
+        Override to bypass assertion that BOS is not present, and additionally insert
```
EricMichaelSmith, line 36: nit: maybe "Overriding TGA.decode_forced() to bypass"? At first I thought this was a command, i.e. "You, the user, should override this method if you want to bypass..."

klshuster:
@EricMichaelSmith I addressed your comment however I also reverted one of the changes to TGA API (see updated description)

klshuster:
the gpu test i am concerned about is the BART one, which I ran locally:

```
$ pytest -k TestBartModel
=============================================================================== test session starts ===============================================================================
rootdir: /private/home/kshuster/ParlAI, inifile: pytest.ini, testpaths: tests
plugins: requests-mock-1.7.0
collected 359 items / 358 deselected / 1 selected

tests/nightly/gpu/test_bart.py .                                                                                                                                            [100%]

============================================================================ slowest 10 test durations ============================================================================
97.41s call     tests/nightly/gpu/test_bart.py::TestBartModel::test_bart

=========================================================== 1 passed, 358 deselected, 30 warnings in 101.19s (0:01:41) ============================================================
```

Going to merge

PR #2842, opened 2020-07-16
Abstract wrapper

dianaglzrico:
**Patch description**
Creation of an abstract teacher that will work as base for any wrapper teacher; existent `LabelToTextTeacher` now uses this class.

**Logs**
```
$ python examples/display_data.py -t wrapper:labelToText -wt blended_skill_talk

- - - NEW EPISODE: blended_skill_talk - - -
that is ok.  have any kids?
- - - NEW EPISODE: blended_skill_talk - - -
that is good. I have 2
- - - NEW EPISODE: blended_skill_talk - - -
that is great! you will be ready
10:35:49 INFO | loaded 27018 episodes with a total of 27018 examples
```

### DIFF
--- a/parlai/tasks/wrapper/agents.py
+++ b/parlai/tasks/wrapper/agents.py
@@ -28,7 +28,6 @@ class AbstractWrapperTeacher(Teacher):
     Abstract teacher working as base for wrapper teachers.
     """
 
-    @abstractmethod
     def __init__(self, opt: Opt, shared=None):
         pass
 
--- a/parlai/tasks/wrapper/agents.py
+++ b/parlai/tasks/wrapper/agents.py
@@ -23,37 +23,20 @@ from parlai.core.opt import Opt
 from parlai.core.teachers import create_task_agent_from_taskname, Teacher
 
 
-class LabelToTextTeacher(Teacher):
+class AbstractWrapperTeacher(Teacher):
     """
-    Teacher that will shift message['labels'][0] into message['text'] for whatever task
-    is specified with --label-to-text-task.
-
-    Because the dialogue history is effectively overwritten by this action, all episodes
-    will be flattened into one example each.
+    Abstract teacher working as base for wrapper teachers.
     """
 
-    @classmethod
-    def add_cmdline_args(cls, argparser):
-        # Teacher has no args
-        parser = argparser.add_argument_group('LabelToText args')
-        parser.add_argument(
-            '-l2tt',
-            '--label-to-text-task',
-            type=str,
-            help='The task whose labels will get shifted into the text field',
-        )
-
+    @abstractmethod
     def __init__(self, opt: Opt, shared=None):
-        if ',' in opt['task']:
-            raise ValueError('LabelToTextTeacher cannot be used with multiple tasks!')
-        self.id = opt['task']
-        self.opt = opt
-        if shared and 'task' in shared:
-            self.task = create_agent_from_shared(shared['task'])
-        else:
-            opt_singletask = copy.deepcopy(opt)
-            opt_singletask['task'] = opt['label_to_text_task']
-            self.task = create_task_agent_from_taskname(opt_singletask)[0]
+        pass
+
+    def act(self):
+        """
+        Act on the previous observation.
+        """
+        raise NotImplementedError('Abstract class: user must implement act() method')
 
     def num_examples(self):
         """
@@ -74,27 +57,6 @@ class LabelToTextTeacher(Teacher):
         Make an observation.
         """
         return self.task.observe(observation)
-
-    def act(self):
-        """
-        Act on the previous observation.
-        """
-        act = self.task.act()
-        new_act = copy.deepcopy(act)
-        if 'labels' in act:
-            labels = act['labels']
-            assert len(labels) == 1
-            new_act.force_set('text', labels[0])
-            new_act.force_set('labels', [''])
-        elif 'eval_labels' in act:
-            labels = act['eval_labels']
-            assert len(labels) == 1
-            new_act.force_set('text', labels[0])
-            new_act.force_set('eval_labels', [''])
-        else:
-            assert 'text' not in act and act['episode_done'] is True
-        new_act.force_set('episode_done', True)  # Clear the dialogue history
-        return new_act
 
     def epoch_done(self):
         """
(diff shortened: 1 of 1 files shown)


parlai/tasks/wrapper/agents.py
```diff
@@ -144,3 +105,58 @@ def shutdown(self):
...
+    will be flattened into one example each.
+    """
+
+    @classmethod
+    def add_cmdline_args(cls, argparser):
+        # Teacher has no args
+        parser = argparser.add_argument_group('LabelToText args')
+        parser.add_argument(
+            '-l2tt',
+            '--label-to-text-task',
```
stephenroller, line 125: maybe we can change the argument to `--wrapper-task`  and then `__init__` could be implemented in the abstract teacher. Just an idea.

dianaglzrico: I had that exact name before, lol. But yes, changing it back and implementing `__init__` :) thanks

parlai/tasks/wrapper/agents.py
```diff
@@ -23,37 +23,19 @@
...
-        self.opt = opt
-        if shared and 'task' in shared:
-            self.task = create_agent_from_shared(shared['task'])
-        else:
-            opt_singletask = copy.deepcopy(opt)
-            opt_singletask['task'] = opt['label_to_text_task']
-            self.task = create_task_agent_from_taskname(opt_singletask)[0]
+        pass
+
+    def act(self):
```
EricMichaelSmith, line 34: You can slap a `@abstractmethod` decorator here to signify that this method needs to be defined by all subclasses :)

parlai/tasks/wrapper/agents.py
```diff
@@ -144,3 +105,58 @@ def shutdown(self):
...
+        if shared and 'task' in shared:
+            self.task = create_agent_from_shared(shared['task'])
+        else:
+            opt_singletask = copy.deepcopy(opt)
+            opt_singletask['task'] = opt['label_to_text_task']
+            self.task = create_task_agent_from_taskname(opt_singletask)[0]
+
+
+    def act(self):
+        """
```
EricMichaelSmith, line 144: (good to take a look at these lint errors =P )

parlai/tasks/wrapper/agents.py
```diff
@@ -23,37 +23,19 @@
 from parlai.core.teachers import create_task_agent_from_taskname, Teacher
 
 
-class LabelToTextTeacher(Teacher):
+class AbstractWrapperTeacher(Teacher):
```
emilydinan, line 26: maybe make this an abstract base class (ABC) since it has abstract methods that need to be implemented by children https://docs.python.org/3/library/abc.html

parlai/tasks/wrapper/agents.py
```diff
@@ -144,3 +105,58 @@ def shutdown(self):
...
+            type=str,
+            help='The task whose labels will get shifted into the text field',
+        )
+
+    def __init__(self, opt: Opt, shared=None):
+        if ',' in opt['task']:
+            raise ValueError('LabelToTextTeacher cannot be used with multiple tasks!')
+        self.id = opt['task']
+        self.opt = opt
+        if shared and 'task' in shared:
```
emilydinan, line 135: when would 'task' not be in shared? should this raise an error?

parlai/tasks/wrapper/agents.py
```diff
@@ -144,3 +105,58 @@ def shutdown(self):
...
+            self.task = create_task_agent_from_taskname(opt_singletask)[0]
+
+
+    def act(self):
+        """
+        Act on the previous observation.
+        """
+        act = self.task.act()
+        new_act = copy.deepcopy(act)
+        if 'labels' in act:
```
emilydinan, line 149: can condense these two blocks since only 'labels' or 'eval_labels' will ever be in act at a given time. `labels = 'labels' if 'labels' in act else 'eval_labels'`

parlai/tasks/wrapper/agents.py
EricMichaelSmith: Good point, we could remove `'task' in shared` and possibly raise an assert (or just let it fail if 'task' is not in shared)

parlai/tasks/wrapper/agents.py
```diff
@@ -18,43 +18,49 @@
...
+    def add_cmdline_args(cls, parser):
+        parser = parser.add_argument_group('AbstractWrapper args')
         parser.add_argument(
-            '-l2tt',
-            '--label-to-text-task',
+            '-wt',
+            '--wrapper-task',
             type=str,
-            help='The task whose labels will get shifted into the text field',
+            help='The task which fields will be manipulated.',
```
EricMichaelSmith, line 40: tiny nit: Maybe `'The task whose fields will be manipulated`?

parlai/tasks/wrapper/agents.py
```diff
@@ -144,3 +129,33 @@ def shutdown(self):
...
+    def __init__(self, opt: Opt, shared=None):
+        super().__init__(opt, shared)
+
+    def act(self):
+        """
+        Act on the previous observation.
+        """
+        act = self.task.act()
+        new_act = copy.deepcopy(act)
+        if 'labels' in act or 'eval_labels' in act:
```
EricMichaelSmith, line 152: I like Emily's suggestion - this is cleaner :) In fact, you could clean it further by doing
```
labels_type = 'labels' if 'labels' in act else 'eval_labels'
labels = act[labels_type]            
```

parlai/tasks/wrapper/agents.py
dianaglzrico: woops. fixed it

parlai/tasks/wrapper/agents.py
dianaglzrico: done :)

parlai/tasks/wrapper/agents.py
```diff
@@ -18,43 +18,51 @@
...
-        # Teacher has no args
-        parser = argparser.add_argument_group('LabelToText args')
-        parser.add_argument(
-            '-l2tt',
-            '--label-to-text-task',
+    def add_cmdline_args(cls, parser):
+        agent = parser.add_argument_group('AbstractWrapper args')
+        agent.add_argument(
+            '-wt',
+            '--wrapper-task',
```
EricMichaelSmith, line 38: Ah, so there's one last annoying step needed here - since you're changing the name of an existing flag, you'll need to change this name in all of the places in the ParlAI code where this flag already exists, or otherwise that code will break once this PR is merged in. Fortunately this should be pretty simple - you can search for all places in ParlAI base and ParlAI-Internal where the strings `--label-to-text-task` and `label_to_text_task` appear, and change them to `--wrapper-task` and `wrapper_task`, respectively. (Your IDE should have an easy way to search over the entire repos - lemme know if it's not working for you :) )

parlai/tasks/wrapper/agents.py
```diff
@@ -144,3 +131,33 @@ def shutdown(self):
...
+    def act(self):
+        """
+        Act on the previous observation.
+        """
+        act = self.task.act()
+        new_act = copy.deepcopy(act)
+        if 'labels' in act or 'eval_labels' in act:
+            labels_type = 'labels' if 'labels' in act else 'eval_labels'
+            labels = act[labels_type]
+            assert len(labels) == 1
```
emilydinan, line 157: why this assert here?

emilydinan, line 157: why this assert here? this is not universally true in ParlAI, so if required we should add a more descriptive error message

EricMichaelSmith: Okay cool - this was probably originally my line, because I've never seen a case in which there's more than one label. But yes, generalizing this seems like a good idea, then - it's good to future-proof this teacher

dianaglzrico: Yeah, this teacher is @EricMichaelSmith code. Should I add the descriptive message stating it was created for the specific task you created it for and that it can be comment out if needed?

EricMichaelSmith: hmm - maybe you could do
```
if len(labels) != 1:
    raise ValueError('LabelToTextTeacher can only be used with one label!')
```
and then in the future, if there's a need to use this teacher with more than one label, we can figure out what logic we want to use for that then

PR #2775, opened 2020-06-19
Distributed Evaluation

stephenroller:
**Patch description**
Allow for sped up evaluation via distributed evaluation.

Involves a refactor of distributed.

**Testing steps**
Manual testing. New CI.

PR #2850, opened 2020-07-18
Fix broken distributed.

stephenroller:
**Patch description**
Distributed broken in #2775. We don't have tests for distributed (only multiprocessing) so it wasn't caught.

**Testing steps**
Manually launched a sweep.


stephenroller:
Failed tests are flukes, but this needs to be merged fast.

PR #2851, opened 2020-07-18
Faster amazon QA teacher

### LINKED ISSUE #2848: Building Dict is too slow
**Bug description**
```
from parlai.scripts.train_model import TrainModel

TrainModel.main(
    # similar to before
    task='amazon_qa',
    model='projects.wizard_of_wikipedia.generator.agents:EndToEndAgent',
    model_file='/tmp/end2end_generator/model',

    # initialize with a pretrained model
    init_model='zoo:wizard_of_wikipedia/end2end_generator/model',

    # arguments we get from the pretrained model.
    # Unfortunately, these must be looked up separately for each model.
    # eps
    dict_lower=True,
    dict_tokenizer='bpe',
    n_layers=5,
    n_heads=2,
    dropout=0.20,
    ffn_size=512,
    embedding_size=256,
    log_every_n_secs=10,
    validation_patience=12,
    validation_metric='ppl',
    validation_metric_mode='min',
    validation_every_n_epochs=0.5,
    n_positions=128,
    truncate=128,
    max_knowledge=32,
    knowledge_alpha=0.95,
    knowledge_truncate=32,
    learningrate=5e-4,
    warmup_updates=5000,
    clip=0.1,
    lr_scheduler='invsqrt',
    embedding_type='fasttext',
    beam_size=1,
    skip_generation=False,
    batchsize=64,
)

```
I am trying to train amazon_qa task on wizard of Wikipedia model, just to experiment it out, I am not sure if it will work but when I run this script it says creating a task and goes to next stage building dictionary it just becomes to slow of a process.

Screenshot 2020-07-18 at 10 42 32 PM

Like has been around 2 hrs and it hasn't crossed 0% yet.
Can anyone please point me out the error, thanks.

stephenroller:
**Patch description**
The AmazonQA teacher was silly slow. Fixes #2848.

**Testing steps**
Manual testing.


parlai/tasks/amazon_qa/agents.py
```diff
@@ -26,11 +26,14 @@ def __init__(self, opt, shared=None):
             self._setup_data()
         self.reset()
 
+        self.num_ex = sum([len(x) for x in self.data])
```
emilydinan, line 29: this can also be slow because you're calculating it for every teacher in the batch. might want to just do it in the main teacher and "share" the number.

PR #2853, opened 2020-07-19
Teacher tests.

stephenroller:
**Patch description**
Throwing $$ out the window. What happens if we just aggressively test our teachers by just running display_data on many of them.

**Testing steps**
New CI


parlai/tasks/convai2/agents.py
```diff
@@ -105,93 +105,11 @@ def normalize_replies(self, x):
...
-                        start = False
-                    else:
-                        yield split, False
-                    # reset x in case there is unlabeled data still left
-                    x = ''
-                    reward = 0
-            if x:
-                x = self.normalize_replies(x)
-                yield [x, None, reward], start
+        for (text, labels, reward, candidates), new_episode in super().setup_data(path):
```
EricMichaelSmith, line 108: Hmm - is this this duplicated logic? Great that we can remove so much formatting code...

parlai/tasks/ubuntu/agents.py
```diff
@@ -113,70 +113,5 @@ def share(self):
         return shared
 
 
-class UbuntuTeacherWithNegs(UbuntuTeacher):
```
EricMichaelSmith, line 116: This is not needed anymore?

parlai/utils/testing.py
```diff
@@ -364,3 +364,96 @@ def display_model(opt) -> Tuple[str, str, str]:
...
+            task=self.task, datatype=datatype, display_verbose=True, **kwargs
+        )
+
+    def test_train(self):
+        """
+        Test --datatype train.
+        """
+        return self._run_display_data('train')
+
+    def test_train_stream(self):
```
EricMichaelSmith, line 383: This especially is good to have, because I've had teachers that would specifically break in the streaming case...

parlai/tasks/ubuntu/agents.py
stephenroller: Untested and didn’t work. So kill it lol

PR #2857, opened 2020-07-20
Try to fix codecovbot.

stephenroller:
**Patch description**
I don't know what's going on here. I'm just throwing stuff at the wall.

**Testing steps**
Maybe it posts in this PR. Probably we have to merge to master.


stephenroller:
Talked to support; it's on their side.

PR #2879, opened 2020-07-26
Tie off release version 0.8.0

stephenroller:
**Patch description**
Tie off a release version, 0.8.0. We are done with date-based versioning. The number was an arbitrary start.

**Testing steps**
Nada.


setup.py
```diff
@@ -30,7 +27,7 @@
 if __name__ == '__main__':
     setup(
         name='parlai',
-        version='0.1.{DATE}{BUILD}'.format(DATE=DATE, BUILD=BUILD),
+        version='0.6.0',
```
klshuster, line 30: shouldn't we be pulling this from the `__init__` file?

stephenroller: i couldn't figure out how. setup.py happens _before_ parlai is installed, so it can't actually read the init.py. pytorch distributes a "version.txt" that would be nice but

klshuster: i'm fine with whatever, just would need to make sure we update both versions each time (perhaps leave a comment in `__init__` and `setup.py` indicating that?)

stephenroller: Done

PR #2881, opened 2020-07-27
Support logging in interactive script

jonathan-conder:
**Patch description**
The interactive script ostensibly supports logging, but some of the options required to make this work are missing. The first patch adds these missing options and (rather than also adding `--report-filename`) replaces `--save-world-logs` with `--outfile`.
After this, logging sort of works, but the last exchange in an episode is logged twice. The second patch fixes that. The third patch adds an interactive logging test (which would have failed before the second patch). The fourth patch just fixes a typo.

**Testing steps**
Run `tests/test_interactive.py` or try `python -m parlai.scripts.interactive -mf <model> --outfile /tmp/log.jsonl && cat /tmp/log.jsonl | jq .`


stephenroller:
Hmm yeah the only troublesome issue is in distributed, we have to add the rank to the filename somewhere

jonathan-conder:
thanks!

PR #2884, opened 2020-07-27
[Bug] Fix `log.warn`

klshuster:
**Patch description**
The correct syntax (now) for `logging.warning` is `logging.warn`

**Testing steps**
Tested locally with agents that defined `self.temp_history`

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -299,7 +299,7 @@ class History(object):
         Return the string version of the history.
         """
         if self.temp_history and self.history_strings:
-            logging.warning('temporary history strings now include the delimiter.')
+            logging.warn('temporary history strings now include the delimiter.')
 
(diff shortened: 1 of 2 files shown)


parlai/core/torch_agent.py
```diff
@@ -299,7 +299,7 @@ def get_history_str(self):
         Return the string version of the history.
         """
         if self.temp_history and self.history_strings:
-            logging.warning('temporary history strings now include the delimiter.')
+            logging.warn('temporary history strings now include the delimiter.')
```
stephenroller, line 302: i keep trying to do both.

We should just make a helper method in logging that allows warning

klshuster: done

klshuster:
failing tests are unrelated (TestSelfChat and TestWrapper)

PR #2889, opened 2020-07-27
Fix bug where override wasn't playing with subparsers.

stephenroller:
**Patch description**
Use a dirty hacks so that are subparsers are propagated to the supercommand parser when we need it. This allows us to reuse our existing (terrible) method for `_process_args_to_opts` that we use to identify which commands were set via command line, and which are coming from defaults.

Alternatives considered:
- Run `parse_args([])` to identify defaults. This fails if the user manually sets a value to the default, but which differs from the opt.
- Trying to hunt through the subparsers to find the one used
- Temporarily overriding `argparse.SUPRESS` so that the returned Namespace only is filled with values manually set by the user. This one fails for the same reason as the original bug.

A potentially cleaner solution future solutions:
- One could imagine would be that `add_argument` would override the default of the action to `argparse.SUPRESS` and update `_defaults` with the true default, which is manually setattr'd by argparse. we could then temporarily hide all of `_defaults` to get the ones set by argparse
- Modifying the argparse Actions `__call__` method to set a "touched" attribute in the namespace, so that the namespace would clearly tell you what actions were touched.
- Perhaps adjusting the values of `_defaults` to have magic attributes? Many would be special python types that you cannot setattr on these.
- Probably the cleanest: override `_parse_known_args` to remember namespace before all the actions are utilized.

**Testing steps**
Manual testing 🙈

```
devfair0237 parlai fixoverridesubparser $ parlai eval_model -t convai2 -mf zoo:blender/blender_90M/model -bs 4 --inference topk
19:44:20 | Overriding opt["task"] to convai2 (previously: internal:blended_skill_talk,wizard_of_wikipedia,convai2,empathetic_dialogues)
19:44:20 | Overriding opt["model_file"] to /private/home/roller/working/parlai/data/models/blender/blender_90M/model (previously: /checkpoint/edinan/20200210/baseline_BST_retnref/lr=7.5e-06_attention-dropout=0.0_relu-dropout=0.0/model)
19:44:20 | Overriding opt["batchsize"] to 4 (previously: 16)
19:44:20 | Overriding opt["inference"] to topk (previously: beam)


devfair0237 parlai fixoverridesubparser $ python -m parlai.scripts.eval_model -t convai2 -mf zoo:blender/blender_90M/model -bs 4 --inference topk
19:44:44 | Overriding opt["datatype"] to valid (previously: train)
19:44:44 | Overriding opt["task"] to convai2 (previously: internal:blended_skill_talk,wizard_of_wikipedia,convai2,empathetic_dialogues)
19:44:44 | Overriding opt["model_file"] to /private/home/roller/working/parlai/data/models/blender/blender_90M/model (previously: /checkpoint/edinan/20200210/baseline_BST_retnref/lr=7.5e-06_attention-dropout=0.0_relu-dropout=0.0/model)
19:44:44 | Overriding opt["batchsize"] to 4 (previously: 16)
19:44:44 | Overriding opt["inference"] to topk (previously: beam)
```


emilydinan:
can you explain what the problem was before / when it happeneD?

klshuster:
@emilydinan i'll give some insight - suppose you run 

```
$ parlai interactive -mf zoo:blender/blender_90M/model --inference beam --beam-size 5
``` 

In the current setup, these args are parsed directly into the `opt` dictionary, (i.e. `opt['inference'] == 'beam'`). When we go to [create an agent from the opt file](https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/agents.py#L308), we *only* override args in the opt file that are specified in `opt['override']`. Unfortunately, `inference` and `beam_size`, as they are parsed directly into `opt`, are not in `opt['override']`, and thus the cmd line args are not respected

PR #2905, opened 2020-07-30
Pytest README

emilydinan:
**Patch description**
Instructions for running tests locally with pytest.


stephenroller:
Yes, that option also exists.

PR #2907, opened 2020-07-30
Add return in report function of wrapper agent

dianaglzrico:
**Patch description**
Agent was breaking because of missing `return` when calling `report()`.

**Testing steps**
Unit test:

```
$ python -m unittest tests.test_wrapper
16:06:36 INFO | creating task(s): wrapper:labelToTextTeacher
.
Ran 1 test in 0.092s

OK
```

As stated [here](https://github.com/facebookresearch/ParlAI/pull/2642) this teacher doesn't (and is not expected to) pass `python tests/datatests/test_new_tasks.py` because of the nature of this task (read link's PR description for a more detailed explanation).

### DIFF
--- a/parlai/tasks/wrapper/agents.py
+++ b/parlai/tasks/wrapper/agents.py
@@ -93,7 +93,7 @@ class AbstractWrapperTeacher(Teacher, ABC):
         """
         Report metrics for the subtask.
         """
-        self.task.report()
+        return self.task.report()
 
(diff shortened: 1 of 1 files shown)


dianaglzrico:
> Great catch! (Make sure to paste in the output from `python tests/datatests/test_new_tasks.py` :) )

added unittest and explanation!

PR #2445, opened 2020-03-03
Bug fixes for multiworker validation.

stephenroller:
**Patch description**
Previous multiworker validation synced the reports returned by train_model, but logs contained the incorrect number of examples. This meant what was being written to train stats was correct, but not what was being output to stdout.

Also caught a small bug where many workers would smash the .valid and .test files.

For review ease, only submitting the bugfix for review, and then I'll include the reverted commit before I land.

**Testing steps**
Manually trained and evaluated a small convai2 model.

```
[ time:136.0s total_exs:253952 epochs:1.93 ] {'exs': 3840, 'loss': 5.35, 'ppl': 210.5, 'token_acc': 0.2312, 'tokens_per_batch': 3366.0, 'gnorm': 1.079, 'clip': 1.0, 'updates': 15, 'lr': 1.0, 'gpu_me
m_percent': 0.5557, 'total_train_updates': 992}
[ running eval: valid ]
[ eval completed in 2.47s ]
valid:{'exs': 7801, 'loss': 5.398, 'ppl': 221.0, 'token_acc': 0.2336, 'tokens_per_batch': 3203.0, 'lr': 1.0, 'gpu_mem_percent': 0.5557, 'total_train_updates': 992}
[ new best ppl: 221.01678489721164 (previous best was 242.06570622832945) ]
```

PR #2433, opened 2020-02-25
Agent metrics mw

mwillwork:
**Patch description**
Updated distributed training to allow all the workers to do evaluation (validation and test sets). Changes made in both DialogData and StreamDialogData for streaming mode.

**Testing steps**
Run the distributed tests which test both streaming and non-stream mode and max examples flag. 
python tests/test_distributed.py

**Logs**
```
[Setting up test test_distributed_eval_max_exs]
[creating task(s): integration_tests:nocandidate]
[ running dictionary over data.. ]
Building dictionary: 100%|██████████████████████████████████████████████████████████████████████████████████| 500/500 [00:00<00:00, 32.6kex/s]
Dictionary: saving dictionary to /tmp/tmpixzqvtnc/model.dict
[ dictionary built with 11 tokens in 0s ]
multiprocess_train: rank: 0, port: 31337
multiprocess_train: rank: 1, port: 31337
[rank:  1] Distributed group initialized
[rank:  0] Distributed group initialized
[rank:  0] [ building dictionary first... ]
[rank:  0] [ dictionary already built .]
[rank:  1] [ building dictionary first... ]
[rank:  1] [ dictionary already built .]
[rank:  0] [ no model with opt yet at: /tmp/tmpixzqvtnc/model(.opt) ]
[rank:  1] [ no model with opt yet at: /tmp/tmpixzqvtnc/model(.opt) ]
[rank:  0] [ Using CUDA ]
[rank:  0] Dictionary: loading dictionary from /tmp/tmpixzqvtnc/model.dict
[rank:  0] [ num words =  11 ]
[rank:  1] [ Using CUDA ]
[rank:  1] Dictionary: loading dictionary from /tmp/tmpixzqvtnc/model.dict
[rank:  1] [ num words =  11 ]
[rank:  1] Total parameters: 83104
[rank:  0] Total parameters: 83104
[rank:  1] Trainable parameters:  17568
[rank:  0] Trainable parameters:  17568
[rank:  0] [creating task(s): integration_tests:nocandidate]
[rank:  1] [creating task(s): integration_tests:nocandidate]
[rank:  0] [ training... ]
[rank:  1] [ training... ]
/private/home/marywilliamson/.conda/envs/conda_parlai/lib/python3.7/site-packages/torch/cuda/memory.py:294: DeprecationWarning: torch.cuda.memory_cached has been renamed to torch.cuda.memory_reserved
  DeprecationWarning)
[...]

PR #2909, opened 2020-07-31
Keep dynamic batches about the same shape

stephenroller:
**Patch description**
This patch aims to improve distributed training with dynamic batches. In distributed training, we are always bottlenecked by our slowest worker, as we need to `sync()` on every parley and every SGD step.

*On main branch*:
Dynamic batching introduces variability into the training procedure. When building a dynamic batch, we start with a random example, and then fill the batch with similar length examples until we run out of space. One issue here is that the workers may choose differently sized batches. For example, worker A might build a batch with 100 short examples, while worker B might build a batch with only a 5 very long examples. Both workers will forward/backward on these batches and both use roughly the same amount of memory, so there is actually _less_ variability in their performance. Sounds great, right?

However, tokenization is one of the most expensive parts of training (and one major reason we pay a performance cost relative to fairseq). Once both workers finish the current batch they are working on, it will be necessary for them to tokenize the _next_ sentences of the conversations. Here lies the straggler issue: worker A will now need to tokenize 100 sentences, but worker B only needs to tokenize 5. We can reduce this variability by ensuring _all_ workers attempt to build similarly-shaped batches (making the same many-short or few-long decision at the same time), so they all have roughly the same number of items to tokenize.

*With this patch*:
This patch adds this heuristic to training. While we can't guarantee that all workers make the same choice of batch size, we can ensure that they begin building from the same position when starting with the sorted list:

https://github.com/facebookresearch/ParlAI/blob/6728a9ad3d3a36e7d8f9315d027f51500c14cf47/parlai/core/worlds.py#L1197-L1207

Note that we need a separate RNG for this, because not all workers will pick the exact same sized batch, just similarly sized ones. This means the rng usage above could cause them to get out of sync.

However there is an issue: this introduces a significant amount of determinism into the conversation selection process, and separate workers may then begin to select the exact same conversation over. This reduces our stochasticity, and hinders learning. To remedy this, we expand upon our recent changes (#2433, #2445), which stratifies evaluation across many workers by having each of the N workers take only every Nth episode. We expand this to also include training, so the data loaded into memory by each worker is separate, and therefore they can never pick different examples.

**Testing steps**
Correctness tests:
- Added a print statement to all workers (using 8 workers) to print their examples on each step. Before the stratification change, I confirmed that many workers were picking the same conversation. I introduced the stratification change in order to fix this, and verified this was the case.
- Added a print statement to all workers to print their batch sizes. Verified that workers were creating roughly equal batch lengths.
- Trained small models and verified the PPL looked right. Before the stratification change, there was a performance hit (though it was somewhat moderate)

Timing tests:
- Trained on `large_secret_internal_teacher` (a ChunkTeacher) for 300 seconds. Measured the number of updates.
- Trained on `-t convai2 -dt train` for 300 seconds. Measured number of updates
- Trained on `-t convai2 -dt train:valid` for 300 seconds. Measured the number of updates.

**Logs**
Command run (via srun)
```
python -u -m parlai.scripts.distributed_train  -m transformer/generator -esz 128 --ffn-size 512 --n-heads 4 --dict-tokenizer bytelevelbpe --dict-file zoo:blender/blender_3B/model.dict -tr 512 -bs 32 --skip-generation true -opt adam -lr 1e-4 -vmt ppl --warmup-updates 1000 -bs 32 -dynb full -nl 8 -ttim 300 --distributed-world-size GPUS  -dt DATATYPE -t TEACHER
```

Legend: upd (m) = updates on main branch; upd (b) = updates on this branch.

 GPUs | Teacher |     Datatype |     Upd (m) | Upd (b) 
----- | ------- | ------------ | ----------- | --------
8     |  secret | train:stream |       601   | **679** 
16    |  secret | train:stream |       581   | **667** 
32    |  secret | train:stream |       543   | **617** 
64    |  secret | train:stream |       500   | **584** 
8     | ConvAI2 | train:stream |       799   | **817** 
16    | ConvAI2 | train:stream |     **762** |   748   
8     | ConvAI2 | train        |       801   | **867** 
16    | ConvAI2 | train        |       738   | **829** 
       
The one where we regressed slightly is explained to I/O thrashing: because ConvAI2 is so small, using 16 GPUs caused it to be loaded about 12 times in the experiment. Solution is to not use streaming with small data.

### DIFF
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -951,7 +951,8 @@ class StreamDialogData(DialogData):
         idx = 0
         while True:
             for episode in self._read_episode(data_loader(datafile)):
-                yield episode
+                if idx % self.dws == self.rank:
+                    yield episode
                 idx += 1
             while not self.cycle:
                 yield self._END_OF_EPOCH
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -731,7 +731,7 @@ class DialogData(object):
         # about the stratifications we're doing.
         self._num_examples_cache = 0
         self._num_episodes_cache = 0
-        for i, episode in enumerate(self._read_episode(data_loader(datafile))):
+        for episode in self._read_episode(data_loader(datafile)):
             self._num_episodes_cache += 1
             self._num_examples_cache += len(episode)
             self.data.append(episode)
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -967,7 +967,7 @@ class StreamDialogData(DialogData):
             for episode in self._read_episode(data_loader(datafile)):
                 # We only shard the data set at evaluation time, as training is
                 # done using sampling-with-replacement.
-                if idx % self.num_workers == self.rank
+                if idx % self.num_workers == self.rank:
                     yield episode
                 idx += 1
             while not self.cycle:
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -45,7 +45,7 @@ from parlai.core.opt import Opt
 from parlai.utils.conversations import Conversations
 from parlai.utils.data import DatatypeHelper
 from parlai.utils.misc import AttrDict, no_lock, str_to_msg, warn_once
-from parlai.utils.distributed import get_rank, num_workers, is_distributed
+from parlai.utils.distributed import get_rank, num_workers
 import parlai.utils.logging as logging
 
 from abc import ABC, abstractmethod
@@ -676,9 +676,6 @@ class DialogData(object):
         # in case we need to shard the dataset
         self.rank = get_rank()
         self.num_workers = num_workers()
-        self.is_distributed_and_is_eval = is_distributed() and any(
-            x in opt['datatype'] for x in ('valid', 'test', 'train:evalmode')
-        )
 
         # self.data is a list of episodes
         # each episode is a tuple of entries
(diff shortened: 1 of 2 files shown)


parlai/core/teachers.py
```diff
@@ -396,29 +405,6 @@ def next_example(self):
 
         return ex, epoch_done
 
-    def next_batch(self):
```
stephenroller, line 399: this was never called anywhere. Just deleting some old code I found.

stephenroller:
Putting this back as a draft. It's turned out to be deviously trickly to get right. I think the Teacher rewrite may need to come first.

parlai/core/teachers.py
```diff
@@ -356,10 +359,16 @@ def next_episode_idx(self, num_eps=None, loop=None):
         if loop is None:
             loop = self.training
         if self.random:
-            new_idx = random.randrange(num_eps)
+            new_idx = (
+                random.randrange(
+                    num_eps // self.dws + int(num_eps % self.dws > self.rank)
```
klshuster, line 364: what is this computation exactly?

stephenroller: we split across dws (distributed world size) workers.

if there are 1000 episodes and 16 workers, then 8 workers get 63 examples, and 8 workers get 62.

8 * 62 + 8 * 63 = 1000.

The > rank mess is identifying who gets the the 63 (the first workers) and who gets the 62 (the last workers)

parlai/core/teachers.py
```diff
@@ -705,6 +692,9 @@ def share(self):
             'cands': self.cands,
             'image_loader': self.image_loader,
         }
+        if hasattr(self, '_num_examples_cache'):
+            shared['num_examples_cache'] = (self._num_examples_cache,)
```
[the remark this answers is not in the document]
stephenroller: bug

klshuster:
@stephenroller is this something we should look back into? from your experiments it seemed to have worked pretty well in terms of speedup

stephenroller:
Yeah I'd very much like to land this one day, but it's just not any priority. It requires reworking teachers.

klshuster:
applying `donotreap`

stephenroller:
Finding a different way to do this.

PR #2911, opened 2020-07-31
[Image S2S] Standardize Image Seq2Seq Image Processing

klshuster:
**Patch description**
In the current Image Seq2Seq implementation, dummy image features are only concatenated to token embeddings if there is at least one valid image in a batch. However, this means that during interactive, the model would never see dummy image features. This fix updates to *always* concatenate dummy image features.  

**Testing steps**
Relying on existing CI tests for Image Seq2seq. I additionally tried locally in interactive mode.

### DIFF
--- a/parlai/agents/image_seq2seq/image_seq2seq.py
+++ b/parlai/agents/image_seq2seq/image_seq2seq.py
@@ -128,7 +128,7 @@ class ImageSeq2seqAgent(TransformerGeneratorAgent, TorchImageAgent):
                 images.append(img)
             batch.image = images
         else:
-            images = [None for _ in range(len(batch.valid_indices))]
+            images = [None] * len(batch.valid_indices)
             batch.image = images
         return batch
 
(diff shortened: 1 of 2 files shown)


parlai/agents/image_seq2seq/image_seq2seq.py
```diff
@@ -120,13 +120,16 @@ def batchify_image_features(self, batch: Batch) -> Batch:
...
-        if type(batch.image) == list and any(b is not None for b in batch):
+        if type(batch.image) == list and any(b is not None for b in batch.image):
             images = []
             for img in batch.image:
                 if isinstance(img, torch.Tensor):
                     img = self._process_image_features(img)
                 images.append(img)
             batch.image = images
+        else:
+            images = [None for _ in range(len(batch.valid_indices))]
```
EricMichaelSmith, line 131: nit: this can also be `[None] * len(batch.valid_indices)`, and similarly for below

parlai/agents/image_seq2seq/modules.py
```diff
@@ -251,6 +250,16 @@ def encode_images(
...
+        else:
+            image_encoded = torch.stack(
+                [self.dummy_image_enc for _ in range(len(images))]
+            ).reshape(
+                len(images),
+                self.n_image_tokens * self.n_image_channels,
+                self.embedding_size,
+            )
+            image_masks = torch.stack([~self.ones_mask for _ in range(len(images))])
+            assert image_masks.shape == image_encoded.shape[:2]
```
EricMichaelSmith, line 262: Hmm we can move this line out of the `else` block because it's also applied in the `if` block

PR #2913, opened 2020-07-31
Add style-controlled generation agent

EricMichaelSmith:
**Patch description**
Add code to train/evaluate style-controlled generative models. Some of this code was also in https://github.com/facebookresearch/ParlAI/pull/2829, but the classifier code in that PR will be open-sourced at a later date. Also add a model trained with this agent.

**Testing steps**
```pytest tests/nightly/gpu/test_style_gen.py```

### DIFF
--- a/projects/style_gen/README.md
+++ b/projects/style_gen/README.md
@@ -1,0 +1,5 @@
+# Style-controlled generation
+
+Agent for training and evaluating generative models conditioned on a style token, for instance, the "personality" string attached to each example of the Image-Chat dataset.
+
+- `--model style_gen`: Subclass of `TransformerGeneratorAgent` for which style tokens can be appended to the history for every training/evaluation example. Found in `style_gen.py`.
--- a/projects/style_gen/README.md
+++ b/projects/style_gen/README.md
@@ -8,9 +8,9 @@ Agent for training and evaluating generative models conditioned on a style token
 
 Evaluating a style-controlled generation model on the BlendedSkillTalk dataset, labeled with personalities from Image-Chat:
 ```
-python examples/eval_model.py \
+parlai eval_model \
 --model-file zoo:style_gen/c75_labeled_dialogue_generator/model \
---model projects.style_gen \
+--model projects.style_gen.style_gen:StyleGenAgent \
 --skip-generation True \
 --task style_gen:LabeledBlendedSkillTalk \
 --datatype test \
--- a/projects/style_gen/README.md
+++ b/projects/style_gen/README.md
@@ -2,7 +2,7 @@
 
 Agent for training and evaluating generative models conditioned on a style token, for instance, the "personality" string attached to each example of the Image-Chat dataset.
 
-- `--model style_gen`: Subclass of `TransformerGeneratorAgent` for which style tokens can be appended to the history for every training/evaluation example. Found in `style_gen.py`.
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by EricMichaelSmith** [comment]
@stephenroller My CI test for my BlenderBot3B-sized style-gen model is timing out given the slow CI GPUs, even though it runs in 72 sec on one of my devfair GPUs. It doesn't look like we have any CI tests for BlenderBot3B itself, so I'm not sure if it's a solved problem how to test these models with CI.

**Comment by stephenroller** [comment]
Unsolved problem.

A few thoughts:
- add support for multiprocessing eval. I created an issue for this.
- turn off fp16. Emulation of fp16 is probably very slow for these old gpus
- turn off generation

**Comment by EricMichaelSmith** [comment]
> Unsolved problem.
> 
> A few thoughts:
> 
> * add support for multiprocessing eval. I created an issue for this.
> * turn off fp16. Emulation of fp16 is probably very slow for these old gpus
> * turn off generation

Okay, makes sense - I've turned FP16 off (and generation is already skipped), so let's see if this gets the test to pass on these GPUs

File: projects/style_gen/modules.py [hunk 1]
```diff
@@ -0,0 +1,89 @@
...
+
+import random
+
+from parlai.core.torch_agent import History
+
+
+STYLE_SEP_TOKEN = ' STYLE '
+
+
+class StyleHistoryMixin(History):
```
emilydinan, line 18: could we potentially use the TorchAgent funciton `get_temp_history` to do this instead? it was added after you wrote this code initially, but makes the changes much simpler

stephenroller: Yeah I would love that

EricMichaelSmith: Yes, good idea! So, I refactored `StyleAgentMixin` to use `get_temp_history()`, and I re-ran the command that I listed in `model_list.py`: `parlai eval_model --datatype test --model projects.style_gen.style_gen:StyleGenAgent --model-file zoo:style_gen/c75_labeled_dialogue_generator/model --skip-generation True --task style_gen:LabeledBlendedSkillTalk --use-style-frac 1.00`. The resulting ppl is 0.026 higher than before, and in fact I see that there are slight differences in some of the generations:

Screen Shot 2020-08-03 at 5 10 29 PM

However, clearly the agent is still setting the style correctly, because if it weren't, the perplexities would be *way* higher than what I'm seeing (around 10.15, instead of 9.47). Thus, I'll chalk this up to fluctuations in generation

projects/style_gen/README.md
```diff
@@ -0,0 +1,18 @@
+# Style-controlled generation
+
+Agent for training and evaluating generative models conditioned on a style token, for instance, the "personality" string attached to each example of the Image-Chat dataset.
+
+- `--model style_gen`: Subclass of `TransformerGeneratorAgent` for which style tokens can be appended to the history for every training/evaluation example. Found in `style_gen.py`.
```
klshuster, line 5: nit: maybe include full path here `--model projects.style_gen.style_gen:StyleGenAgent`

EricMichaelSmith: wow - good catch, changed

EricMichaelSmith:
All CI checks have passed except for `long_gpu_tests`, which timed out due to known issues; I'll merge this in

PR #2916, opened 2020-08-01
 Aligning header elements to the center

sawravchy:


stephenroller:
Sorry I missed this and then it got lost in the fray. Thanks for this!

Issue #2142, opened 2019-11-09
rouge-L metric should be renamed to rouge_L


stephenroller:
The hyphen makes it difficult to use with jq.

hadasah:
@stephenroller where does this cause a problem? just in the output?

stephenroller:
Yeah in .trainstats, when you try to use “jq” to read the file.

We should just rename the keys from hyphens to underscores.

PR #2926, opened 2020-08-03
rename rouge metrics to use underscores

hadasah:
**Patch description**
rename rouge-{x} to rouge_{x} in the train stats outputs, to play nicer with jq, as requested in #2142.

**Testing steps**
`pytest -v tests/test_eval_model.py` passes all tests


hadasah:
@stephenroller I only changed it in the output - everywhere else, I retained the hyphen (the rationale being that it's easier in the code to retain that convention). Let me know if you think it would be better to use an underscore everywhere. 

Also only made this change to rouge-L, not to rouge-1 or rouge-2, since they weren't listed in #2142.

stephenroller:
Thanks. Can you do the others too?

PR #2929, opened 2020-08-03
Show all stdout on GPU tests

EricMichaelSmith:
**Patch description**
Show stdout on all long GPU tests, which will help prevent tests from timing out with no output

**Testing steps**
CI check


stephenroller:
LGTM. Not approving since this isn't actually supposed to land.

PR #2940, opened 2020-08-05
Tweak colors.

stephenroller:
**Patch description**
Slightly tweak colors of logs. I hadn't realized it was like this because my console was ignoring 256 colors.

**Testing steps**
Before:

image

After:
image

### DIFF
--- a/parlai/utils/logging.py
+++ b/parlai/utils/logging.py
@@ -39,7 +39,7 @@ COLORED_LEVEL_STYLES = {
     'debug': {'faint': True},
     'verbose': {'color': 'blue'},
     'error': {'color': 'red'},
-    'info': {'faint': True},
+    'info': {},
     'report': {'bold': True},
     'success': {'bold': True, 'color': 'green'},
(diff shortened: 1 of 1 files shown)


emilydinan:
i kind of liked the faint logs!

stephenroller:
It’s still there for debug :p

jaseweston:
we have color option settings so we could code different styles depending
on what you want couldn't we?
https://github.com/facebookresearch/ParlAI/blob/master/parlai/utils/strings.py#L101

klshuster:
a little late but fwiw i also liked the faint logs lol

stephenroller:
Okay let's talk about it tuesday. We can turn them back.

jaseweston:
As I said earlier in the thread, I added color styles, so we can make
everybody happy ...

stephenroller:
The color styles need work to work with logs. It's different logic.

stephenroller:
(We can do it, i'm totally down. It just isn't off the shelf)

PR #2941, opened 2020-08-05
[Dynamic Batching] Add world logging capability

### LINKED ISSUE #2912: World logging not working with eval_model.py with dynamic batching
Use this to open other questions or issues, and provide context here.
When I tried `multiprocessing_eval.py --dynamic-batching full`, it gave me this error `AttributeError: 'DynamicBatchWorld' object has no attribute 'acts'`. It seems dynamic batching is not supported in `world_logging`?

emilydinan:
**Patch description**
Currently dynamic batching fails with world logging because it doesn't log `self.acts` in the way that other worlds do. Fixes #2912.  I added a few attributes to track this. 

I'll note that saving `self._task_acts` and `self._obs` is a little redundant and perhaps a waste of memory. The alternative is to remove all extraneous fields (like text_vec, label_vec, etc.) from the acts in `self._obs` that are added post observation before we add to `self.act`, but it's a little hacky. Thoughts @stephenroller ?

**Testing steps**
Wrote a test

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -1185,8 +1185,12 @@ class DynamicBatchWorld(World):
                 self.worlds[i].parley_init()
 
             act = self.worlds[i].get_task_agent().act()
-            self._task_acts[i] = act
+
+            # we log the task act and the index of the act
+            # in the buffer for world logging purposes
+            self._task_acts[i] = act  # for world logging
             self._task_acts[i]['dyn_batch_idx'] = i
+
             obs = self.worlds[i].get_model_agent().observe(act)
             self._obs[i] = obs
 
@@ -1258,8 +1262,11 @@ class DynamicBatchWorld(World):
             self.worlds[i].get_model_agent().self_observe(act)
             # move these worlds forward
             act = self.worlds[i].get_task_agent().act()
+            # we log the task act and the index of the act
+            # in the buffer for world logging purposes
             self._task_acts[i] = act
             self._task_acts[i]['dyn_batch_idx'] = i
+            # save the observations to form a batch
             obs = self.worlds[i].get_model_agent().observe(act)
             self._scores[i] = self._score(obs)
             self._obs[i] = obs
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -1110,8 +1110,10 @@ class DynamicBatchWorld(World):
 
     def reset(self):
         super().reset()
+        self._task_acts = [None for _ in range(self._BUFFER_SIZE)]
         self._obs = [None for _ in range(self._BUFFER_SIZE)]
         self._scores = [None for _ in range(self._BUFFER_SIZE)]
+        self.acts = [None, None]
 
         self.number_parleys = 0
         self.total_exs = 0
@@ -1183,6 +1185,7 @@ class DynamicBatchWorld(World):
                 self.worlds[i].parley_init()
 
             act = self.worlds[i].get_task_agent().act()
+            self._task_acts[i] = act
             obs = self.worlds[i].get_model_agent().observe(act)
             self._obs[i] = obs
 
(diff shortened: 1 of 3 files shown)


stephenroller:
A test is really crucial for this one imho. I'm nervous about it. There's an integration teacher that has a variable width, and there's the `test_agents/repeat` or something similar that you could use.

emilydinan:
> Hmm, so a few things:
> 
> * you missed line 1260 which would also need this change
> * act and obs are almost certainly the same thing. Our models do in place editing
> 
> Does DialogPartnerWorld.acts work?
> you could define a dynamic property like
> 
> ```
> @property
> def acts(self):
>     return [self.worlds[i].acts for i in self.last_incices]
> ```
> 
> and set self.last_indices on line 1250, with the batch act. That might be cleaner

i don't see why line 1260 needs this change

emilydinan:
resolved the issues @stephenroller, what do you think of this solution?

i will write a test soon

parlai/utils/world_logging.py
```diff
@@ -89,9 +89,10 @@ def _log_batch(self, world):
         batch_act = world.get_acts()
         parleys = zip(*batch_act)
         for i, parley in enumerate(parleys):
-            self._add_msgs(parley, idx=i)
+            idx = parley[0]['dyn_batch_idx'] if 'dyn_batch_idx' in parley[0] else i
```
stephenroller, line 92: leave a comment please

stephenroller:
This solution seems good to me. Would just like to see a test and some comments in both DynamicBatchWorld and WorldLogger pointing out the coupling between them.

emilydinan:
> why would we need to remove the post-processing stuff from an obs prior to logging (tensors etc.)?

when we log, we typically don't include that info because we want to see exactly what was handed to the model from the teacher. plus it gets messy/hard to read when you log all of that stuff

parlai/utils/world_logging.py
```diff
@@ -81,15 +81,21 @@ def _add_episode(self, episode):
...
         batch_act = world.get_acts()
         parleys = zip(*batch_act)
         for i, parley in enumerate(parleys):
-            self._add_msgs(parley, idx=i)
+            # in dynamic batching, we only return `batchsize` acts, but the
+            # 'dyn_batch_idx' key in the task act corresponds the episode index
+            # in the buffer
+            idx = parley[0]['dyn_batch_idx'] if 'dyn_batch_idx' in parley[0] else i
+            self._add_msgs(parley, idx=idx)
             if world.worlds[i].episode_done():
```
klshuster, line 97: does this need to be `world.worlds[idx]` as well?

parlai/core/worlds.py
```diff
@@ -1245,15 +1253,20 @@ def parley(self):
...
             # we need to make sure that the teachers saw the result
             self.worlds[i].get_task_agent().observe(act)
             # and that the agent copies saw their own voice
             self.worlds[i].get_model_agent().self_observe(act)
-
             # move these worlds forward
             act = self.worlds[i].get_task_agent().act()
+            # we log the task act and the index of the act
+            # in the buffer for world logging purposes
+            self._task_acts[i] = act
```
klshuster, line 1267: sorry, could you explain again why we do this twice in `parley`?

parlai/utils/world_logging.py
emilydinan: yes great catch thank you

parlai/core/worlds.py
emilydinan: ah yes so for the indices we selected, we have to move the subworlds forward and get the next task act from the teacher

stephenroller: yeah it's dynamic batching. Dynb doesn't in parley() ends with all subworlds ready for the model's turn.

PR #2954, opened 2020-08-07
Cache data directory for tests

stephenroller:
**Patch description**
Tests now cache ~/ParlAI/data to reduce slow downloads

**Testing steps**
CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -415,6 +415,7 @@ jobs:
           paths:
             - "~/venv/bin"
             - "~/venv/lib"
+      - <<: *findtests
       - restore_cache:
           key: data-20200904-gpu16-{{ checksum "teststorun.txt" }}
       - run:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -535,23 +601,20 @@ workflows:
       - long_gpu_tests:
           requires:
             - cleaninstall_36
-            - unittests_38
+            - unittests_gpu14
```
EricMichaelSmith, line 604: Hmm what's the reason for the requirements change?

.circleci/config.yml
```diff
@@ -535,23 +601,20 @@ workflows:
...
             - unittests_38
             - unittests_37
             - unittests_36
-            - unittests_osx
-            - mturk_tests
       - finalize_coverage_report:
           requires:
             - teacher_tests
             - long_gpu_tests
+            - mturk_tests
```
EricMichaelSmith, line 616: Same here

stephenroller:
This is still draft and doesn't work right.

Issue #2951, opened 2020-08-07
Expected object of device type cuda but got device type cpu for argument #1 'self' in call to _th_mm


curehabit:
**Bug description**
Hi, when use GPU to run personality_caption project, may cause this error in project/personality_captions/transresnet/modules:402-407:
Runtime Error: Expected object of device type cuda but got device type cpu for argument #1 'self' in call to _th_mm

**Reproduction steps**
Set config: no_cuda = False and run project personality_captions

**Expected behavior**
Variable 'context_encoding' and 'candidates_encoded[img_index]' should be in same GPU device.

**Additional context**
I change code like this and it works:

scores = torch.mm(
                candidates_encoded[img_index].to(context_encoding.device)
                if not one_cand_set
                else candidates_encoded.to(context_encoding.device),
                context_encoding.transpose(0, 1),
            )
 
Thanks

stephenroller:
Thanks for filing. Can you share the command you ran for context?

Assigning to @klshuster and @EricMichaelSmith. I think the patch above is probably enough.

klshuster:
I am having trouble reproducing on my end, could you please share the exact command you ran @curehabit?

stephenroller:
I think the key is —no-cuda

klshuster:
I tried with and without `--no-cuda` to no avail

curehabit:
Sorry, I found I commented  projects/personality_captions/interactive line:278-279 because I want to run with cuda.

#opt['no_cuda'] = True
#opt['override']['no_cuda'] = True

Than run command 

`python projects/personality_captions/interactive.py -mf models:personality_captions/transresnet/model`

and my config:(hidden some path information with XXX)

[ optional arguments: ]
[  activation: relu ]
[  attention_dropout: 0.2 ]
[  display_examples: False ]
[  display_ignore_fields: label_candidates,text_candidates ]
[  display_prettify: False ]
[  dropout: 0.0 ]
[  embedding_size: 300 ]
[  embeddings_scale: True ]
[  ffn_size: 1200 ]
[  fixed_cands_path: None ]
[  freeze_patience: -1 ]
[  interactive_task: True ]
[  learn_positional_embeddings: False ]
[  model_parallel: False ]
[  n_decoder_layers: -1 ]
[  n_encoder_layers: -1 ]
[  n_heads: 2 ]
[  n_layers: 2 ]
[  n_positions: 1000 ]
[  n_segments: 0 ]
[...]

PR #2963, opened 2020-08-10
Personality Captions interactive mode fix

EricMichaelSmith:
**Patch description**
Implements the proposed fix at https://github.com/facebookresearch/ParlAI/issues/2951 to get interactive mode for Personality Captions to work with CUDA enabled.

**Testing steps**
```
python projects/personality_captions/interactive.py -mf models:personality_captions/transresnet/model
python projects/personality_captions/interactive.py -mf models:personality_captions/transresnet/model --no-cuda
```

### DIFF
--- a/projects/personality_captions/interactive.py
+++ b/projects/personality_captions/interactive.py
@@ -275,8 +275,6 @@ def setup_interactive():
         )
     opt['task'] = 'parlai.agents.local_human.local_human:LocalHumanAgent'
     opt['image_mode'] = 'resnet152'
-    opt['no_cuda'] = True
-    opt['override']['no_cuda'] = True
     SHARED['opt'] = opt
     SHARED['image_loader'] = ImageLoader(opt)
 
--- a/projects/personality_captions/transresnet/modules.py
+++ b/projects/personality_captions/transresnet/modules.py
@@ -399,9 +399,9 @@ class TransresnetModel(nn.Module):
         for img_index in range(len(context_encoded)):
             context_encoding = context_encoded[img_index : img_index + 1, :]
             scores = torch.mm(
-                candidates_encoded[img_index]
+                candidates_encoded[img_index].to(context_encoding.device)
                 if not one_cand_set
-                else candidates_encoded,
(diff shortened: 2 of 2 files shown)


projects/personality_captions/transresnet/modules.py
```diff
@@ -399,9 +399,9 @@ def choose_best_caption(
         for img_index in range(len(context_encoded)):
             context_encoding = context_encoded[img_index : img_index + 1, :]
             scores = torch.mm(
-                candidates_encoded[img_index]
+                candidates_encoded[img_index].to(context_encoding.device)
```
klshuster, line 402: Nit: I believe you can just do `.to(context_encoding`

projects/personality_captions/interactive.py
```diff
@@ -275,8 +275,6 @@ def setup_interactive():
         )
     opt['task'] = 'parlai.agents.local_human.local_human:LocalHumanAgent'
     opt['image_mode'] = 'resnet152'
-    opt['no_cuda'] = True
```
klshuster, line 278: I'm happy with this but was it contributing to the bug?

projects/personality_captions/transresnet/modules.py
stephenroller: rly? that's new to me.

klshuster:
```python
In [1]: import torch

In [2]: torch.__version__
Out[2]: '1.4.0'

In [3]: a = torch.LongTensor([1, 2, 3]).cuda()

In [4]: a
Out[4]: tensor([1, 2, 3], device='cuda:0')

In [5]: b = torch.LongTensor([4, 5, 6]).to(a)

In [6]: b
Out[6]: tensor([4, 5, 6], device='cuda:0')

```

EricMichaelSmith: Yes, changed

projects/personality_captions/interactive.py
klshuster: totally fine with the change, can't remember myself why it was in there in the first place

Issue #2962, opened 2020-08-10
Multiprocessing_train blocked


aslu98:
I started a multiprocessing training with the following command:
`parlai multiprocessing_train -t fromfile:parlaiformat \
      --fromfile_datapath ../parlai_utilities/covid_qa+FAQ_parlaiformat.txt -m transformer/generator \
        --init-model zoo:blender/blender_3B/model --dict-file zoo:blender/blender_3B/model.dict \
        -veps 0.25 --attention-dropout 0.0 --batchsize 32 --embedding-size 2560 \
         --ffn-size 10240 --variant prelayernorm --n-heads 32 --n-positions 128 \
        --n-encoder-layers 2 --n-decoder-layers 24 \
        --history-add-global-end-token end --delimiter ' ' \
        --dict-tokenizer bytelevelbpe --dropout 0.1 --fp16 True \
        --label-truncate 128 --log_every_n_secs 10 -lr 7e-06 --lr-scheduler reduceonplateau \
        --lr-scheduler-patience 3 --optimizer adam --relu-dropout 0.0 --activation gelu \
        --model-parallel true --text-truncate 128 --truncate 128 \
        --warmup_updates 100 --fp16-impl mem_efficient --update-freq 2 --gradient-clip 0.1 \
        --skip-generation True -vp 10 -vmt ppl -vmm min \
        --save-after-valid True --model-file  covidmodel1`

Using two 12GB GPUs but it is has been stuck on `07:54:19 INFO | num words = 8008` for almost two hours.
I checked nvidia-smi and it is using the GPUs and `python -c 'import torch; print(torch.cuda.is_available())'` returns True.

multiprocessing_train output: https://imgur.com/T0GEUMl
nvidia-smi: https://imgur.com/CuREqRn

From the Recipe page I read that I could try to fine-tune 3B model with multiple GPUs, is it stuck because I don't have enough RAM? What else could it be? I already tried to stop the container and restart it.

stephenroller:
You can’t mix multiprocessing train with model parallel true. The two processes just try to fight over both GPUs.

I’ll try to add an error message.

aslu98:
Thanks! So for my understanding using the multiprocessing_train script is the same as using train_model with --model-parallel true, right?

stephenroller:
[...]

PR #2964, opened 2020-08-10
Prevent mixing model parallel and multiprocessing.

stephenroller:
**Patch description**
Follow up to #2962. If the user attempts to combine model parallelism and multiprocessing, throw an error and point them to that issue.

**Testing steps**
Manual testing. New CI.

### DIFF
--- a/parlai/utils/torch.py
+++ b/parlai/utils/torch.py
@@ -337,8 +337,8 @@ class PipelineHelper(object):
         """
         Check compatibility for opts.
 
-        Really just used to raise an error message if the user mixes
-        multiprocessing and model parallelism.
+        Really just used to raise an error message if the user mixes multiprocessing and
+        model parallelism.
         """
         if opt.get('multiprocessing') and not os.environ.get('PARLAI_FORCE_MP'):
             raise RuntimeError(
--- a/parlai/utils/torch.py
+++ b/parlai/utils/torch.py
@@ -333,6 +333,24 @@ class PipelineHelper(object):
             self.devices.append(d)
             self.__device_allocations[d] = 0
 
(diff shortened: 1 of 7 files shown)


parlai/utils/torch.py
```diff
@@ -333,6 +333,24 @@ def __init__(self):
...
+    def check_compatibility(self, opt):
+        """
+        Check compatibility for opts.
+
+        Really just used to raise an error message if the user mixes multiprocessing and
+        model parallelism.
+        """
+        if opt.get('multiprocessing') and not os.environ.get('PARLAI_FORCE_MP'):
+            raise RuntimeError(
+                "It looks like you are trying to mix multiprocessing data "
```
dianaglzrico, line 345: Nice message 👍

tests/test_distributed.py
```diff
@@ -169,6 +169,23 @@ def test_chunked_teacher(self):
...
+    def test_no_model_parallel(self):
+        """
+        Checks that we throw an error when combining mp_train with.
+
+        --model-parallel true.
+        """
+        config = copy.deepcopy(self._base_config)
+        config['model_parallel'] = True
+        for m in [
+            'transformer/generator',
```
dianaglzrico, line 181: Why these models only?

stephenroller:
> This looks great! once I read the issue it all made sense, good idea to add this check. Left a few comments only :)

These are the only models that support it.

tests/test_distributed.py
stephenroller: These are the only models that support it, and it covers TRA, TCA, and TGA, so all the abstract classes are covered.

PR #2967, opened 2020-08-11
Renaming save function in tokenizer

dianaglzrico:
**Patch description**
Save function in tokenizers was renamed from `save()` to `save_model()` in recent version of Tokenizers.

### DIFF
--- a/parlai/utils/bpe.py
+++ b/parlai/utils/bpe.py
@@ -826,7 +826,7 @@ class HuggingFaceBpeHelper(BPEHelper):
         :param file_name:
             file to save.
         """
-        self.tokenizer.save(dir_name, file_name)
+        self.tokenizer.save_model(dir_name, file_name)
 
 
 class SlowBytelevelBPE(Gpt2BpeHelper):


klshuster:
I think we'll need to update our `requirements.txt` to have the latest version of `transformers`, otherwise this will break for people with old `transformers` installs

also will automatically bump the circleCI cache and fix the failing tests

stephenroller:
Yeah we might need to check the version too. I am not sure if tokenizers we want to force 0.8.0 or not.

dianaglzrico:
> Yeah we might need to check the version too. I am not sure if tokenizers we want to force 0.8.0 or not.

 `requirements.txt` updated

klshuster:
@stephenroller were you saying we _dont_ want to force 0.8.0? i.e. would we rather have a check in `bpe.py` of which `tokenizers` version the user has?

stephenroller:
What's your opinion, Kurt?

klshuster:
I don't see a downside in keeping up with the latest `tokenizers`, as long as it doesn't break anything significant. I think our only reason for _not_ upgrading package versions in the past was to prevent any significant breaking changes to our codebase (or changes that required non-trivial updates). given this is just a one-liner, I think i'm ok with it. doesn't seem to break any existing CI tests anyway

stephenroller:
Okay, Diana can you update requirements again to be >= instead of ==?

PR #2972, opened 2020-08-12
Add a custom metrics tutorial.

stephenroller:
**Patch description**
Add a tutorial on how to use custom metrics, including both in teachers and agents.

**Testing steps**
Walked through the tutorial.

**Logs**
In the tutorial.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -20,6 +20,7 @@ ParlAI is a one-stop-shop for dialog research.
    tutorial_worlds
    tutorial_torch_generator_agent
    tutorial_torch_ranker_agent
+   tutorial_metrics
    tutorial_fast
    tutorial_tipsntricks
    tutorial_mturk
--- a/docs/source/tutorial_metrics.md
+++ b/docs/source/tutorial_metrics.md
@@ -1,0 +1,370 @@
+# Understanding and adding new metrics
+
+Author: Stephen Roller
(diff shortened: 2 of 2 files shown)


stephenroller:
Okay this has blown up into a full blown update of the autodocs and switching to a new markdown parser.

image

docs/source/agents_list.rst
```diff
@@ -1,15 +1,14 @@
...
 state-of-the-art with careful tuning and experimentation. Each of these contains
 a README with examples of usage and description of the agent.
 
-```eval_rst
 .. toctree::
    :maxdepth: 1
    :caption: List of Agents
    :glob:
 
    agent_refs/*
```
spencerp, line 14: Reading up on myst_parser, it seems like it's built to augment Markdown with all the features that rst has.

So it might be more straightforward if we keep all the files md, and use the myst syntax to replace eval_rst. What do you think?

For instance, toctree can be specified with:
```
```{toctree}
```

Reference: https://myst-parser.readthedocs.io/en/latest/using/faq.html

I'm not actually familiar with rst or myst, though, so feel free to ignore if you've already thought about this.

stephenroller: Hm good point. I would like us to move to fully md somehow, but I was observing that myst didn’t have a way to trigger autodocs

PR #2975, opened 2020-08-12
Add custom metrics tutorial

stephenroller:
**Patch description**
Add a tutorial on how we do metrics in ParlAI.

This depends on #2972 

**Testing steps**
Read tutorial. Assigning to @mojtaba-komeili since he is the least familiar with this process, so the best positioned to offer constructive criticism.

image


emilydinan:
what does (real) mean lol

stephenroller:
I originally had this in the other PR, but then realized I needed to make some updates to the doc system, and then decided the previous PR would be where I do that, and this is where I would write the new tutorial.

stephenroller:
I originally had this in the other PR, but then realized I needed to make some updates to the doc system, and then decided the previous PR would be where I do that, and this is where I would write the new tutorial.

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
+# Understanding and adding new metrics
+
+Author: Stephen Roller
+
+## Introduction and Standard Metrics
+
+ParlAI contains a number of built-in metrics that are automatically computed when
+we train and evaluate models. Some of these metrics are _text generation_ metrics,
+which happen any time we generate a text: this includes F1, BLEU and Accuracy.
```
mojtaba-komeili, line 9: Like you did below, maybe after introducing the metrics here, say that more info about them is below. Also, I think it catches the eye better (if you are glancing the doc) to have these bold.

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
...
+  the predicted response and the reference response. It is measured on
+  tokenized text, and uses NLTK to compute it.
+- F1: This is the [Unigram](https://en.wikipedia.org/wiki/N-gram) F1 overlap
+  between your text and the reference response.
+- exs: the number of examples we have evaluated
+
+If you don't see the BLEU-4 score, you may need to install NLTK with
+`pip install nltk`.
+
+We can also measure ROUGE. Note that we need to `pip install py-rouge` for this
```
mojtaba-komeili, line 39: Could you add a reference to ROUGE please?

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
...
+...
+14:54:14 | Evaluating task dailydialog using datatype valid.
+14:54:14 | creating task(s): dailydialog
+...
+15:26:19 | Finished evaluating tasks ['dailydialog'] using datatype valid
+    accuracy  bleu-4  ctpb  ctps  exps  exs    f1  gpu_mem  loss      lr  ltpb  ltps   ppl  token_acc  total_train_updates   tpb   tps
+           0 .002097 14202 442.5 6.446 8069 .1345    .0384 2.979 7.5e-06  3242   101 19.67      .4133               339012 17445 543.5
+```
+
+Here we see a number of extra metrics, each of which we explain below:
```
mojtaba-komeili, line 70: It seems to me that some of these are metrics related to the performance of the model/algorithm. And the rest are related to complexity of it and how that is handled. If that's the case, would you mind separating the two?

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
...
+## Adding custom metrics
+
+Of course, you may wish to add your own custom metrics: whether this is because
+you are developing a special model, special dataset, or otherwise want other
+information accessible to you. Metrics can be computed by either _the teacher_ OR
+_the model_. Within the model, they may be computed either _locally_ or _globally_.
+There are different reasons for why and where you would want to choose each
+location:
+
+- __Teacher metrics__: This is the best spot for computing metrics that depend
```
mojtaba-komeili, line 96: I know you explain them below. But would you mind adding an example for each of these? It again helps with seeing them all in one place, and comparing them in one look.

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
...
+        self,
+        teacher_action: Message,
+        labels: Optional[Tuple[str]],
+        model_response: Message,
+    ) -> None:
+        pass
+```
+
+The signature for this method is as follows:
+- `teacher_action`: this is the last message the teacher sent to the model. This likely
```
mojtaba-komeili, line 130: Are the teacher message sent to the model or the world?

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
...
+cases), please see the [metrics API documentation](metrics_api).
+
+## Agent (model) level metrics
+
+In the above example, we worked on a metric defined by a Teacher. However,
+sometimes our models will have special metrics that only they want to compute,
+which we call an Agent-level metric. Perplexity is one example.
+
+To compute model-level metrics, we can define either a global metric, or a
+local metric. Global metrics can be computed anywhere, and are easy to use,
```
mojtaba-komeili, line 221: I think making 'global' and 'local' bold (or any other emphasis) might make this read better. And prepare reader for what comes next.

mojtaba-komeili:
How these metrics are related to model metrics in PyTorch? Is there any visualization tool, for example Tensorboard, that could visualize these metrics?

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
...
+        task='dailydialog',
+        model='GlobalHelloCounter',
+        model_file='zoo:blender/blender_90M/model',
+        batchsize=32,
+    )
+```
+
+Running the script, we see that our new metric appears. Note that it is
+different than the metric in the first half of the tutorial: that is because
+previously we were counting the number of times the model said hello (a lot),
```
mojtaba-komeili, line 271: Previously, we had a teacher agent and metric was calculated on its actions. If I am not wrong, those are coming from dataset. I don't understand why it says here that the metric counts the number of times the model says hello then.

stephenroller:
All of these metrics are automatically logged to tensorboard if you enable it. That’s a good thing to say.

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
...
+
+@register_agent('LocalHelloCounter')
+class LocalHelloCounterAgent(TransformerGeneratorAgent):
+    def batchify(self, observations):
+        batch = super().batchify(observations)
+        if hasattr(batch, 'text_vec'):
+            num_hello = ["hello" in o['text'] for o in observations]
+            self.record_local_metric(
+                'local_hello',
+                # AverageMetric.many(seq1) is shorthand for
```
jxmsML, line 334: nit: seq1 -> seq same naming as the next line.

docs/source/tutorial_metrics.md
jxmsML: Those were counting how many 'text' in model_response which model_response is The full model response, including any extra fields the model may have sent. (see L135) What I believe the difference is the previous one is counting how frequent the model says hello  given that it sees the dataset, and here it counts how often the model observe "hello" the context. I could be a little confusing as the naming of the metrics could suggest otherwise.

stephenroller: Thanks for flagging, I agree. The confusion is worth making unambiguous

docs/source/tutorial_metrics.md
```diff
@@ -0,0 +1,372 @@
...
+
+Teacher metrics are useful for items that depend on a specific dataset.
+For example, in some of our task oriented datasets, like
+[`google_sgd`](https://github.com/facebookresearch/ParlAI/blob/master/parlai/tasks/google_sgd/agents.py),
+we want to additionally compute metrics around slots.
+
+Teacher metrics can be added by adding the following method to your teacher:
+
+```python
+    def custom_evaluation(
```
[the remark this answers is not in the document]
stephenroller: Works great :)

docs/source/tutorial_metrics.md
stephenroller: ty

docs/source/tutorial_metrics.md
stephenroller: You mean bold F1, BLEU, and Accuracy?

docs/source/tutorial_metrics.md
stephenroller: I've added some clarifications and extra tip section emphasizing this.

PR #2977, opened 2020-08-14
Add debug messages to measure generation time

EricMichaelSmith:
**Patch description**
With a `--log-generation-time` flag, log the amount of time to do forward passes through the encoder and decoder during generation, as well as the number of passes through the decoder. `CUDA_LAUNCH_BLOCKING` should be set for this.

**Testing steps**
```pytest -s tests/test_tga.py```

### CODE REVIEW

File: parlai/core/torch_generator_agent.py [hunk 1]
```diff
@@ -1114,15 +1117,21 @@ def _generate(
...
         incr_state = None
 
+        total_decoder_elapsed_time = 0
+        num_decoder_forwards = 0
         for _ts in range(max_ts):
             if all((b.is_done() for b in beams)):
                 # exit early if possible
                 break
 
+            decoder_start_time = time.time()
```
stephenroller, line 1127: https://stackoverflow.com/questions/17579357/time-time-vs-timeit-timeit

fyi

parlai/core/torch_generator_agent.py
```diff
@@ -1114,15 +1117,21 @@ def _generate(
...
             score, incr_state = model.decoder(decoder_input, encoder_states, incr_state)
             # only need the final hidden state to make the word prediction
             score = score[:, -1:, :]
             score = model.output(score)
```
stephenroller, line 1131: we might need to do a flag anyway. this code isn't going to be deterministic. pytorch gpu execution is asynchronous and so you can easily be significantly undercounting the model effort here. I believe we have to use CUDA_LAUNCH_BLOCKING to get accurate measurements.

parlai/core/torch_generator_agent.py
jxmsML: interesting to know!!!!

EricMichaelSmith: good point, changing

parlai/core/torch_generator_agent.py
EricMichaelSmith: Yes, good point - changed

EricMichaelSmith:
Will integrate this into a standalone benchmarking script soon

PR #2989, opened 2020-08-18
Add new HRED agent and integration tests

NathanShuster:
**Patch description**

Implements a basic HRED agent agent that extends seq2seq. I originally implemented a heavier version that had a lot of dataset specific features for SIMMC and stripped out all the dataset specific stuff for this PR. Is a general implementation but has some room for added functionality (for example, currently only supports LSTM, not GRU) that is not in this PR.

**Testing steps**
As part of my intern project I reproduced the SIMMC (https://github.com/facebookresearch/simmc) HRED baseline in ParlAI, and successfully reproduced the original results. To generalize for broader use in ParlAI, I stripped out all the dataset specific features and consulted with Stephen in the process. I ran some follow up testing with the DailyDialog set, shared my findings with Stephen and my manager, and the numbers were deemed sufficiently close to send out a PR.

**Logs**
Can be run with any of the ParlAI datasets with -m seq2seq/hred


parlai/agents/seq2seq/hred.py
```diff
@@ -0,0 +1,397 @@
+#!/usr/bin/env python3
```
stephenroller, line 1: how do you feel about making this its own agent?

parlai/agents/hred/ ? I think it would be better, and highlight it as its own thing.

additionally, separate the HredAgent into hred.py, and put all the pytorch modules into a `modules.py`.

stephenroller: Can you also add an integration test for this? Something like these two:

https://github.com/facebookresearch/ParlAI/blob/4b257b8509a309870f1037b61fbf268797443741/tests/test_transformers.py#L252-L304

Additionally, maybe we can try to do a pretrained model for `-t integration_tests:multiturn_candidate`. Something with like 1 layer, 16 esz, 32 hidden size

parlai/agents/seq2seq/hred.py
```diff
@@ -0,0 +1,397 @@
...
+    OutputLayer,
+    RNNEncoder,
+    Seq2seq,
+    _transpose_hidden_state,
+    opt_to_kwargs,
+)
+from .seq2seq import Seq2seqAgent
+
+
+class HredModel(Seq2seq):
```
stephenroller, line 16: Why are we inheriting from seq2seq, when many of the options I think we don't support?

NathanShuster: Ah, I structured it like this because it originally contained most of this functionality--wasn't sure if it was going to be added back in in some form later--I will switch it to TorchGeneratorModel and update the PR. Thanks!

parlai/agents/seq2seq/hred.py
NathanShuster: Yup I can definitely make it its own agent and add the integration test--how do I get the baseline valid['metric'] figures? Just run it myself?

stephenroller:
Test that is failing is that you need to add an `__init__.py` into your hred folder, with its only contents being the shebang and copyright comment

stephenroller:
Tests will pass on master, depended on #3019.

PR #2779, opened 2020-06-19
[Transformer] Factor out TransformerEncoder.forward

klshuster:
**Patch description**
Nothing functionally changing here - just factoring out the `forward` function in `TransformerEncoder`.

Allows for more focused overrides in potential subclasses, as different elements of the forward pass are self-contained; `forward` now becomes:

1. `forward_embedding`
2. `forward_layers`
3. `reduce_output` 

**Testing steps**
Local testing and mostly relying on CI here to make sure nothing breaks.

**Other information**
I also added some type annotation 😄

PR #3009, opened 2020-08-25
[Transformer] Minor refactor of transformer decoder forward

klshuster:
**Patch description**
Similar to change in encoder (see #2779) we factor out the `forward_layers` of a TF decoder into its own function.

**Testing steps**
CircleCI


klshuster:
Flaky tests be flaky

PR #3020, opened 2020-08-28
[Squad2] Add option to change label for impossible answers

emilydinan:
**Patch description**
Squad2 features questions with impossible answers. Currently, the label for these answers is an empty string. One might one to change this, say, for example, to train a model that responds "I don't know" to impossible answers. 

Data can be viewed by:
```
parlai dd -t squad2 --impossible-answer-string "I don't know" -ne 1000
```

### DIFF
--- a/parlai/tasks/squad2/agents.py
+++ b/parlai/tasks/squad2/agents.py
@@ -90,6 +90,7 @@ class DefaultTeacher(DialogTeacher):
 
     For SQuAD, this does not efficiently store the paragraphs in memory.
     """
+
     @classmethod
     def add_cmdline_args(cls, argparser):
         agent = argparser.add_argument_group('Squad2 teacher arguments')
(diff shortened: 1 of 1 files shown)


parlai/tasks/squad2/agents.py
```diff
@@ -23,6 +33,10 @@ class IndexTeacher(FixedDialogTeacher):
     location of the answer in the context.
     """
 
+    @classmethod
+    def add_cmdline_args(cls, argparser):
+        add_common_cmdline_args(argparser)
```
EricMichaelSmith, line 38: Hmm are there params from this that we need to use in this teacher?

emilydinan: yes --impossible-answer-string

EricMichaelSmith: ha, makes sense :)

PR #3022, opened 2020-08-28
Upgrade to pytorch 1.6

stephenroller:
**Patch description**
(Almost) get us to pytorch 1.6. Turns out we covered most of it already. Additionally sort of needs #3025.

**Testing steps**
CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,7 @@ installtorchgpu16: &installtorchgpu16
   run:
     name: Install torch GPU and dependencies
     command: |
-        python -m pip install --progress-bar off torch==1.6.0+cu101 torchvision==0.7.0+cu101 -f https://download.pytorch.org/whl/torch_stable.html
+      python -m pip install --progress-bar off torch==1.6.0+cu101 torchvision==0.7.0+cu101 -f https://download.pytorch.org/whl/torch_stable.html
       python -m pip install --progress-bar off 'subword-nmt==0.3.7'
       python -m pip install --progress-bar off pytorch-pretrained-bert
       python -m pip install --progress-bar off 'torchtext==0.7.0'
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -89,47 +89,51 @@ installdeps: &installdeps
...
-      python -m pip install --progress-bar off 'torch==1.3.1'
-      python -m pip install --progress-bar off 'subword-nmt==0.3.7'
-      python -m pip install --progress-bar off pytorch-pretrained-bert
-      python -m pip install --progress-bar off transformers
-      python -m pip install --progress-bar off fairseq
+      python -m pip install --progress-bar off 'torch==1.5.1'
+      python -c 'import torch; print("Torch version:", torch.__version__)'
+      python -m torch.utils.collect_env
+
+installtorchgpu14: &installtorchgpu14
```
EricMichaelSmith, line 113: Hmm we still want to support 1.4 instead of just supporting the last 2 versions?

parlai/utils/testing.py
```diff
@@ -30,6 +30,13 @@
     TORCH_AVAILABLE = False
     GPU_AVAILABLE = False
 
+try:
+    import torchvision  # noqa: F401
+
+    VISION_AVAILABLE = True
+except ImportError:
+    VISION_AVAILABLE = False
```
EricMichaelSmith, line 38: Hmm should we log this or no?

requirements.txt
```diff
@@ -8,7 +8,7 @@ flake8==3.7.8
 fvcore==0.1.1.post20200716
 gitdb2==2.0.5
 GitPython==3.0.3
-torch==1.4.0
+torch
```
EricMichaelSmith, line 11: no `torch>=1.4.0`?

parlai/utils/testing.py
stephenroller: Nah,  the test says "skipped because"

.circleci/config.yml
stephenroller: I don't want to force the entire team to upgrade right now.

PR #3024, opened 2020-08-29
Allow colored logging in jupyter notebooks.

stephenroller:
**Patch description**
Right now, jupyter/colab users see these full verbose logs that look like we're logging to files. It's quite annoying. This improves their experience.

**Testing steps**
image


stephenroller:
Lint is fixed in master

PR #3025, opened 2020-08-29
LR Scheduler fixes

stephenroller:
**Patch description**
Loading a model trained with `cosine` or `linear` LR schedules was throwing an exception on PyTorch 1.6. This is because pytorch 1.6 broke our method for forwarding an LR schedule with `step(num_epochs)`.

List of changes:
- Migrates us to the new sanctioned LR loading
- Works on both pytorch 1.4 and 1.6 (didn't bother testing 1.5.1)
- Adds a new `test_agents/unigram` model, which is a torch generator agent that always outputs fixed distributions. This is useful to have a TA that has extremely few parameters (tens or hundreds) and can be trained extremely fast. (Also added its own CI for this)
- Add end2end tests for select LR schedulers
- Add resume LR checks for select LR schedulers.

**Testing steps**
- Ran select unit tests in both 1.4 and 1.6.
- New CI

### DIFF
--- a/parlai/agents/test_agents/unigram.py
+++ b/parlai/agents/test_agents/unigram.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+
+
+"""
+UnigramAgent always predictions the same vocabulary distribution.
+
+It is a full TorchGeneratorAgent model, so it can be used heavily in testing,
+while being very quick to optimize.
+"""
+
+import torch
+import torch.nn as nn
+from parlai.core.torch_generator_agent import TorchGeneratorAgent, TorchGeneratorModel
+
+
+class UnigramEncoder(nn.Module):
+    def forward(self, x):
+        return None
+
+
+class UnigramDecoder(nn.Module):
+    def forward(self, x, encoder_state, incr_state=None):
+        return x, None
+
+
+class UnigramModel(TorchGeneratorModel):
+    def __init__(self, dictionary):
+        super().__init__()
+        self.encoder = UnigramEncoder()
+        self.decoder = UnigramDecoder()
+        self.v = len(dictionary)
+        self.p = nn.Parameter(torch.randn(self.v))
+
+    def output(self, do):
+        desired = tuple(list(do.shape) + [self.v])
+        return self.p.unsqueeze(0).unsqueeze(0).expand(desired)
+
+    def reorder_encoder_states(self, *args):
+        return None
+
+    def reorder_decoder_incremental_state(self, *args):
+        return None
+
+
+class UnigramAgent(TorchGeneratorAgent):
+    def build_model(self):
+        return UnigramModel(self.dict)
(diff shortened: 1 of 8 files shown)


parlai/agents/test_agents/unigram.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+UnigramAgent always predictions the same vocabulary distribution.
```
klshuster, line 8: nit: `always predicts`

parlai/agents/test_agents/unigram.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
```
klshuster, line 1: nice example of how easy it is to subclass TGA

parlai/nn/lr_scheduler.py
```diff
@@ -48,21 +48,32 @@ def __init__(self, hard_reset, warmup_updates, warmup_rate):
...
         else:
             self.warmup_scheduler = None
 
+    def get_last_lr(self):
+        s = self.warmup_scheduler if self._is_lr_warming_up() else self.scheduler
+        try:
+            # pytorch 1.5 or newer
+            return s.get_last_lr()[0]
+        except AttributeError:
+            # pytorch 1.4 or older
```
klshuster, line 74: maybe add a TODO here to remove this check if we ever stop supporting 1.4?

parlai/nn/lr_scheduler.py
```diff
@@ -87,12 +98,17 @@ def load_state(self, states):
         """
         Load state of scheduler from states.
         """
-        if self.scheduler and 'lr_scheduler' in states:
-            self.scheduler.load_state_dict(states['lr_scheduler'])
         if states.get('warmup_scheduler') and getattr(self, 'warmup_scheduler', None):
             self.warmup_scheduler.load_state_dict(states['warmup_scheduler'])
+        if self.scheduler and 'lr_scheduler' in states:
```
klshuster, line 103: is there any reason for moving this?

parlai/nn/lr_scheduler.py
```diff
@@ -480,13 +501,13 @@ def __init__(
     def _linear_lr(self, step):
         # this multiplicative factor ensures linear decay rate
         # lr_mult = float(self.max_lr_steps - step - 1) / float(self.max_lr_steps - step)
-        lr_mult = max(0.0, 1.0 - step / self.max_lr_steps)
+        lr_mult = max(0.0, 1e-6 + (1.0 - step / self.max_lr_steps) * (1 - 1e-6))
```
klshuster, line 504: what does this do?

klshuster (approved):
thanks for adding a bunch of tests too

parlai/nn/lr_scheduler.py
stephenroller: Ya idk something crashed for me but I don't remember.

parlai/nn/lr_scheduler.py
stephenroller: It makes it so the LR never hits exactly 0 (at most 1e-6 * the peak LR)

PR #3040, opened 2020-09-02
[DOC DAY]TGA doc

dexterju:
**Patch description**
This adds TGA docs on:
* Custom loss functions
*  Adding fields to the Batch
* Adding temp history string

### DIFF
--- a/docs/source/tutorial_torch_generator_agent.md
+++ b/docs/source/tutorial_torch_generator_agent.md
@@ -121,7 +121,7 @@ class Encoder(nn.Module):
 
 The decoder is initialized in the same way as the encoder, but now the forward pass reflects the fact that the input tokens need to be passed through the embedder and LSTM one token at a time rather than all at once. If this is the first pass through the decoder, we pass a tuple `encoder_state` to the LSTM that consists of the initial hidden and cell state, as taken from the output of the encoder. If this is a subsequent pass through the decoder, the LSTM will have given us the current values of the hidden and cell states, so we pass that back in to the LSTM, after potentially having reindexed the states with `ExampleModel().reorder_decoder_incremental_state()`.
 
-```
+```python
 class Decoder(nn.Module):
 
     def __init__(self, embeddings, hidden_size):
@@ -142,7 +142,6 @@ class Decoder(nn.Module):
         return output, incr_state
 ```
 
-
 ### Training
 
 The full code for the agent can be seen [here](https://github.com/facebookresearch/ParlAI/tree/master/parlai/agents/examples/seq2seq.py). To call training:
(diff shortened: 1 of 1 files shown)


dexterju:
image

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -154,3 +153,79 @@ parlai train_model -m examples/seq2seq \
...
+            loss = (
+                1 - self.knowledge_alpha
+            ) * token_loss + self.knowledge_alpha * know_loss
+        if return_output:
+            return (loss, model_output)
+        else:
+            return loss
+```
+
+### Adding more fields to the batch
```
stephenroller, line 202: This is really part of Torch agent.

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -154,3 +153,79 @@ parlai train_model -m examples/seq2seq \
...
+
+Batch is a `namedtuple` containing data sent to an agent. This class is the input type of the train_step and eval_step functions. Agents can override the `batchify` function to return an extended namedtuple with additional fields if they would like. However, we recommend calling the parent function to set up these fields as a base. An example is:
+
+```python
+    def batchify(self, obs_batch):
+        batch = super().batchify(obs_batch)
+        reordered_observations = [obs_batch[i] for i in batch.valid_indices]
+
+        checked_sentences = []
+        for obs in reordered_observations:
```
stephenroller, line 212: can we use a simpler example. This one has 3 fields and makes it more complicated than it needs to be.

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -154,3 +153,79 @@ parlai train_model -m examples/seq2seq \
...
+            checked_sentence = '{} {} {}'.format(
+                obs.get('title', ''), TOKEN_KNOWLEDGE, obs.get('checked_sentence', '')
+            )
+            checked_sentences.append(checked_sentence)
+
+        batch['checked_sentence'] = checked_sentences
+        return batch
+```
+
+### Adding tokens to dialogue history
```
stephenroller, line 222: also torch agent. But we should discuss `_set_text_vec`

docs/source/tutorial_torch_generator_agent.md
```diff
@@ -154,3 +153,79 @@ parlai train_model -m examples/seq2seq \
...
 You should get a perplexity of around 140 and a token accuracy of around 28% on the ConvAI2 validation/test set.
+
+### Using custom loss function
+
+It is common in downstream tasks that people want to use a custom loss function on top of the original cross-entropy loss used to train the generative model. You can achieve this by overriding the `compute_loss` function. And a concrete example can be found [here](https://github.com/facebookresearch/ParlAI/blob/0c25349ac91abe2a5f232c63985aed14a37f5dcf/projects/wizard_of_wikipedia/generator/agents.py#L121).  Please note `compute_loss` changes its output type based on the `return_output` parameter. It either returns only the loss or returns the loss and whatever the model's `output.forward()` was. Be mindful of that. 
+
+Let us walk though an example:
+
+```python
+    def compute_loss(self, batch, return_output=False):
```
stephenroller, line 164: instead of copy pasting, can you give a simple example?

dexterju: Can you make up an use case? I

stephenroller: Label smoothing would be really nice, but you could do something silly like "Also minimize the sum of all the logits"

stephenroller: how about a regularizer on the encoder states.

`loss = nll_loss + encoder_states.square().sum()`

PR #3049, opened 2020-09-02
[Doc Day] Totally rewrite Worlds docs.

stephenroller:
**Patch description**
I completely rewrote the Worlds docs. They try to focus on a _high level_ explanation of _why_ we have to do sharing and _how_ batching is implemented. It tries to move away from what the user needs to implement, since no one ever writes `batch_act` manually.

**Testing steps**
Local render.


docs/source/tutorial_basic.md
```diff
@@ -171,15 +171,13 @@ Another simple world we include is MultiAgentDialogWorld, which is
...
-we include both a BatchWorld and a HogwildWorld. These worlds are
-automatically used when either the `numthreads` parameter or the
-`batchsize` parameter are set to greater than one. Some extra
-functionality is needed to get these to work on the side of both the
-teacher and the learner, but we'll cover that in a different tutorial
-(see: tutorial\_worlds).
+:::{note} Advanced Worlds
+We also include a few more advanced "container" worlds: in particular, we
+include both a BatchWorld and a DyamicBatchWorld. These worlds may be used when
+certain options are sent. See the [tutorial\_worlds](Worlds) tutorial to
```
mojtaba-komeili, line 177: I think this was supposed to be a link. The link here is not active in my doc.

docs/source/tutorial_basic.md
```diff
@@ -171,15 +171,13 @@ Another simple world we include is MultiAgentDialogWorld, which is
...
-We also include a few more advanced "container" worlds: in particular,
-we include both a BatchWorld and a HogwildWorld. These worlds are
-automatically used when either the `numthreads` parameter or the
-`batchsize` parameter are set to greater than one. Some extra
-functionality is needed to get these to work on the side of both the
-teacher and the learner, but we'll cover that in a different tutorial
-(see: tutorial\_worlds).
+:::{note} Advanced Worlds
+We also include a few more advanced "container" worlds: in particular, we
+include both a BatchWorld and a DyamicBatchWorld. These worlds may be used when
```
jaseweston, line 176: Dynamic

docs/source/tutorial_basic.md
jaseweston: same, link doesnt wort

docs/source/tutorial_worlds.md
```diff
@@ -1,9 +1,9 @@
-Data Handling and Batching
+Worlds, Sharing & Batching
 ==========================
 
-__Authors__: Alexander Holden Miller, Kurt Shuster
+__Authors__: Alexander Holden Miller, Kurt Shuster, Stephen Roller
 
-:::{tip} Before you begin
+:::{important} Before you begin
```
jaseweston, line 6: this doesn't work  in the github "rich diff" mode but maybe it isn't supposed to? hard to see if it works tho from here in that case..

docs/source/tutorial_worlds.md
```diff
@@ -1,9 +1,9 @@
...
 If you are unfamiliar with the basics of displaying data or calling train or
 evaluate on a model, please first see the [getting started](tutorial_basic)
 section. If you are interested in creating a task, please see
```
jaseweston, line 9: please see.... ?

jaseweston: oh .. maybe none of the links work from here.. that's awkward

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
-Shared Worlds.
+This document discusses the overview of [__Worlds__](parlai.core.worlds.World), a
+core concept within ParlAI. It aims to be a high level overview of
+_how_ some concepts are implemented in ParlAI, so that you can know what's
+happening behind the scenes, but you are unlikely to write any custom worlds
+yourself.
+
+Worlds are where agents live, and the world defines the communication flow
+between agents.  If you are familiar with
+[Reinforcement Learning](https://en.wikipedia.org/wiki/Reinforcement_learning),
```
jaseweston, line 24: sad we lost the executable world because you need that for proper RL with something like LIGHT... but guess we can always restore when finally get that

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+processing multiple items at the same time.
 
-We create shared agents by instantiating them in the following way:
+In the remainder of this document, we will cover how Batching works, and eventually
+discuss the implementation of Dynamic Batching.
+
+## Agent Sharing
+
+_Agent Sharing_ is the primary mechanism by which we implement batching, model
+serving, and many other ParlAI tricks.  Agent sharing works by creating
```
jaseweston, line 84: parlAI tricks makes it sound bad.. maybe say it nicer?

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+_clones_ of an Agent. Each clone of an Agent has _shared state_ and
+_independent state_. Independent state includes things like the current
+dialogue context (conversation history). Shared state includes things like
```
jaseweston, line 87: (conversation history, which is different for every batch element)

jaseweston (commented):
couple more comments

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+        super().__init__(opt, shared)
+        if shared is None:
+            # we are the original. We need to initialize any state
+            self.model = self.build_model()
+        else:
+            # we are a clone. use the shared dict to set any variables.
+            # this is where we incorporate any shared state
+            self.model = shared['model']
+
+    def share(self):
```
mojtaba-komeili, line 112: Does clone call this, or should we be calling this after cloning?

docs/source/tutorial_worlds.md
stephenroller: it works

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
-
-![image](_static/img/world_batchagent.png)
-
-Now, the agent can process the entire batch at once. This is especially
-helpful for GPU-based models, which prefer to process more examples at a
-time.
+Clones are why most agents have this rough structure in their initialization
+procedures.
 
-:::{tip} Implementing batch\_act and act
```
mojtaba-komeili, line 122: I do not understand how we are treating `shared` object. Do we pass `None` to the first agent, and then retrieve the `shared` attribute of that and pass to the agents? In other words, is the first agent in charge of creating the shared object, and the clones receive it in their constructor?

docs/source/tutorial_worlds.md
stephenroller: Yeah I've sketched how it should be for Margaret. But if we're going to add it, it needs to be something that's actually supported -- not test-less code without READMEs or documentation that's 3 years old.

docs/source/tutorial_worlds.md
stephenroller: Clone calls this. Will clarify.

docs/source/tutorial_worlds.md
stephenroller: Exactly. Will clarify.

docs/source/tutorial_basic.md
stephenroller: Thanks. We have an automatic link checker in the CI tests. Will fix.

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+            self.model_copies[i].self_observe(model_acts[i])
+            # compute any metrics of the response
+            self.teacher_copies[i].observe(model_acts[i])
+```
+
+This logic is more complicated, but enables us to efficiently implement batched
+operations. The new logic can be encapsulated in this graph:
+
+<center>
+<img src="_static/img/world_batchbasic.png" alt="Parley with batch_act" width="100%"/>
```
mojtaba-komeili, line 246: This is identical to the image above. Did you mean the other image to go here?

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+            # perform preprocessing, vectorizing, etc.
+            self.model_copies[i].observe(teacher_act)
+            # produce a model response
+            model_act = self.model_copies[i].act()
+            # compute any metrics of the response
+            self.teacher_copies[i].observe(model_act)   
+```
+
+<center>
+<img src="_static/img/world_batchbasic.png" alt="Independent acts" width="100%"/>
```
mojtaba-komeili, line 202: Could you remove the spell checker line from under the title?

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+        # make batchsize copies of all the models
+        self.teacher_copies = []
+        self.model_copies = []
+        self.batchsize = opt['batchsize']
+        for i in range(self.batchsize):
+            self.teacher_copies.append(self.teacher.clone())
+            self.model_copies.append(self.model.clone())
+
+```
+This initialization code is then represented by this graphic:
```
mojtaba-komeili, line 179: Is agent0 the manager (master) here and the rest its workers? Does agent0 itself run the conversations, or is it just handling the clones in the batch?

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+This is called the __batchsort__ algorithm. Note that our batch size remains
+fixed at 4, but we've grouped our worst examples so that they happen at the
+same time.
+
+:::{tip}
+You can use this your training or evaluation runs with `-dynb batchsort` or
+`--dynamic-batching batchsort`.
+:::
+
+::{note}
```
mojtaba-komeili, line 329: `:::{not}`

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+Notice how the conversations no longer play in order. ParlAI handles all of this
+extra complexity for you. Remember, you only need to implement state tracking
+for _individual_ conversations, and along with `batch\_act`.
+::
```
mojtaba-komeili, line 333: `:::`

docs/source/tutorial_worlds.md
```diff
@@ -13,120 +13,353 @@ section. If you are interested in creating a task, please see
...
+8  |   1    |  May the Force be with you. |       6
+11 |   1    |  I'm king of the world! |       5
+3  |   1    |  Here's Johnny! |       2
+4  |   1    |  Expecto patronum! | 2
+
+With this algorithm, we tried to get as close to 80 words as possible, without
+going over. The first batch ends up with exactly 80 words. The batch ends up
+with only 51. But we've now reduced the number of batches we need to process
+from 4 to only 2! This is the trick of how dynamic batching can provide
+[massive speed ups](tutorial_fast).
```
mojtaba-komeili, line 360: I wonder if this has any effect on the performance of the model: since short or long text may contain different topics, this may make gradients more homogenous compared to regular (fully random) batching. Is there any reference that shows this does not happen? Could you add that here, if one exists? Reading this I get the impression that this might be a trade-off between cost of training and accuracy.

mojtaba-komeili (commented):
How do you feel about replacing images with dot diagrams? With the style changes that we discussed.

docs/source/tutorial_worlds.md
stephenroller: Internal tests show that it converged about 2.5x faster with a difference of like 0.05 ppl worse.

It's actually a little more complex than presented here. Namely, for fp16 reasons, we treat all examples the same length up to a divisor of 8. So we actually get a fair amount of stochasticity.

stephenroller:
I didn't think the dot diagrams were complete yet. I'd rather merge this and then have a separate PR for the dot diagrams.

docs/source/tutorial_worlds.md
mojtaba-komeili: Sorry, I meant `:::{note}`, LOL.

docs/source/tutorial_worlds.md
stephenroller: Ran a spell check, not what you mean here

mojtaba-komeili (approved):
Thanks for clarifying some of the explanations, particularly on the cloning process.

PR #3052, opened 2020-09-02
Test and details in convo render

dianaglzrico:
**Patch description**
Test was added for `convo_render.py` that will create a self-chat, save it to file to then render it to HTML. Also fixed a typo and updated field names.

**Logs**

Tested with:
```
pytest -s tests/test_convo_render.py
```


stephenroller (approved):
Thanks. Just autoformat. This looks awesome.

vedantpuri (approved):

PR #3055, opened 2020-09-04
[tests] Switch to OverfitTeacher for some model tests.

stephenroller:
**Patch description**
We have many flaky tests. They've started to get even flakier lately. This has become quite annoying, and made the value of the red X much smaller.

In reality, in most of these tests, we're just trying to ensure that they can optimize and learn. For this, the classical overfit-on-a-single-batch can cover most of our use cases, while being considerably less flaky, and considerably faster.

This patch does the following:
- Move a bunch of tests to overfit-on-batch. They can generally be learned much faster, and without retry as much
- A bunch of generation tests have been moved to generating a single example and verifying its results
- Adjust parallelism levels to reduce overall usage cost, exploiting the fact that the longgpu tests take a long time and so we can afford to run the other-version tests in parallel a bit slower
- Delete the BadCandidateTeacher. I've decided I just don't care about this test.

**Testing steps**
New CI.


tests/test_distributed.py
```diff
@@ -183,8 +189,14 @@ def test_no_model_parallel(self):
             'transformer/classifier',
         ]:
             config['model'] = m
-            with self.assertRaises(RuntimeError):
```
mojtaba-komeili, line 186: Is there a reason for replacing this? Is throwing the exception here indeterministic? Because what you do next is checking for this error again.

stephenroller: It's just to get the `finally` clause which is necessary tearDown. I can't use tearDown itself because the retry decorator doesn't call it.

mojtaba-komeili (commented):
It looks like harmless changes to me, if they pass the tests.

parlai/core/worlds.py
```diff
@@ -145,7 +145,7 @@ def clone(self):
         """
         Create a duplicate of the world.
         """
-        return type(self)(opt=copy.deepcopy(self.opt), agents=None, shared=self.share())
+        return type(self)(opt=self.opt, agents=None, shared=self.share())
```
klshuster, line 148: why were we doing this before?

tests/nightly/gpu/test_dialogpt.py
```diff
@@ -16,20 +16,21 @@ class TestDialogptModel(unittest.TestCase):
...
         valid, test = testing_utils.train_model(
             dict(
-                task='integration_tests:nocandidate',
+                task='integration_tests:overfit',
                 model='hugging_face/dialogpt',
                 add_special_tokens=True,
                 add_start_token=True,
-                optimizer='sgd',
-                learningrate=1,
+                optimizer='adam',
```
klshuster, line 26: this fits in memory and all that?

tests/test_distributed.py
```diff
@@ -97,11 +93,13 @@ def test_multitask_distributed(self):
         # Tests that DialogData.get() is doing the right thing
         # Ensure no duplication of examples among workers
         # It would be 200 if each worker did all the examples
-        self.assertEqual(valid['exs'].value(), 500)
-        self.assertEqual(test['exs'].value(), 500)
+        self.assertEqual(valid['exs'].value(), 4 * 4 + 4)
```
klshuster, line 96: should this `4` be a constant somewhere? assuming it relates to `corpus_size`

tests/test_memnn.py
```diff
@@ -7,30 +7,28 @@
 import unittest
 import parlai.utils.testing as testing_utils
 
-BATCH_SIZE = 1
-NUM_EPOCHS = 3
-LR = 1
 
-
-@unittest.skip
```
klshuster, line 15: 😮

tests/test_other_scripts.py
```diff
@@ -61,20 +60,6 @@ def test_verify_data(self):
         assert report['missing_labels'] == 0
         assert report['missing_text'] == 0
 
-    def test_verify_bad_data(self):
```
klshuster, line 64: was this taking a long time?

klshuster (approved):
really like this

parlai/core/worlds.py
stephenroller: Abundance of caution. But the worlds were making their own deep copy on init

tests/nightly/gpu/test_dialogpt.py
stephenroller: Passes in CI :p

tests/test_distributed.py
stephenroller: Yes it should be.

tests/test_other_scripts.py
stephenroller: No. I was finding it nondeterministic.

stephenroller: I think Paul was complaining about verify data breaking. Maybe that's what I was observing. 🤔

(Either way, I just decided to yeet this test, lol, maybe that's bad reasoning)

klshuster: that's kinda odd that a verify data script was nondeterministic? I'm curious what was breaking if that's what was going on - nevertheless fine with yeeting as it looks like we have tests to make sure verify data works on good data

PR #3060, opened 2020-09-08
[BART] Don't override init if it exists

emilydinan:
**Patch description**
We should not override the init model if it exists. This causes
- issues with load from checkpoint when you are pre-empted
- the BART model to download, even when you're fine-tuning an already trained model


stephenroller (approved):
use pathmanager please

klshuster (approved):
🤕

PR #3063, opened 2020-09-09
fromfile warning for using valid/test

jaseweston:


parlai/tasks/fromfile/agents.py
```diff
@@ -83,6 +84,9 @@ def __init__(self, opt, shared=None):
         datafile = opt['fromfile_datapath']
         if self.opt['fromfile_datatype_extension']:
             datafile += "_" + self.opt['datatype'].split(':')[0] + '.txt'
+        else:
+            if self.opt['datatype'] == 'valid' or self.opt['datatype'] == 'test':
+                logging.warn('You are using this fromfile data as a valid or test set without setting fromfile_datatype_extension to true. Please be aware this uses directly the file you indicated, make sure this is not the same as your training file.')
```
stephenroller, line 89: I think it's also going to print Batchsize number times. can you add a "if shared is None" protection.

Also I don't think it will handle valid:stream right but that's unusual in fromfile anyway.

jaseweston: added

stephenroller (approved):
Cheers

PR #3065, opened 2020-09-09
[BART] allow overriding BART args

emilydinan:
**Patch description**
Allow override of BART args. For example, @wyshi  is hacking on some extra positional embeddings so we can deal with longer contexts, so we need to be able to override `--n-positions 2048`. This allows us to do so.


klshuster (approved):
thanks, i'd just let long gpu tests run to make sure this doesn't break BART tests

stephenroller (approved):

PR #3066, opened 2020-09-09
[History] Global end token bug

emilydinan:
**Patch description**
The global end token in history was not respecting the argument set by the flag `--history-add-global-end-token` and instead was adding `self.end_idx` always.


stephenroller:
Is this dead?

PR #3068, opened 2020-09-10
Fix masked_fill no-op

EricMichaelSmith:
Fix issue with using the out-of-place version of .masked_fill()


stephenroller (approved):
Hopefully no CI breaks

PR #3073, opened 2020-09-11
Two inaugural FAQ questions

EricMichaelSmith:
Adds 2 questions to the FAQ that might be useful to others. (Not sure if we would want to hold off on adding FAQs until we have a critical mass of them, but this could be a start.)

### DIFF
--- a/docs/source/faq.md
+++ b/docs/source/faq.md
@@ -3,3 +3,11 @@
 This document contains a number of questions that are regularly asked on GitHub
 Issues.
 
+**Why can't I reproduce the results of an evaluation on a task with a pretrained model?**
+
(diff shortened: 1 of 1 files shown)


stephenroller:
Yay! You could also ask why is my model not saying anything (skip generation)

stephenroller (approved):

klshuster (approved):
looks great! agree with stephen u can add that one more question "Why is my model not generating a response?" "make sure skip_generation is set to false" but fine without it

EricMichaelSmith:
> Yay! You could also ask why is my model not saying anything (skip generation)

Agreed - added :)

PR #3077, opened 2020-09-13
Disable zipfile serialization option of torch.save

maryamdaneshi:
Due to torch versioning, the option is needed when loading the models from manifold
This will fix the error 
raise RuntimeError('New zip file serialization is currently not supported '
RuntimeError: New zip file serialization is currently not supported by Manifold serialization. Please set _use_new_zipfile_serialization=False when doing torch.save

**Testing steps**
With the patch, I was able to successfully load the model from manifold and use it for inference on the messaging_internal

the model was saved in manifold here:
manifold ls cair_models/tree/test_model/v8
  2041509510 model
  2041509510 model.checkpoint
       64601 model.checkpoint.dict
        3593 model.checkpoint.dict.opt
        3474 model.checkpoint.opt
       39400 model.checkpoint.trainstats
       64577 model.dict
        3567 model.dict.opt
        3474 model.opt
         245 model.test
       39401 model.trainstats
         248 model.valid

test:
.
Ran 1 test in 0.031s

OK

### DIFF
--- a/parlai/utils/torch.py
+++ b/parlai/utils/torch.py
@@ -55,10 +55,12 @@ def atomic_save(state_dict: Any, path: str) -> None:
 
     if io_util.USE_ATOMIC_TORCH_SAVE:
         with open(path + ".tmp", "wb") as f:
+            # TODO: Upgrade the manifold torch version. See T75615407
             torch.save(state_dict, f, _use_new_zipfile_serialization=False)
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
Can we leave a TODO linking to a task to upgrade torch manifold?

PR #3081, opened 2020-09-14
Self-chat batch size assertion error

dianaglzrico:
**Patch description**
(Bug found during docday) Self-chat breaks whenever `-bsz` is greater than 1, adding `AssertionError` for now to prevent from breaking.


jxmsML:
thanks for the fix!!!

jxmsML (approved):

Issue #3064, opened 2020-09-09
Error in self_chat with 2 seed utterances and no context


shaigrt:
Hi,

I tried to run a model that I fine-tuned over some dataset (using the blender-3B training command) in self-chat mode. The SelfChatWorld world of this task has 2 seed utterances and no context:

```
class SelfChatWorld(SelfChatBaseWorld):
    def __init__(self, opt, agents, shared=None):
        super().__init__(opt, agents, shared)

    def get_openers(self, episode_num: int) -> Optional[List[str]]:
        return ['Hi', 'Hi, how are you?']

    def get_contexts(self):
        return None
```

In this setting I am getting the following error:

```
File "debater-parlai/parlai/scripts/self_chat.py", line 80, in _run_self_chat_episode
    world.parley()
  File "debater-parlai/parlai/tasks/self_chat/worlds.py", line 168, in parley
    self.agents[i].self_observe(self.acts[i])
  File "debater-parlai/parlai/core/torch_agent.py", line 1656, in self_observe
    self._validate_self_observe_invariants()
  File "debater-parlai/parlai/core/torch_agent.py", line 1728, in _validate_self_observe_invariants
    "You're self_observing without having observed something. Check if "
RuntimeError: You're self_observing without having observed something. Check if you're missing a step in your observe/act/self_observe loop.
```

As I understand, the self_observe function has a condition that self.observation is already set, but this is empty for the first seed utterance. Am I missing something?

stephenroller:
Can you share what command you are running?

shaigrt:
After I modify the SelfChatWorld of blended_skill_talk/worlds.py with the above implementation, running the command `parlai/scripts/self_chat.py -t blended_skill_talk -mf zoo:blender/blender_3B/model` fails with this error.

stephenroller:
@dianaglzrico can you take a look?

dianaglzrico:
Hi @shaigrt a fix has been added to attend this issue, I've tested it and seems to work!

stephenroller:
Closing, but feel free to reopen if there are still issues.

PR #3082, opened 2020-09-15
Add ED test to Self-chat

dianaglzrico:
**Patch description**
Self-chat was breaking when  seed utterances but not context were provided like [in here](https://github.com/facebookresearch/ParlAI/issues/3064), as `self_observe()` was being called without an `observe()`. An ED test was added that resembles this case to verify the fix.


parlai/tasks/empathetic_dialogues/worlds.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class SelfChatWorld(SelfChatBaseWorld):
+    def __init__(self, opt, agents, shared=None):
+        super().__init__(opt, agents, shared)
+
+    def get_contexts(self):
+        return None
+
+    def share(self):
```
stephenroller, line 17: This is a no-op

parlai/tasks/empathetic_dialogues/worlds.py
```diff
@@ -0,0 +1,19 @@
...
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from parlai.tasks.self_chat.worlds import SelfChatWorld as SelfChatBaseWorld
+
+
+class SelfChatWorld(SelfChatBaseWorld):
+    def __init__(self, opt, agents, shared=None):
```
stephenroller, line 11: So is this

tests/test_self_chat.py
```diff
@@ -29,3 +29,16 @@
...
+        opt = pp.parse_args(
+            [
+                '-mf',
+                'zoo:blender/blender_3B/model',
+                '-t',
+                'empathetic_dialogues',
+                '--seed-messages-from-task',
+            ]
+        )
+        self_chat.self_chat(opt)
```
stephenroller, line 44: Can we use SelfChat.main(task='empathetic_dialogues', ...)

tests/test_self_chat.py
```diff
@@ -29,3 +29,16 @@
             ]
         )
         self_chat.self_chat(opt)
+
+    def test_ed(self):
+        pp = self_chat.setup_args()
+        opt = pp.parse_args(
+            [
+                '-mf',
+                'zoo:blender/blender_3B/model',
```
stephenroller, line 38: This is a very heavy model for a small test. Can we use the tutorial generator?

tests/test_self_chat.py
```diff
@@ -29,3 +29,16 @@
...
+                '-t',
+                'empathetic_dialogues',
+                '--seed-messages-from-task',
```
stephenroller, line 41: Does this not need a true argument?

stephenroller (commented):
A few questions

parlai/tasks/empathetic_dialogues/worlds.py
dianaglzrico: fixed

parlai/tasks/empathetic_dialogues/worlds.py
dianaglzrico: fixed

tests/test_self_chat.py
dianaglzrico: Why or what would be the difference to use this?

tests/test_self_chat.py
dianaglzrico: Yeah you're right about that, will change it.

tests/test_self_chat.py
dianaglzrico: no this is good, it's a `store_true` flag

tests/test_self_chat.py
dianaglzrico: Yeah you're right, I'll change it

tests/test_self_chat.py
stephenroller: Cleaner and newer syntax. Functionally equivalent.

tests/test_self_chat.py
stephenroller: Can we also change that to not be a store_true then? Store_true is impossible to override once it's been set in an opt file

dianaglzrico: Yeah, that's true, I'll change it.

dianaglzrico: Addressed this [here](https://github.com/facebookresearch/ParlAI/pull/3113).

emilydinan (commented):
i don't really understand the difference between openers and context.. can you explain @dianaglzrico ? should we change teh default in the base world?

dianaglzrico:
> eally understand the difference between openers and context.. can y

As I understand in self chat, context is used to provide info depending on the task (personas, topics, etc) and openers which can be taken from the task too but are seeded messages used to open the conversation. The issue that I linked in the description reported an error when you provide openers and no context at all.

stephenroller (approved):
Thank you

PR #3083, opened 2020-09-15
[ImageFeaturizer] Detectron feature extraction

klshuster:
## Patch description
Add a Detectron Image Featurizer to ParlAI.

To make abundantly clear the image type, this will be featured under `--image-mode faster_r_cnn_152_32x8d`. One will need to install a few external dependencies (opencv, and a separate repo) in order to use the featurizer.

A few other minor details:
- no longer lazy import torch in the `ImageLoader` as it's now a requirement for ParlAI
- I've included a few more tests for the `ImageLoader`

CC @apsdehal, @vedanuj 
## Testing steps
Added test for extracting image features with the new image mode.

Note that I'm not sure I did the CircleCI stuff correctly... mostly tagging @stephenroller in that regard

## Logs
```
$ pytest -k TestImageLoader
=============================================================================== test session starts ===============================================================================
platform linux -- Python 3.6.10, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
rootdir: /private/home/kshuster/ParlAI, configfile: pytest.ini, testpaths: tests
plugins: requests-mock-1.8.0
collected 388 items / 384 deselected / 4 selected

tests/test_image_featurizers.py ....                                                                                                                                        [100%]

============================================================================== slowest 10 durations ===============================================================================
35.87s call     tests/test_image_featurizers.py::TestImageLoader::test_resnext
24.60s call     tests/test_image_featurizers.py::TestImageLoader::test_faster_r_cnn
4.54s call     tests/test_image_featurizers.py::TestImageLoader::test_resnet

(7 durations < 0.005s hidden.  Use -vv to show these durations.)
============================================================ 4 passed, 384 deselected, 15 warnings in 69.87s (0:01:09) ============================================================
```


stephenroller:
Can you rebase?

.circleci/config.yml
```diff
@@ -406,12 +415,13 @@ jobs:
           paths:
             - "~/nvidia-downloads/"
       - restore_cache:
-          key: deps-20200828-gpu16-{{ checksum "requirements.txt" }}
+          key: deps-20200915-gpu16-{{ checksum "requirements.txt" }}
```
stephenroller, line 418: Bump all the caches with a global string replace please. It's a pain when they get out of sync.

.circleci/config.yml
```diff
@@ -138,6 +138,15 @@ installtorchcpu: &installtorchcpu
       python -c 'import torch; print("Torch version:", torch.__version__)'
       python -m torch.utils.collect_env
 
+
+installdetectrondeps: &installdetectrondeps
+  run:
+    name: Install opencv, vqa-maskrcnn-benchmark
+    command: |
+      python -m pip install opencv-python==4.2.0.34
+      git clone https://gitlab.com/vedanuj/vqa-maskrcnn-benchmark.git maskbench
```
apsdehal, line 147: I would suggest depending on a fixed commit. Just in case, we make some changes for our projects, we don't want this to break down.

.circleci/config.yml
klshuster: i'm assuming you just mean the ones related to `requirements.txt`? as there are plenty of out-of-sync dates currently with other files (specifically looking at `nvidia-downloads`)

stephenroller: I mean all the `deps-` ones

parlai/core/image_featurizers.py
```diff
@@ -39,6 +46,7 @@
     'resnext101_32x16d_wsl_spatial': ['resnext101_32x16d_wsl', -2],
     'resnext101_32x32d_wsl_spatial': ['resnext101_32x32d_wsl', -2],
     'resnext101_32x48d_wsl_spatial': ['resnext101_32x48d_wsl', -2],
+    'faster_r_cnn_152_32x8d': ['', -1],
```
EricMichaelSmith, line 49: Why is this string empty?

parlai/core/image_featurizers.py
```diff
@@ -251,4 +256,229 @@ def load(self, path):
...
+
+        if im.shape[-1] > 3:
+            im = np.array(img.convert("RGB")).astype(np.float32)
+
+        # IndexError: too many indices for array, grayscale images
+        if len(im.shape) < 3:
+            im = np.repeat(im[:, :, np.newaxis], 3, axis=2)
+
+        im = im[:, :, ::-1]
+        im -= np.array([102.9801, 115.9465, 122.7717])
```
EricMichaelSmith, line 347: Hmm - is it stated in the MMF code what this is? Some constant to center the distribution probably?

parlai/utils/testing.py
```diff
@@ -53,6 +53,14 @@
 except ImportError:
     BPE_INSTALLED = False
 
+try:
+    import maskrcnn_benchmark  # noqa: F401
+    import cv2  # noqa: F401
+
+    DETECTRON_AVAILABLE = True
```
EricMichaelSmith, line 60: Random thought - if these import statements break in the future, does that mean that our Detectron CI checks will stop silently because `skipUnlessDetectron()` will skip them?

EricMichaelSmith (commented):
Seems sensible! Will let others with more knowledge of the MMF code approve

parlai/utils/testing.py
klshuster: yes, though that could be said for every import statement in this file

parlai/core/image_featurizers.py
klshuster: I didn't see any documentation around it, perhaps @apsdehal or @vedanuj could offer some insight

parlai/core/image_featurizers.py
klshuster: this is a mapping from ParlAI `--image-mode` to the PyTorch cnn instantiation string & layer to use. This dict is where we get our image mode choices from, so i've included the faster r cnn features in here, however the extractor does not use a PyTorch CNN like the other image modes do (it has its own extractor) so there's no relevant mapping

parlai/utils/testing.py
EricMichaelSmith: Yeah, I see that a ton of them have the same try/except logic

stephenroller:
Yeah looks like this broke something in CI but not obvious what

stephenroller:
> error: torch 1.4.0 is installed but torch==1.6.0 is required by {'torchvision'}

klshuster:
Yeah that was because I added `torchvision` to the requirements.txt file, however it kinda breaks CI because you need to install torchvision with the correct torch build.

klshuster:
CI passes!

stephenroller (approved):
Can I assign you a future (different than this PR) task of adding a test for parlai/scripts/extract_image_feature.py?

klshuster:
Yeah sure, i'll probably want to clean up that script as well

PR #3088, opened 2020-09-17
[Chunk Teacher] Add sleep before enqueuing next load

emilydinan:
**Patch description**
Add a sleep before enqueueing the next load when the buffer is filled. This allows the thread to be released. I've been seeing issues where the thread is not released and this seems to help.

### DIFF
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -2223,6 +2223,7 @@ class ChunkTeacher(FixedDialogTeacher, ABC):
             if self.is_train or self.tot_samples_loaded % self.dws == self.rank:
                 self.samples.put(sample)
             self.tot_samples_loaded += 1
+        time.sleep(0.01)  # release thread
         # and start loading the next chunk
         self._enqueue_request()
 


stephenroller:
This is a gross hack. I'd like to see proof it's necessary first.

It also deserves a paragraph long comment explaining it.

klshuster (commented):
i too am a little curious where the deadlock is happening

as hacky as it is, I did a similar thing for the chat services, but that involved threads fighting with the GIL after releasing for cuda ops, so just wondering what's causing the threads to lock up here

stephenroller:
We think that is happening here too

klshuster:
why is the sleep outside the while loop then?

stephenroller:
We're sleeping inside the infinite loop of enqueuing/processing requests

emilydinan:
proof/explanation: i have a teacher (internal only) for which display_data (ne=10) hangs after completion if i do not add this sleep here. i tried sleeping for 0.001 and it still hangs, so 0.01 it is.

klshuster:
I don't have an issue with this change but indeed I believe it'd be good to document in the code two things:

1. why we think the deadlock is happening (seems that we know this)
2. why this sleep solves it (perhaps this goes in tandem with 1)

Not looking for any lengthy explanation, just maybe that we think there's thrashing going on trying to reacquire GIL after CUDA ops?

Anyway, seemed that @stephenroller felt a little more strongly about this so will leave final approval to him

stephenroller:
The code should link to this thread, and this thread should contain information about the issue.

PR #3094, opened 2020-09-18
Move torch.load/torch.save to PathManager.

stephenroller:
**Patch description**
See D23791495 for internal rational for this patch. tldr, manifold doesn't work as well with torch as previously expected.

**Testing steps**
Internal testing for manifold. CI here.


klshuster (approved):
looks to be very similar if not the same as internal diff; lgtm

PR #3095, opened 2020-09-18
[dist] Allow non-tcp based distributed setup.

stephenroller:
**Patch description**
See D23761809 for internal rationale of this patch.

In some settings, we may wish to use something other than tcp for distributed initialization. In particular, in internal cases, distributed initialization is supported by the `env://` method. This patch refactors out the initialization method so that it is directly a parameter, rather than demanding the TCP based method.

**Testing steps**
See internal patch. CI here.


EricMichaelSmith (approved):
Makes sense, good to have!

PR #3097, opened 2020-09-18
Fix for flaky DialoGPT test

EricMichaelSmith:
**Patch description**
Fix for flaky DialoGPT unittest

**Testing steps**
tests/nightly/gpu/test_dialogpt.py


stephenroller:
Don't double batch size. That actually makes the task harder, bc the overfit teacher uses different examples per batchsize.

Instead, increase epochs to correspond to number of updates.

EricMichaelSmith:
> Don't double batch size. That actually makes the task harder, bc the overfit teacher uses different examples per batchsize.
> 
> Instead, increase epochs to correspond to number of updates.

Whoops, doubled the bs instead of # epochs, yes. Changed

stephenroller (approved):
Thanks for this

PR #3103, opened 2020-09-22
Porting ACUTE-Eval task to Hydra

EricMichaelSmith:
Making ACUTE-Eval Mephisto task compatible with the latest version of Mephisto, which uses Hydra. Unit task will be added in a future PR, because the infrastructure for running unit tests of Mephisto blueprints hasn't been built out yet.

(2 unit tests fail due to issue fixed by https://github.com/facebookresearch/ParlAI/pull/3105 .)


EricMichaelSmith:
(@JackUrb @jxmsML this PR is no longer WIP and is ready for review.)

parlai/crowdsourcing/tasks/acute_eval/README.md
```diff
@@ -76,29 +74,29 @@ Note that we assume that "dialogue" consists of strictly alternating turns (e.g.
...
+- `task_description`: Includes detailed information about the kind of task that the HIT contains. On the Amazon Mechanical Turk web site, the HIT description appears in the expanded view of search results, and in the HIT and assignment screens.
+- `task_tags`: One or more words or phrases that describe the HIT, separated by commas. On MTurk website, these words are used in searches to find HITs.
+- `additional_task_description`: Additional text to show in the left-hand pane of the chat window.
 
 
-### CLI arguments
+### YAML and CLI arguments
 
-A comprehensive list of settings specific to ACUTE-Eval can be found in `add_args_to_group()` in `acute_eval_blueprint.py`. For the arguments most likely to be useful for running ACUTE-Eval, see `example_script.py`.
+A comprehensive list of settings specific to ACUTE-Eval can be found in `AcuteEvalBlueprintArgs` in `acute_eval_blueprint.py`. For examples of how these arguments can be set in practice, see `conf/example.yaml`. For modifying the `task_reward` argument (for example) on the command line, append `mephisto.task.task_reward=${DESIRED_TASK_REWARD}`.
```
JackUrb, line 102: I'd note as well that people can specify their own yaml files in `conf/<anything>.yaml` and then use those instead of `example` with `conf=<anything>`.

parlai/crowdsourcing/tasks/acute_eval/acute_eval_blueprint.py
```diff
@@ -35,123 +89,31 @@ class AcuteEvalBlueprint(Blueprint):
...
 
     @classmethod
-    def assert_task_args(cls, opts: Any) -> None:
+    def assert_task_args(cls, args: DictConfig, shared_state: SharedTaskState) -> None:
         """
         Ensure that the data can be properly loaded.
         """
-        if opts.get("pairings_filepath") is not None:
-            pairings_filepath = os.path.expanduser(opts["pairings_filepath"])
+        if args.blueprint.get("pairings_filepath") is not None:
```
JackUrb, line 101: Note - a `get` from a `OmegaConf` `DictConfig` is unlike a regular get. If you don't know if there will be a value (like times when you've specified the default as `MISSING`, you should provide a default value. `args.blueprint.get("pairings_filepath", None)` is what you want here.

parlai/crowdsourcing/tasks/acute_eval/acute_eval_blueprint.py
```diff
@@ -35,123 +89,31 @@ class AcuteEvalBlueprint(Blueprint):
...
-                len(opts.get("pairings_task_data")) > 0
-            ), "Length of data dict provided was 0"
         else:
             raise AssertionError(
                 "Must provide one of a data csv, json, or a list of tasks"
             )
 
-        if opts.get("block_on_onboarding_fail") is True:
-            if opts.get("block_qualification") is None:
+        if args.blueprint.get("block_on_onboarding_fail") is True:
```
JackUrb, line 111: If you do provide a default value however, you can just access directly without running the risk of crashing due to the variable not being set. `args.blueprint.block_on_onboarding_fail` works here.

You already do this elsewhere, so this one is just a nit really.

parlai/crowdsourcing/tasks/acute_eval/acute_eval_blueprint.py
```diff
@@ -174,12 +136,12 @@ def get_initialization_data(self) -> Iterable["InitializationData"]:
         # TODO(#99) once we can release HITs over time, configure this to
         # release as many as needed thusfar and top off when
         # onboardings fail
-        print(self.opts)
+        print(self.args)
```
JackUrb, line 139: I think this was leftover debug, idk if we still want it.

parlai/crowdsourcing/tasks/acute_eval/example_script.py
```diff
@@ -3,97 +3,70 @@
...
+
+    world_opt = {}
+
+    custom_bundle_path = cfg.mephisto.blueprint.get("custom_source_bundle", None)
+    if custom_bundle_path is not None:
+        assert os.path.exists(custom_bundle_path), (
+            "Must build the custom bundle with `npm install; npm run dev` from within "
+            f"the {TASK_DIRECTORY}/webapp directory in order to demo a custom bundle "
+        )
+        world_opt["send_task_data"] = True
```
JackUrb, line 64: This isn't relevant for your task, was just for ParlAI custom tasks

JackUrb (commented):
This is generally looking good! Some comments inline.

How was the experience transitioning to Hydra?

.gitignore
```diff
@@ -134,6 +134,8 @@ tmp/
 # Heroku things
 heroku-cli-*
 *.db
+# Mephisto
```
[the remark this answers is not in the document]
EricMichaelSmith: Good point, changing this

parlai/crowdsourcing/tasks/acute_eval/README.md
EricMichaelSmith: Great, added

parlai/crowdsourcing/tasks/acute_eval/acute_eval_blueprint.py
EricMichaelSmith: Cool, changed

parlai/crowdsourcing/tasks/acute_eval/acute_eval_blueprint.py
EricMichaelSmith: Hmm, but we call `block_on_onboarding_fail` on the first line here and `block_qualification` on the second line, so we still need to check if `block_on_onboarding_fail` is set

parlai/crowdsourcing/tasks/acute_eval/acute_eval_blueprint.py
EricMichaelSmith: Good point, removed

parlai/crowdsourcing/tasks/acute_eval/example_script.py
EricMichaelSmith: Oy, yes, good catch

parlai/crowdsourcing/tasks/acute_eval/acute_eval_blueprint.py
JackUrb: But in the dataclass you provide a default value:
```
block_on_onboarding_fail: bool = field(
    default=True, metadata={"help": "whether to block on onboarding failure"}
)
```
Due to this, unless the user is doing something that they really shouldn't be (like using a configuration for this that doesn't derive from your `AcuteEvalBlueprintArgs`, this should always be set.

Realistically though this doesn't need to change, I was more using it as a discussion point to convey some of how hydra defaults work.

EricMichaelSmith: Ah, yes, I see what you mean that a default value has been set here - yes, it's good in that case to assume that the attribute does exist. Changed

EricMichaelSmith:
> This is generally looking good! Some comments inline.
> 
> How was the experience transitioning to Hydra?

Great thanks! Hmm, the experience wasn't too bad once I got a sense of the scope of what needed to be changed (i.e. that most of the scripts needed to be tweaked at least somewhat). Thanks again for the docs on this!

JackUrb (approved):
LGTM! Thanks for addressing the relevant concerns and taking the time to make this upgrade.

PR #3105, opened 2020-09-22
Prevent MTurk config folders from needing __init__.py

EricMichaelSmith:
**Patch description**
Exclude all folders named `task_config` from needing `__init__.py` files, and rename the folders of config files in the turn-annotation MTurk task and the turn_annotations_static Mephisto task to `task_config`

**Testing steps**
`pytest tests/nightly/gpu/test_crowdsourcing.py`


stephenroller (approved):
thanks eric

PR #3106, opened 2020-09-22
Update precommit config to include git-secrets

stephenroller:
**Patch description**
Add git-secrets pre-commit hook. Also re-enable the black one.

Unfortunately, git-secrets is not easy to use in a per-remote fashion. Unlike other, more standard hooks, adding it to the config remotely causes it to not work quite right; furthermore, it requires every user run the `git secrets --register-aws` command at setup, which is also more likely to be forgotten than correctly done. My solution was to copy over git secrets set it up right for our repo.

Developers will need to run once `pip install pre-commit; pre-commit install`.

New github autocheck too. Required me to black a bunch of files that have grown stale.

**Testing steps**
Local testing. Will make a fresh clone and try there too, and get a volunteer.

### DIFF
--- a/.github/scripts/git-secrets
+++ b/.github/scripts/git-secrets
@@ -1,0 +1,379 @@
+#!/usr/bin/env bash
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+#
+# ParlAI's modifications from original:
+# - automatically calls aws-provider, and hardcodes AWS parameters
+# - doesn't require seperate installs for everyone
+#
+# ---------------------------------------------------------------------------
+#
+# Copyright 2010-2013 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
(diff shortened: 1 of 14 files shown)


.github/scripts/git-secrets
```diff
@@ -0,0 +1,379 @@
+#!/usr/bin/env bash
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+#
+# ParlAI's modifications from original:
+# - automatically calls aws-provider, and hardcodes AWS parameters
```
klshuster, line 8: thanks for including these comments - could you also note in the code explicitly where the changes are?

.pre-commit-config.yaml
```diff
@@ -1,6 +1,13 @@
 repos:
--   repo: https://github.com/ambv/black
-    rev: stable
+-   repo: https://github.com/psf/black
```
klshuster, line 2: new black repo??

klshuster (approved):
cool, glad to have this in ParlAI

.pre-commit-config.yaml
stephenroller: Idk I think we were referring to a fork...

stephenroller: Ah it moved from a developer's individual account to the python software foundation. The previous is an alias for the new correct name

PR #3111, opened 2020-09-23
Mephisto dep cleanup

JackUrb:
**Patch description**
Resolves a number of dependency issues by removing unneeded ones and updating the rest.

**Testing steps**
`npm install; npm audit; npm run dev`

Tasks in the Mephisto repo are building and rendering without these, so I don't see them being necessary here.


EricMichaelSmith (approved):
Seems reasonable, thanks! That's quite a large number of dependencies, ha...

PR #3112, opened 2020-09-23
Allow missing args when using `--init-opt`

EricMichaelSmith:
**Patch description**
Set `--allow-missing-init-opts True` to be able to load in opt args with `--init-opt` even if some of the args no longer exist. Also added unit testing of `--init-opt`

**Testing steps**
`python -m unittest tests.test_opt`


tests/test_opt.py
```diff
@@ -126,3 +127,51 @@ def test_compare_opts_load_raw(self):
...
+            init_opt = Opt({'made_up_arg': 'foo'})
+            init_opt.save(init_opt_path)
+
+            # Assert that the opt file normally can't be loaded in
+            with self.assertRaises(RuntimeError):
+                _ = ParlaiParser(True, True).parse_kwargs(init_opt=init_opt_path)
+
+            # Assert that the opt file *can* be loaded in if we set
+            # --allow-missing-init-opts
+            _ = ParlaiParser(True, True).parse_kwargs(
```
stephenroller, line 175: do you want to add a check that the resulting opt has the `made_up_arg` in it?

stephenroller (approved):

PR #3114, opened 2020-09-23
Add test for interactive_web

stephenroller:
**Patch description**
Add a test for `interactive_web.py`. Works by launching the server in a background thread, then polling the server, and then shutting it down.

Not sure if I want this approach being used widely. It seems to work reasonably well in a limited capacity, but unsure how reliable it is.

**Testing steps**
New CI.


klshuster (approved):
really great

PR #3117, opened 2020-09-24
Test parlai/core/script.py

stephenroller:
**Patch description**
Add tests for the abstract class in script.py, and the supercommand.

**Testing steps**
New CI.
```
$ coverage report parlai/core/script.py
Name                    Stmts   Miss  Cover
-------------------------------------------
parlai/core/script.py     137      2    99%
```

### DIFF
--- a/tests/test_script.py
+++ b/tests/test_script.py
@@ -54,7 +54,7 @@ class TestScriptRegistry(unittest.TestCase):
         assert opt.get('bar') == 'sneaky'
 
     def test_main_noargs(self):
-        with patch.object(sys, 'argv', ['test_script.py']):
+        with patch.object(sys, 'argv', ['test_script.py']):  # argv[0] doesn't matter
             opt = _TestScript.main()
         assert opt.get('foo') == 'defaultvalue'
         assert opt.get('bar') == 'sneaky'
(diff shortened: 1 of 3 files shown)


klshuster (approved):
big fan of this, hopefully this will catch the super subtle differences that sometimes come up between e.g. `parlai tm` and `python -m parlai.scripts.train_model` - ideally we can continue adding tests to this for maximum robustness

tests/test_script.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_supercommand(self):
+        opt = script.superscript_main(args=['test_script', '--foo', 'test'])
+        assert opt.get('foo') == 'test'
+
+    def test_no_setup_args(self):
+        with self.assertRaises(NotImplementedError):
+            script.superscript_main(args=['no_setup_args'])
+
+    def test_help(self):
+        with testing_utils.capture_output() as output:
```
jxmsML, line 91: is this redundant w.r.t. L101 - 104

jxmsML (approved):

tests/test_script.py
stephenroller: lol yes it is, will drop

tests/test_script.py
```diff
@@ -0,0 +1,104 @@
...
+        assert opt.get('foo') == 'test'
+        assert opt.get('bar') == 'sneaky'
+
+    def test_main_args(self):
+        opt = _TestScript.main('--foo', 'test')
+        assert opt.get('foo') == 'test'
+        assert opt.get('bar') == 'sneaky'
+
+    def test_main_noargs(self):
+        with patch.object(sys, 'argv', ['test_script.py']):
```
stephenroller, line 57: wait why does this work

stephenroller: ah, argv[0] doesn't matter

PR #3122, opened 2020-09-25
swap lines in eval to have partial results

wyshi:
move `aggregate_unnamed_reports` to below `world_logger.write`, so that we don't have to wait for all nodes to finish to write logs.


stephenroller (approved):
lgtm, thanks!

PR #3123, opened 2020-09-28
Return text for all hyps in the beam from TorchGenerator

akhti:
**Patch description**
1. Return texts for all hyps in the beam.
2. Fix name collision in _generate()

**Testing steps**

Testing with BART model:
```
        ex = {"text": input_seq, "episode_done": True}
        self.parlai_agent.observe(ex)
        response = self.parlai_agent.act()
        print(response["top_n_texts"])
```

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -882,31 +882,23 @@ class TorchGeneratorAgent(TorchAgent, ABC):
                 )
 
         preds = None
+        top_n_texts = []
         if self.skip_generation:
             warn_once("--skip-generation true produces limited metrics")
         else:
             maxlen = self.label_truncate or 256
-            # print("BEAM:", self.beam_size)
             beam_preds_scores, beams = self._generate(batch, self.beam_size, maxlen)
             preds, scores = zip(*beam_preds_scores)
-            # print(beam_preds_scores)
-            # print(beams)
             self._add_generation_metrics(batch, preds)
 
-            n_best_beam_preds_scores = [b.get_rescored_finished() for b in beams]
-
-            # get the top prediction for each beam (i.e. minibatch sample)
-            beam_preds_scores = [n_best_list[0] for n_best_list in n_best_beam_preds_scores]
-            top_n_texts = []
-            for i, beams2 in enumerate(n_best_beam_preds_scores):
+            for beam in beams:
                 top_n_texts.append([])
-                for b, (tokens, score) in enumerate(beams2):
+                for tokens, score in beam.get_rescored_finished():
                     try:
                         top_n_texts[-1].append((self._v2t(tokens), score))
-                    except:
-                        print("Bad tokens:", tokens, flush=True)
-                        logging.error("Bad tokens: %s", tokens)
-                        raise
+                    except KeyError:
+                        logging.error("Decoding error: %s", tokens)
+                        continue
 
         cand_choices = None
         # TODO: abstract out the scoring here
@@ -936,7 +928,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
             # compute additional bleu scores
             self._compute_fairseq_bleu(batch, preds)
             self._compute_nltk_bleu(batch, text)
-        return Output(text, cand_choices, token_losses=token_losses)
+        return Output(text, cand_choices, token_losses=token_losses, top_n_texts=top_n_texts)
 
     def _treesearch_factory(self, device):
         method = self.opt.get('inference', 'greedy')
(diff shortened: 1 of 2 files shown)


akhti:
Seems that mypy dies with some unrelated error

parlai/core/torch_generator_agent.py
```diff
@@ -882,14 +882,24 @@ def eval_step(self, batch):
                 )
 
         preds = None
+        top_n_texts: List[List[Tuple[str, float]]] = []
```
stephenroller, line 885: Add a comment saying # bsz x beamsize

parlai/core/torch_generator_agent.py
```diff
@@ -1168,8 +1180,8 @@ def _generate(
         # get the top prediction for each beam (i.e. minibatch sample)
         beam_preds_scores = [n_best_list[0] for n_best_list in n_best_beam_preds_scores]
         if self.opt.get('verbose'):
```
stephenroller, line 1182: Let's drop this section now that it's in the return value

parlai/core/torch_generator_agent.py
```diff
@@ -918,7 +928,9 @@ def eval_step(self, batch):
             # compute additional bleu scores
             self._compute_fairseq_bleu(batch, preds)
             self._compute_nltk_bleu(batch, text)
-        return Output(text, cand_choices, token_losses=token_losses)
+        return Output(
+            text, cand_choices, token_losses=token_losses, top_n_texts=top_n_texts
+        )
```
stephenroller, line 933:
```suggestion
        retval = Output(text, cand_choices, token_losses=token_losses)
        if not self.skip_generation:
            retval.top_n_texts = top_n_texts
        return retval
```

I would rather it only provide the return value if we have skip generation on.

stephenroller (commented):
Just a couple minor things.

@klshuster can you weigh in on naming?

stephenroller:
Also, can we have a unit test please?

You can model a test off of this (just generate from the model with beam size > 1 and assert the return value has that field and it's not empty, and that the first one is the "true" generation):
https://github.com/facebookresearch/ParlAI/blob/7bcaa0e12886d21110e51212ab266e7ad8a0b1a0/tests/test_transformers.py#L278-L319

parlai/core/torch_generator_agent.py
```diff
@@ -882,14 +882,24 @@ def eval_step(self, batch):
...
             preds, scores = zip(*beam_preds_scores)
             self._add_generation_metrics(batch, preds)
 
+            for beam in beams:
+                top_n_texts.append([])
+                for tokens, score in beam.get_rescored_finished():
+                    try:
+                        top_n_texts[-1].append((self._v2t(tokens), score.item()))
+                    except KeyError:
+                        logging.error("Decoding error: %s", tokens)
```
klshuster, line 900: i know i was pulled in for naming but I am curious when this error arises?

klshuster (commented):
perhaps it's more fitting to have it be `beam_texts` as opposed to `top_n_texts` (if that's the input you're looking for @stephenroller). I like `top_n_texts` in theory but realize that it's not entirely clear the differentiation between that and `cand_choices`; including something about `beam` or like `top_n_gen_texts` would differentiate better (that last one is a bit wordy though)

parlai/core/torch_generator_agent.py
akhti: During beam search with high large size on model with low PPL that predicts not NLP, but structured messages sometimes things like `__FP16_PAD_0__` are produced that cause exception in v2t that I can't nicely avoid.

akhti:
@stephenroller I've added a test, but due to reasons beyond my understanding the beam contains `beam_size + 1` elements instead `beam_size` (see comment in the test). For topk sampling it's fine. This is not related to this PR, so just flagging

Also, renamed the field as per Kurt's suggestion.

tests/test_transformers.py
```diff
@@ -275,6 +275,32 @@ def test_beamsearch(self):
...
+
+        agent = create_agent_from_model_file(
+            'zoo:unittest/beam_blocking/model',
+            opt_overrides={"beam_size": size, "inference": "beam"},
+        )
+        agent.observe({'text': '5 5 5 5 5 5 5', 'episode_done': True})
+        response = agent.act()
+        self.assertTrue("beam_texts" in response)
+        # This line fails!!! It returns 4 options instead of 4
+        # self.assertEqual(len(response["beam_texts"]), size)
```
stephenroller, line 293: I guess let's make it >=, instead of ==

stephenroller: Can you also assert the types are correct?

stephenroller (approved):
Thank you!

stephenroller:
IDK what's wrong with CI. I don't see any obvious lint issues in this latest ocmmit, so feel free to ignore that.

klshuster:
I found CI for lint/mypy to almost never work on my PRs from my fork fwiw

akhti:
Updated the test. Seems that I can't merge even with approval. So please feel free to merge anyone :)

How often is `parlai` package updated?

stephenroller:
Okay will land after tests pass.

It's released whenever I feel there's a reason for a release. Next one is probably in a week. You can always `python -m pip install git+https://github.com/facebookresearch/ParlAI.git@master` to install the latest.

PR #3128, opened 2020-09-29
changed use_reply default from 'label' to 'none'

candacelax:
This is a small change in the command line args for TorchAgent. This changes default value for the argument _use_reply_ from "label" to "none". This ensures the ground-truth labels won't, by default, be included in the history for multi-turn agents.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -618,7 +618,7 @@ class TorchAgent(ABC, Agent):
         )
         agent.add_argument(
             '--use-reply',
-            default='none',
+            default='label',
             hidden=True,
             choices=['label', 'model', 'none'],
             help='Which previous replies to use as history. If label, use '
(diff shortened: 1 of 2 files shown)


stephenroller:
That will break a large number of models. We want to limit the change to just `TorchClassifierAgent`. You can do this by adding `parser.set_defaults(use_reply='none')` to `TorchClassifierAgent.add_cmdline_args`.

emilydinan:
> That will break a large number of models. We want to limit the change to just `TorchClassifierAgent`. You can do this by adding `parser.set_defaults(use_reply='none')` to `TorchClassifierAgent.add_cmdline_args`.

yep just move this to TorchClassifierAgent instead of TorchAgent

candacelax:
Whoops good catch, sorry! Will update

stephenroller (approved):
This looks right to me. Thanks!

PR #3131, opened 2020-09-29
Add gender bias agent

EricMichaelSmith:
Add trait for appending gender token to the context, as well as two agents that use it


klshuster (approved):
lgtm

PR #3132, opened 2020-09-29
Flag to enable polarity categories

EricMichaelSmith:
**Patch description**
Add the ability to have a fraction of personalities in the Image-Chat teacher be replaced by either "positive/neutral" or "negative", depending on the personality


klshuster (approved):
awesome!

PR #3134, opened 2020-09-30
Add jga metric to taskmaster2

moyapchen:
Fairly straightforward.

Test Plan:

Ran an eval model with the model that I had trained previously on taskmaster2 w/ PPL as a metric locally.

### DIFF
--- a/parlai/tasks/taskmaster2/agents.py
+++ b/parlai/tasks/taskmaster2/agents.py
@@ -171,6 +171,9 @@ class _Abstract(DialogTeacher):
                     f'{domain}_slot_r',
                     AverageMetric(correct, len(teacher_action['slots'])),
                 )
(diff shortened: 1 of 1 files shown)


parlai/tasks/taskmaster2/agents.py
```diff
@@ -171,6 +171,9 @@ def custom_evaluation(
                     f'{domain}_slot_r',
                     AverageMetric(correct, len(teacher_action['slots'])),
                 )
+                self.metrics.add(
+                    'jga', AverageMetric(correct == len(teacher_action['slots']))
```
stephenroller, line 175: It might be more readable for us to just compare the dictionaries (see FB:multiwoz code), but this works.

stephenroller (approved):

PR #3135, opened 2020-09-30
Add taskmaster2 command-line arg to specify subset of domains

moyapchen:
**Patch description**
Title 

**Testing steps**
Before this change, `parlai dd -t taskmaster2 --display-verbose` displayed a bunch of sports conversations.

After this change, running `parlai dd -t taskmaster2 --display-verbose --domains music` displays music conversations. Tried on a few other domains to validate; also had a print in the `_load_data()` function.

Also verified that no argument case that all domains were used +
defining `domains` multiple times only used the last one.

### DIFF
--- a/parlai/tasks/taskmaster2/agents.py
+++ b/parlai/tasks/taskmaster2/agents.py
@@ -53,9 +53,9 @@ class _Abstract(DialogTeacher):
         argparser.add_argument(
             '--domains',
             nargs='+',
-            default=[],
+            default=DOMAINS,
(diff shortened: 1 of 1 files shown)


moyapchen:
`python tests/datatests/test_new_tasks.py` is broken on taskmaster2, so skipping.

stephenroller (approved):
does choices magically work  with nargs='+'?

moyapchen:
Yeah, it yells at me if I spell things wrong.

PR #3137, opened 2020-09-30
Using the ParlAI color

JackUrb:
**Patch description**
@moyapchen raised the concern that the parl.ai website was not using the same color throughout as the color of the icon, and now the discrepancy can't be unseen. 

This PR replaces all references to the wrong `#ce3239` with the correct `#ec1454`

**Testing steps**
Directly edited the HTML on the site, this now looks like it matches:
Screen Shot 2020-09-30 at 1 53 44 PM


stephenroller (approved):

moyapchen (approved):
yayyayyay

(dammit, github doesn't have macros)

PR #3140, opened 2020-10-01
Default datafile to datatype if not provided.

### LINKED ISSUE #3139: Document that when subclassing DialogTeacher `opt["datafile"]` needs to be populated
Documentation should indicate that `opt["datafile"]` needs to be populated (if `shared` is not) before calling `__init__.super()` when subclassing DialogTeacher as the current error/traceback doesn't clearly indicate what the issue is (example below) without diving deeply into the code.

Alternatively catch this in DialogTeacher and produce a more meaningful error, or provide a default in DialogTeacher if `opt["datafile"]` isn't set.

```
  File ".../parlai/core/worlds.py", line 1211, in _create_task_agents
    task_agents = my_module.create_agents(opt)  # type: ignore
AttributeError: module 'parlai_fb.tasks.work_in_progress.agents' has no attribute 'create_agents'

During handling of the above exception, another exception occurred:
...
<snip>
...
    super().__init__(opt, shared)
  File ".../parlai/core/teachers.py", line 546, in __init__
    **kwargs,
  File ".../parlai/core/teachers.py", line 678, in __init__
    self._load(data_loader, opt['datafile'])
KeyError: 'datafile'
```

stephenroller:
**Patch description**
Change the default value of `opt['datafile']` to be `opt['datatype']` (mostly).

It doesn't break anything, but I would like opinions since I'm making a large default change.

On master, `opt['datafile']` MUST be set before initing your teacher. This is a bit weird, and it's documented, but very poorly (mostly by examples, it's only briefly touched in the data tutorial).

So options:
- Leave as is, but document better
- Make the error message friendly, hinting at what the user should do
- (Implemented in this PR): Just default to the datatype without suffixes. I usually find this convenient when creating a teacher with `setup_data`. I don't see much benefit in setting the "datafile" opt value.

Fixes #3139.

**Testing steps**
New CI.


parlai/utils/data.py
```diff
@@ -13,6 +13,21 @@ class DatatypeHelper:
     Helper class to determine properties from datatype strings.
     """
 
+    @classmethod
+    def fold(cls, datatype: str) -> str:
```
EricMichaelSmith, line 17: Good to have! Surprised we made it long without something like this...

tests/test_utils.py
```diff
@@ -139,5 +140,57 @@ def test_uppercase(self):
         assert string_utils.uppercase("tEst") == "TEst"
 
 
+class TestDatatypeHelper(unittest.TestCase):
```
EricMichaelSmith, line 143: Really like these tests!

EricMichaelSmith (commented):
I think this makes sense, but this is the first time I've heard of this param so I'll let others with more context approve. I must say - `opt['datafile']` did throw me for a loop for a bit because it seems to be one of the rare `opt` params that can't be set via the command line. Is there a usecase for which it *would* be different than `DatatypeHelper.fold(opt['datatype'])`? If not, is it even worth having it as a separate param at all?

klshuster (commented):
while I understand that it's mostly irrelevant what goes in `opt['datafile']` if it's not specified, it's perhaps even more confusing that we populate it with `opt['datatype']`

while not documented adequately, it semantically makes sense that the `datafile` is the file from which we load data. it seems like the issue here is that `setup_data` is now overloaded with two types of args - one is a path to the file, the other is the data fold. 

perhaps we can rectify this by defining `def setup_data(self, file_or_fold)`, and handling both cases? (I see that it is documented in the docstring correctly, but changing the name might be good)

emilydinan:
i'm a bit confused by this change, could you add some more context?

stephenroller:
Context of the change: @skiingpacman recently implemented a teacher, and found that it took a long time to identify what the problem was. It was poorly documented, and not obvious what needed to happen.

So obviously we need to document it. That's very clear.

But there's also a question if there's a more reasonable "default" behavior. I personally never found the `datafile` thing to make sense: why can't I just figure out the filename in setup_data instead of in `__init__`? Especially if you have a dataset that has multiple files (e.g. domains) and you have to iterate over them separately. A single datafile doesn't make much sense there... Similarly for our fake teachers; there we just set opt['datafile'] to the fold.

I will note that in this diff is that `datafile` != `datatype`. It's only the first part, so you really just have to focus on creating what fold of the data.

An alternative is to throw loud errors saying, "You must set `opt['datafile']` in the initialization of your teacher. This may point to the file we will load, or you can alternatively set it to `DatatypeHelper.fold(opt['datatype'])`." Seems a bit weird. If we don't mention the last part, we risk another common error: users assume parlai automatically creates train/val/test folds for them. But at least this is better than `KeyError: datafile`.

@klshuster ya, agreed, the parameter can change in the abstract implementation.

stephenroller:
I will note that as implemented, if `datafile` is not set, then `datafile` never appears anywhere, so there is no entry in the opt that would confuse someone. Only the parameter name.

skiingpacman:
Given that `opt['datafile']` can be set to any arbitrary string, I'm wondering what is the value of keeping it? What is it intended to be used for? Logging?

As @stephenroller points out, except where there is a neat division of data into one file per fold, the more common pattern of implementation that I've seen is that Teachers are implemented with their own private pointers to the data file structures, e.g. `self.dpath` and not passing that information to `setup_data(.)` through `opt['datafile']`. Given that what is lost by removing this all together?

Based on the error I hit it's only required here:
```
    super().__init__(opt, shared)
  File ".../parlai/core/teachers.py", line 546, in __init__
    **kwargs,
  File ".../parlai/core/teachers.py", line 678, in __init__
    self._load(data_loader, opt['datafile'])
KeyError: 'datafile'
```
```
Class DialogData(object):
...
        # self.data is a list of episodes
        # each episode is a tuple of entries
        # each entry is a tuple of values for the action/observation table
        if shared:
            self.image_loader = shared.get('image_loader', None)
            self.data = shared.get('data', [])
            self.cands = shared.get('cands', None)
        else:
            self.image_loader = ImageLoader(opt)
            self.data = []
>>>         self._load(data_loader, opt['datafile'])
            self.cands = None if cands is None else set(c for c in cands)
```

stephenroller:
Nothing ostensibly, except that it's a large break in backwards compatibility.

emilydinan:
I would opt (tehe) for getting rid of `opt['datafile']` if possible

stephenroller:
Discussed in internal meeting, decided to just throw a very loud and descriptive error message, rather than deprecating such a longstanding piece of template code. New push reflects that error message. Seeking reviews.

klshuster (approved):
thanks for the very descriptive warnings

parlai/core/teachers.py
```diff
@@ -539,6 +539,16 @@ def __init__(self, opt, shared=None):
...
             self.data = data_class(opt, shared=shared['data'], **kwargs)
         else:
+            if 'datafile' not in self.opt:
+                raise KeyError(
+                    f"{self.__class__.__name__} is expected to set "
+                    f"self.opt['datafile'] inside `__init__` before calling "
+                    f"`super().__init__`. This will passed to setup_data,"
+                    f"indicating what data to load. If you don't know what to"
+                    f"use, set `opt['datafile'] = "
+                    f"parlai.utils.data.DatatypeHelper.fold(opt['datatype'])` "
```
EricMichaelSmith, line 549: 👍

parlai/core/teachers.py
```diff
@@ -675,6 +705,17 @@ def __init__(self, opt, data_loader=None, cands=None, shared=None, **kwargs):
         else:
             self.image_loader = ImageLoader(opt)
             self.data = []
+
+            if 'datafile' not in opt:
+                raise KeyError(
```
EricMichaelSmith, line 710: nit: maybe abstract out this error message instead of repeating it in 3 different places?

EricMichaelSmith (approved):
Seems reasonable

PR #3143, opened 2020-10-01
Make sure folder is always created when unzipping COCO

EricMichaelSmith:
**Patch description**
The members of the COCO Captions ZIP file, as listed by `ZipFile.namelist()`, include several files in an `annotations` subfolder but not the `annotations` subfolder itself, leading to a `FileNotFoundError` when `.built_data._unzip()` tries to copy these nested files to a subfolder that does not exist. This fix will create the subfolder if it does not exist.  

**Testing steps**
`parlai dd -t coco_caption`

### DIFF
--- a/parlai/core/build_data.py
+++ b/parlai/core/build_data.py
@@ -348,10 +348,19 @@ def _unzip(path, fname, delete=True):
     with zipfile.ZipFile(PathManager.open(fullpath, 'rb'), 'r') as zf:
         for member in zf.namelist():
             outpath = os.path.join(path, member)
+            dirname = os.path.dirname(member)
             if zf.getinfo(member).is_dir():
+                # Member is a subfolder, so make it if it does not exist
                 logging.debug(f"Making directory {outpath}")
                 PathManager.mkdirs(outpath)
                 continue
+            elif dirname != '':
+                # Member is a file *in* a subfolder, so make the subfolder if it does
+                # not exist
+                outfolder = os.path.join(path, dirname)
+                if not PathManager.isdir(outfolder):
+                    logging.debug(f"Making directory {outfolder}")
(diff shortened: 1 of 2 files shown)


stephenroller:
This is a very aggressive change, and I worry a little bit about the effects there could be on a large number of tasks.

Indeed, the sort of standard is that zip files don't have a folder typically...

EricMichaelSmith:
> This is a very aggressive change, and I worry a little bit about the effects there could be on a large number of tasks.
> 
> Indeed, the sort of standard is that zip files don't have a folder typically...

Hmm, a more lightweight workaround could be to just catch the error if you try to copy the file and its dirname doesn't exist, and then to create that missing dir reactively instead of actively. What would you think of this approach instead?

stephenroller:
https://github.com/facebookresearch/ParlAI/blob/0522f47c2d1be77bd49fd31f82942881175c2da8/parlai/tasks/redial/build.py#L21

See like, this teacher the file doesn't have the folder, and it just says "unzip into the redial" folder. If you change this behavior, it will break this teacher.

Why not have the COCO teacher just unzip into the correct folder, rather than updating `unzip` to add silently one?

EricMichaelSmith:
> https://github.com/facebookresearch/ParlAI/blob/0522f47c2d1be77bd49fd31f82942881175c2da8/parlai/tasks/redial/build.py#L21
> 
> See like, this teacher the file doesn't have the folder, and it just says "unzip into the redial" folder. If you change this behavior, it will break this teacher.
> 
> Why not have the COCO teacher just unzip into the correct folder, rather than updating `unzip` to add silently one?

Yes that works - changed

stephenroller (approved):

PR #3145, opened 2020-10-01
Add ability for Metrics to add metrics from another Metrics object

moyapchen:
See https://github.com/facebookresearch/ParlAI/pull/3138 for context and use

(This is also me testing out if stacked branches work as much like stacked commits in mercurial as I'm hoping...)

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -643,7 +643,17 @@ class Metrics(object):
         return {'data': self._data}
 
 
-class TeacherMetrics(Metrics):
+class MetricsCollection(Metrics):
+    """
+    Helper container for Metrics that can have other metrics added to it.
+    """
+
+    def add(self, other: Metrics) -> None:
+        for k, v in other.report():
+            self.add(k, v)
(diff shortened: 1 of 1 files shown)


parlai/core/metrics.py
```diff
@@ -643,7 +643,17 @@ def share(self):
         return {'data': self._data}
 
 
-class TeacherMetrics(Metrics):
+class MetricsCollection(Metrics):
```
stephenroller, line 646: let's just add the add_metrics directly into the root `Metrics` class.

parlai/core/metrics.py
```diff
@@ -643,7 +643,17 @@ def share(self):
...
+    """
+    Helper container for Metrics that can have other metrics added to it.
+    """
+
+    def add_metrics(self, other: Metrics) -> None:
+        for k, v in other.report().items():
```
stephenroller, line 652: Once we add, we could use the internals to add

stephenroller (approved):

PR #3148, opened 2020-10-02
Update google_sgd to use slot metrics

moyapchen:
Do note that this branch is stacked on top of the `slot_teachers` branch, which implements the relevant slot metrics. 

Tested with a pretrained model on taskmaster2 + validated that the response made sense. 

Note that I did have to comment out the "resp.startswith" bit in the slot parsing to use the call token from taskmaster2 since the model expected a different token. The values there were all, unsurprisingly, zeroes, but this seemed okay for now for the purposes of validation.


moyapchen:
Abandoning this; will do internally and sync once it's all nice.

PR #3154, opened 2020-10-05
Adding option for file with annotation indices per conversation and other features to turn annotations static

mwillwork:
Adding features to turn annotations static task:
- Optional params handled in run_task (though this will get handled differently upon merge to Hydra)
- New params:
    - Setting random seed
    - Option to supply a second file specifying which utterance indices to annotation in each conversation
    - Conversation count: only take the first N conversations of the data file
    - Being able to customize the annotation question

@EricMichaelSmith I think this should get merged in before you merge your PR to convert to Hydra


parlai/crowdsourcing/tasks/turn_annotations_static/run_task.py
```diff
@@ -82,6 +83,32 @@ def setup_mephisto(launch_config):
         f'--onboarding-qualification turn-ann-s-onb '
         f"-use-onboarding True "
     )
+    # Optional flags:
+    try:
```
EricMichaelSmith, line 87: This exception handling will not be needed with Hydra because it'll just use the default value of the param, but this is fine for now

EricMichaelSmith (approved):
This seems fine - if you merge this in, I can revise my Hydra PR

stephenroller:
Can we have a slightly more descriptive PR title before merging?

parlai/crowdsourcing/tasks/turn_annotations_static/run_task.py
mwillwork: Agree

PR #3158, opened 2020-10-06
Check directory depending on size param for gpt2 files

ankitade:
**Patch description**
Earlier distilgpt2 model was trying to load the gpt2 weights and hence failing during runtime. Now we load the files according to the type of the model being loaded

**Testing steps**
with-proxy python3 parlai/scripts/train_model.py -t integration_tests -m hugging_face/gpt2 -mf /tmp/model --max-train-time 100 -vmt ppl

### DIFF
--- a/parlai/agents/hugging_face/dict.py
+++ b/parlai/agents/hugging_face/dict.py
@@ -8,6 +8,7 @@ from abc import ABC, abstractmethod
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 from parlai.core.dict import DictionaryAgent
+from parlai.utils.io import PathManager
 
 
 try:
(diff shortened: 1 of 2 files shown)


parlai/agents/hugging_face/dict.py
```diff
@@ -74,22 +74,23 @@ def get_tokenizer(self, opt):
...
+            model_key = "distilgpt2"
+        else:
+            model_key = f"gpt2-{model_sz}"
         # check if datapath has the files that hugging face code looks for
+        hf_dir = os.path.join(opt["datapath"], "hf", model_key)
         if all(
-            os.path.isfile(
-                os.path.join(opt["datapath"], "models", "gpt2_hf", file_name)
-            )
+            PathManager.exists(os.path.join(hf_dir, file_name))
```
stephenroller, line 87: lint seems to have caught a bug

stephenroller (approved):

PR #3159, opened 2020-10-06
[Polyencoder] Update README

klshuster:
**Patch description**
We've gotten lots of questions about dstc7/ubuntu. Just adding a few lines to the project README to give more info on training with those datasets.


stephenroller (approved):
Questions in the README are questions answered only once 😁

PR #3160, opened 2020-10-06
Note about metrics in AbstractWrapperTeacher.act()

EricMichaelSmith:
Add to the docstring that `self.task.act()` must be called in the wrapper's `.act()` method for metrics to be recorded correctly

### DIFF
--- a/parlai/tasks/wrapper/agents.py
+++ b/parlai/tasks/wrapper/agents.py
@@ -60,9 +60,8 @@ class AbstractWrapperTeacher(Teacher, ABC):
     def act(self):
         """
         Act on the previous observation.
-        
-        Note that self.task.act() must be called in this method for metrics to be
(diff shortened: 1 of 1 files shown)


stephenroller (commented):
wait, most metrics are implemented in observe tho, I don't understand.

EricMichaelSmith:
> wait, most metrics are implemented in observe tho, I don't understand.

Yes, but assuming that the wrapped teacher is a `FixedDialogTeacher`, if `self.task.act()` is not called, `self.lastY` will not get set and `self.task.observe()` will not be able to record metrics. But not calling `self.task.act()` would also lead to other problems like (most importantly) not getting the act Message from the wrapped teacher, so I've just generalized this language.

stephenroller (approved):
maybe s/will/should

PR #3162, opened 2020-10-06
Another opt bug fix ('init_model' not being set correctly)

emilydinan:
**Patch description**
`create_agent_from_opt_file` was not properly copying over `'init_model'` as set by `opt`. This causes issues, when, for example, we have a model that has been (1) initialized with some model file (2) requeued and has (3) `load_from_checkpoint = True`. When requeued, the `'init_model'` will revert back to the model set in (1) inside `create_agent_from_opt_file`, instead of using the checkpoint file set by the train script (which is contained in `opt['init_model']`). This in turn leads to TorchAgent loading the model file from the best valid instead of the checkpoint.

Thanks to @wyshi for finding this bug!!

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -324,9 +324,12 @@ def create_agent_from_opt_file(opt: Opt):
         if k not in opt_from_file:
             opt_from_file[k] = v
 
-    opt_from_file['model_file'] = model_file  # update model file path
+    # update model file path to the one set by opt
+    opt_from_file['model_file'] = model_file
+    # update init model path to the one set by opt
+    # NOTE: this step is necessary when for example the 'init_model' is
+    # set by the Train Loop (as is the case when loading from checkpoint)
     if opt.get('init_model') is not None:
-        # update the init model path
         opt_from_file['init_model'] = opt['init_model']
 
     # update dict file path
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -325,6 +325,9 @@ def create_agent_from_opt_file(opt: Opt):
             opt_from_file[k] = v
 
     opt_from_file['model_file'] = model_file  # update model file path
+    if opt.get('init_model') is not None:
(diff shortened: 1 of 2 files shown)


parlai/core/agents.py
```diff
@@ -325,6 +325,9 @@ def create_agent_from_opt_file(opt: Opt):
             opt_from_file[k] = v
 
     opt_from_file['model_file'] = model_file  # update model file path
+    if opt.get('init_model') is not None:
```
klshuster, line 328: a couple quick questions:

1) why is this not handled in lines 309-315?
2) why is this not handled in lines 323-325?

stephenroller (commented):
Could we add a test that would catch this?

parlai/core/agents.py
emilydinan: 1. when train_model.py sets opt['init_model'] to the checkpoint, it does not add it into override. adding to override could be one possible solution, what do you think?
2. it isn't handled here in the specific case where init_model *did* exist in opt_from_file (i.e. (1) the PR description). ex: finetuning a model on BST that's initialized with Reddit and it gets requeued.

klshuster: I dont know which is better, but i'm very slightly leaning towards adding it to `override` in train model, it feels marginally better than carving out a special case for `init_model` in these `opt` functions. do you have an opinion either way?

emilydinan: I don't really know if I like either. I think adding to override is hacky and can't be trusted lol. But also carving out this special exception isn't very nice either... @stephenroller ?

stephenroller: Both seem equally gross to me. This solution, with comments explaining wtf is going on, and a test checking that we do it correctly, seems sufficient.

Should we kill override, then we'll at least catch doing it wrong in the future.

stephenroller (approved):
Add the comment discussing what's going on and why please

PR #3164, opened 2020-10-06
add a --version flag to the parlai command #3163

### LINKED ISSUE #3163: Add --version flag to parlai command
**Is your feature request related to a problem? Please describe.**
Make parlai "super command" conform the unix idiom of a --version flag.

Usecase: A user switching between virtual environments or reading about the release of a new feature should be able to easily identify which version they have on their path.

**Describe the solution you'd like**
```bash
$ parlai --version
parlai version X.X.X
# Optionally any relevant install info (not probably not any)
# Optionally copyright/license info (up to the maintainers whether you want this here or not)
```

**Describe alternatives you've considered**
```bash
$ python -c "import parlai; print(parlai.__version__)"
0.9.2
```
A little less natural

DNGros:
**Patch description**
Fixes #3163 

Some utilities (like GNU core utilities) also print copyright/license info. Other utilities (like git or firefox) only print program name and the version.

I wasn't sure which the maintainers would prefer, so just left it at program name and version info, but this could be updated.

**Testing steps**
manual test of:
```
$ parlai --version
ParlAI version 0.9.3
```
meets expected behavior.

It's a small change and I didn't bother running the test suit locally. Hopefully the CI will do that...

### DIFF
--- a/parlai/core/script.py
+++ b/parlai/core/script.py
@@ -169,7 +169,6 @@ class _SupercommandParser(ParlaiParser):
         self.print_help()
 
 
-
 class _SubcommandParser(ParlaiParser):
     """
     ParlaiParser which always sets add_parlai_args and add_model_args to False.
--- a/parlai/core/script.py
+++ b/parlai/core/script.py
@@ -210,7 +210,7 @@ def superscript_main(args=None):
         action='helpall',
         help=helpall_help_str,
     )
-    versioninfo_help_str = 'Prints version info and exit.srlai '
+    versioninfo_help_str = 'Prints version info and exit.'
     parser.add_argument(
         '--version',
         action='version',
--- a/parlai/core/script.py
+++ b/parlai/core/script.py
@@ -204,18 +204,16 @@ def superscript_main(args=None):
     parser = _SupercommandParser(
         False, False, formatter_class=_SuperscriptHelpFormatter
     )
-    helpall_help_str = 'List all commands, including advanced ones.'
     parser.add_argument(
         '--helpall',
         action='helpall',
-        help=helpall_help_str,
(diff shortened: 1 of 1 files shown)


DNGros:
@facebook-github-bot  I did that. (though right before submitting PR). You might not have noticed yet 😉
image

(I'll that listed email if doesn't autoresolve itself...)

parlai/core/script.py
```diff
@@ -203,10 +204,18 @@ def superscript_main(args=None):
     parser = _SupercommandParser(
         False, False, formatter_class=_SuperscriptHelpFormatter
     )
+    helpall_help_str = 'List all commands, including advanced ones.'
     parser.add_argument(
         '--helpall',
         action='helpall',
-        help='show all commands, including advanced ones.',
```
stephenroller, line 209: nit: I prefer the inline strings, if you don't mind.

stephenroller (approved):
Thanks for adding. If you could just `black` and then I'll merge!

parlai/core/script.py
DNGros: ok, I made the change as string was repeated in L225 but slightly different text.

DNGros: i'll change it back

DNGros: ok 99be94f85b makes it an inline string, but still edits to make the help string and description consistent

stephenroller: Great, thanks for the PR mate!

DNGros:
@stephenroller thanks for reviewing and for releasing the project. I had ran the autoformat.sh which should have run black. Let me know if any issues / merge blockers

PR #3165, opened 2020-10-07
Add "fb" option to autoformat script for `parlai_fb`

moyapchen:
'f' was already taken by flake, so I just made the single letter code `z`

NOTE: I also fixed a bug where running the autoformatter didn't seem to do anything locally by removing the ellipses after "master" in
```
    CHECK_FILES="$(git $REPO diff --name-only master... | grep '\.py$' | reroot $ROOT | onlyexists | tr '\n' ' ')"
```

I have no idea if there are side effects on CircleCI or whatnot from doing this.

**Testing steps**
Add a bunch of empty new lines to the start of files in `parlai`, `parlai_internal`, and `parlai_fb`. Verify that autoformat appropriately removes these lines while running autoformat with no flag, `-i`, and `-z`, respectively.

### DIFF
--- a/autoformat.sh
+++ b/autoformat.sh
@@ -28,6 +28,7 @@ Optional Arguments
   -f, --flake8       only run flake8.
   -h, --help         print this help message and exit
   -i, --internal     run within parlai_internal
+  -z, --fb     run within parlai_fb
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
Cheers

PR #3167, opened 2020-10-07
Make tests/datatests/test_new_tasks.py work

moyapchen:
Values were there, just the format changed slightly.

Tested/debugged by changing
  `parlai/tasks/jsonfile/agents.py`
to load in the example json file colocated in the same directly.

Printed the results of `verify(opt)` as well as the `KEYS` object; that made it clear-ish what ought to be happening.

**Patch description**

**Testing steps**

**Logs**

```
```

**Other information**


stephenroller (approved):
Thank you!

PR #3168, opened 2020-10-07
Fix convai2:normalized task's candidate list

shuuki4:
**Patch description**
- Fix the list of candidates generated by `convai2:normalized` task.
  - AS-IS code returns normalized labels rather than candidates, which was ruining the evaluation of retrieval based models on this task. (hits@1 always hits 1.0)


stephenroller (approved):
Good catch, thank you.

PR #3177, opened 2020-10-09
R3f Implementation

moyapchen:
Update: Implementation here works now.

Tested with both BART (`--model bart`) + GPT2 (`--model hugging_face/gpt2`) and all combinations of encoder/decoder on. Verified that the correct number of encoder/decoder noise adds actually occurred per iteration. 

~~Don't fully know what I'm doing yet; pushing up code for feedback purposes.~~

~~I *think* doing the code this way is actually sort of irrelevant, since the noise needs to be added to the input embeddings, not to just the generator.~~


parlai/agents/transformer/module_extentions.py
```diff
@@ -0,0 +1,87 @@
...
+                "sum",
+            )
+        ) / noised_logits.size(0)
+
+
+class R3FNoiseEncoderTrait(object):
+    """
+    Implements "Better Fine-Tuning by Reducing Representational Collapse"
+ 
+    Paper: https://arxiv.org/abs/2008.03156
```
stephenroller, line 66: Add a full citation please. 

`Aghajanyan et al. (2020). https://arxiv.org/ans/2008.03156`

moyapchen:
I'm not 100% sure that there isn't some detail I haven't fudged since I haven't run a full training sweep on this yet, but it's in good enough of a state to be worth getting some comments. 

Notably, I am interested in getting some comments on class organization + use of python language idioms, since I haven't done too much with more advanced python functionality. I ended up trying but deciding against using Mixins — I kept getting things like 
```
torch.nn.modules.module.ModuleAttributeError: 'TransformerEncoder' object has no attribute '_modules'
```
and eventually gave up. 

Basic neural network/implementation correctness feedback also greatly appreciated. I tried to line things up well with https://github.com/pytorch/fairseq/pull/2455 + D22942118 + D23444863, but can fully believe if some detail got overlooked. 

(Doing only integration with BART for now; this should be easy enough to drop into `TransformerGeneratorAgent` as well.)

stephenroller:
After reading this, I feel like that we could implement this with forward hooks and a context manager extremely cleanly.

moyapchen:
Getting some OOMs while running longer trains -- putting this back as a draft. 

(Some low hanging fruit here, so will do that first.)

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+                "Trying to use an R3F loss when this has not been allowed."
+            )
+        regular_output = model(*args, **kwargs)
+        with R3FNoiseEmbeddingContextManager(self, model) as model:
+            noised_output = model(*args, **kwargs)
+        symm_kl = self._calculate_symm_kl(noised_output[0], regular_output[0])
+        return regular_output, self.r3f_lambda * symm_kl
+
+
+class R3FNoisedEmbedding(nn.Module):
```
stephenroller, line 129: I don't think this is a module, is it?

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+        self.encoder_hook = None
+        self.decoder_hook = None
+        self.context = context
+        for name, layer in module.named_modules():
+            if context.noise_encoder and name == "encoder":
+                self._find_embedding_and_set_hook(layer, "encoder")
+            if context.noise_decoder and name == "decoder":
+                self._find_embedding_and_set_hook(layer, "decoder")
+        self.module = module
+        self.context.had_first_pass = True
```
stephenroller, line 159: tricky

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+    def __init__(self, noise_sampler):
+        super(R3FNoisedEmbedding, self).__init__()
+        self.noise_sampler = noise_sampler
+
+    def __call__(self, module, input, output):
+        noise = self.noise_sampler.sample(sample_shape=output.shape).to(output)
+        return output + noise
+
+
+class R3FNoiseEmbeddingContextManager(AbstractContextManager):
```
stephenroller, line 139: Dope, I think we can merge all 3 of these classes into one now.

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    This class finds the first embedding(s) under the encoder/decoder, depending on what
+    is selected, and adds R3F noise to their output.
+
+    Note that this does not discrinate between the dictionary embedding lookup +
+    position embedding classes. However, seeing as how addition is commutative, this
```
stephenroller, line 145: I think it'd be better if we did lock in on word embeddings.

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+            setattr(
+                parent, embedding_name, copy.deepcopy(getattr(parent, embedding_name))
+            )
+        embedding = getattr(parent, embedding_name)
+        val = R3FNoisedEmbedding(self.context.noise_sampler)
+        setattr(self, name + "_hook", embedding.register_forward_hook(val))
+
+    def _find_embedding(self, parent):
+        for child_name, child in parent.named_children():
+            if child.__class__.__name__ == "Embedding":
```
stephenroller, line 179: isinstance(child, nn.Embedding)

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+    Note that this does not discrinate between the dictionary embedding lookup +
+    position embedding classes. However, seeing as how addition is commutative, this
+    should not be an issue.
+    """
+
+    def __init__(self, context, module):
+        self.encoder_hook = None
+        self.decoder_hook = None
+        self.context = context
+        for name, layer in module.named_modules():
```
stephenroller, line 153: At this point we have two searches into named modules (one is named_children and one is this one). I think we can simplify and use the full name to match the pattern.

parlai/core/torch_generator_agent.py
```diff
@@ -705,7 +708,13 @@ def compute_loss(self, batch, return_output=False):
         """
         if batch.label_vec is None:
             raise ValueError('Cannot compute loss without a label.')
-        model_output = self.model(*self._model_input(batch), ys=batch.label_vec)
+        r3f_adjustment = None
```
stephenroller, line 711: Now that we've got the context manager, I think we can implement this as a mixin pretty well, approximately:

```python
class R3FMixin:
    def compute_loss(self, batch, return_output)
          loss, standard_output = super().compute_loss(batch, return_output=True)
          standard_scores, *_ = standard_output
          with R3FNoiseEmbeddingContextManager(self.model) as r3f:
              perturbed_output = self.model(*self._model_input(batch), ys=batch.label_vec)
              perturbed_scores, *_ = perturbed_output
              r3floss = r3f.compute_sym_loss(standard_scores, perturbed_scores)
              loss += r3floss[notnull].mean()
         if return_output:
              return loss, standard_output
         else:
              return loss

class R3FTransformerGeneratorAgent(R3FMixin, TransformerGeneratorAgent):
      pass
```

I've got in my mind a possible way that we could have losses be more generically implemented (basically migrate us to how fairseq/classyvision do it).

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+                f"R3F noise: Could not find an embedding as a child of {name}. Does the model you are training use this type of layer? (Default R3F settings noises the encoder only, which some models - ex. GPT2 - do not use.)"
+            )
+        if not self.context.had_first_pass:
+            # need to do a deep copy because models oftentimes share embedding object for encoder + decoder; breaks noise encoder/decoder only experiments
+            setattr(
+                parent, embedding_name, copy.deepcopy(getattr(parent, embedding_name))
+            )
+        embedding = getattr(parent, embedding_name)
+        val = R3FNoisedEmbedding(self.context.noise_sampler)
+        setattr(self, name + "_hook", embedding.register_forward_hook(val))
```
stephenroller, line 175: Do we need to store this as a variably-named attribute, rather than just accumulating into a list like `self._hooks` or similar?

parlai/agents/transformer/module_extensions.py
stephenroller: (I could see future models may not treat positional embeddings as additive)

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+                self._find_embedding_and_set_hook(layer, "decoder")
+        self.module = module
+        self.context.had_first_pass = True
+
+    def _find_embedding_and_set_hook(self, layer, name):
+        try:
+            parent, embedding_name = self._find_embedding(layer)
+        except TypeError:
+            raise RuntimeError(
+                f"R3F noise: Could not find an embedding as a child of {name}. Does the model you are training use this type of layer? (Default R3F settings noises the encoder only, which some models - ex. GPT2 - do not use.)"
```
stephenroller, line 166: nit: wraps please lol

parlai/agents/transformer/module_extensions.py
```diff
@@ -0,0 +1,195 @@
...
+        self.noise_encoder = opts.get("r3f_encoder_noise")
+        self.noise_decoder = opts.get("r3f_decoder_noise")
+
+        if self.use_r3f and not self.noise_encoder and not self.noise_decoder:
+            raise RuntimeError(
+                "R3FContext: Noise must be added to at least one of the encoder or decoder."
+            )
+        self.had_first_pass = (
+            False
+        )  # necessary because nn.Embedding for encoder/decoder on some models are the same. Use this to only make any relevant deep copies once.
```
stephenroller, line 93: I'm hoping we can reach a conclusion on which mode is best in your experiments (encoder only, decoder only, both), and then make an opinionated implementation which does only that.

parlai/agents/transformer/module_extensions.py
moyapchen: I think these 3 still need to be separate. Namely, 

1. `R3FNoisedEmbedding` implements `__call__` which is necessary for the forward hook that needs to be separate from the ContextManager to do it (Unless there's some lambda magic that I haven't figured out the incantation for that can be used there)
2. The `AbstractContextManager` needs to be separate from whatever has `forward_pass_with_r3f` (unless having an object use itself in a `with` works?)
3. 

I guess one thing that might make this more clear is if I underscored `R3FNoisedEmbedding` + `R3FNoiseEmbeddingContextManager` and moved them as inner classes to `R3FNoiseContext`? (And change the naming, cause it's kind of confusing as is.)

parlai/agents/transformer/module_extensions.py
moyapchen: Ah yeah, nice catch. Will remove. 

It's a remnant from before I figured out the "deep copy to get rid of shared embedding between encoder/decoder".

parlai/agents/transformer/module_extensions.py
moyapchen: 👍

parlai/agents/transformer/module_extensions.py
moyapchen: Yeah, I realized today that just using a map or something is probably way nicer than this `setattr` mess I was doing last night.

parlai/agents/transformer/module_extensions.py
moyapchen: yeah, idk why the linter didn't format that

parlai/agents/transformer/module_extensions.py
moyapchen: Makes sense. 

This would still be necessary in the encoder only case (unless doing a deep copy of an embedding each train loop isn't too bad?)

parlai/agents/transformer/module_extensions.py
moyapchen: Nvm, just saw your other comment about mixins.

parlai/agents/transformer/module_extensions.py
moyapchen: Makes sense.

parlai/agents/transformer/module_extensions.py
moyapchen: D'oh, yeah, that would make it easier.

parlai/agents/transformer/module_extensions.py
moyapchen: Tricky thing is I'm not sure if it's possible to do that unless it's called out by name, and different models use different names.

For example, adding a 

```
        for name, layer in module.named_modules():
            print(name, layer.__class__.__name__)
```

gets 

```
encoder.embeddings Embedding
encoder.position_embeddings Embedding
```

on BART and

```
decoder.transformer.wte Embedding
decoder.transformer.wpe Embedding
```

on GPT2. 

(Unless there's some Pytorch magic here that'll do this that I don't know about.)

parlai/agents/transformer/module_extensions.py
stephenroller: I think a deep copy is very expensive

If we end up going encoder-only, the hook can set the first pass to True to turn any future calls to the hook into the identity function

parlai/agents/transformer/module_extensions.py
moyapchen: I did consider at one point adding a `set_word_embedding_name` function which might be a reasonable workaround?

parlai/agents/transformer/module_extensions.py
stephenroller: The hook can be a member of the context manager like this. `self.` does this automatic currying in python.

```python
class MyContextManger(object):
    def __init__(self, module):
         self.module = module
    def _hook_implementation(self, input, output):
        return output + noise
    def __enter__(self):
        self.module.register_hook(self._hook_implementation)
  ```

File: parlai/agents/transformer/module_extensions.py (hunk 11 above)
Comment (moyapchen) [reply parlai/agents/transformer/module_extensions.py:93]: Yeah, makes sense as a workaround.

File: parlai/agents/transformer/module_extensions.py (hunk 4 above)
Comment (stephenroller) [reply parlai/agents/transformer/module_extensions.py:139]: (For this to combine with the other discussion thread, I think `__enter__` will ultimately have to `return self` to get that access to context manager object)

File: parlai/core/torch_generator_agent.py (hunk 8 above)
Comment (moyapchen) [reply parlai/core/torch_generator_agent.py:711]: Thinking about it a little more, I'm not actually quite sure this works in the context of BART.

Specifically, BART overrides `compute_loss` with its own loss function to remove the start token. Implementing R3F as a Mixin makes the `super()` line in the code above sort of wonky.

Comment (moyapchen) [reply parlai/core/torch_generator_agent.py:711]: (In general, I do think "composition over inheritance" might be a decent pattern to go for here, but we can chat offline in more depth. I'm especially curious about how fairseq/classyvision currently has it done, if you have code pointers.)

File: parlai/agents/transformer/module_extensions.py (hunk 5 above)
Comment (stephenroller) [reply parlai/agents/transformer/module_extensions.py:145]: Given that there's only two, I just suggest we do if endswith(".embedding") or endswith(.wte)

File: parlai/core/torch_generator_agent.py (hunk 8 above)
Comment (stephenroller) [reply parlai/core/torch_generator_agent.py:711]: Dope, would love to talk about how we can make this a composition.

Comment (moyapchen) [reply parlai/core/torch_generator_agent.py:711]: Update: Chatted offline with @stephenroller about what the inheritance tree here would look like. 

Doing 

```
class BartAgent_R3F(R3FMixin, BartAgent):
    pass
```

so that's the final class works.

moyapchen:
Redone at https://github.com/facebookresearch/ParlAI/pull/3200

moyapchen:
Closing/abandoning in favor of https://github.com/facebookresearch/ParlAI/pull/3204

PR #3188, opened 2020-10-12
Tweaks to static turn annotation task

EricMichaelSmith:
**Patch description**
Tweaks to the static turn annotations task to fix issues when placing the annotation buckets correctly. The main change I've made is to default to annotating all bot utterances if annotation indices are not present, as Mary suggested

**Testing steps**
Check out the `turn-annotations-testing` branch and run the following command: `python parlai/crowdsourcing/tasks/turn_annotations_static/run_task.py --launch-config parlai.crowdsourcing.tasks.turn_annotations_static.launch_config`
- If you run this command as-is, it'll test the code without an `annotation_indices_jsonl` file. Flip the two commented-out lines, one in `launch_config.py` and one in `run_task.py`, to test this *with* a sample `annotation_indices_jsonl` file.


EricMichaelSmith:
I've already incorporated these changes into my PR to port this task to Hydra at https://github.com/facebookresearch/ParlAI/pull/3149, so I'm closing this and focusing on that PR

PR #3189, opened 2020-10-12
Make the log rate an adjustable parameter

EricMichaelSmith:
**Patch description**
Allow the log rate to be adjusted.

**Testing steps**
Tested by adding a command-line argument (`log_rate=60`) as well as a line in `example.yaml` (`log_rate: 10`)


JackUrb (approved):
Made an internal comment about the naming of this (`log_rate` is pretty vague), but otherwise this is a nice fix.

EricMichaelSmith:
> Made an internal comment about the naming of this (`log_rate` is pretty vague), but otherwise this is a nice fix.

Yes - changed this to `monitoring_log_rate` for more specificity (it should be assumed that it's for a crowdsourcing task because that's all that we use Hydra for currently)

PR #3193, opened 2020-10-13
[gpt2] Fix GPT-2 generation when using batchsize >1

stephenroller:
**Patch description**
@chinnadhurai recently observed that when using batchsize > 1, the generations of GPT2 don't match those when using batchsize = 1, resulting in a significant loss of performance.

Upon inspection, the bug was around how positional embeddings were being treated with padding. HF's GPT-2 implementation defaults to using `arange`, rather than counting non-padding tokens, which caused incongruences in the positional embeddings used for generations with unequal number of context tokens. This was particularly cumbersome after the first generation, and exacerbated by our beam search's speedup which only computes output softmax on the final timestep.

Fortunately, HF's implementation allows for manually specifying both the padding and position embeddings as input. This diff reworks our wrapper around HF's code so that these things are computed on our side. The result is actually a cleaner implementation, in my humble opinion.

**Testing steps**
Trained a model for one epoch on [redacted], then tested with:
- bs 1
- bs 2, 4, 8, 16
- dynamic batching bs 8
- multiprocessing eval with bs 8

Final results were _slightly different_ (fourth decimal place), but in the final ballpark.

Also added new CI for GPT2 which tests generation with different batchsizes.

### DIFF
--- a/parlai/agents/hugging_face/gpt2.py
+++ b/parlai/agents/hugging_face/gpt2.py
@@ -117,7 +117,7 @@ class GPT2Decoder(torch.nn.Module):
 
         if incr_state is None:
             # pull out only the hidden states for the label tokens
-            output = hidden_states[:, -(input.size(1) + 1) :]
+            output = hidden_states[:, -input.size(1) - 1 + int(self.add_start_token) :]
             # hack: we need the last state of the encoder-side to be the first
             # element of the decoder-side
             lengths = (input != self.NULL_IDX).sum(dim=-1)
--- a/parlai/agents/hugging_face/gpt2.py
+++ b/parlai/agents/hugging_face/gpt2.py
@@ -253,6 +253,15 @@ class Gpt2Agent(TorchGeneratorAgent):
 
     def __init__(self, opt, shared=None):
         if not opt["add_special_tokens"] and opt.get('batchsize', 1) > 1:
+            # *** STOP ***
+            # You may be a future researcher who has stumbled upon this odd
+            # restriction, and is tempted to comment this out. After all, the
+            # code still runs when it's uncommented, why shouldn't you?
+            # You should know this has serious implications, as gpt2 doesn't have
+            # padding tokens. This is incompatible with ParlAI's batching,
+            # which puts conversations of different length in the same
+            # batch. Without a padding token, nonsense will be inserted into
+            # the context, and the generations & PPL you get will be wrong.
             raise RuntimeError(
                 "If using batchsize > 1, --add-special-tokens must be True."
             )
--- a/parlai/agents/hugging_face/gpt2.py
+++ b/parlai/agents/hugging_face/gpt2.py
@@ -65,7 +65,9 @@ class GPT2Decoder(torch.nn.Module):
         return GPT2Model.from_pretrained(fle_key)
 
     def forward(self, input, encoder_state, incr_state=None):
+        # __import__("ipdb").set_trace()  # FIXME
         attention_mask = None
+        position_ids = None
         if incr_state is None:
             # first step
             if (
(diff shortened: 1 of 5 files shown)


stephenroller:
I've also slipped in some slightly better initialization to the special tokens. There's little improvement after as few as 200 steps, but it keeps the first couple prints from being out of control.

parlai/agents/hugging_face/gpt2.py
```diff
@@ -35,12 +34,21 @@ def __init__(self, opt, dict):
...
-            self.add_start_token = opt["add_start_token"]
+            with torch.no_grad():
+                # first reduce the random jitter of the initialization
+                self.transformer.wte.weight[size_before:] *= 0.1
+                # next center it on the endoftext token
+                self.transformer.wte.weight[
+                    size_before:
+                ] += self.transformer.wte.weight[size_before - 1].unsqueeze(0)
+
+        self.add_start_token = opt["add_start_token"]
```
emilydinan, line 48: should this go above? will the transformer need to be resized depending on this value?

parlai/agents/hugging_face/gpt2.py
```diff
@@ -253,11 +252,27 @@ def add_cmdline_args(cls, argparser):
         return agent
 
     def __init__(self, opt, shared=None):
-        if not opt["add_special_tokens"] and opt["batchsize"] > 1:
+        if not opt["add_special_tokens"] and opt.get('batchsize', 1) > 1:
+            # *** STOP ***
```
emilydinan, line 256: lmao

tests/nightly/gpu/test_gpt2.py
```diff
@@ -0,0 +1,97 @@
...
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import unittest
+from parlai.core.agents import create_agent
+
+
+class TestGpt2(unittest.TestCase):
```
emilydinan, line 11: nice

parlai/agents/hugging_face/gpt2.py
```diff
@@ -66,33 +74,62 @@ def _init_from_pretrained(self, opt):
...
             else:
                 # forced decoding: concatenate the context
                 # with the labels
-                model_input, _ = concat_without_padding(
-                    encoder_state, input, use_cuda=self.use_cuda, null_idx=self.null_idx
-                )
-                attention_mask = model_input != self.null_idx
+                model_input = torch.cat([encoder_state, input], dim=-1)
+            attention_mask = model_input != self.NULL_IDX
+            position_ids = (
```
emilydinan, line 92: nice

emilydinan (approved):
woof this was a doozy but lgtm

parlai/agents/hugging_face/gpt2.py
stephenroller: Add start token just says whether we use the start token or not, not whether we add it.

parlai/agents/hugging_face/gpt2.py
```diff
@@ -35,12 +34,21 @@ def __init__(self, opt, dict):
...
         self.transformer = self._init_from_pretrained(opt)
         # add special tokens
-        self.start_idx = dict.start_idx
-        self.null_idx = dict.null_idx
-        self.add_start_token = False
         if opt["add_special_tokens"]:
+            size_before = self.transformer.wte.weight.size(0)
             self.transformer.resize_token_embeddings(len(dict.tokenizer))
-            self.add_start_token = opt["add_start_token"]
+            with torch.no_grad():
```
chinnadhurai, line 40: Wondering as to why we need this ?

chinnadhurai (approved):

parlai/agents/hugging_face/gpt2.py
stephenroller: The torch nograd or the initialization changes?

PR #3194, opened 2020-10-13
[Poly] Allow Polyencoder to resize token embs

klshuster:
**Patch description**
Similar to how we support this for the `transformer/generator`, I've enabled the ability to add special tokens to the poly-encoder dictionary via adding the function to resize token embeddings.

1) Mainly copied `TransformerGenerator._resize_embeddings` and fit to polyencoder setup.
2) Shamelessly used an existing test for CI

Note that it might look a bit like copy and paste, and there probably is a way to make this slightly more readable, however it's a bit model-dependent (i.e. which embeddings specifically are resized, the model attrs/state dict keys). If we find ourselves adding this to several more models it'll probably be better to standardize the function in `TorchAgent`

**Testing steps**
Local testing as well as CI.

**Logs**

```
$  pytest -k TestPolyencoder
======= test session starts =======
platform linux -- Python 3.6.10, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
rootdir: /private/home/kshuster/ParlAI, configfile: pytest.ini, testpaths: tests
plugins: requests-mock-1.8.0
collected 414 items / 413 deselected / 1 selected

tests/test_transformers.py .                                                                                                                                                   [100%]

======= slowest 10 durations =======
11.41s call     tests/test_transformers.py::TestPolyencoder::test_resize_embeddings

(2 durations < 0.005s hidden.  Use -vv to show these durations.)
======= 1 passed, 413 deselected, 15 warnings in 16.24s =======
```

### DIFF
--- a/parlai/agents/transformer/polyencoder.py
+++ b/parlai/agents/transformer/polyencoder.py
@@ -14,6 +14,9 @@ import torch
 
 from parlai.core.opt import Opt
 from parlai.core.torch_ranker_agent import TorchRankerAgent
+from parlai.utils.misc import recursive_getattr
+from parlai.utils.logging import logging
+
 from .biencoder import AddLabelFixedCandsTRA
 from .modules import (
     BasicAttention,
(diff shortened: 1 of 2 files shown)


parlai/agents/transformer/polyencoder.py
```diff
@@ -236,6 +239,35 @@ def load_state_dict(self, state_dict):
             state_dict['codes'] = self.model.codes
         super().load_state_dict(state_dict)
 
+    def _resize_token_embeddings(self, state_dict, msg=None):
+        """
+        Resize the token embeddings when adding extra special tokens.
+
+        H/t TransformerGenerator._resize_token_embeddings for inspiration.
```
stephenroller, line 246: this is subtly different from the TransGA one innit? 😂

tests/test_transformers.py
```diff
@@ -560,48 +609,7 @@ def test_temperature(self):
 
     @pytest.mark.nofbcode
     def test_resize_embeddings(self):
```
stephenroller, line 611: tbh I think i would rather have the separate tests i think, and not try to DRY this...

stephenroller (approved):

emilydinan (approved):
nice...it will be awesome when we eventually ahve this with all of our models and dictionaries...

tests/test_transformers.py
klshuster: not sure what you mean - are you supporting current setup? or would prefer more specialized testing for each model

stephenroller: I'd prefer specialized tests and not doing this refactor here. But I yield to you.

klshuster: i think this test right now is adequate for the functionality being tested - i.e., that an agent is able to handle additional special tokens to the dictionary. As alluded to in the PR description, this ideally would be a top level `TorchAgent` function, which would indeed only require a set of tests for TA specifically - including this test. 

So, for now, i think this CI is appropriate. but happy to add more testing once this becomes top level functionality

PR #3202, opened 2020-10-15
adding task multiwoz_trade (multiwoz2.1 with TRADE preprocess), including both dialog state tracking and dialog response generation

qbetterk:
**Patch description**
adding a new task for multiwoz2.1 dataset with TRADE preprocess, including two teacher model, one for dialog state tracking and one for dialog response generation (seq-to-seq without dst nor db search, similar with multiwoz_v2.1); besides, reformat data as each turn would be an individual episode, and the input sequence includes all dialog history in previous turns. 

**Testing steps**
Before training model, please use 
`parlai display_data -t multiwoz_trade`
to generate and preprocess data.
Then, for DST task:
`parlai multiprocessing_train \
    -m hugging_face/gpt2 -t multiwoz_trade -eps 20 -bs 8 -opt adam -lr 5e-4 \
    --eval-batchsize 8 \
    --fp16 true \
    --warmup_updates 100 \
    --warmup_rate 1e-5 \
    --log-every-n-secs 100 \
    --validation-every-n-epochs 1 \
    --model-file experiment/gen_gpt2/model \
    --validation-metric 'joint goal acc' \
    --validation-metric-mode max \
    --add-special-tokens True`
For dialog response generation task:
`parlai multiprocessing_train \
    -m hugging_face/gpt2 -t multiwoz_trade -eps 20 -bs 8 -opt adam -lr 5e-4 \
    --eval-batchsize 8 \
    --fp16 true \
    --warmup_updates 100 \
    --warmup_rate 1e-5 \
    --log-every-n-secs 100 \
    --validation-every-n-epochs 1 \
    --model-file experiment/gen_gpt2/model \
    --add-special-tokens True`

### DIFF
--- a/parlai/tasks/multiwoz_trade/__init__.py
+++ b/parlai/tasks/multiwoz_trade/__init__.py
@@ -1,0 +1,5 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
--- a/parlai/tasks/multiwoz_trade/agents.py
+++ b/parlai/tasks/multiwoz_trade/agents.py
@@ -10,85 +10,103 @@ from parlai.core.metrics import AverageMetric
 from parlai.utils.io import PathManager
 from .build import build
 from .utils.trade_proc import trade_process
-from .utils.reformat import reformat_dst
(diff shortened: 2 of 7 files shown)


stephenroller:
Thanks for this. Let's move this to internal repos (parlai-fb) so we can iterate more before committing to public interfaces.

PR #3204, opened 2020-10-17
Implement R3F, with token level + options for different combinations of encoder/decoder noise.

moyapchen:
Title.

Ran locally with debug prints. Also added in a test system cause I didn't fully trust the encoder/decoder logic, and figured I'd rather have it be well tested now than curse things later. (It's a little bit bulky for my tastes — cause of some of the implicit extras that `forward` and `compute_batch` require — but it does seem to work.)


parlai/nn/r3f.py
```diff
@@ -0,0 +1,192 @@
...
+            # add in loss
+            loss += self.r3f_lambda * r3f_loss
+        if return_output:
+            return (loss, standard_output)
+        else:
+            return loss
+
+    def _calculate_symm_kl(self, noised_logits, input_logits):
+        return F.kl_div(
+            F.log_softmax(noised_logits, dim=-1, dtype=torch.float32),
```
stephenroller, line 109: shoot, you're already doing it in fp32?!

parlai/agents/bart/bart.py
```diff
@@ -237,3 +237,12 @@ def _construct_token_losses(self, labels, model_output):
                 )
             )
         return token_losses
+
+
+class BartAgent_R3F(R3FMixin, BartAgent):
```
stephenroller, line 242: Hm I think if this is `R3fAgent` then we can use `-m bart/r3f`

parlai/nn/r3f.py
```diff
@@ -0,0 +1,190 @@
...
+R3F_NOISE_NORMAL = "normal"
+R3F_NOISE_UNIFORM = "uniform"
+
+
+class R3FMixin(object):
+    """
+    Class that helps implement "Better Fine-Tuning by Reducing Representational
+    Collapse", Aghajanyan et al.
+
+    (2020). https://arxiv.org/ans/2008.03156.
```
stephenroller, line 29: Formatting

parlai/nn/r3f.py
```diff
@@ -0,0 +1,190 @@
...
+        # Find embedding values and store locally so we don't have to find them each turn
+        self.r3f_embeddings = {}
+        self.is_generative_model = True  # assume true until we find otherwise
+        self._find_embeddings(self.model)
+
+    def compute_loss(self, batch, return_output=False):
+        if not hasattr(self, 'r3f_lambda'):
+            self.set_r3f_settings_from_opts(self.opt)
+        loss, standard_output = super().compute_loss(batch, True)
+        with R3FNoiseEmbeddingContextManager(self, self.model) as r3f:
```
stephenroller, line 88: love this final verison

parlai/nn/r3f.py
```diff
@@ -0,0 +1,190 @@
...
+            noised_result = self.model(*self._model_input(batch), ys=batch.label_vec)
+            noised_scores, _, *_ = noised_result
+            standard_scores, _, *_ = standard_output
+            r3f_loss = r3f._calculate_symm_kl(noised_scores, standard_scores)
```
stephenroller, line 92: this looks like it should be self._calculate_symm_kl

But I actually think r3f.calculate_symm_kl will be more convenient if/when we need to apply this to a non-generative model.

parlai/nn/r3f.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+        group.add_argument(
+            '--r3f-decoder-noise',
+            type=bool,
+            default=False,
+            help='Add noise to decoder.',
+        )
+        return argparser
+
+    def set_r3f_settings_from_opts(self, opts):
```
stephenroller, line 63: make private?

parlai/nn/r3f.py
```diff
@@ -0,0 +1,190 @@
...
+                re.search(f"decoder.*norm_embeddings$", name)
+                or re.search("decoder.*wte$", name)  # cause GPT2 does it differently
+            ):
+                self.r3f_embeddings["decoder"] = layer
+        if (self.noise_encoder is True and "encoder" not in self.r3f_embeddings) or (
+            self.noise_decoder is True and "decoder" not in self.r3f_embeddings
+        ):
+            raise RuntimeError(
+                """
+                R3F: Embedding to add noise to not found in model. (Does the part, encoder/decoder, exist in the model you are running? Is the name for the embedding accounted for in `_find_embeddings()`?)
```
stephenroller, line 136: wrap please.

stephenroller:
```python
(
   "This is a wrapped "
   "single line string "
   "in python."
)
```

parlai/nn/r3f.py
```diff
@@ -0,0 +1,190 @@
...
+    def __enter__(self):
+        return self.context
+
+    def __exit__(self, type, value, traceback):
+        for key in self.hooks.keys():
+            if self.hooks[key]:
+                self.hooks[key].remove()
+            self.hooks[key] = None
+
+    def _noise_embedding_this_pass(self):
```
stephenroller, line 173: Comment for poor future generation trying to figure out why you did this.

tests/test_r3f.py
```diff
@@ -0,0 +1,161 @@
...
+        self.decoder = _DummyModuleWithEmbedding(
+            opt.get('r3f_decoder_noise'), 'decoder', DECODER_ONLY
+        )
+
+    def forward(self, *args, **kwargs):
+        self.decoder()
+        return DUMMY_MODEL_RETURN
+
+
+######################### Following are wrapper abstractions to work with the fact that R3F is tightly coupled with how models in Parlai are set up.
```
stephenroller, line 89: :'(

tests/test_r3f.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class TestR3FEmbeddingHook(unittest.TestCase):
+    """
+    Test that R3F embedding hook combinations work appropriately.
+    """
+
+    def _get_base_opt(self):
+        opt = {
+            'r3f_lambda': 1000000000,  # High enough to overwhelm 1.0
```
stephenroller, line 133: sorry I don't understand

parlai/nn/r3f.py
stephenroller: Like not sure how tests pass without this

stephenroller (commented):
One concerning comment, rest is nonblocking.

parlai/nn/r3f.py
moyapchen: I don't get why the autoformatter doesn't just... do this. 

Maybe I'll take a look at it at some point.

tests/test_r3f.py
moyapchen: Brain fart, lol.

tests/test_r3f.py
moyapchen: Yeah, couldn't figure out how to do it without mocking everything in between.

parlai/nn/r3f.py
```diff
@@ -0,0 +1,190 @@
...
+            None,
+            "sum",
+        )
+
+    def _find_embeddings(self, module):
+        self.is_generative_model ^= not re.search(".*GPT.*", module.__class__.__name__)
+
+        for name, layer in module.named_modules():
+            self.is_generative_model ^= not re.search(".*GPT.*", name)
+            if self.noise_encoder and (re.search(f"encoder.*norm_embeddings$", name)):
```
stephenroller, line 124: oh this isn't right, i don't think it's norm embeddings. just `\.embeddings$`

parlai/nn/r3f.py
```diff
@@ -0,0 +1,190 @@
...
+            F.softmax(input_logits, dim=-1, dtype=torch.float32),
+            None,
+            None,
+            "sum",
+        ) + F.kl_div(
+            F.log_softmax(input_logits, dim=-1, dtype=torch.float32),
+            F.softmax(noised_logits, dim=-1, dtype=torch.float32),
+            None,
+            None,
+            "sum",
```
stephenroller, line 116: I think to implement per-token correctly, we would need to use reduction none, and then mask out the padding, and then do .sum() / num_tokens

parlai/agents/bart/bart.py
moyapchen: `bart/bart/r3f` is apparently what works, but yeah, changing.

stephenroller: I think we just gotta move it to bart/r3f.py

stephenroller: I hate the import logic in parlai...

tests/test_r3f.py
stephenroller: Given that you don't do a test of correctness, we could just do an end to end test? Train for 5 sgd steps...

parlai/nn/r3f.py
moyapchen: To be fair, this looks pretty much the same as what was initially planned out?

tests/test_r3f.py
moyapchen: Not sure if we can check for adding noise to encoder/decoder that way.

Actually, I suppose I might be able to string something together via capturing log output, hmm....

parlai/nn/r3f.py
stephenroller: 👀

tests/test_r3f.py
stephenroller: We do have a capture_stdout and with.assertLogs but I hate using them. Your mock is good.

moyapchen:
Going to go for internal implementation first

PR #3205, opened 2020-10-20
Unify display-verbose and verbose

jxmsML:
**Patch description**
Unify --display-verbose and --verbose

**Testing steps**

**Logs**

```
```

**Other information**

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -738,6 +738,9 @@ class ParlaiParser(argparse.ArgumentParser):
             choices={None, 'full', 'batchsort'},
             help='Use dynamic batching',
         )
+        parlai.add_argument(
+            '--verbose', dest='verbose', action='store_true', help='Print all messages'
+        )
         self.add_parlai_data_path(parlai)
 
     def add_distributed_training_args(self):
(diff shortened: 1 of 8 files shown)


stephenroller:
Discussed in group meeting

parlai/core/params.py
```diff
@@ -738,6 +738,19 @@ def add_parlai_args(self, args=None):
...
             help='Use dynamic batching',
         )
+        parlai.add_argument(
+            '-v',
+            '--verbose',
+            dest='verbose',
+            action='store_true',
+            help='Print all messages',
+        )
+        parlai.add_argument(
```
stephenroller, line 748: This should actually be just in display_model/interactive, not in generic params.

parlai/core/params.py
```diff
@@ -738,6 +738,19 @@ def add_parlai_args(self, args=None):
             choices={None, 'full', 'batchsort'},
             help='Use dynamic batching',
         )
+        parlai.add_argument(
+            '-v',
+            '--verbose',
+            dest='verbose',
+            action='store_true',
```
stephenroller, line 745: use `type="bool"` instead of `action="store_true"`. The problem with store true is that once we load it from an opt file, it's hard to unset.

stephenroller (approved):
Looks great to me, thanks!

parlai/core/params.py
jxmsML: Thanks. will edit that.

PR #3207, opened 2020-10-20
Fix error with using GPT2 + DistributedDataParallel

moyapchen:
**Patch description**

DistributedDataParallel adds another name to the module path. Fix this with an if check. 

**Testing steps**

Tested by running GPT2 training with DistributedDataParallel.
Also
```
pytest test_gpt2.py::TestDistributed::test_multitask_distributed
```
both with and without the change; verifies it breaks in latter and passes in the former.

### DIFF
--- a/tests/nightly/gpu/test_gpt2.py
+++ b/tests/nightly/gpu/test_gpt2.py
@@ -161,7 +161,7 @@ class TestDistributed(unittest.TestCase):
     def test_distributed(self):
         config = copy.deepcopy(self._base_config)
         config['num_epochs'] = 50
-        config['task'] = 'integration_tests:overfit'
+        config['task'] = 'integration_tests:short_fixed'
         config['dynb'] = 'full'
         valid, test = self._distributed_train_model(config)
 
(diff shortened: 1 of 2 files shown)


tests/nightly/gpu/test_gpt2.py
```diff
@@ -95,3 +102,76 @@ def test_nospecialtok(self):
...
+
+    def _distributed_train_model(self, opt):
+        with testing_utils.tempdir() as tmpdir:
+            if 'model_file' not in opt:
+                opt['model_file'] = os.path.join(tmpdir, 'model')
+            if 'dict_file' not in opt:
+                opt['dict_file'] = os.path.join(tmpdir, 'model.dict')
+
+            parser = mp_train.setup_args()
+            popt = self._forced_parse(parser, opt)
```
stephenroller, line 147: please add a TODO over here and in the multiprocessing tests to kill forced_parse after dictionaries build correctly

stephenroller (approved):

klshuster (approved):
awesome! as a note to @stephenroller I standardized the `distributed_train_model` command/moved it to testing_utils in #2371 so once that lands we can update this to use that as well

PR #3213, opened 2020-10-21
Support special tokens in non-HF BPE dictionaries.

stephenroller:
**Patch description**
Presently, ByteLevelBPE tokenizer is the only one that supports special tokens, which are interpreted before BPE tokenization. This diff adds support for the feature in SlowByteLevelBPE tokenizer, GPT2 tokenizer, and SubwordBPE tokenizer.

As an additional bit (for testing, but also general usage), this diff also adds new model zoo files for the dictionaries/tokenizers we used in BlenderBot. I find myself using them frequently, and want them distributed without having to download the full models. Additionally, they're used as tests here.

**Testing steps**
New CI. Experiment launched using this new feature.


parlai/core/dict.py
```diff
@@ -333,12 +333,13 @@ def add_additional_special_tokens(self, additional_special_tokens: List[str]):
         """
         self.additional_special_tokens = additional_special_tokens
 
-        if (
-            self.additional_special_tokens
-            and not self.supports_additional_special_tokens()
-        ):
-            raise RuntimeError(
-                f'{self.tokenizer} does not currently support adding additional special tokens'
+        if hasattr(self, 'bpe'):
```
emilydinan, line 336: not true. you can add extra special tokens with split or space

emilydinan: see lines you deleted below

parlai/utils/bpe.py
```diff
@@ -219,6 +229,17 @@ def sync_with_dict(self, dict_agent):
             agent with which we are syncing the dictionary
         """
 
+    def add_special_tokens(self, dict_agent, special_tokens: List[str]):
+        """
+        Add special tokens to the tokenizer.
+
+        These tokens are treated as discreet tokens, and prioritized over the BPE
```
emilydinan, line 236: nit: discrete ? but i'm also not sure what you're trying to say here

emilydinan (commented):
nice this will be useful

parlai/core/dict.py
stephenroller: They weren't properly handled.

parlai/utils/bpe.py
stephenroller: ya bad english

parlai/core/dict.py
stephenroller: With space or split, special tokens were still being split up

emilydinan:
is this ready for re-review?

stephenroller:
You can review. Only thing left is change the flag and add a warning.

parlai/utils/bpe.py
```diff
@@ -127,7 +129,7 @@ def add_cmdline_args(argparser):
             '--hf-skip-special-tokens',
             hidden=True,
             type='bool',
-            default=True,
+            default=False,
```
EricMichaelSmith, line 132: Hmm will this default change break compatibility with anything?

parlai/utils/bpe.py
```diff
@@ -219,6 +248,18 @@ def sync_with_dict(self, dict_agent):
...
         """
 
+    def add_special_tokens(self, dict_agent, special_tokens: List[str]):
+        """
+        Add special tokens to the tokenizer.
+
+        These tokens are treated as discreet tokens, and prioritized over the BPE
+        tokenization.
+        """
+        # note, HF ByteLevelBPE tokenizer handles special tokens itself in
```
EricMichaelSmith, line 258: Hmm what is the special way - is it something we should be concerned about if using this?

parlai/utils/bpe.py
```diff
@@ -219,6 +248,18 @@ def sync_with_dict(self, dict_agent):
...
+        # a special way
+        for token in special_tokens:
+            self._special_tokens[token] = 1
```
EricMichaelSmith, line 261: hmm why a dict and not a set, if the val is always 1? Do we need a fixed sort order?

parlai/zoo/blender/dict_3B.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+3B parameter Blender dictionary file: please see <parl.ai/project/blender>.
```
EricMichaelSmith, line 8: hmm, I've never seen this URL notation before - https://parl.ai/project/blender gives me a "Page not found"

tests/test_dict.py
```diff
@@ -544,3 +544,48 @@ def _run_prefix_space_test(self, agent):
...
+            opt = parser.parse_kwargs(**kwargs)
+            da = DictionaryAgent(opt)
+            before = da.tokenize(string)
+            da.add_additional_special_tokens(["SPECIAL_TOKENS"])
+            after = da.tokenize(string)
+            assert before != after
+            assert len(before) > len(after)
+            assert after[-1] == "SPECIAL_TOKENS"
+            # we need to let the dictionary handle the tokenid mappings
+            assert da.vec2txt(da.txt2vec(string)) == string
```
EricMichaelSmith, line 570: ❤️

EricMichaelSmith (commented):
Looks good to me, with minor comments - not approving yet because Emily had outstanding comments

parlai/utils/bpe.py
stephenroller: Adding a warning about this is what's left in the PR.

parlai/utils/bpe.py
stephenroller: No, I just wanted to point out this is all no-ops if using bytelevelbpe.

parlai/utils/bpe.py
stephenroller: I'm exploiting the fact that dictionaries in python 3.5+ are insert-ordered to emulate an ordered set. sets are unordered.

stephenroller:
Made an aggressive decision to just kill the hf_skip_special_tokens flag. No other tokenizer had behavior like this, and even the `transformers.GPT2Tokenizer` implementation had a different default (but supported both). After rummaging around, I couldn't find any examples of models trained with BlenderBot as a base that had special in the labels, which is the only place this affects.

So bringing it in line with all the other tokenizers.

stephenroller:
This is finished.

EricMichaelSmith (approved):
Approving given comments

parlai/core/dict.py
stephenroller: Okay, supported now.

PR #3219, opened 2020-10-23
Stabilized distributed gpt2 test

stephenroller:
**Patch description**
Attempts to make the new test more reliable in CI.

**Testing steps**
Ran CI manually, plus the auto run here.

### DIFF
--- a/tests/nightly/gpu/test_gpt2.py
+++ b/tests/nightly/gpu/test_gpt2.py
@@ -157,14 +157,18 @@ class TestDistributed(unittest.TestCase):
 
         return (valid, test)
 
+    @testing_utils.retry()
+    def test_distributed(self):
+        config = copy.deepcopy(self._base_config)
+        config['num_epochs'] = 50
+        config['task'] = 'integration_tests:overfit'
+        config['batchsize'] = 2
+        config['dropout'] = 0.0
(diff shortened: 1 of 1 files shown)


tests/nightly/gpu/test_gpt2.py
```diff
@@ -157,14 +157,18 @@ def _distributed_train_model(self, opt):
...
+    def test_distributed(self):
+        config = copy.deepcopy(self._base_config)
+        config['num_epochs'] = 50
+        config['task'] = 'integration_tests:overfit'
+        config['batchsize'] = 2
+        config['dropout'] = 0.0
+        config['attention_dropout'] = 0.0
+        config['learningrate'] = 1.0
+        config['momentum'] = 0.90
+        config['skip_generation'] = True
```
moyapchen, line 170: Worth it to make some of these the defaults in other distributed tests as well?

emilydinan (approved):

tests/nightly/gpu/test_gpt2.py
stephenroller: Maybe, I personally would rather just move on for the moment. A few of these are just low-memory speedups since we can't use `-opt adam` in GPT2 tests.

PR #3228, opened 2020-10-27
Torch 1.8 Compatibility

klshuster:
**Patch description**
External version of internal fix. Ping me for details.


stephenroller (approved):
hm, wonder why the detach_ above doesn't work?

klshuster:
actually it does, the `detach()` below is redundant. will remove

PR #3229, opened 2020-10-27
[TCA] Classifiers load LR Scheduler state

emilydinan:
**Patch description**
TorchClassifierAgent was not updated to load the LR scheduler state when we added LRScheduler stuff. This adds it. Thanks to @wyshi for noticing something was wrong.

### DIFF
--- a/parlai/core/torch_classifier_agent.py
+++ b/parlai/core/torch_classifier_agent.py
@@ -341,7 +341,7 @@ class TorchClassifierAgent(TorchAgent):
             self.threshold = None
 
         # set up model and optimizers
-
+        states = {}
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
Tests for these fixes would be nice...

stephenroller:
I also still firmly believe that if someone has a dependency on bert classifier, this is the right time to kill AdamWithDecay

emilydinan:
> I also still firmly believe that if someone has a dependency on bert classifier, this is the right time to kill AdamWithDecay

i think that should be a separate PR as this PR is not specfici to BERT

PR #3231, opened 2020-10-27
Implement BART incremental decoding

stephenroller:
**Patch description**
Adds support for incremental decoding with bart.

**Testing steps**
Ran an evaluation of a model with -bs 1, -bs 8, and mp_eval -bs 16 -dynb full. Verified results were the same. Discussed speedups with @EricMichaelSmith and they were congruent with his experiences.

### DIFF
--- a/parlai/agents/bart/modules.py
+++ b/parlai/agents/bart/modules.py
@@ -56,7 +56,19 @@ class BartModel(TransformerGeneratorModel):
         """
         Incremental state is weird to handle when we seed decoder with two inputs
         initially.
-
-        We leave as a future exercise.
         """
-        return None
+        # we only have this method called when it's actually being used
+        assert incremental_state is not None
+        assert len(incremental_state) > 0
+
+        for layer in incremental_state.keys():
+            incr_state_l = incremental_state[layer]
+            assert 'self_attn' in incr_state_l
+            assert 'prev_mask' in incr_state_l['self_attn']
+            self_attn_mask = incr_state_l['self_attn']['prev_mask']
+            # check this is on the very first run with incremental state
+            if self_attn_mask.ndim == 3 and tuple(self_attn_mask.shape[1:]) == (2, 2):
+                # cut off the inappropriate incremental state
+                incr_state_l['self_attn']['prev_mask'] = self_attn_mask[:, -1:, :]
+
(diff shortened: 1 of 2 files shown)


parlai/agents/bart/modules.py
```diff
@@ -56,7 +56,19 @@ def reorder_decoder_incremental_state(
...
         initially.
-
-        We leave as a future exercise.
         """
-        return None
+        # we only have this method called when it's actually being used
+        assert incremental_state is not None
+        assert len(incremental_state) > 0
+
+        for layer in incremental_state.keys():
```
EricMichaelSmith, line 64: Nit: could probably just do `for incr_state_l in incremental_state.values():` here, right?

parlai/agents/bart/modules.py
```diff
@@ -56,7 +56,19 @@ def reorder_decoder_incremental_state(
...
+            incr_state_l = incremental_state[layer]
+            assert 'self_attn' in incr_state_l
+            assert 'prev_mask' in incr_state_l['self_attn']
+            self_attn_mask = incr_state_l['self_attn']['prev_mask']
+            # check this is on the very first run with incremental state
+            if self_attn_mask.ndim == 3 and tuple(self_attn_mask.shape[1:]) == (2, 2):
+                # cut off the inappropriate incremental state
+                incr_state_l['self_attn']['prev_mask'] = self_attn_mask[:, -1:, :]
```
EricMichaelSmith, line 72: Hmm so I suppose the problem was that dim=1 of `prev_mask` included both start tokens, whereas `prev_key` and `prev_value` only had 1 each?

EricMichaelSmith (approved):
Makes sense - great to have!

parlai/agents/bart/modules.py
stephenroller: Problem was that prev mask was a Bx2x2 tensor instead of a Bx2x1. The mask here appears the be accumulating what the new token is allowed to attend to (everything)

klshuster: Assuming you mean Bx2xD vs. Bx1xD in this explanation? (fits the line too)

klshuster (approved):
thank you for completing my future exercise

PR #3232, opened 2020-10-27
[bpe] Support BPE dropout

stephenroller:
**Patch description**
Only works for BART and SlowByteLevelBPE. Still need to add support for HuggingFace Tokenizers.

**Testing steps**
Internal experiments.

### DIFF
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -21,6 +21,15 @@ import re
 import parlai.utils.logging as logging
 from parlai.utils.io import PathManager
 from typing import List
+import enum
+
+
+class TokenizationMode(enum.Enum):
+    TRAIN_TIME_TEXT = 0
+    TRAIN_TIME_LABEL = 1
+    TEST_TIME_TEXT = 2
+    TEST_TIME_LABEL = 3
+
 
 RETOK = re.compile(r'\w+|[^\w\s]|\n', re.UNICODE)
 
@@ -236,7 +245,7 @@ class DictionaryAgent(Agent):
         ).split(",")
 
         # used to signal whether we should use training time tricks, like bpe droput
-        self._is_training_mode = False
+        self._tokenization_mode = TokenizationMode.TEST_TIME_LABEL
 
         try:
             self.tokenizer_fun = getattr(self, self.tokenizer + '_tokenize')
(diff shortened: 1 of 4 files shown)


parlai/core/dict.py
```diff
@@ -694,21 +697,6 @@ def sort(self, trim=True):
         assert len(self.freq) == len(self.ind2tok) == len(self.tok2ind)
         return sorted_pairs
 
-    def parse(self, txt_or_vec, vec_type=list):
```
stephenroller, line 697: This is an unrelated change, but just killing a bad method.

stephenroller:
Should be finished now.

parlai/core/torch_agent.py
```diff
@@ -1660,14 +1660,26 @@ def observe(self, observation):
             # make sure we note that we're expecting a reply in the future
             self.__expecting_to_reply = True
 
+        # keep around the observation for updating history based on label
         self.observation = observation
+
+        # possibly change tokenization methodology based on if this is a
+        # training example
+        if hasattr(self.dict, 'set_training_mode'):
```
moyapchen, line 1668: I'm lightly confused by this piece of code. `set_training_mode` is a function in DictAgent but this `if` seems to be looking for it as a member variable? 

(Noticing that this is stacked on top of another diff so maybe it's written there, but makes it a tad tricky to review if so.)

parlai/core/dict.py
moyapchen: Nit: Probably best to put this in a separate change for rubber stamping?

parlai/core/torch_agent.py
stephenroller: Functions are also attributes

parlai/utils/bpe.py
```diff
@@ -124,8 +127,20 @@ def add_cmdline_args(argparser):
...
         )
+        parser.add_argument(
+            '--bpe-dropout',
+            type=float,
+            default=None,
+            help='Use BPE dropout during training.',
+        )
         return parser
 
+    def set_training_mode(self, mode: bool):
```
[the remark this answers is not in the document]
stephenroller: That was intentional. BPE dropout is just one training time decision you might make.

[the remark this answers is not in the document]
moyapchen: If this can take multiple values beyond just a setting BPE to true, either it should be changed away from a bool to an enum or `mode` should be `use_bpe` or something. 

(All kind of nits, but from a "someone else looking at the code quickly figuring out what's supposed to happen here" perspective...)

parlai/core/torch_agent.py
moyapchen: Are there cases where `self.dict` isn't `DictionaryAgent`? Cause otherwise, this will always return `True`, no?

moyapchen (approved):
Nits, but otherwise seems fine

moyapchen (approved):

chinnadhurai (approved):

PR #3234, opened 2020-10-28
[CrowdSourcing] Turn annotation edits

jxmsML:
**Patch description**
A few issues for enabling open sourcing mephisto task dependent on turn_annotation_static task:
- unable to parse any quoted string (e,g,<span style="color: XXX") in task_description or task_title as it's re-processed in the jsx file since """ from Python is carrying over from ARG_STRING during pre-hydra era.
- Allow checkbox input type to be configurable, like the following (use radio button instead multi-label checkboxes)
Screen Shot 2020-10-28 at 2 34 55 PM
- Disable submit-button until the user input is validated. Before the change, the submit-button is grayish and turkers can still click to proceed to the next subtask even if the user input is NOT valid.
- Display correct closing left pane messages (before the change, it would show "You are currrently at conversation 4 / 3 " at the left pane after they submit the answers to the last subtask and before they return the HIT. Instead, the desired message would be:
Screen Shot 2020-10-28 at 2 20 54 PM

This pr also add a safety mephisto task that is dependent on the turn_annotation_static task.

**Testing steps**
CI

**Logs**

``` python projects/safety_recipes/human_safety_evaluation/run.py ```

Screen Shot 2020-10-29 at 3 40 16 PM
Screen Shot 2020-10-29 at 3 40 30 PM

**Other information**

### CODE REVIEW

File: projects/safety_recipes/human_safety_evaluation/run.py [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+    STATIC_BLUEPRINT_TYPE,
+)
+from parlai.crowdsourcing.tasks.turn_annotations_static.util import run_static_task
+from parlai.crowdsourcing.utils.mturk import MTurkRunScriptConfig
+
+TASK_DIRECTORY = os.path.dirname(os.path.abspath(__file__))
+PARLAI_HOME = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
+)
+BASE_TASK_DIRECTORY = os.path.join(
```
EricMichaelSmith, line 25: you can define this more easily (and in a way that is less likely to break in the future if the paths change) by importing the `run.py` file from `turn_annotations_static` and then setting `BASE_TASK_DIRECTORY = os.path.dirname(os.path.abspath(run.__file__))`

EricMichaelSmith (commented):
Thanks Jing - this is great to have! I had a few tiny cleanup comments first but I think it's almost ready to go

projects/safety_recipes/human_safety_evaluation/run.py
```diff
@@ -0,0 +1,59 @@
...
+    PARLAI_HOME, 'parlai/crowdsourcing/tasks/turn_annotations_static'
+)
+
+defaults = [
```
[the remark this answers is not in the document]
jxmsML: good pt!

parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
```diff
@@ -120,6 +120,7 @@ var handleSubtaskSubmit = function (subtaskIndex, numSubtasks, initialTaskData,
     mephistoSubmit(window.workerAnswers);
   }
   showDisableCssNextButton();
+  setIndex(subtaskIndex + 1);
```
jxmsML, line 123: moving setIndex from old L233 to here (in old version, index (subtask index get bumped to next +1 regardless of whether the turker input is valid or not)

projects/safety_recipes/human_safety_evaluation/run.py
jxmsML: good pt!

EricMichaelSmith (approved):
Looks good to me - thanks for those tweaks!

emilydinan (approved):
LGTM, thanks @jxmsML 🥳  

Can you update the projects page README with instructions for how to run this task with the safety stufF?

PR #3235, opened 2020-10-28
[BERT] Deprecate Adam with Decay

emilydinan:
**Patch description**
Deprecate Adam with Decay as Adam with weight decay is available in Pytorch now. 

I also added a small warning about the dictionary due to some confusion CC @wyshi


parlai/agents/bert_ranker/bert_dictionary.py
```diff
@@ -26,6 +27,9 @@ class BertDictionaryAgent(DictionaryAgent):
     def __init__(self, opt):
         super().__init__(opt)
         # initialize from vocab path
+        logging.warn(
```
stephenroller, line 30: warn_once?

stephenroller (approved):
Love that red.

PR #3236, opened 2020-10-28
[Opt] Do not save load_from_checkpoint

emilydinan:
**Patch description**
Do not save `load_from_checkpoint` in opt. #3162 which fixed issues with loading from checkpoint caused this issue, as the init_model saved on disk was changed to a `.checkpoint` file during requeue. This caused torch agent to load checkpoint files, even when we are evaluating (as noticed by @klshuster).


stephenroller (approved):
Cheeky

klshuster:
😍

PR #3238, opened 2020-10-29
ACUTE-Eval Mephisto test

EricMichaelSmith:
**Patch description**
Adds a test to check the outputs of the ACUTE-Eval Mephisto task, by subclassing a supervisor testing class that is being introduced as part of https://github.com/facebookresearch/Mephisto/pull/296. The test will be skipped for all CI checks other than `crowdsourcing_tests`. A lot of this code will be modularized into a common testing framework when future `parlai.crowdsourcing.tasks` tests are written

**Testing steps**
`pytest tests/crowdsourcing/tasks/test_acute_eval.py`


stephenroller:
image


very nice

stephenroller:
I think this could potentially benefit in brevity and readability using the regressions stuff I'm exploring now.

EricMichaelSmith:
> I think this could potentially benefit in brevity and readability using the regressions stuff I'm exploring now.

Okay cool! Would you be able to point me to that? Yeah, it'd be great to make this more readable

tests/crowdsourcing/tasks/test_acute_eval.py
```diff
@@ -0,0 +1,359 @@
...
+            args = self.BlueprintClass.ArgsClass()
+            args.timeout_time = 5
+            args.is_concurrent = False
+            task_runner = task_runner_class(
+                self.task_run, self.config.mephisto, EMPTY_STATE
+            )
+            sup.register_job(self.architect, task_runner, self.provider)
+            channel_info = list(sup.channels.values())[0]
+            task_runner = channel_info.job.task_runner
+            sup.launch_sending_thread()
```
JackUrb, line 319: Ultimately by using the `AbstractTestSupervisor` rather than relying on the `Operator` more directly to handle configuration, I think you're slightly overcomplicating the setup flow, and requiring test writers to be utilizing different information than they've needed to use for your runs. The only setup you should really need inside of a test case is something like:
```
db, cfg = load_db_and_process_config(cfg)
operator = Operator(db)
operator.validate_and_run_config(cfg.mephisto, shared_state)
server = operator.supervisor.channels[-1].job.architect.server
# ...rest of test...
```
This also means that you can have slightly different compositions and confs for different kinds of tests you want to run, and those configuration arguments can exist in a regular `conf/*.yaml` file rather than needing to manually set args inline as here.

EricMichaelSmith: Yes - changed to using an `Operator`, as discussed offline

.circleci/config.yml
```diff
@@ -148,6 +148,18 @@ installdetectrondeps: &installdetectrondeps
       cd maskbench; git checkout 4c168a637f45dc69efed384c00a7f916f57b25b8 -b stable
       python setup.py develop; cd -
 
+installcrowdsourcingdeps: &installcrowdsourcingdeps
+  run:
+    name: Install Mephisto
+    command: |
+      cd ..
+      git clone git@github.com:facebookresearch/Mephisto.git Mephisto
+      cd Mephisto; git checkout a9abd9deb2d2a5e0d6fde7dd6873fb73b4cb4872 -b stable
```
JackUrb, line 157: I suppose there's going to need to be some stable version of Mephisto that ParlAI is set to work with? Not sure the best way to get that in here.

tests/crowdsourcing/tasks/test_acute_eval.py
```diff
@@ -0,0 +1,355 @@
...
+                        f'+current_time={int(time.time())}',
+                    ],
+                )
+                # TODO: when Hydra 1.1 is released with support for recursive defaults,
+                #  don't manually specify all missing blueprint args anymore, but
+                #  instead define the blueprint in the defaults list directly.
+                #  Currently, the blueprint can't be set in the defaults list without
+                #  overriding params in the YAML file, as documented at
+                #  https://github.com/facebookresearch/hydra/issues/326 and as fixed in
+                #  https://github.com/facebookresearch/hydra/pull/1044.
```
JackUrb, line 309: ❤️

JackUrb (approved):
This is a great first step to real crowdsourcing tests across all of our tasks!

.circleci/config.yml
EricMichaelSmith: hmm, if Mephisto supports `pip install` and version numbering in the future, I could just add a row for it in `requirements.txt` :)

tests/crowdsourcing/tasks/test_acute_eval.py
```diff
@@ -0,0 +1,361 @@
...
+            )
+
+            # Check that the inputs and outputs are as expected
+            agent = self.db.find_agents()[0]
+            state = agent.state.get_data()
+            self.assertEqual(DESIRED_INPUTS, state['inputs'])
+            self.assertEqual(DESIRED_OUTPUTS, state['outputs'])
+
+
+except ImportError:
```
klshuster, line 357: sorry for the late review - we usually handle these `ImportError`s via a wrapper, e.g. `@testing_utils.skipUnlessMephisto` --> see https://github.com/facebookresearch/ParlAI/blob/master/parlai/utils/testing.py#L72. Is there a reason we did not do this here?

EricMichaelSmith: Yeah, good catch - I tried to do that here, but in this case, if I can't import objects from `mephisto`, (say, `LocalMephistoDB`, `Operator`, etc.), I won't be able to define `TestAcuteEval` at all, even with a `@testing_utils.skipUnlessMephisto` decorator, because those unimportable objects are called in the test class. That's not the case with checks like `skipUnlessTorch`, where we don't try to do `import torch` in the test module itself

klshuster: Ahh true - you could always import within the class and set it as an attr:

```
import mephisto
self.mephisto = mephisto
```

which isn't exactly super clean per se, so what you did works for me

PR #3244, opened 2020-10-30
[Chunk Teacher] Reset bug fix

emilydinan:
**Patch description**
Chunk teacher experiences issues when reset is called in the middle of loading a chunk. Since concurrent futures is unable to cancel futures, the way we fix this is to add a counter to track the number of resets and remove stale examples at `get` time. `self.tot_samples_loaded` also has to be keyed on the number of resets to be tracked properly.

**Testing**
Extensive testing with example that triggered the error. We see that indeed stale examples are removed:
```
2020-11-04 13:53:10,946 INFO     | We removed 14896 stale examples from the queue
2020-11-04 13:57:29,989 INFO     | We removed 948 stale examples from the queue
```
I also wrote a test in which stale examples are removed due to slow loading. 

Also made a change to only reset the agent at the end of validation if we do not share valid/train agents. This is unnecessary in that case because we reset the valid world at the beginning of validation.


parlai/utils/misc.py
```diff
@@ -311,6 +311,21 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+class SimpleCounter:
+    """
+    Simple counter object.
+    """
+
+    def __init__(self, value=0):
+        self.val = value
+
+    def increment(self, value=1):
+        self.val += value
```
stephenroller, line 323: I think this needs a lock :/

emilydinan: I added a lock and it caused a deadlock actually. I got rid of it because I figured that the actual value doesn't matter as much as the fact that it increased in value

emilydinan: I can try with a reentrant lock?

parlai/scripts/train_model.py
```diff
@@ -516,7 +516,8 @@ def _run_single_eval(self, opt, valid_world, max_exs):
             cnt = valid_world.report().get('exs') or 0
 
         valid_report = valid_world.report()
-        valid_world.reset()  # make sure world doesn't remember valid data
+        if opt.get('validation_share_agent', False):
```
klshuster, line 519: so we're sure this is only necessary if `validation_share_agent` is `True`?

parlai/utils/misc.py
klshuster, line 323: so we don't need a lock here after all?

klshuster (approved):
cool, approving assuming a test comes later 😄

parlai/utils/misc.py
emilydinan: Nope because the increment/access only happens on the main thread now, so we are good

stephenroller (approved):

PR #3245, opened 2020-10-30
CircleCI refactor

stephenroller:
**Patch description**
For a while, we've been using the legacy 2.0 syntax of CircleCI, rather than the newer 2.1 syntax. For a while, this was because we used a feature that was only available in 2.0 (dynamically running jobs), but now we run our full CI on every commit.

This PR refactors us to the newer 2.1 syntax. This also allows us to fix a small bug, where data caches were being indexed on unit tests, rather than teachers, and causing invalid caches more frequently than necessary.

**Testing steps**
CI


.circleci/config.yml
```diff
@@ -239,459 +333,146 @@ jobs:
...
-      - save_cache:
-          key: nvidia-downloads-20200516
-          paths:
-            - "~/nvidia-downloads/"
-      - restore_cache:
-          key: deps-20200921-gpu14-{{ checksum "requirements.txt" }}
-      - <<: *setup
-      - <<: *installtorchgpu14
-      - <<: *installdeps
-      - <<: *installdetectrondeps
```
EricMichaelSmith, line 357: Hmm how do things like `installdetectrondeps` get run now if they're not explicitly called?

.circleci/config.yml
```diff
@@ -239,459 +333,146 @@ jobs:
...
+          more_installs:
+            - installtorchgpu16
+          install_cuda: true
+          cachename: nightly
+          marker: nightly_gpu
+          pytest_flags: -v -s
 
   crowdsourcing_tests:
-    <<: *standard_cpu37
+    executor: small_cpu37
```
EricMichaelSmith, line 428: hmm - the change is just to conserve resources? If so, that's fine with me - you have a better sense of how much each size CPU can handle

tests/test_quickstart.sh
```diff
@@ -1,48 +0,0 @@
-#!/bin/bash
-# Copyright (c) Facebook, Inc. and its affiliates.
-# This source code is licensed under the MIT license found in the
-# LICENSE file in the root directory of this source tree.
-
-set -e -x  # error and exit on any failure; print the commands being run
-
-# test installation & help commands
```
EricMichaelSmith, line 8: Hmm I guess this was redundant?

EricMichaelSmith (approved):
Seems like a good idea! I had a few questions

.circleci/config.yml
stephenroller: Yeah, just lowering our bill

.circleci/config.yml
stephenroller: This was missed, it needs to be added back in

tests/test_quickstart.sh
stephenroller: I threw it away!

PR #3247, opened 2020-11-03
Ease-of-use improvements to wrapper teacher

EricMichaelSmith:
**Patch description**
In most cases, `AbstractWrapperTeacher` is used to edit the fields of the action returned by the teacher that it wraps around. Assuming that this inner teacher is a `FixedDialogTeacher`, however, editing the fields of this action is tricky because it must be done before the `FixedDialogTeacher` performs further processing on the action and registers it for metrics and evaluations, as part of `FixedDialogTeacher.process_action()`. This, this PR gives `AbstractWrapperTeacher` a concrete implementation of `.act()` and an abstract method `._edit_action()`, in order to give subclasses of `AbstractWrapperTeacher` an easy way to edit the action at the appropriate point in the pipeline without worrying about side effects.

This change is back-compatible with all existing subclasses of `AbstractWrapperTeacher`, which are not required to use the new `.act()` method or to provide an implementation of `._edit_action()`. However, two particular subclasses of `AbstractWrapperTeacher` have been upgraded to use `._edit_action()` to demonstrate how this new method is used.

**Testing steps**
This change has test coverage through `tests/nightly/gpu/test_style_gen.py` and `tests/test_wrapper.py`.

***test_new_tasks.py***
As expected, an error is produced when running `python tests/datatests/test_new_tasks.py` on `AbstractWrapperTeacher`, for the reason given in https://github.com/facebookresearch/ParlAI/pull/2642

### DIFF
--- a/parlai/tasks/style_gen/agents.py
+++ b/parlai/tasks/style_gen/agents.py
@@ -6,8 +6,9 @@
 
 import os
 
+from parlai.core.message import Message
 from parlai.core.opt import Opt
-from parlai.core.teachers import FixedDialogTeacher, ParlAIDialogTeacher
+from parlai.core.teachers import ParlAIDialogTeacher
 from parlai.tasks.style_gen.build import (
     build_personality_list,
     build_style_labeled_datasets,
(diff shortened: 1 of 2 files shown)


klshuster (approved):
lgtm!

dianaglzrico:
@EricMichaelSmith sorry, just saw this but this looks really nice! it will help to understand better for any future users, thanks for doing this!

PR #3250, opened 2020-11-04
Make directory at start of dict.save()

spencerp:
**Patch description**
If `self.bpe.copy_codecs_file` is hit before the directory is created, a "path does not exist" exception will be thrown. This won't happen until validation, which can waste a lot of time.

**Testing steps**
```
parlai train_model -t convai2 --model transformer/generator --dict-file /path/to/existing/model.dict --model-file /path/to/new/model --embedding-size 32 --n-positions 32 --truncate 32 --validation-every-n-epochs 0.01 --dict-tokenizer bpe
```
Where `/path/to/existing/model.dict` exists, but `/path/to/new/` doesn't exist yet.

### DIFF
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -626,6 +626,7 @@ class DictionaryAgent(Agent):
         If ``sort`` (default ``True``), then first sort the dictionary before saving.
         """
         filename = self.opt['dict_file'] if filename is None else filename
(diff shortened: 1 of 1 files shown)


klshuster:
good catch, lgtm thanks!

klshuster (approved):

PR #3253, opened 2020-11-06
[Safety Recipes] Open source Sensitive Topics classifier and data

emilydinan:
**Patch description**
Open source sensitive topics classifier and data. 

In the process, also made a change to automatically flatten tar files if `flatten_tar=True`, which is False by default so as not to break existing downloads.


jxmsML (approved):
💯

PR #3256, opened 2020-11-10
fix dialogpt dual usage of END_IDX

jxmsML:
**Patch description**
- Problem description: Unable to reproduce the model generation of DialoGPT under `-bs 1`  after this pr landed (https://github.com/facebookresearch/ParlAI/pull/3207)  even though it does not intend to change anything for `-bs 1`.

- It turns out in DialoGPT, both pad token and global history end token share the same `|<endoftext>|` (thus `NULL_IDX = END_IDX`), which after this pr landed, would cause the attention mask NOT to mask away end token.(https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/hugging_face/gpt2.py#L93). 

- In this pr, I reset NULL_IDX within the DialoGPTModel if ` NULL_IDX == END_IDX and bs == 1 and add_special_token=False`

- The test examples for reproducibility of DialoGPTModel generation before and after (https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/hugging_face/gpt2.py#L93) are
``` 
test_cases = [
            ("What a nice weather!", "I'm in the UK and it's raining here."),
            ("Nice to meet you!", "Hello! I'm from the future!"),
        ]
```
 with each being a (input, output) pair.

**Testing steps**
CI

**Logs**

```
```

**Other information**


parlai/agents/hugging_face/dict.py
```diff
@@ -125,6 +125,13 @@ def override_special_tokens(self, opt):
 
 
 class DialoGPTDictionaryAgent(Gpt2DictionaryAgent):
+    def _define_special_tokens(self, opt):
+        super()._define_special_tokens(opt)
+        if not opt["add_special_tokens"]:
+            # the original pad token '|<endoftext>|' has another usage in global history end token.
+            self.tokenizer.add_special_tokens({"pad_token": DIALOGPT_PAD_TOKEN})
```
stephenroller, line 132: you could also just make the mask in model.forward do nothing if NULL_IDX == END_IDX, idk.

jxmsML: yea that also can solve my problem. I'm just wondering adding pad token here regardless would be closer to fixing a root cause? or changes here would have any corner cases?

stephenroller: I mean, even simpler is just set NULL_IDX = -1 and don't add any special tokens. It already says bs=1 is mandatory for `not add_special_tokens`, and all we really need is for the `|<endoftext>|` token to not be masked out.

stephenroller:
I think we need a test

parlai/agents/hugging_face/dialogpt.py
```diff
@@ -26,6 +26,12 @@ class DialoGPTDecoder(GPT2Decoder):
     This decoder is initialized with the pretrained model from Hugging Face.
     """
 
+    def __init__(self, opt, dict):
+        super().__init__(opt, dict)
+        if opt.get('batchsize', 1) == 1 and self.END_IDX == self.NULL_IDX:
```
emilydinan, line 31: when is the latter condition not going to be true? if you are inheriting from this model but changing things?

jxmsML: when ```-bs 1 --add_special_token True ``` ? basically I only want to override the NULL_IDX if it's the same as END_IDX.

tests/nightly/gpu/test_dialogpt.py
```diff
@@ -16,6 +22,64 @@ class TestDialogptModel(unittest.TestCase):
...
+
+        print(f'results_single = {results_single}')
+        print(f'results_batched = {results_batched}')
+        assert results_single == results_batched
+
+    def test_batchsize(self):
+        """
+        Ensures dialogpt provides the same generation results regardless of batchsize.
+        """
+        for batchsize in [2, 2, 4, 2]:
```
mojtaba-komeili, line 71: * I do not understand, why the batch size 2 is repeated many times?
* Since you have 4 utterances, I think it is not a bad idea to test with a batch size that results in the last batch being less than batch size (for example 3).

jxmsML: Ah This is me testing generation consistensy on randomized initialization. The pr is work in progress.

tests/nightly/gpu/test_dialogpt.py
```diff
@@ -6,6 +6,12 @@
 
 import unittest
 import parlai.utils.testing as testing_utils
+from parlai.core.agents import create_agent
+import sys
+import warnings
+
+if not sys.warnoptions:
+    warnings.simplefilter("ignore")
```
stephenroller, line 14: Let's not have this in the diff right now

parlai/agents/hugging_face/dialogpt.py
```diff
@@ -26,6 +26,25 @@ class DialoGPTDecoder(GPT2Decoder):
...
+    def _get_special_tokens(opt, dict):
+        null_idx = dict.null_idx
+        if (
+            opt.get('batchsize', 1) == 1
+            and not opt['add_special_tokens']
+            and null_idx == dict.end_idx
+        ):
+            # get around the dual usage of end_idx that would otherwise mask endtoken during forward pass.
+            null_idx = -1
+            warn_once("WARNING: null_idx is set to -1 otherwise null_idx = end_idx")
```
stephenroller, line 45: Do we need the warning? IDTS, no?

tests/nightly/gpu/test_gpt2.py
```diff
@@ -12,9 +12,14 @@
 import parlai.scripts.build_dict as build_dict
 import os
 import copy
+from parlai.utils.misc import warn_once
 
 
 class TestGpt2(unittest.TestCase):
+    warn_once(
```
stephenroller, line 19: just a comment, not a warning plz

stephenroller (approved):
lgtm

PR #3259, opened 2020-11-10
Revert "[Safety Recipes] Open source Sensitive Topics classifier and data"

stephenroller:
Reverts facebookresearch/ParlAI#3253

Broke teachers
```
        if ".zip" in fname:
>           return _unzip(path, fname, delete=delete, flatten=flatten)
E           TypeError: _unzip() got an unexpected keyword argument 'flatten'
```

PR #3260, opened 2020-11-10
[Safety Recipes] Open source Sensitive Topics classifier and data

emilydinan:
Reverts facebookresearch/ParlAI#3259, which reverts another PR. Problem was that the keyword arg `flatten` was passed to the `unzip` function, when it should only be passed to the `untar` function. Tests didn't catch due to stale cache.


stephenroller (approved):
Only reviewed the fresh build

PR #3263, opened 2020-11-12
Analysis script for static turn annotations

EricMichaelSmith:
**Patch description**
Generalize and open-source the script to analyze static-turn-annotation Mephisto runs. This code is now a class, `TurnAnnotationsStaticResultsCompiler`, that can be easily subclassed in order to change settings or functionality. Buckets are defined in the `PROBLEM_BUCKETS` attribute.

**Testing steps**
`parlai tests/crowdsourcing/tasks/turn_annotations_static/test_analysis.py`: this tests all major parts of the code, including checking against gold annotations and in-flight QA.


parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
```diff
@@ -0,0 +1,558 @@
...
+            opt = {}
+        if 'results_folders' in opt:
+            self.results_folders = opt['results_folders'].split(',')
+        else:
+            self.results_folders = None
+        self.output_folder = opt.get('output_folder')
+        self.onboarding_in_flight_data_file = opt.get('onboarding_in_flight_data_file')
+        self.gold_annotations_file = opt.get('gold_annotations_file')
+
+    def get_data_paths_mephisto(self, task_run_id_folder):
```
jxmsML, line 80: why dont we use the databrowser? b/c read run id folder is faster?

tests/crowdsourcing/tasks/turn_annotations_static/test_analysis.py
```diff
@@ -0,0 +1,172 @@
...
+none_all_good: 58.3% (1 gold problem samples)
+Average agreement problem samples only with 4 total gold utterances annotated was:
+bucket_0: nan% (0 gold problem samples)
+bucket_1: 100.0% (1 gold problem samples)
+bucket_2: nan% (0 gold problem samples)
+bucket_3: nan% (0 gold problem samples)
+bucket_4: 100.0% (2 gold problem samples)
+none_all_good: 33.3% (1 gold problem samples)
+Calculating agreement on 8 annotations.
+Fleiss' kappa for bucket_0 is: -0.410.
```
jxmsML, line 127: somehow I'm confused at the calculation of fleiss' kappa. I thought the fleiss kappa is calculated the agreement among the same set of raters. Given how the turn annotation task is set up, each assignment can have different set of raters, so I'm expecting to see fless kappa calculated PER assignment PER bucket?

jxmsML (approved):
everything else lgtm!!! Thanks for standardizing this so nicely in the public repo! I'll direct the comments to maybe other reviewers in this pr.

parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
JackUrb: Good question - I'd consider direct folder access to be something of an anti-pattern, especially if this code intends to work with future Mephisto setups (internal may not use the MephistoDB that stores files locally, and the possible distributed MephsitoDB definitely wont have easy access like this).

EricMichaelSmith: Good point, thanks! There are a lot of things left to refactor for these analysis scripts, so I've added this to the list of todos, and I've added a TODO on this line about that

tests/crowdsourcing/tasks/turn_annotations_static/test_analysis.py
EricMichaelSmith: Hmm interesting - I've read over the definition of Fleiss' kappa at https://en.wikipedia.org/wiki/Fleiss%27_kappa and it looks like by default it sums over assignments (which I think this is what they call "subjects" in this article), but I agree that it might be useful to break this metric out by assignment as well. I've added a TODO line for that in this script, as well as a note in my list of future improvements.

PR #3266, opened 2020-11-13
fix: call custom dictionary agent's add_cmdline_args()

ihcho9088:
When build_dict.py is executed, an argument '**--dict-class**' can be given to specify which dictionary class to use when building dictionary.
However, this custom **dict-class** is not able to add any command line arguments since DictionaryAgent's static method add_cmdline_args is always called instead of custom **dict-class**'s one.

This pull request fixes this by parsing given python arguments before argparser takes an effect.
This it sort of dirty, but I could not find a better solution for this.


stephenroller:
We will be able to do this more cleanly with #3328

PR #3269, opened 2020-11-16
MDGender Open Source Pt. 3 (FINAL!)

emilydinan:
**Patch description**
Final MDGender open source: adds the Opensubtitles data. The way it works is that we redownload open subtitles to include character information. This data is then further filtered using several criteria. THEN the data is processed for gender info.

Left to do:
- [ ] This code path hasn't been tested because the opensubtitles website is busy. Leaving this PR open for now...


ledw (approved):
Looks good!

PR #3285, opened 2020-12-02
Updating old Mephisto import reference

JackUrb:
**Patch description**
Found a reference to `mephisto.core` while investigating how much of ParlAI is `v0.3.0` ready.


EricMichaelSmith (approved):
Good catch!

PR #3305, opened 2020-12-14
Analyze results from multiple ACUTE-Eval runs

EricMichaelSmith:
Support performing analysis of ACUTE-Evals over **multiple** Mephisto runs simultaneously and then combining the results from all of them. Can be enabled by passing multiple Mephisto run IDs when calling `analysis.py`, in the format `--run-id acute_eval_1607692844,acute_eval_1607701999,acute_eval_1607709138`. Results will be saved in a new timestamped folder located inside `--outdir`. Will throw an error if runs contain multiple different eval questions (because then the results would not be all comparable with each other).


parlai/crowdsourcing/tasks/fast_acute/analysis.py
```diff
@@ -648,6 +654,66 @@ def save_results(self, path: str = None):
                 f.write(compiled_text)
 
 
+class MultiRunAcuteAnalyzer(AcuteAnalyzer):
```
jxmsML, line 657: It's great to have this!!! Btw can `MultiRunAcuteAnalyzer` handle different run_ids with different eval questions? It seems like the win dataframe is calculated per model pair excluding the eval question?

parlai/crowdsourcing/tasks/fast_acute/analysis.py
```diff
@@ -648,6 +654,66 @@ def save_results(self, path: str = None):
...
+    """
```
jxmsML, line 658: Given that this would be really useful, especially we might be collecting acute eval sequentially. Do you think it is necessary to enforce that across different runs they must use the exact same pairing files. (people must be alerted that pairing files will be overriden if they launch multiple runs using the same command).

parlai/crowdsourcing/tasks/fast_acute/analysis.py
```diff
@@ -46,17 +48,18 @@
...
-
 
 def setup_args():
     """
     Setup appropriate args.
     """
     parser = ParlaiParser(False, False)
     parser.add_argument(
-        '-id', '--run-id', type=str, default=None, help='run id to analyze'
+        '-id',
```
klshuster, line 58: nit: `'-ids', '--run-ids'` since this can now be more than one?

klshuster (approved):
looks good to me!

stephenroller:
Can you beef up the PR description a bit please.

parlai/crowdsourcing/tasks/fast_acute/analysis.py
EricMichaelSmith: Makes sense - changing this

parlai/crowdsourcing/tasks/fast_acute/analysis.py
EricMichaelSmith: Hm, not necessarily - for instance, run 1 might use models A and B and run 2 might use models A,B, and C; the pairings filepath will be different but we may still want to combine the results from the two runs

parlai/crowdsourcing/tasks/fast_acute/analysis.py
EricMichaelSmith: Yeah, good point - I'm now checking that only one eval question is used across all results, both within one run and across runs

EricMichaelSmith:
@jxmsML @klshuster @stephenroller Finished addressing all PR comments

jxmsML (approved):
LGTM!! :)

PR #3306, opened 2020-12-14
Remove package-lock files

EricMichaelSmith:
Remove two unnecessary package-lock files, which are automatically created from building the package and are highly liable to change - they are not needed by the user and may lead to vulnerability warnings


JackUrb (approved):
You should add all `package-lock.json` files underneath crowdsourcing to the `.gitignore`.

EricMichaelSmith:
> You should add all `package-lock.json` files underneath crowdsourcing to the `.gitignore`.

Just added a line to ignore these files across the entire ParlAI repo - we'd probably want to ignore them for any npm packages that aren't in the `crowdsourcing/` folder as well, right?

JackUrb:
Potentially - it's more that if ParlAI ever creates its own packages, you _do_ want to commit those. This is how it works in Mephisto, so I choose to only exclude in places where I know those won't be, but either is valid.

EricMichaelSmith:
> Potentially - it's more that if ParlAI ever creates its own packages, you _do_ want to commit those. This is how it works in Mephisto, so I choose to only exclude in places where I know those won't be, but either is valid.

Yes, fair enough, it's good to not block future packages throughout the code - changed this

PR #3315, opened 2020-12-18
[torch_generator] sort the special token by length

wyshi:
sort the special token list by length, because in https://github.com/facebookresearch/ParlAI/blob/master/parlai/utils/bpe.py#L159-L167, 
`split = text.split(special_token)`.

e.g. if text = "ABCDEF", and special_token=['A', 'AB'], if they are not sorted, then text would become ['A', 'BCDEF'], instead of ['AB', 'CDEF'].

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -865,7 +865,10 @@ class TorchAgent(ABC, Agent):
         Made easily overridable for special cases.
         """
         if self.opt.get('special_tok_lst'):
-            return self.opt['special_tok_lst'].split(',')
+            special_tok_lst = self.opt['special_tok_lst'].split(',')
+            # sort the tokens by length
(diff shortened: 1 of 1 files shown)


stephenroller (changes requested):
I'm not fond of this. The user has total control over the ordering of the special tokens as of right now, and you're removing that. Just input a sorted list for the effect you want.

wyshi:
> I'm not fond of this. The user has total control over the ordering of the special tokens as of right now, and you're removing that. Just input a sorted list for the effect you want.

ok I agree. Then maybe we should add an explanation somewhere on how the special tokens are used? It took me a while and many pdbs to realize what's happening and why some special tokens are not split and some special tokens are...

stephenroller:
Sounds good

stephenroller:
Replaced with #3349.

PR #3316, opened 2020-12-18
[BART] Add BART-Base model

emilydinan:
**Patch description**
Add BART-base (smaller version, only 6 enc/dec layers, emb size of 768) to ParlAI. Due to some dictionary weirdness, we add an additional conversion step to truncate the embedding size, explained in the comments.

### DIFF
--- a/parlai/zoo/bart/build.py
+++ b/parlai/zoo/bart/build.py
@@ -26,7 +26,7 @@ CONVERSION_ARGS = {
     'history_add_global_end_token': None,
 }
 
-BART_ARGS = {
+BART_LARGE_ARGS = {
     'embedding_size': 1024,
     'ffn_size': 4096,
     'dropout': 0.1,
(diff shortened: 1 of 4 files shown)


stephenroller (approved):
test and lint plz

parlai/zoo/bart/build.py
```diff
@@ -57,14 +65,18 @@ def download(datapath, version='v1.0'):
...
-        ConversionScript.main(**args)
+        models = ['bart.large', 'bart.base']
+        for model_name in models:
+            url = f'http://dl.fbaipublicfiles.com/fairseq/models/{model_name}.tar.gz'
+            build_data.download(url, dpath, f'{model_name}.tar.gz')
+            build_data.untar(dpath, f'{model_name}.tar.gz')
+            args = CONVERSION_ARGS.copy()
+            if model_name == 'bart.base':
+                args['model'] = 'bart/base'
+            args['orig_dict_file'] = os.path.join(dpath, 'bart.large/dict.txt')
```
klshuster, line 76: should this be f`{model_name}/dict.txt'? if not, i'm not sure the added option is necessary (since it doesn't seem to differ for either model)

klshuster (approved):
super clean, glad this was not super difficult to do (or at least doesn't appear to have been!)

echoing stephen's test and lint requests as well 😄

stephenroller:
This was abandoned

PR #3319, opened 2020-12-21
Move README and add references to it

EricMichaelSmith:
This PR moves the main ParlAI crowdsourcing README from `parlai/crowdsourcing/tasks/` to `parlai/crowdsourcing/` for greater visibility. Within each crowdsourcing task's own README, a link has been added to refer the user back to the main README, for general info on how to specify a custom Hydra YAML file, etc.


JackUrb (approved):

mwillwork (approved):
This is fine but I actually meant writing it in the run.py file like this: 

# To run the task with your own config outside this folder (recommended!)
# Use the command formulation below. For more info, 
# check the README in parlai/crowdsourcing/
# python turn_annotations_static/run.py conf=<conf name sans yaml> --config-dir <path to directory with a conf/ folder

EricMichaelSmith:
> This is fine but I actually meant writing it in the run.py file like this:
> 
> # To run the task with your own config outside this folder (recommended!)
> # Use the command formulation below. For more info,
> # check the README in parlai/crowdsourcing/
> # python turn_annotations_static/run.py conf= --config-dir <path to directory with a conf/ folder

Hmm - @mwillwork the reason I didn't add those conf instructions to the run.py file itself is because I'd like to keep the instructions for a task centralized in the README, not in the run.py file or anywhere else: I think that'd be a good practice for users to check the README when they start using a task, because that should be the one source of truth. Having multiple sets of documentation there and in every run.py file can easily lead to conflicts as they become out of sync with each other, we forget to update one, etc. But if you feel strongly about this lemme know and I'm happy to reconsider

PR #3321, opened 2020-12-22
Fix for crowdsourcing CI checks

EricMichaelSmith:
Fix crowdsourcing CI checks that failed due to a last-minute flag name change in https://github.com/facebookresearch/ParlAI/pull/3305. (Note: it's possible that there's some flakiness in one of the tests that's causing the later ones to fail, but given its supposed flakiness, I can't reproduce. I'll keep an eye out for this on future merges to master.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -500,7 +500,10 @@ workflows:
           requires:
             - unittests_gpu16
             - unittests_37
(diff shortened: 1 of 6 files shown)


klshuster (approved):
thanks for fixing! also sorry, i think that might've been my comment that changed the flag name 😳

PR #3326, opened 2020-12-27
[transformer] Faster model parallel

stephenroller:
**Patch description**
Speed up model parallel generations for 9.4B model by ~40% on 8 GPUs.

The implementation of model parallel caused incremental state to be moved from device to device while generating. However, incremental state is computed per-layer, and isn't needed in later layers. As such, we were transferring the incremental state to GPUs that didn't need it, causing a significant slowdown.

This patch explicitly avoids transferring incremental state. Happily, this seems to remove most of the bottleneck we had with using >1 gpu in model parallel during generation. Previously, the generations got slower with 8 gpus than with 2 gpus. Now, generations seem to be about

**Testing steps**
test.py:
```python
#!/usr/bin/env python

import timeit
from parlai.core.agents import create_agent_from_model_file

agent = create_agent_from_model_file(
    "zoo:blender/blender_9B/model", {'model_parallel': True}
)

agent.observe({"text": "Hello!", "episode_done": True})
start = timeit.default_timer()
print(agent.act()['text'])
end = timeit.default_timer()
print(f"Time: {end - start:.2f}")
```

Patch | Model | # GPUs | Timing
------|-----|----|---
Before patch | 2.7B | 4 GPUs | 16.3s
After patch | 2.7B | 4 GPUs | 11.8s
Before patch | 9.4B | 4 GPUs | 20.8s
After patch | 9.4B | 4 GPUs | 16.4s

Patch | Model | # GPUs | Timing
------|-----|----|---
Before patch | 2.7B | 8 GPUs | 22.0s
After patch | 2.7B | 8 GPUs | 12.1s
Before patch | 9.4B | 8 GPUs | 26.8s
After patch | 9.4B | 8 GPUs | 16.0s

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -907,23 +907,29 @@ class TransformerDecoder(nn.Module):
         )
         work_items = PipelineHelper.schedule_work_items(self.layers, chunks)
 
-        new_incr_state = [{} for _ in chunks]
+        new_incr_state = {i: [] for i, _ in enumerate(self.layers)}
 
         for chunk_idx, layer_nos, next_device in work_items:
             s_tensor, s_enc_out, s_enc_mask, s_incr_state = chunks[chunk_idx]
             for layer_no in layer_nos:
-                s_tensor, new_incr_state[chunk_idx][layer_no] = self.layers[layer_no](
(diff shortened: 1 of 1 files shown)


klshuster (approved):
wow, i had noticed generation was slower when using model parallel, good catch

PR #3328, opened 2020-12-28
Migrate add_cmdline_args to include partial_opt

stephenroller:
# Are you coming here because an error message told you to?

I'm guessing you received an error message saying that an agent or a teacher's signature of `add_cmdline_args` changed.

Probably the agent and/or teacher you're trying to use has code like one of these:

```python
@staticmethod
def add_cmdline_args(parser):
     # .... stuff here ....

@classmethod
def add_cmdline_args(cls, parser):
     # .... stuff here ....
```

This needs to be upgraded to:
```python
@classmethod
def add_cmdline_args(cls, parser, partial_opt: Optional[Opt] = None) -> ParlaiParser:
     # .... stuff here ....
     return parser
```

The old version may use a static method, or not have a return value. In an effort to keep things tidy, please upgrade to a classmethod and return the parser, if your method doesn't already.

**Then the error message should go away. This PR should have fixed all official agents, but if we missed one, please file a new issue.**

-------

# Normal PR info

**Patch description**
Provide an official path for `partial_opt` to be ingested by `add_cmdline_args`, so that we can have conditional arguments. Also a codemod to fix as many agents and teachers as I could automatically. Some manual fixes were also required.

_Still needed_:
- [x] PRs for our internal-only code that need to be upgraded.

**Testing steps**
Old CI. New CI. Everybody CI CI.


stephenroller:
For posterity, here's the codemod script.

Invocation:
` git co -- . ; git ls-files | grep py$ | parallel python sub.py | tee checkup ; git st | grep modified: | awk '{print $2}' | xargs black`

Source: `sub.py`

```python
#!/usr/bin/env python3

FINAL = "def add_cmdline_args(cls, parser: ParlaiParser, partial_opt: Optional[Opt] = None) -> ParlaiParser:"

import sys

fn = sys.argv[1]
with open(fn, "r") as f:
    oldcode = code = f.read()

code = code.replace("def add_cmdline_args(self, *args, **kwargs):", FINAL)
code = code.replace("def add_cmdline_args(argparser: ParlaiParser):", FINAL)
code = code.replace("def add_cmdline_args(self, argparser):", FINAL)

code = code.replace("def add_cmdline_args(cls, parser):", FINAL)
code = code.replace("def add_cmdline_args(cls, argparser):", FINAL)

code = code.replace("def add_cmdline_args(parser):", FINAL)
code = code.replace("def add_cmdline_args(argparser):", FINAL)

code_lines = code.split("\n")

if code != oldcode:
    first_import = [
        i
        for i, line in enumerate(code_lines)
        if "import" in line and '    ' not in line
    ][0]
    if "import Opt" not in code:
        code_lines.insert(first_import, "from parlai.core.opt import Opt")
    if "import ParlaiParser" not in code:
        code_lines.insert(first_import, "from parlai.core.params import ParlaiParser")
    if 'import Optional' not in code:
        code_lines.insert(first_import, "from typing import Optional")

code_lines.append("===END===")

classname = None
methstart = None
lastmethstart = None
toadd = []
sawreturn = False
for i, line in enumerate(list(code_lines)):
    if line.startswith('class '):
        classname = line
    if 'def add_cmdline_args' in line or 'def setup_args' in line:
        lastmethstart = methstart = line
    elif methstart and ' return ' in line:
        sawreturn = True
    elif (
        'def ' in line
        or line.startswith('class ')
        or ' @' in line
        or '===END===' in line
    ) and methstart:
        # missing a return!
        if not sawreturn:
            if 'class' in line or ' @' in line:
                ln = i - 1
            else:
                ln = i
            toadd.insert(0, (ln, '        return parser'))
        methstart = None
        sawreturn = False

    line = line.replace(
        'dictionary = BPEHelper.add_cmdline_args(dictionary)',
        'dictionary = BPEHelper.add_cmdline_args(dictionary, partial_opt=partial_opt)',
    )

    if 'def add_cmdline_args' in line and '@staticmethod' in code_lines[i - 1]:
        code_lines[i - 1] = code_lines[i - 1].replace('staticmethod', 'classmethod')
    elif 'def add_cmdline_args' in line and '@classmethod' not in code_lines[i - 1]:
        toadd.insert(0, (i, '    @classmethod'))

    line = line.replace("argparser", "parser")

    if '.add_cmdline_args' in line and 'partial_opt' not in line:
        if lastmethstart is None or 'partial_opt' not in lastmethstart:
            line = line.replace(
                ".add_cmdline_args(parser)",
                ".add_cmdline_args(parser, partial_opt=None)",
            )
            line = line.replace(
                ".add_cmdline_args(argparser)",
                ".add_cmdline_args(parser, partial_opt=None)",
            )
        else:
            line = line.replace(
                ".add_cmdline_args(parser)",
                ".add_cmdline_args(parser, partial_opt=partial_opt)",
            )
            line = line.replace(
                ".add_cmdline_args(argparser)",
                ".add_cmdline_args(parser, partial_opt=partial_opt)",
            )
        if classname and 'super()' not in line and 'cls.' not in line:
            searchname, *_ = line.strip().split(".")
            if (
                'DictionaryAgent.add_cmdline_args' in line
                and lastmethstart
                and 'cls' in lastmethstart
            ):
                line = line.replace(
                    'DictionaryAgent.add_cmdline_args',
                    'cls.dictionary_class().add_cmdline_args',
                )
            elif ',' in searchname and 'hasattr' not in searchname:
                line = line.replace(searchname, 'super()')
            elif (
                searchname in classname
                and 'dictionary_class' not in line
                and ',' not in classname
            ):
                line = line.replace(searchname, 'super()')
            elif (
                searchname in classname
                and 'dictionary_class' in line
                and lastmethstart
                and 'cls' in lastmethstart
            ):
                line = line.replace(searchname, 'cls')
            else:
                print(f'{fn}:{i+1}:1: {line}')
    code_lines[i] = line

for i, val in toadd:
    code_lines.insert(i, val)

# remove END
code_lines.pop(-1)

code = "\n".join(code_lines)
code = code.replace("\n\n        return parser", "\n        return parser")


with open(fn, "w") as f:
    f.write(code)

```

parlai/agents/hred/hred.py
```diff
@@ -60,7 +65,7 @@ def add_cmdline_args(cls, argparser):
             help="Probability of replacing tokens with UNK in training.",
         )
 
-        super(HredAgent, cls).add_cmdline_args(argparser)
+        super().add_cmdline_args(parser, partial_opt=partial_opt)
         return agent
```
klshuster, line 69: kind of a nit (this applies to other classes as well), but `return agent` here returns the argument group, not a `ParlaiParser` (unless my understanding of the return value of `add_argument_group` is incorrect)

tests/test_params.py
```diff
@@ -173,5 +174,99 @@ def test_bool(self):
...
+                argparser.add_argument("--no-find", default=True)
+                return argparser
+
+        with self.assertLogs(logger=logging.logger, level='ERROR') as cm:
+            ParlaiParser(True, True).parse_kwargs(
+                model="repeat_query", task="bad_addcmdlineargs_teacher"
+            )
+            self.assertIn("add_cmdline_args(argparser)", "\n".join(cm.output))
+
+    def test_good_teacher(self):
```
klshuster, line 240: super nit: `test_good_task` to keep in line with above

klshuster (approved):
really glad to have this functionality

parlai/agents/hred/hred.py
stephenroller: This is worth debating. We use the `add_argument_group` bit quite a bit, in order to add sub-arguments of groups based on classes.

We could ban returning argument groups: every subclass should make a new group. Not sure? It's unclear to me that the class hierarchy is what should be reflected in the config options.

klshuster: i think the argument groups improve readability tremendously, especially when using e.g. `parlai tm -m my_agent -h`, as you can easily read through args based on who added them

where do we use this returned parser anyway?

stephenroller: In superclass:

group  = parser.add_argument_group()
return group

In subclass:

parser = super().add_argument_group()
parser.add_argument('--subclass-option')
return parser

stephenroller: Anyway I thought about this and figured our typechecker isn't strict, let's just violate it

parlai/core/params.py
```diff
@@ -299,7 +299,7 @@ class ParlaiParser(argparse.ArgumentParser):
     modules by passing this object and calling ``add_arg()`` or
     ``add_argument()`` on it.
 
-    For example, see ``parlai.core.dict.DictionaryAgent.add_cmdline_args``.
+    super().core.dict.DictionaryAgent.add_cmdline_args``.
```
EricMichaelSmith, line 302: hmm was this on purpose? The opening tick marks are getting removed, for example

tests/test_params.py
```diff
@@ -173,5 +174,99 @@ def test_bool(self):
         assert opt['bar'] is False
 
 
+class TestAddCmdlineArgs(unittest.TestCase):
+    def test_bad_agent(self):
+        from parlai.core.agents import register_agent
+
+        @register_agent("bad_addcmdlineargs_agent")
```
EricMichaelSmith, line 181: Would it be worth adding a comment here explaining why this is a bad agent and why the later one is a good agent?

EricMichaelSmith (approved):
Yay for standardization / partial opts! Minor comments

parlai/agents/hred/hred.py
klshuster: if it's just a matter of violating the typechecker (and nothing functionally is incorrect) I agree

stephenroller: Maybe the really clean thing would be to have the return value always be an ArgumentGroup but idk if I can handle that codemod

klshuster: you could modify the return to be `Union[ParlaiParser, ArgumentGroup]`, seems like an easier codemod

stephenroller: I've opted not to do that because it just is a lot of noise. We'll put this off for future better eng.

stephenroller:
Offline discussion deciding that we will raise a TypeError instead of logging an error message.

PR #3330, opened 2020-12-28
Testing of the QA data collection task

EricMichaelSmith:
**Patch description**
- Adds a CI check for the QA data collection task
- Removes the old pre-Mephisto QA data collection task
- Minor tweaks of the new QA data collection task to get it to match the structure of other `parlai.crowdsourcing` tasks
- Adds documentation

**Testing steps**
`pytest tests/crowdsourcing/tasks/qa_data_collection/test_qa_data_collection.py`


EricMichaelSmith:
@stephenroller @JackUrb @klshuster @mwillwork @jxmsML I had briefly made this a WIP earlier today, but I do think it's ready to review. The 2 things I have to change before merging in are:

1. Reverting removing the prerequisites for the crowdsourcing CI check, once I know all the checks pass
2. Fixing an issue with inference between CI checks. @JackUrb I'm still seeing a weird issue with these CI checks where an issue with one check will cause later ones to fail, even though I call `self.operator.shutdown()` after every task. Is there a better way to ensure isolation of these tasks? These interactions between tests are really subtle and quite tricky to debug - I've spent quite a while digging into it, but you have a better sense of what objects might be carrying over from one test to the next without being shutdown correctly

tests/crowdsourcing/tasks/qa_data_collection/test_qa_data_collection.py
```diff
@@ -0,0 +1,102 @@
...
+        SharedParlAITaskState,
+        BLUEPRINT_TYPE,
+    )
+
+    from parlai.crowdsourcing.tasks.qa_data_collection.run import TASK_DIRECTORY
+    from parlai.crowdsourcing.tasks.qa_data_collection.util import get_teacher
+    from parlai.crowdsourcing.utils.frontend import build_task
+    from parlai.crowdsourcing.utils.tests import AbstractParlAIChatTest
+
+    class TestQADataCollection(AbstractParlAIChatTest, unittest.TestCase):
```
klshuster, line 35: sorry i think i've said this in a previous PR but can you remind me why we can't have a wrapper for e.g. `skipUnlessMephisto` instead of this big try/except?

klshuster (commented):
looks good to me, but also looks like tests are failing

tests/crowdsourcing/tasks/qa_data_collection/test_qa_data_collection.py
EricMichaelSmith: Ah, because that'd only decorate the `TestQADataCollection` class, but here it's the Mephisto-based imports (including `AbstractParlAIChatTest`, which this class inherits from) that have to be checked for `ImportError`s

EricMichaelSmith (commented):
> looks good to me, but also looks like tests are failing

Thanks for reviewing! Yeah, the failing crowdsourcing CI check is a known issue - will wait for Jack to advise before merging in (see my comment above for context)

tests/crowdsourcing/tasks/qa_data_collection/test_qa_data_collection.py
klshuster: you could maybe wrap the imports in a try/except, set a variable like `MEPHISTO_AVAILABLE`, and then use the standard `unittest.skipUnless(MEPHISTO_AVAILABLE)`. i don't feel particularly strongly either way

EricMichaelSmith: Yes, that's a good idea - but maybe that'd be better as a separate PR, since I have a couple of PRs open and a clean implementation of this would affect all of them? But I've added this as a TODO on the feature-tracker list

klshuster: sounds good to me!

JackUrb (approved):
Overall the testing here is looking good to me!

@EricMichaelSmith Generally what you're doing in here for shutdown is already good, but the shutdown issue has a lot to do with Mephisto wanting threads to rejoin at specific times, and waiting for human input. I have a PR in the works for resolving a number of improved `Operator.shutdown` flows, including being able to force different levels of it, and perhaps that will help resolve what you're after.

JackUrb:
Following up here, try to run on https://github.com/facebookresearch/Mephisto/pull/348 and use the `Operator.force_shutdown()` method instead. This may resolve the crowdsourcing tests issue.

klshuster (approved):

Issue #3339, opened 2021-01-03
Babi - agents.py


klrkt:
Windows 10 system. CMD prompt as admin.
Trying to follow this:
https://parl.ai/docs/tutorial_quick.html#view-a-task-train-a-model
Tried:
git clone https://github.com/facebookresearch/ParlAI.git ~/ParlAI
cd ~/ParlAI
python setup.py develop
parlai display_data -t squad
all works fine - until I try  parlai display_data -t babi:task10k:1
This reports: AttributeError: module 'parlai.tasks.babi.agents' has no attribute 'create_agents'
This is CORRECT: 
  File "c:\users\bundy\chat\parlai-master\parlai\tasks\babi\agents.py" does not have 'create_agents' module.
THIS SEEMS TO BE THE PROBLEM?
I also tried  the create  simple parrot example. Same problem (because it uses the same agents.py

C:\Users\bundy\chat\ParlAI-master>parlai display_model -t babi:task10k:1 -m parrot
12:52:00 | creating task(s): babi:task10k:1
[building data: C:\Users\bundy\chat\ParlAI-master\data\bAbI]
12:52:00 | Downloading http://parl.ai/downloads/babi/babi.tar.gz to C:\Users\bundy\chat\ParlAI-master\data\bAbI\babi.tar.gz
Downloading babi.tar.gz: 0.00B [00:00, ?B/s]
Traceback (most recent call last):
  File "c:\users\bundy\chat\parlai-master\parlai\core\worlds.py", line 1212, in _create_task_agents
    task_agents = my_module.create_agents(opt)  # type: ignore
AttributeError: module 'parlai.tasks.babi.agents' has no attribute 'create_agents'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\bundy\AppData\Local\Programs\Python\Python38\Scripts\parlai-script.py", line 33, in <module>
    sys.exit(load_entry_point('parlai', 'console_scripts', 'parlai')())
  File "c:\users\bundy\chat\parlai-master\parlai\__main__.py", line 14, in main
    superscript_main()
  File "c:\users\bundy\chat\parlai-master\parlai\core\script.py", line 307, in superscript_main
    return SCRIPT_REGISTRY[cmd].klass._run_from_parser_and_opt(opt, parser)
  File "c:\users\bundy\chat\parlai-master\parlai\core\script.py", line 90, in _run_from_parser_and_opt
    return script.run()
[...]

Issue #3311, opened 2020-12-16
Error: File being used by another process when downloading datasets


Auxodevio:
**Bug description**
When downloading datasets (babi.tar.gz and personachat.tgz) I encounter the following error on Windows 10:
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'E:\\thesis\\ParlAI\\data\\Persona-Chat\\personachat.tgz'

**Reproduction steps**
Running **parlai display_data -t babi:task1k:1** produces:
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'E:\\thesis\\ParlAI\\data\\bAbI\\babi.tar.gz'

Running **parlai train_model -t personachat -m transformer/ranker -mf /tmp/model_tr6** produces:
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'E:\\thesis\\ParlAI\\data\\Persona-Chat\\personachat.tgz'

**Expected behavior**
The console output indicates that the file was successfully downloaded in both cases. I am not accessing the file using any other process. 

**Logs**
Please paste the command line output:

```
PS E:\thesis> parlai display_data -t babi:task1k:1
11:52:33 | Opt:
11:52:33 |     allow_missing_init_opts: False
11:52:33 |     batchsize: 1
11:52:33 |     datapath: E:\thesis\ParlAI\data
11:52:33 |     datatype: train:ordered
11:52:33 |     dict_class: None
11:52:33 |     display_add_fields:
11:52:33 |     download_path: None
11:52:33 |     dynamic_batching: None
11:52:33 |     hide_labels: False
11:52:33 |     ignore_agent_reply: True
11:52:33 |     image_cropsize: 224
11:52:33 |     image_mode: raw
11:52:33 |     image_size: 256
11:52:33 |     init_model: None
11:52:33 |     init_opt: None
11:52:33 |     loglevel: info
11:52:33 |     max_display_len: 1000
11:52:33 |     model: None
11:52:33 |     model_file: None
11:52:33 |     multitask_weights: [1]
11:52:33 |     num_examples: 10
11:52:33 |     override: "{'task': 'babi:task1k:1'}"
11:52:33 |     parlai_home: E:\thesis\ParlAI
11:52:33 |     starttime: Dec16_11-52
11:52:33 |     task: babi:task1k:1
11:52:33 |     verbose: False
[...]

Issue #3185, opened 2020-10-10
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 


SEMTEX99:
Using pycharm with venv py 3.6 sticking with the quickstart tutorial that is in the docs, but when i try to run:  parlai display_data -t babi:task10k:1 in the terminal i get PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'J:\\AI\\ParlAI\\data\\bAbI\\babi.tar.gz' error. it first downloads the file and then it shows the error it also happnes even when i change the interpreter to conda python 3.6 

cloned the repo, installed the requirements and torch 1.4, ran the setup.py develop command and then when i try the command above or even go the python parlai/scripts/safe_interactive.py -t blended_skill_talk -mf zoo:blender/blender_3B/model command it just downloads the file and always gives the same error also the same error happens on my laptop as well.

full terminal paste: 
22:47:26 | Opt:
22:47:26 |     allow_missing_init_opts: False
22:47:26 |     batchsize: 1
22:47:26 |     datapath: J:\AI\ParlAI\data
22:47:26 |     datatype: train:ordered
22:47:26 |     dict_class: None
22:47:26 |     display_ignore_fields: agent_reply
22:47:26 |     display_verbose: False
22:47:26 |     download_path: None
22:47:26 |     dynamic_batching: None
22:47:26 |     hide_labels: False
22:47:26 |     image_cropsize: 224
22:47:26 |     image_mode: raw
22:47:26 |     image_size: 256
22:47:26 |     init_model: None
22:47:26 |     init_opt: None
22:47:26 |     loglevel: info
22:47:26 |     max_display_len: 1000
22:47:26 |     model: None
22:47:26 |     model_file: None
22:47:26 |     multitask_weights: [1]
22:47:26 |     num_examples: 10
22:47:26 |     override: "{'task': 'babi:task10k:1'}"
22:47:26 |     parlai_home: J:\AI\ParlAI
22:47:26 |     starttime: Oct10_22-47
22:47:26 |     task: babi:task10k:1
22:47:26 | creating task(s): babi:task10k:1
[building data: J:\AI\ParlAI\data\bAbI]
22:47:26 | Downloading http://parl.ai/downloads/babi/babi.tar.gz to J:\AI\ParlAI\data\bAbI\babi.tar.gz
Downloading babi.tar.gz: 0.00B [00:00, ?B/s]
Traceback (most recent call last):
[...]

PR #3340, opened 2021-01-04
Suppress PermissionErrors in windows deletions.

stephenroller:
**Patch description**
Suppresses the error message in #3185, #3311, #3339, maybe others. Given that the workaround seems to work for people, just incorporate it with an error message.

**Testing steps**
None. :/


EricMichaelSmith (approved):
Seems reasonable

PR #3344, opened 2021-01-05
[minor] Clarify conversation format of ACUTE

moyapchen:
**Patch description**
Super minor, but going through things I had on my todo list. Makes it clear that the ACUTE-Eval format is different than the Conversations format and the relationship of `speakers_to_eval` to `dialogue_dict`.  (Latter of this apparently already existed, but was put far away enough where I apparently thought it had more to do with onboarding than the general format...)

For FB internal folks: https://fb.workplace.com/groups/766121277488755/permalink/844079596359589/?comment_id=844140446353504&reply_comment_id=844397979661084

### DIFF
--- a/parlai/crowdsourcing/tasks/acute_eval/README.md
+++ b/parlai/crowdsourcing/tasks/acute_eval/README.md
@@ -64,7 +64,7 @@ This is a template of the expected format with the minimal expected fields:
       ]
     }
 
-Note that we assume that "dialogue" consists of strictly alternating turns (e.g. speakers a, b, a, b, a...). Speakers that we would like to evaluate in the dialogues of `dialogue_dicts` should appear in the same order as `speakers_to_eval`; consequently, the number of entries should be the same between the values of these fields. See `task_config/pairings.jsonl` for examples of the format required.
+Note that we assume that "dialogue" consists of strictly alternating turns (e.g. speakers a, b, a, b, a...). Speakers that we would like to evaluate in the dialogues of `dialogue_dicts` should appear in the same order as `speakers_to_eval`. (Consequently, the number of dialogues in `dialogues_dicts` will be the same as the number of speakers in `speakers_to_eval`.) See `task_config/pairings.jsonl` for examples of the format required.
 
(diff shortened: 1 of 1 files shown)


parlai/crowdsourcing/tasks/acute_eval/README.md
```diff
@@ -62,6 +64,8 @@ This is a template of the expected format with the minimal expected fields:
       ]
     }
 
+Note that we assume that "dialogue" consists of strictly alternating turns (e.g. speakers a, b, a, b, a...). Speakers that we would like to evaluate in the dialogues of `dialogue_dicts` should appear in the same order as `speakers_to_eval`; consequently, the number of entries should be the same between the values of these fields. See `task_config/pairings.jsonl` for examples of the format required.
```
EricMichaelSmith, line 67: Hmm - could you clarify "the number of entries should be the same between the values of these fields"? i.e. each turn should have the same number of speakers?

EricMichaelSmith (approved):
Great, thanks for the clarification/reordering! Minor comment

parlai/crowdsourcing/tasks/acute_eval/README.md
moyapchen: Hmm, how about "the number of dialogues in `dialogues_dicts` should be the same as the number of speakers in `speakers_to_eval`? 

Was trying to figure out a way to make 

> Additionally, `speakers_to_eval` must be in the same order as the dialogue_dicts.

a bit more clear relative to

```
                eval_speakers = [
                    s
                    for d in convo_pair["dialogue_dicts"]
                    for s in d["speakers"]
                    if s in convo_pair["speakers_to_eval"]
                ]
                # make sure order is preserved
                assert eval_speakers == convo_pair["speakers_to_eval"]
              
```

though not sure if the wording I settled on was much better, lol.

EricMichaelSmith: Yeah, I think that's good! Or maybe even "Consequently, each speaker listed in `speakers_to_eval` will correspond to a different dialogue in `dialogue_dicts`", whichever you prefer

PR #3346, opened 2021-01-05
Increase wait time for parlai.crowdsourcing check

EricMichaelSmith:
**Patch description**
Occasionally the model-chat test will fail due to the Blender 90M bot agent taking surprisingly long to respond to our fake human (maybe ~20 s per message in the case of https://app.circleci.com/pipelines/github/facebookresearch/ParlAI/8282/workflows/b59066f6-ddae-489d-89f5-922970f7989b/jobs/67149/steps ). If the bot agent keeps taking arbitrarily long to respond, we will have to find another solution, but we can perhaps fix the flakiness of the test by giving the conversation 2.5 minutes to complete, up from 1 minute currently

**Testing steps**
`pytest tests/crowdsourcing/tasks/model_chat/test_model_chat.py::TestModelChat::test_base_task`

### DIFF
--- a/parlai/crowdsourcing/utils/tests.py
+++ b/parlai/crowdsourcing/utils/tests.py
@@ -298,8 +298,8 @@ class AbstractParlAIChatTest(AbstractCrowdsourcingTest):
         # # Check that the inputs and outputs are as expected
 
         # Wait until all messages have arrived
-        wait_time = 1  # In seconds
-        max_num_tries = 60  # max_num_tries * wait_time is the max time to wait
+        wait_time = 5.0  # In seconds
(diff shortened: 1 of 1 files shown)


JackUrb (approved):
Looks good to me - if this doesn't resolve in the long term we'll definitely need to examine why the model is slow on the test server sometime.

stephenroller:
Does it have to be BB? If any agent will do, we have a "fixed_response" agent that is particularly good for tests.

EricMichaelSmith:
> Does it have to be BB? If any agent will do, we have a "fixed_response" agent that is particularly good for tests.

Hmm, I like having it be BB because then you get a confirmation that the response is contextually appropriate, but if these tests get way too long then maybe we could revisit using a faster model

PR #3348, opened 2021-01-06
[parlaiparser] Fix bad parse of `-opt` in python 3.8

stephenroller:
**Patch description**
In Python 3.8, `-opt adam` is misparsed initially as `-o pt` and causes a FileNotFoundError. This hack works around this.

**Testing steps**
New test fails before this patch, passes after. Manual testing.

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -875,7 +875,12 @@ class ParlaiParser(argparse.ArgumentParser):
         parsed = vars(self.parse_known_args(args, nohelp=True)[0])
         # Also load extra args options if a file is given.
         if parsed.get('init_opt') is not None:
-            self._load_known_opts(parsed.get('init_opt'), parsed, first_pass=True)
+            try:
+                self._load_known_opts(parsed.get('init_opt'), parsed)
+            except FileNotFoundError:
+                # don't die if -o isn't found here. See comment in second call
+                # later on
+                pass
         parsed = self._infer_datapath(parsed)
 
         # find which image mode specified if any, and add additional arguments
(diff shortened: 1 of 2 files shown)


parlai/core/params.py
```diff
@@ -875,7 +875,12 @@ def add_extra_args(self, args=None):
         parsed = vars(self.parse_known_args(args, nohelp=True)[0])
         # Also load extra args options if a file is given.
         if parsed.get('init_opt') is not None:
-            self._load_known_opts(parsed.get('init_opt'), parsed)
+            try:
+                self._load_known_opts(parsed.get('init_opt'), parsed)
+            except FileNotFoundError:
+                # don't die if -o isn't found here. See comment in second call
```
EricMichaelSmith, line 881: hmm is it worth adding a comment for why a FileNotFoundError would be raised if the string gets parsed as `-o`?

parlai/core/params.py
```diff
@@ -904,6 +909,13 @@ def add_extra_args(self, args=None):
                 parsed.get('selfchat_task', False),
             )
 
+        # reparse args now that we've inferred some things.  specifically helps
+        # with a misparse of `-opt` as `-o pt` which may be ambigiuous until
+        # after we have added the model args.
```
EricMichaelSmith, line 914: why does adding the model args help with not misparsing?

EricMichaelSmith (approved):
Sounds reasonable! But added two small comments regarding clarification

parlai/core/params.py
stephenroller: Because before model args are added -opt isn't an argument

PR #3354, opened 2021-01-06
Make convenience script for dumping out chat to conversations format

moyapchen:
Suggested by @stephenroller in D25544140

**Testing steps**
```
    python parlai/scripts/dump_data_to_conversations.py -t google_sgd --save-world-logs True --report-filename hi
```
has output as expected


parlai/scripts/dump_data_to_conversations.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Alias script around `parlai eval_model -m repeat_label`  to dump data to conversations
```
jaseweston, line 7: i think this naming needs to be clearer, because 'dump data' sounds like converting the format of a task, not involving a model (as in eval model).

stephenroller: That's what it seems to be doing though...

stephenroller: Convert is more in line with our other script though.

jaseweston: oh i see.. the eval_model threw me off.  yes, the naming doesn't match the other script that does the parlai_format (https://github.com/facebookresearch/ParlAI/blob/master/parlai/scripts/convert_data_to_parlai_format.py) 

question: .. is conversations format something different to the json format (https://github.com/facebookresearch/ParlAI/tree/master/parlai/tasks/jsonfile) ?  (I've always wanted a converter for that too ..)

moyapchen: 'extract_task_to_conversations_format'? 

I have zero preferences on naming.

stephenroller: @jaseweston it's the same. Probably consistent naming is worthwhile here (convert_to_json_format)

I agree, let's hide this implementation detail from the docstring where it might confuse a user, and delegate that to comments of implementation. The docstring should discuss usage.

We need Kaushik to implement to the world logger changed

jaseweston: snds good

stephenroller:
I've asked @kauterry to do the changes to world logger that are a blocker for landing this (distinct filename outputs without mandating reports)

parlai/scripts/dump_data_to_conversations.py
kauterry: The world logger fix is pushed, it's no longer blocking.

kauterry:
The world logger fix is pushed, it's no longer blocking.

stephenroller:
Can we finish this?

stephenroller (approved):

PR #3355, opened 2021-01-06
Fix second flaky crowdsourcing test

EricMichaelSmith:
**Patch description**
Occasionally a crowdsourcing test will fail because a mock worker is registered but then cannot be found, as in https://app.circleci.com/pipelines/github/facebookresearch/ParlAI/8306/workflows/ee4a4ec5-4187-479c-b079-1152972579d2/jobs/67325/tests. This fix utilizes the existing try/except block to catch this source of flakiness as well

**Testing steps**
Crowdsourcing CI check


JackUrb (approved):
Seems like a good fix to me. Eventually if these keep coming up we can move to a testing utility function, but for now this gets the tests working.

PR #3356, opened 2021-01-06
Add user response option to Turn Annotations Static Task (new)

meganung:
This PR adds the option to have a response field textbox for each utterance so we can collect the user’s response (chat history is not updated with the user’s response).
- Validation of the user’s text input: requires more than 10 characters, at least 1 vowel, and at least 2 words

Annotation buckets can also be optional, if no annotation buckets file path is passed in there will be no annotation buckets.

Screen Shot 2021-01-04 at 2 35 09 PM

(take 2 of previous PR 3341)


EricMichaelSmith (approved):
Looks good, thanks for getting the tests working :)

PR #3363, opened 2021-01-08
Fix more flaky crowdsourcing tests

EricMichaelSmith:
**Patch description**
This PR attempts to correct two flaky crowdsourcing test runs from today:
- https://app.circleci.com/pipelines/github/facebookresearch/ParlAI/8342/workflows/7fc190f5-6ba1-460e-b451-cc8f318d8690/jobs/67720/tests
- https://app.circleci.com/pipelines/github/facebookresearch/ParlAI/8352/workflows/8407a260-67ba-4dcd-a2d6-7a3413da297b/jobs/67820/steps

The first of these fails because a registered worker agent could not be found, and so I've tried to address this by re-trying to register the agent if the agent could not be found the first time. The second of these fails due to slowness with BlenderBot90M on the CI box; I've remedied this by switching to `-m fixed_response` to speed up model response.

In order to enable human+model chat with the `fixed_response` model agent, I've also removed the ability to set model parameters for human+model chat by specifying a common model folder in which all models reside, and instead I've changed this so that the only possible way to do any kind of human+model chat is to pass in a YAML file of models and ParlAI flags. This makes the method of specifying models identical for human+model chat both with and without images.

**Testing steps**
`tests/crowdsourcing/tasks/model_chat/test_model_chat.py`


parlai/crowdsourcing/utils/tests.py
```diff
@@ -133,22 +133,30 @@ def _register_mock_agents(self, num_agents: int = 1) -> List[str]:
...
 
-            # Register the worker
             mock_worker_name = f"MOCK_WORKER_{idx:d}"
             max_num_tries = 6
             initial_wait_time = 0.5  # In seconds
             num_tries = 0
             wait_time = initial_wait_time
             while num_tries < max_num_tries:
                 try:
+
```
JackUrb, line 143: Unrelated, but you can get rid of mypy errors in this file by providing typing for self, like:
```python
def _check_output_key(self: Union[<testcase class you're mixing in with>, 'AbstractParlAIChatTest'], key: str, actual_value: Any, expected_value: Any):
```

JackUrb (approved):
I think both the yaml changes and the flakey bugfix change are great to have in! (Though they may have been better as separate PRs, as here it could be seen as an operational change to usage of `model_chat` being hidden in a PR about flakey tests).

In any case, LGTM and follows what was discussed in the ParlAI meeting this week.

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
```diff
@@ -445,10 +446,14 @@ def __init__(
...
-        _ = args  # Not needed
-        models_needed = list(self.conversations_needed.keys())
-        active_models = [m for m in models_needed if self.conversations_needed[m] > 0]
-        return TurkLikeAgent.get_bot_agents(args=args, active_models=active_models)
+        with open(args.blueprint.model_opt_path) as f:
+            all_model_opts = yaml.load(f.read())
+        active_model_opts = {
+            model: opt
+            for model, opt in all_model_opts.items()
+            if self.conversations_needed[model] > 0
```
klshuster, line 454: oh great i just suggested this in the form

klshuster (approved):
cool, thanks for looping in the model_opts change as well!

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
EricMichaelSmith: Yeah - your suggestion still applies to image chat, though, for which `self.conversations_needed` isn't used

parlai/crowdsourcing/utils/tests.py
EricMichaelSmith: Good point - changed

PR #3372, opened 2021-01-12
Diagnose crowdsourcing test flakiness

EricMichaelSmith:
**Patch description**
ParlAI chat task crowdsourcing tests have been flaky again, but in a way that's hard to diagnose. This PR attempts to make understanding test flakiness easier:
- If the actual and expected number of messages don't match, print all actual messages to see what's going wrong
- If a key in the state dict has the wrong value, print the name of the key, what the value should be, and what it actually is. (I'd like to switch these tests over to pytest regressions at some point, but that's a much bigger refactor.)

**Testing steps**
crowdsourcing CI tests


JackUrb (approved):
Getting to the root of this will definitely make writing these tests cleaner in the future!

PR #3373, opened 2021-01-12
How to switch Worlds

EricMichaelSmith:
**Patch description**
Add notes to the READMEs of 2 ParlAI chat tasks describing how to specify a custom chat World and onboarding World independently of setting a new Blueprint. This is useful for demonstrating the modularity of the Mephisto and `parlai.crowdsourcing` code 

**Testing steps**
(NA - only README changes)


JackUrb (approved):
Great to have expanded documentation on this!

As an aside, this can also be set by directly passing the imported module to `SharedParlAITaskState.world_module` in the run script, preventing the need to extract paths, but removing the ability to set the module directly from the command line or config yamls.

PR #3380, opened 2021-01-14
Fixing world logger filename issue

kauterry:
**Patch description**
We now take an argument "world-logs" which takes a string indicating the file path to save the world logs. We scrap the "save-world-logs" boolean argument.


stephenroller (approved):
LGTM, thanks! Next time, keep these external PR descriptions/testing steps as full as the internal ones, and link to the internal diff when there is one.

Issue #3367, opened 2021-01-08
Dynamic Batching x Chunk Teacher x Distributed Training hangs during evaluation


emilydinan:
In certain settings Dynamic Batching x Chunk Teacher x Distributed Training hangs during evaluation. Hanging happens before first eval step. Cannot reproduce with <=16 gpus and without dynamic batching.

stephenroller:
Suspicion is buffer doesn't fill on such wide distribution

emilydinan:
That is my suspicion as well, although it has worked in the past with a smaller number of validation examples.

PR #3382, opened 2021-01-14
Temporary bug fix: chunk teacher x distributed training x dynamic batching

emilydinan:
**Patch description**
See #3367. This is a temporary fix to unblock folks using this setting. Since the bug currently cannot be reproduced with fewer than 16 gpus during distributed training, it will take some time to debug.

**Testing steps**
```
parlai tm -t babi -m transformer/generator -bs 20 --dynamic-batching full --eval-dynamic-batching off -mf /tmp/testdynbatch --truncate 30 -vtim 5 -vme 100
```

### DIFF
--- a/parlai/scripts/train_model.py
+++ b/parlai/scripts/train_model.py
@@ -84,10 +84,10 @@ def setup_args(parser=None) -> ParlaiParser:
         type='nonestr',
         choices={None, 'off', 'full', 'batchsort'},
         help=(
-        	'Set dynamic batching at evaluation time. Set to off for '
-	        'train-only dynamic batching. Set to none (default) to use same '
-	        'setting as --dynamic-batching.'
-	    ),
+            'Set dynamic batching at evaluation time. Set to off for '
+            'train-only dynamic batching. Set to none (default) to use same '
+            'setting as --dynamic-batching.'
+        ),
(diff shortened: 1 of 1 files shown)


parlai/scripts/train_model.py
```diff
@@ -78,6 +78,14 @@ def setup_args(parser=None) -> ParlaiParser:
         hidden=True,
         help='Eval time batch size (defaults to same as -bs)',
     )
+    train.add_argument(
+        '--eval-dynamic-batching',  # FIXME: see https://github.com/facebookresearch/ParlAI/issues/3367
+        default=None,
+        type='nonestr',
+        choices={None, 'off', 'full', 'batchsort'},
+        help='Temporary fix for an issue tracked in #3367. Eval dynamic batching; '
+        'if None, defaults to whatever is set in --dynamic-batching. If "off", sets dynamic batching to None',
```
stephenroller, line 87:
```suggestion
        help=(
        	'Set dynamic batching at evaluation time. Set to off for '
	        'train-only dynamic batching. Set to none (default) to use same '
	        'setting as --dynamic-batching.'
	    ),
```

stephenroller (approved):
Help string suggestion

PR #3386, opened 2021-01-15
Check chat data folder validity

EricMichaelSmith:
**Patch description**
Ensure that a ~ doesn't exist in the model chat chat_data_folder param, because it currently can't be parsed

**Testing steps**
Existing CI checks

### DIFF
--- a/parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
+++ b/parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
@@ -188,6 +188,17 @@ class BaseModelChatBlueprint(ParlAIChatBlueprint, ABC):
             full_path
         ), f"Target left pane text path {full_path} doesn't exist"
 
+        if args.blueprint.get("chat_data_folder") == '':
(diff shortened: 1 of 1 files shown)


parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
```diff
@@ -188,6 +188,15 @@ def assert_task_args(
             full_path
         ), f"Target left pane text path {full_path} doesn't exist"
 
+        if args.blueprint.get("chat_data_folder") == '':
+            raise ValueError('Must provide a valid chat data folder')
+        else:
```
klshuster, line 193: you probably dont need the `else` here

klshuster (approved):
thanks for the quick fix!

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
EricMichaelSmith: fair - changed

PR #3387, opened 2021-01-15
Fix analysis arg help

EricMichaelSmith:
**Patch description**
For the model chat task, save the results to a date folder (i.e. `2021_01_15`) inside the root folder, so that the user doesn't have to create a date folder manually, since a date folder is expected by the analysis code. Also clean up a misleading arg help string and document the use of the arg

**Testing steps**
CI checks have good coverage of the change


stephenroller (approved):
I see no issues

klshuster (approved):
thanks!

PR #3397, opened 2021-01-19
Update turn_annotations_static frontend to use React state instead of global variables/document accessors

spencerp:
**Patch description**
Updates the turn_annotations_static UI so that state is stored in MainTaskComponent instead of a global variable.
#3396

**Testing steps**
Used some test data and verified that the box description and submit button updated correctly:
Screen Shot 2021-01-19 at 12 00 26 PM
Screen Shot 2021-01-19 at 12 49 30 PM

### DIFF
--- a/parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
+++ b/parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
@@ -6,28 +6,10 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
- // NOTE: this frontend uses document accessors rather than React to control state,
- // and may not be compatible with some future Mephisto features
-
 import React from "react";
 import { ErrorBoundary } from './error_boundary.jsx';
 import { Checkboxes } from './checkboxes.jsx';
 import { FormControl } from 'react-bootstrap';
-
-// HACK global variable (fix at some point)
-// Array of arrays (by subtask and then by turn)
-window.workerAnswers = [];
-
-var showDisableCssNextButton = function () {
-  // Have to disable the Next button in CSS b/c if React doesn't allow you to do
-  // it in JS Note: person can actually click button even though it doesn't
-  // look like they can! HACK: This is a hack; should tie it to state really
-  document.getElementById('submit-button').style = 'cursor:not-allowed;background-color: #c8c8c8 !important;';
-}
-
-var showEnabledCssNextButton = function () {
-  document.getElementById('submit-button').style = '';
-}
 
 var validateFreetextResponse = function (response, charMin, wordMin, vowelMin) {
   // Requires the response to contain at least charMin characters, wordMin words, and vowelMin vowels.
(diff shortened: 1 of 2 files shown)


parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
```diff
@@ -184,125 +49,352 @@ function SubtaskCounter({ subtaskIndex, numSubtasks }) {
...
+
+    if (subtaskIndex == this.props.taskData.length - 1) {
+      this.props.onSubmit(this.state.answers);
+    }
+    this.setState(prevState => ({index: prevState.index + 1}));
+  };
+
+  setChecked = (turnIdx, bucket, value) => {
+    var answers = this.state.answers;
+    answers[this.state.index][turnIdx].buckets[bucket] = value;
```
meganung, line 345: The bucket value is null if it isn't checked, would be nice if you could explicitly set the answer data to true if value is true and false if value is null/false.

parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
```diff
@@ -184,125 +49,352 @@ function SubtaskCounter({ subtaskIndex, numSubtasks }) {
...
+          display: 'table',
+          minWidth: `${agentIdx == 0 ? '30%' : '80%'}`,
+          marginTop: `${turnIdx == 1 ? '40px' : 'auto'}`,
+        }}>
+        <span>
+          <b>{turnIdx % 2 == 0 ? 'YOU' : 'THEM'}:</b> {turnAnswers.text}
+          <ErrorBoundary>{extraElements}</ErrorBoundary>
         </span>
       </div>
-      {responseInputElement}
```
meganung, line 242: this line / the divs around it needs to be ported over so the `responseInputElement` aka the freeform textbox will show up!

parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
```diff
@@ -184,125 +49,352 @@ function SubtaskCounter({ subtaskIndex, numSubtasks }) {
...
+        var answersForTurn = {
+          text: subTaskData[j].text,
+          agent_idx: subTaskData[j].agent_idx,
+          other_metadata: subTaskData[j].other_metadata,
+          buckets: {},
+          do_annotate: subTaskData[j].do_annotate,
+          input_reason: null,
+          response: null,
+        };
+        Object.keys(this.props.taskConfig.annotation_buckets.config).forEach(
```
meganung, line 280: Null check around this since annotation_buckets field is optional so it can be null.

parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
```diff
@@ -184,125 +49,352 @@ function SubtaskCounter({ subtaskIndex, numSubtasks }) {
...
+          type="text"
+          name="input_response"
+          id={"input_response_" + turnIdx}
+          style={{
+              fontSize: "14px",
+              resize: "none",
+              marginBottom: "40px"
+          }}
+          onChange={this.handleResponseFieldChange}
+          value={turnAnswers.response}
```
meganung, line 159: nit- there's a warning that `'value' prop on 'input' should not be null. Consider using an empty string` when you initialize the `response` variable set it to "" instead maybe. (also might want to check `input_reason` variable)

parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
```diff
@@ -184,125 +49,352 @@ function SubtaskCounter({ subtaskIndex, numSubtasks }) {
...
+//   taskData: TaskData,
+//   taskTitle: string,
+//   taskDescription: string,
+//   taskConfig: TaskConfig,
+//   onSubmit: WorkerAnswers => void,
+// };
+// type MainTaskComponentState = {
+//   index: number,
+//   answers: WorkerAnswers,
+// };
```
meganung, line 254: this is super helpful!

meganung (changes requested):
I think you need to update the onboarding_components.jsx as well to use react state and pass in the necessary props to the checkboxes component since in `checkboxes.jsx` the component is expecting certain props to be passed in such as `turnAnswers`.

parlai/crowdsourcing/tasks/turn_annotations_static/webapp/src/components/task_components.jsx
```diff
@@ -184,125 +49,352 @@ function SubtaskCounter({ subtaskIndex, numSubtasks }) {
...
-    if (annotationBuckets !== null) { 
-      extraElements = (<span key={'extra_' + turnIdx}><br /><br />
-        <span style={{ fontStyle: 'italic' }} >
-          <span dangerouslySetInnerHTML={{ __html: annotationQuestion }}></span>
-          <br />
-          <Checkboxes turnIdx={turnIdx} annotationBuckets={annotationBuckets} askReason={askReason} onUserInputUpdate={onUserInputUpdate} />
-        </span>
-      </span>)
-    }
-    if (responseField !== null) {
```
meganung, line 214: `responseField` is true/false so we should also condition on if it is true. (I think this is a bug from before)

PR #3400, opened 2021-01-19
Save model chat rating in output file

EricMichaelSmith:
**Patch description**
Currently, the model chat crowdsourcing task does not save the final rating to the summary output file (which typically has a name like `"20210119_161519_102_sandbox.json"`). This PR remedies that, and then modifies the analysis script to work with the slightly changed way in which the final rating is now saved. The logic for adding the final set of turn annotations has also been cleaned up.

**Testing steps**
Crowdsourcing CI checks for model chat: regular model chat, model image chat, and model chat analysis. Also ran sandbox HIT and confirmed that the rating is now saved in the output file


klshuster (approved):
awesome, thanks!

PR #3405, opened 2021-01-25
Release BST human annotations

### LINKED ISSUE #3338: BST dataset: where is 'mode' for each utterances?
Hi, I have a question about BST dataset (https://arxiv.org/pdf/2004.08449.pdf).
스크린샷 2021-01-03 오후 9 48 39

While the paper and Fig. 1 tell us that each utterance are annotated with the 'mode' (e.g., _PB,K,E,S_),
 I cannot find it in raw BST dataset (i.e., `train.json, valid.json, test.json`) unfortunately :(

Without it, one cannot even reproduce **MT Two-Stage model** in the paper because the model is based on a dialog manager that classifies the mode given utterance.

Is the 'mode' only used for private perhaps?

Thanks.

EricMichaelSmith:
**Patch description**
Open-sources the human annotations of the BST validation set, and adds them to the set of files downloaded when using a BST teacher. Fixes #3338 

**Testing steps**
`parlai dd -t blended_skill_talk` to re-download the BST files


stephenroller (approved):

stephenroller:
please add a Fixes #number in your description to refer to the original issue

PR #3413, opened 2021-01-26
[TFIDF] Remove duplicate arg, add cosine similarity utility function

spencerp:
**Patch description**
I found cosine similarity useful in a project, so I thought I'd add it here!

**Testing steps**
Set pdb trace and tested in interactive:
```
>>> q = tfidf_ranker.text2spvec(label)
>>> c = tfidf_ranker.text2spvec(sentence)
>>> cosine_similarity(q, c)
0.5348931739406508
```

### DIFF
--- a/parlai/agents/tfidf_retriever/tfidf_retriever.py
+++ b/parlai/agents/tfidf_retriever/tfidf_retriever.py
@@ -82,12 +82,6 @@ class TfidfRetrieverAgent(Agent):
             help='How many docs to retrieve.',
         )
         parser.add_argument(
-            '--retriever-mode',
-            choices=['keys', 'values'],
-            default='values',
-            help='Whether to retrieve the stored key or the stored value.',
-        )
-        parser.add_argument(
             '--remove-title',
             type='bool',
             default=False,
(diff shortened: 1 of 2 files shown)


parlai/agents/tfidf_retriever/tfidf_retriever.py
```diff
@@ -81,12 +81,6 @@ def add_cmdline_args(
             type=int,
             help='How many docs to retrieve.',
         )
-        parser.add_argument(
-            '--retriever-mode',
```
klshuster, line 85: why are we deleting this key?

klshuster (approved):
cool, lgtm

klshuster (changes requested):
actually want to clarify the key removal prior to merge

parlai/agents/tfidf_retriever/tfidf_retriever.py
spencerp: It's a repeat of the arg on line 96 (90 in the new file)

klshuster (approved):

parlai/agents/tfidf_retriever/tfidf_retriever.py
klshuster: good catch!

spencerp:
Checked `unittests_gpu16` and it doesn't appear to be related to these changes, so ignoring for the scope of this PR.

PR #3419, opened 2021-01-26
World logger outputs metrics.

stephenroller:
**Patch description**
Ensures that "metrics" field is not dropped from the world logger outputs.

**Testing steps**
Manual testing. CI


kauterry (approved):
Looks good.

PR #3420, opened 2021-01-27
Delete some dead code from teachers

stephenroller:
**Patch description**
Locks no longer make sense. They were a hogwild thing, and they don't apply to Chunk teacher. Kill the dead code!

**Testing steps**
CI


EricMichaelSmith (approved):
Yeah sounds good!

PR #3422, opened 2021-01-28
Abstract class for compiling crowdsourcing data with MephistoDataBrowser

mojtaba-komeili:
**Patch description**
Added a general function and an arg to `AbstractResultsCompiler` class. They can be used by all the crowdsourcing tasks that use Mephisto.

**Testing steps**
Have a private class that is derived from `AbstractResultsCompiler`. I can load the Mephisto task data from it.


EricMichaelSmith:
Note: from discussions, this new flag will get added to a new `AbstractDataBrowserResultsCompiler` subclass, to be created after https://github.com/facebookresearch/ParlAI/pull/3423 is merged in

stephenroller:
Nit: better PR title

parlai/crowdsourcing/utils/analysis.py
```diff
@@ -79,3 +82,65 @@ def __init__(self, opt: Dict[str, Any]):
             raise ValueError(
                 'There must be a "none_all_good" category in self.problem_buckets!'
             )
+
+    @abstractmethod
+    def compile_results(self) -> pd.DataFrame:
```
mojtaba-komeili, line 87: I missed this method by mistake while merging. I will remove this as it is already in the parent class.

parlai/crowdsourcing/utils/analysis.py
```diff
@@ -79,3 +82,65 @@ def __init__(self, opt: Dict[str, Any]):
...
+        return db.get_worker(worker_id)["worker_name"]
+
+    def get_task_units(self, task_name: str) -> List[Unit]:
+        """
+        Retrieves the list of work units from the Mephisto task.
+        """
+        data_browser = self.get_mephisto_data_browser()
+        return data_browser.get_units_for_task_name(task_name)
+
+    def get_units_data(self, task_units: List[Unit]) -> List[Dict]:
```
EricMichaelSmith, line 132: supernit: you can also do `List[dict]` here

parlai/crowdsourcing/utils/analysis.py
```diff
@@ -79,3 +82,65 @@ def __init__(self, opt: Dict[str, Any]):
...
+        """
+        Retrieves task data for a lits of Mephisto task units.
```
EricMichaelSmith, line 134: nit: typo in "list"

parlai/crowdsourcing/utils/analysis.py
```diff
@@ -79,3 +82,65 @@ def __init__(self, opt: Dict[str, Any]):
...
+    @abstractmethod
+    def compile_results(self) -> pd.DataFrame:
+        """
+        Method for returning the final results dataframe.
+        """
+
+
+class AbstractDataBrowserResultsCompiler(AbstractResultsCompiler):
+    """
+    Results compiler subclass to provide utility code for crowdsourcing tasks.
```
EricMichaelSmith, line 89: nit: would be good to say something specifically about the fact that this is a DataBrowser-based subclass in the first line of the docstring, because otherwise this first line on its own doesn't say anything about what the subclass is

EricMichaelSmith (approved):
Yeah, looks good, once that extra `compile_results()` method in `AbstractTurnAnnotationResultsCompiler` is removed - it's great to have this! But it looks like the unittests_gpu16 and unittests_37 CI checks are failing - it'd be good to figure out what's going wrong there, by SSHing into the CI box and re-running the failing command. Ping me if you'd like more info on how to do this

mojtaba-komeili:
I figured out the reason for test crashing. It is because of Mephisto imports in `analysis.py`. Since this happens both for `unittests_gpu16` and `unittests_37`, should I add something like this to the `.cricleci/config.yml` for both of them?
```
          more_installs:
            - installcrowdsourcingdeps
            - ...
```

Or should I import Mephisto as the part of the function?

parlai/crowdsourcing/utils/analysis.py
```diff
@@ -79,3 +78,69 @@ def __init__(self, opt: Dict[str, Any]):
...
             )
+
+
+# Defining the class only if Mephisto is installed, since it relies on Mephisto
+try:
+    from mephisto.abstractions.databases.local_database import LocalMephistoDB
+    from mephisto.data_model.unit import Unit
+    from mephisto.tools.data_browser import DataBrowser as MephistoDataBrowser
+
+    class AbstractDataBrowserResultsCompiler(AbstractResultsCompiler):
```
stephenroller, line 89: if you drop the type annotations, or switch them to strings, the class definition doesn't need to be in the try

stephenroller: Let's discuss structure before we merge this.

stephenroller: Actually I think Python 3.7 solves it pretty easy for us. The following works fine:

```python
#!/usr/bin/env python3

try:
    from nonexistant import Nonimported
except ImportError:
    pass

class MyClass:
    def foo(self) -> Nonimported:
        pass
```

PR #3423, opened 2021-01-28
Turn-annotations results compiler class

EricMichaelSmith:
**Patch description**
Split off all attributes in `AbstractResultsCompiler` that are specific to turn annotations tasks into a new `AbstractTurnAnnotationResultsCompiler` subclass. This way, code that compiles results that *aren't* based on turn annotations can subclass `AbstractResultsCompiler` without inheriting turn-annotations attributes.

One use case of this will be to create a new subclass of `AbstractResultsCompiler` called `AbstractDataBrowserResultsCompiler` that will provide utility code for pulling results directly from Mephisto's `DataBrowser`. (Eventually, it'll be good to convert all existing analysis code to pull from `DataBrowser`, since this is the canonical way of reading in results.) 

**Testing steps**
Existing CI checks provide test coverage of analysis scripts

### DIFF
--- a/parlai/crowdsourcing/utils/analysis.py
+++ b/parlai/crowdsourcing/utils/analysis.py
@@ -30,9 +30,9 @@ class AbstractResultsCompiler(ABC):
         self.output_folder = opt.get('output_folder')
 
     @abstractmethod
-    def compile_results(self) -> pd.DataFrame:
+    def compile_results(self) -> Any:
         """
-        Method for returning the final results dataframe.
+        Method for returning the final results object.
         """
 
 
@@ -77,3 +77,11 @@ class AbstractTurnAnnotationResultsCompiler(AbstractResultsCompiler):
             raise ValueError(
                 'There must be a "none_all_good" category in self.problem_buckets!'
             )
+
+    @abstractmethod
+    def compile_results(self) -> pd.DataFrame:
+        """
+        Method for returning the final results dataframe.
+
+        Each row of the dataframe consists of one utterance of one conversation.
+        """
--- a/parlai/crowdsourcing/utils/analysis.py
+++ b/parlai/crowdsourcing/utils/analysis.py
@@ -22,10 +22,34 @@ class AbstractResultsCompiler(ABC):
     def setup_args(cls):
         parser = argparse.ArgumentParser('Compile crowdsourcing results')
         parser.add_argument(
-            '--results-folders', type=str, help='Comma-separated list of result folders'
-        )
-        parser.add_argument(
             '--output-folder', type=str, help='Folder to save output files to'
+        )
+        return parser
+
+    def __init__(self, opt: Dict[str, Any]):
+        self.output_folder = opt.get('output_folder')
+
+    @abstractmethod
+    def compile_results(self) -> pd.DataFrame:
+        """
+        Method for returning the final results dataframe.
+        """
+
+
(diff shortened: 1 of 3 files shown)


parlai/crowdsourcing/utils/analysis.py
```diff
@@ -22,10 +22,34 @@ class AbstractResultsCompiler(ABC):
...
-            '--results-folders', type=str, help='Comma-separated list of result folders'
+            '--output-folder', type=str, help='Folder to save output files to'
         )
+        return parser
+
+    def __init__(self, opt: Dict[str, Any]):
+        self.output_folder = opt.get('output_folder')
+
+    @abstractmethod
+    def compile_results(self) -> pd.DataFrame:
```
mojtaba-komeili, line 33: I'm not sure if DataFrame is the format of choice for all the data.
I think mine will be json for example.

mojtaba-komeili (changes requested):
Make `compile_results` function suitable for other data formats.

parlai/crowdsourcing/utils/analysis.py
EricMichaelSmith: ah good to know - what will the format of your JSON be? The format of the output of `compile_results` is implicitly expected to be one utterance per row of the dataframe - I'm making a change and adding a comment about this

mojtaba-komeili: For me it could be the context that is provided for that utterance as well. Actually I can have that as a df too, but I still rather having it as a JSON. Might be the case for some other people too.

EricMichaelSmith: Yeah, the context also gets a row in the output dataframe format. What is it in particular that's helpful if you have a JSON format and not a df format? (Presumably you have a structured output that's hard to represent as a dataframe?)

mojtaba-komeili: Not really, I just think that look better and more portable, and readable.

mojtaba-komeili: Lost about `Union[pd.DataFrame, List]`?

EricMichaelSmith: Hmm, if we want to support multiple output types, though, `Union[pd.DataFrame, list]` might still not be permissive enough, because perhaps someone might want to save their output as a single dict instead of a list of dicts, for instance.

My preference would still be for us to have a single output type if possible, because that'll reduce complexity in the future of having to support multiple ways of representing the same thing. That personally matters to me more than whether the return type is a dataframe or JSON format, for instance. I'll bring this up during the meeting next Tuesday, however, because I'm curious where the team as a whole stands on this

mojtaba-komeili: Yeah, I agree with you. It is a good idea to decide on one to reduce the complexity and confusions.

mojtaba-komeili: Meanwhile, I agree with you on keeping it DF for the time being.

mojtaba-komeili (commented):
Keeping the compiled data format as DataFrame for now. But subject to change.

mojtaba-komeili (approved):

parlai/crowdsourcing/utils/analysis.py
EricMichaelSmith: > Meanwhile, I agree with you on keeping it DF for the time being.

Okay cool - merging this in once CI checks pass

PR #3424, opened 2021-01-29
Normalize distillation loss term

EricMichaelSmith:
**Patch description**
Normalize the embedding loss term for knowledge distillation so that it is invariant to the embedding dimension. Will make it easier to generalize finding the best loss coefficient ranges independent of model width

**Testing steps**
(Existing CI checks for distillation code)


stephenroller (approved):
Cool, interested to hear about the results

PR #3428, opened 2021-01-31
Use os.replace instead of os.rename for windows compatibility

vinnik-dmitry07:
**Patch description**
os.rename silently overwrites the destination in linux, but throws an error in windows. os.replace silently overwrites the desitnation in both.

**Testing**
Manual testing (by Stephen), + CI


stephenroller:
I think this is good, just wanted to check carefully. I'm learning something about the standard library today.

stephenroller (approved):
Thanks for fixing this.

PR #3430, opened 2021-02-01
Generalize projection-matrix-removal script

EricMichaelSmith:
When running TinyBERT-style knowledge distillation, projection matrices that are not used for evaluation will be stored in the model checkpoint. This PR generalizes the script used for removing these matrices to make it work on filesystems that do not support `shutil.copyfile()`.


EricMichaelSmith:
@stephenroller Changed to using `PathManager.copy()` (discussed offline)

klshuster (approved):

PR #3431, opened 2021-02-01
Add --no-backup option to vacuum.

stephenroller:
**Patch description**
Add a new option to `vacuum` which allows skipping of backups

**Testing steps**
Manual testing when cleaning up my experiments.

### DIFF
--- a/parlai/scripts/vacuum.py
+++ b/parlai/scripts/vacuum.py
@@ -36,6 +36,9 @@ class Vacuum(ParlaiScript):
             dest='path',
             help="Path to model file.",
         )
+        parser.add_argument(
+            '--no-backup', action='store_true', help="Do not create a backup."
+        )
         return parser
 
     def run(self):
@@ -50,8 +53,9 @@ class Vacuum(ParlaiScript):
             states = torch.load(
                 f, map_location=lambda cpu, _: cpu, pickle_module=parlai.utils.pickle
             )
-        logging.info(f"Backing up {model_file} to {model_file}.unvacuumed")
-        os.rename(model_file, model_file + ".unvacuumed")
+        if not opt['no_backup']:
+            logging.info(f"Backing up {model_file} to {model_file}.unvacuumed")
+            os.rename(model_file, model_file + ".unvacuumed")
         for key in [
             'optimizer',
             'optimizer_type',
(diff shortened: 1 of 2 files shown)


parlai/scripts/vacuum.py
```diff
@@ -53,7 +53,7 @@ def run(self):
             states = torch.load(
                 f, map_location=lambda cpu, _: cpu, pickle_module=parlai.utils.pickle
             )
-        if not opt['no_backup']:
+        if not self.opt['no_backup']:
```
mojtaba-komeili, line 56: Is `opt` safe here if there is no `no_backup`?

tests/test_other_scripts.py
```diff
@@ -113,6 +114,34 @@ def test_vacuum(self):
...
+                    'model': 'transformer/generator',
+                    'ffn_size': 8,
+                    'embedding_size': 8,
+                    'n_layers': 1,
+                }
+            )
+            size_before = os.stat(model_file).st_size
+            Vacuum.main(model_file=model_file, no_backup=True)
+            size_after = os.stat(model_file).st_size
+            assert size_after < size_before
```
mojtaba-komeili, line 142: nit: I usually prefer having that string after `assert` that says what went wrong. Although you can find it later since it is inside a test, and test report tells you what failed. But it might save you a couple of minutes to see it in the stderr. 
A diagnosis tip might help too.

mojtaba-komeili (approved):

tests/test_other_scripts.py
stephenroller: I agree but context is narrow and clear

tests/test_other_scripts.py
```diff
@@ -113,6 +114,34 @@ def test_vacuum(self):
...
+                    'ffn_size': 8,
+                    'embedding_size': 8,
+                    'n_layers': 1,
+                }
+            )
+            size_before = os.stat(model_file).st_size
+            Vacuum.main(model_file=model_file, no_backup=True)
+            size_after = os.stat(model_file).st_size
+            assert size_after < size_before, "Model file did not shrink after vacuum"
+            assert not os.path.exists(model_file + '.unvacuumed')
```
mojtaba-komeili, line 143: nit: a note on this one as well would be great

mojtaba-komeili (approved):

PR #3437, opened 2021-02-04
Fix for OverflowError: out of range integral type conversion attempted in HF BPE

dexterju:
Haven't been here for a long time, feel good!

**Patch description**
The vec2txt function for hugging face bpe is broken by ```OverflowError: out of range integral type conversion attempted```
This error was casued by vector has a list of tensor instead of a list of int, which cased ``` if idx in self.bpe.special_tok_map``` logic to fail. 

**Testing steps**
Tested by running this on any blender dict stuff.


stephenroller (approved):
Thanks!

Wonder if we've got a slow cuda in here (transfering 1 index at a time instead of the entire array)

PR #3438, opened 2021-02-05
Fix valid data file location issue with Natural Questions dataset

moyapchen:
Happened to run

```
parlai display_data -t natural_questions -dt valid:stream
```

and ran into an issue about how it couldn't find the files. Digging into the built script and the output directory at `~/ParlAI/data/NaturalQuestions/valid`, I noticed the files were being stored at `nq-dev-00.jsonl` rather than `nq-valid-00.jsonl` as expected by the agent. This does a quick "move the files over" fix.

Test Plan:
Ran the build script, commenting out the steps to redownload + untar all the data. Verified that

```
parlai display_data -t natural_questions -dt valid:stream
```

was indeed successful.

### DIFF
--- a/parlai/tasks/natural_questions/build.py
+++ b/parlai/tasks/natural_questions/build.py
@@ -81,9 +81,11 @@ def _untar_dataset_files(dpath):
 
 def _move_valid_files_from_dev_to_valid(dpath):
     """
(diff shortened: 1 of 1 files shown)


mojtaba-komeili (approved):
I remember there were previous issues with that naming before. But I missed this one.
Thanks for finding and fixing it 🆒

PR #3440, opened 2021-02-09
Modernize natural_questions with add_cmdline_args; add flag for context

moyapchen:
I want to get rid of the context and figured I should add a new cmdline arg. Natural questions doesn't currently use cmdline arguments, so update it so that it does.

Keep the old versions around for backward compatability purposes.

Test Plan:

Variations of the flag with

```
parlai display_data -t natural_questions -dt valid:stream
```

ensuring that the output is reasonable.

### DIFF
--- a/parlai/tasks/natural_questions/agents.py
+++ b/parlai/tasks/natural_questions/agents.py
@@ -253,57 +253,5 @@ class NaturalQuestionsTeacher(ChunkTeacher):
         return message_dict
 
 
-class NaturalQuestionsTeacherLongAnswerHTML(NaturalQuestionsTeacher):
-    """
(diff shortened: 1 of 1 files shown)


parlai/tasks/natural_questions/agents.py
```diff
@@ -277,5 +305,5 @@ def __init__(self, opt, shared=None):
         super().__init__(opt, shared)
 
 
-class DefaultTeacher(NaturalQuestionsTeacherShortAnswer):
```
mojtaba-komeili, line 280: I am not using this old version. And I not sure if anyone else does. I guess it should be OK to remove, or add a warning for deprecation in like 2 month if someone imports it.

mojtaba-komeili (approved):

PR #3443, opened 2021-02-09
update deprecated api calls to fix warning in fp16 mem efficient

### LINKED ISSUE #3410: Fix warning in FP16 Memory efficient
Right now if you use `--fp16-impl mem_efficient`, you will get a warning from PyTorch about how an API signature has changed. Please find the deprecated call and upgrade it.

Manual testing will suffice.

meganung:
Old usage of `_add` (and `_addcmul`, `_addcdiv`) calls were deprecated so updated  to the new signature.

Fixes #3410.


stephenroller (approved):
Amazing, thank you!!

PR #3452, opened 2021-02-12
Ignore unimportant coverage lines.

stephenroller:
**Patch description**
Stuff in build.py for various teachers only gets called randomly, as we have caches for our data. This patch excludes them from our coverage, as it's up to the cache whether we hit them or not, preventing fluctuations in our coverage.

**Testing steps**
CI


klshuster (approved):
🤑

PR #3455, opened 2021-02-16
[TGA] Abstract TGA candidate ranking and fix ranking for BART

emilydinan:
**Patch description**
Abstract Torch Generator Agent candidate ranking ability to a helper function. 

Then, override this helper function in BART so that we may delete the score for the start token. 

**Testing steps**
for BART:
```
parlai em -t convai2 --rank-candidates True -m bart --skip-generation True
```
for non-BART:
```
parlai em -t convai2 --rank-candidates True -mf zoo:blender/blender_90M/model --skip-generation Tru
```

### DIFF
--- a/parlai/agents/bart/bart.py
+++ b/parlai/agents/bart/bart.py
@@ -244,9 +244,9 @@ class BartAgent(TransformerGeneratorAgent):
         """
         Rank eval label candidates.
 
-        Overridden from TorchGeneratorAgent because BART uses
-        both EOS-BOS tokens to start decoding. During scoring of
-        candidates, we must get rid of the score for the start token.
+        Overridden from TorchGeneratorAgent because BART uses both EOS-BOS tokens to
+        start decoding. During scoring of candidates, we must get rid of the score for
+        the start token.
         """
         cand_choices = []
         cand_choices_scores = []
--- a/parlai/agents/bart/bart.py
+++ b/parlai/agents/bart/bart.py
@@ -25,8 +25,6 @@ from parlai.core.message import Message
 from parlai.core.opt import Opt
 from parlai.core.params import ParlaiParser
 from parlai.core.torch_agent import History
-from parlai.core.torch_generator_agent import PPLMetric
-from parlai.core.metrics import AverageMetric
 from parlai.utils.typing import TShared
 from parlai.utils.io import PathManager
 from parlai.zoo.bart.build import download, CONVERSION_ARGS, BART_ARGS
--- a/parlai/agents/bart/bart.py
+++ b/parlai/agents/bart/bart.py
@@ -15,7 +15,6 @@ or `-mf zoo:bart/bart_large/model` to ensure correct dictionaries are saved.
 """
 import os
 import torch
-import torch.nn.functional as F
 from typing import Optional, Dict, Any
 
 from parlai.agents.bart.convert_fairseq_to_parlai import ConversionScript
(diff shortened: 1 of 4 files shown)


parlai/agents/bart/bart.py
```diff
@@ -238,3 +239,35 @@ def _construct_token_losses(self, labels, model_output):
...
+        cand_choices = []
+        cand_choices_scores = []
+        encoder_states = self.model.encoder(*self._encoder_input(batch))
+        for i in range(batchsize):
+            num_cands = len(batch.candidate_vecs[i])
+            enc = self.model.reorder_encoder_states(encoder_states, [i] * num_cands)
+            cands, _ = self._pad_tensor(batch.candidate_vecs[i])
+            scores, _ = self.model.decode_forced(enc, cands)
+            # ignore the score for the start token
+            scores = scores[:, 1:, :]
```
klshuster, line 260: is this the only line that is different?

I'm wondering now if we can solve all of these issues by just overriding `BartModel.output` to do this computation...

that would allow us to get rid of the duplicate `compute_loss` and `_construct_token_losses` functions as well

emilydinan: the other change is that it needs reshaping...

klshuster: it looks like you just took the `reshape` out of the `cross_entropy` call? i.e. that's already in the base call

emilydinan: not quite -- in the original function it has a call to `scores.view(num_cands * cands.size(1), -1)`, but this breaks with BART unless we call reshape. i think it would probably be fine to change the view call to reshape instead in TGA, but this is a slightly more expensive operation (copy instead of view). what do you think?

emilydinan: since people don't seem to use this functionality too often i'll just change

klshuster: yeah given that this is rarely, if ever, used, it'll be better to just make them both `reshape`

klshuster (approved):
🚀 awesome!

2 things:
1) nit: a couple of lint errors
2) Let's wait for BART tests to pass prior to merging. did you happen to run them locally?

emilydinan:
> 🚀 awesome!
> 
> 2 things:
> 
> 1. nit: a couple of lint errors
> 2. Let's wait for BART tests to pass prior to merging. did you happen to run them locally?

ran bart locally but looks like distillbart, looking into it

emilydinan:
@EricMichaelSmith  can you take a look at the changes to distillbart when you get a chance?

parlai/core/torch_generator_agent.py
```diff
@@ -864,6 +864,37 @@ def _add_generation_metrics(self, batch, preds):
         """
         pass
 
+    def _rank_eval_label_candidates(self, batch, batchsize):
```
stephenroller, line 867: Maybe this should be a public method?

stephenroller (approved):
It's nice to see the cleanup of BART. Did you verify an older checkpoint didn't have results change?

projects/anti_scaling/distillation.py
```diff
@@ -338,6 +341,9 @@ def _perform_forward_passes(self, batch: Batch) -> ForwardPassOutputs:
         mask = self._manipulate_mask(
```
EricMichaelSmith, line 341: nit: maybe `mask` should now be `score_mask` to differentiate it from `decoder_mask`?

EricMichaelSmith (approved):
Seems reasonable, minor comment

emilydinan:
> It's nice to see the cleanup of BART. Did you verify an older checkpoint didn't have results change?

yep! tried training for a bit on convai2 and comparing results

emilydinan:
Circle CI hanging on self chat tests, I ran locally and they passed
Screen Shot 2021-02-22 at 1 21 37 PM

EricMichaelSmith:
> Circle CI hanging on self chat tests, I ran locally and they passed
> Screen Shot 2021-02-22 at 1 21 37 PM

hmm, it looks like `tests/crowdsourcing/tasks/model_chat/test_model_image_chat.py::TestModelImageChat::test_base_task 
` in the crowdsourcing CI check is hanging too - it might be good to run that locally as well. Sometimes there is flakiness with the crowdsourcing CI tests (that's an issue that Jack and I know about), but it might be worth checking that this PR doesn't affect that test in some way

PR #3456, opened 2021-02-18
[Torch Agent] Reset optimizer flag

emilydinan:
**Patch description**
Flag for resetting optimizer. Useful for finetuning when you do not want the previous learning rate to load.

**Testing steps**
Tested finetuning before and after with this falg.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -504,6 +504,12 @@ class TorchAgent(ABC, Agent):
             ),
         )
         optim_group.add_argument(
+            '--reset-optimizer',
+            type='bool',
+            default=False,
+            help='Reset the optimizer during fine-tuning',
+        )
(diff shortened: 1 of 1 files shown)


stephenroller (commented):
https://github.com/facebookresearch/ParlAI/blob/fb5c92741243756516fa50073d34e94ba0b6981e/parlai/core/opt.py#L23-L32

You need to add this flag to this exception list, and add a test.

stephenroller:
https://github.com/facebookresearch/ParlAI/blob/fb5c92741243756516fa50073d34e94ba0b6981e/tests/test_lr_schedulers.py#L42-L77

for the test, just something like this is fine, checking the LR scheduler is reset

stephenroller:
Alternative for test: just create_agent_from_model_file with one of our integration test agents, and capture logs to assert that line of code was hit

PR #3461, opened 2021-02-21
Speed up conversation openers for self chat.

stephenroller:
**Patch description**
Speed up `tests/test_self_chat.py` from ~30s to ~1.5s

Methods:
- Avoid using a real model in the test, which is unnecessarily slow on CPU for unit tests
- Avoid using `repeat_label` when loading conversation openers, which causes metrics to be computed, slowing down things significantly.

**Testing steps**
CI


mojtaba-komeili (approved):

spencerp (approved):
Nice, thanks for doing this!

PR #3462, opened 2021-02-21
Upgrade metrics.py to use new type annotations.

stephenroller:
**Patch description**
Old cleanup that has been possible since we moved to python 3.7. Type annotations no longer need to be fake strings.

**Testing steps**
CI. Mypy passes locally.


EricMichaelSmith (approved):
Great to have! The class-as-a-string thing was always a (necessary) Python hack

PR #3469, opened 2021-02-24
Remove APEX

stephenroller:
**Patch description**
This patch removes (most) of our dependency on APEX by re-implementing the important features.

This was done by:
- Killing the FusedLayerNorm usage. Historically, there was a ~10% speed difference IIRC, but Pytorch has improved and closed the gap to ~2% these days last I heard. This is sufficient for cleaner, simpler code.
- Copying and re-adapting Fairseq's apex-style fp16 optimizer. It has high compatibility with the mem efficient one.

We still have the dependency for FusedAdam. This will be removed in subsequent PRs where we introduce fairscale.

Additionally, there's a small bugfix setting tolerance to 0.0 of the DynamicLossScalar. This was affecting @hadasah and occasionally causing overflows to be ignored in long jobs. Upon investigation, it turns out the default in the _class_ in fairseq is 0.05, but the default in the CLI is 0.0. We move to the better default.

**Testing steps**
Manual testing. CI.

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -29,31 +29,7 @@ from parlai.core.torch_generator_agent import TorchGeneratorModel
 from parlai.utils.misc import warn_once
 from parlai.utils.torch import neginf, PipelineHelper
 
-try:
-    from apex.normalization.fused_layer_norm import FusedLayerNorm as LayerNorm
-
-    APEX_LAYER_NORM = True
-except ImportError:
-    from torch.nn import LayerNorm
-
-    APEX_LAYER_NORM = False
-
 LAYER_NORM_EPS = 1e-5  # Epsilon for layer norm.
-
-
-def _normalize(tensor, norm_layer):
-    """
-    Broadcast layer norm.
-    """
-    is_cpu = tensor.device == 'cpu' or tensor.device.type == 'cpu'
-    if APEX_LAYER_NORM and not is_cpu:
-        # fused_layer_norm has a bug around multi-device networks.
-        # https://github.com/NVIDIA/apex/issues/770
-        # https://github.com/NVIDIA/apex/issues/371
-        with torch.cuda.device(tensor.device):
-            return norm_layer(tensor)
-    else:
-        return norm_layer(tensor)
 
 
 def _create_embeddings(dictionary, embedding_size, padding_idx):
@@ -443,7 +419,7 @@ class TransformerEncoder(nn.Module):
             or self.variant == 'prelayernorm'
             or self.variant == 'bart'
         ):
-            self.norm_embeddings = LayerNorm(self.dim, eps=LAYER_NORM_EPS)
+            self.norm_embeddings = torch.nn.LayerNorm(self.dim, eps=LAYER_NORM_EPS)
         elif self.variant == 'aiayn':
             pass
         else:
@@ -586,7 +562,7 @@ class TransformerEncoder(nn.Module):
         tensor, mask = self.forward_embedding(input, positions, segments)
 
         if self.variant == 'xlm' or self.variant == 'bart':
-            tensor = _normalize(tensor, self.norm_embeddings)
+            tensor = self.norm_embeddings(tensor)
 
         # --dropout on the embeddings
         tensor = self.dropout(tensor)
@@ -597,7 +573,7 @@ class TransformerEncoder(nn.Module):
         tensor = self.forward_layers(tensor, mask)
 
         if self.variant == 'prelayernorm':
-            tensor = _normalize(tensor, self.norm_embeddings)
+            tensor = self.norm_embeddings(tensor)
 
         # reduce output
         tensor, out_mask = self.reduce_output(tensor, mask)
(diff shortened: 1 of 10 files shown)


parlai/core/torch_agent.py
```diff
@@ -1004,31 +1002,30 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None):
...
                 # this is a bit clunky, but alternatives are worse
                 try:
-                    self.optimizer.optimizer.load_state_dict(optim_states)
+                    self.optimizer.load_state_dict(optim_states)
                 except ValueError:
                     warn_once(
                         'WARNING: not loading optim state since model params changed.'
                     )
-                return
+                return True
```
EricMichaelSmith, line 1028: nit: for clarification, maybe it'd be useful to add in the docstring what a return value of True or False indicates?

parlai/core/torch_agent.py
```diff
@@ -1907,6 +1909,11 @@ def upgrade_opt(cls, opt_from_disk: Opt):
         # call the parent upgrades
         opt_from_disk = super(TorchAgent, cls).upgrade_opt(opt_from_disk)
 
+        if 'fp16_impl' in opt_from_disk:
+            # 2021-02-23: we removed apex and replaced it with our own thing
+            if opt_from_disk['fp16_impl'] == 'apex':
```
EricMichaelSmith, line 1914: 👍

parlai/core/torch_generator_agent.py
```diff
@@ -628,6 +630,7 @@ def _init_cuda_buffer(self, batchsize, maxlen, force=False):
 
         This is also used in distributed mode to force a worker to sync with others.
         """
+        return
```
EricMichaelSmith, line 633: Hmm I assume this is a line added for debugging?

EricMichaelSmith (approved):
Seems reasonable - a few comments

parlai/agents/transformer/modules.py
```diff
@@ -29,33 +29,9 @@
 from parlai.utils.misc import warn_once
 from parlai.utils.torch import neginf, PipelineHelper
 
-try:
```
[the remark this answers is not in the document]
klshuster: im so happy right now

parlai/core/torch_generator_agent.py
klshuster: speaking of, can we possibly make this a flag to turn off, i.e. once we make this `debug` mode? i really would appreciate being able to turn this off when debugging...

parlai/utils/fp16.py
```diff
@@ -85,68 +87,186 @@ def has_overflow(grad_norm):
     return False
 
 
-###################################################
-## APEX Wrapper
-###################################################
+class SafeFP16Optimizer(torch.optim.Optimizer):
```
klshuster, line 90: what exactly is a safe fp16 optimizer?

parlai/utils/fp16.py
```diff
@@ -85,68 +87,186 @@ def has_overflow(grad_norm):
...
+            overflow = has_overflow(grad_norm)
+            prev_scale = self.scaler.loss_scale
+            self.scaler.update_scale(overflow)
+            if overflow:
+                self.zero_grad()
+                if self.scaler.loss_scale <= self.min_loss_scale:
+                    # Use FloatingPointError as an uncommon error that parent
+                    # functions can safely catch to stop training.
+                    self.scaler.loss_scale = prev_scale
+                    raise FloatingPointError(
```
klshuster, line 222: thanks for adding this

parlai/utils/fp16.py
```diff
@@ -245,7 +365,7 @@ class MemoryEfficientFP16Optimizer(torch.optim.Optimizer):
     <https://github.com/pytorch/fairseq/blob/master/fairseq/optim/fp16_optimizer.py#L382>
 
     This allows you to train bigger models on a single GPU, but can be unstable.
-    Opt for the APEX implementation if you do not have concerns about memory.
+    Opt for the Pytorch implementation if you do not have concerns about memory.
```
klshuster, line 368: can we say explicitly here `Safe` as well?

tests/test_apex.py
```diff
@@ -4,21 +4,12 @@
...
-
-
 import unittest
 from parlai.core.agents import create_agent
 from parlai.core.params import ParlaiParser
 import parlai.utils.testing as testing_utils
 
 
-@unittest.skipIf(APEX_AVAILABLE, "Apex is installed, can't test its absence.")
 class TestNoApex(unittest.TestCase):
```
klshuster, line 13: why not just delete this test?

klshuster (approved):
🥳

parlai/core/torch_generator_agent.py
stephenroller: On the roadmap to get rid of it entirely

parlai/utils/fp16.py
stephenroller: same as the apex fp16 optimizer, just ours

tests/test_apex.py
stephenroller: bc i abuse it and use it for the safe now. should rename it.

stephenroller: er, actually no. I keep it for backwards compatibility checks

but this test should be running for everyone after removing apex

PR #3470, opened 2021-02-24
[RFC] Initialize TransformerEncoder and TransformerDecoder with opt instead of individual flags

spencerp:
**Patch description**
This is one small step in a larger effort to refactor the transformer code to make it easier to modify and extend. There shouldn't be any functional changes as part of this diff, just organizational.

Some reasons for this change:
- Make these modules less cumbersome to reuse
- Move toward a pattern of using opt (by default) to configure modules instead of explicit values in the code

Let me know if this makes the code more confusing for anyone instead of less!

**Testing steps**
Leaning on the existing unit tests. Will also run some small local runs to sanity check.

### DIFF
--- a/parlai/agents/transformer/crossencoder.py
+++ b/parlai/agents/transformer/crossencoder.py
@@ -8,7 +8,6 @@ from typing import Optional
 from parlai.core.params import ParlaiParser
 from parlai.core.opt import Opt
 from .modules import TransformerEncoder
-from .modules import get_n_positions_from_options
 from parlai.core.torch_ranker_agent import TorchRankerAgent
 from .transformer import TransformerRankerAgent
 from parlai.utils.torch import concat_without_padding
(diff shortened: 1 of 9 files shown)


klshuster (approved):
is there even an argument for *not* using opt here? it's extremely cumbersome otherwise and I readily accept this change. but i'd like to hear other opinions prior to merge.

jaseweston:
having opt all the way down is the good way to go.. in my own use of playing with transformer, i'm doing that..  having opt passed into every module  (e.g. ffn) allows to try experimental flags etc

stephenroller:
Mm one of the disadvantages might be if you use a component outside of the transformer garden, it can be difficult to populate the opt.

The other disadvantage is users may be tempted to set self.opt which can make porting weights more of a pain. We can protect against that though.

Given our goals for this, I agree with using opt.

spencerp:
Abandoning in favor of https://github.com/facebookresearch/ParlAI/pull/3486 which is branched from master instead of `tf-components`.

PR #3474, opened 2021-02-26
Transfer tensors wholesale in Metric.many

stephenroller:
**Patch description**
After discussions with Jing about bottlenecks, realized the list comprehensions in `Metric.many` are probably calling `item()` separately for each bit. Not cool!

**Testing steps**

Testing this change in isolation:
```python
t = torch.rand(30).cuda()
%timeit AverageMetric.many(t)
```
Before: 406 µs ± 5.73 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
After: 51.6 µs ± 439 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

Tested it on an a very small model and observed ~5% speedup of training. Tested it on a 400M model and observed about a 1% speedup. Given that the change is simple and clear, seems worthwhile.

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -163,9 +163,12 @@ class Metric(ABC):
         Useful if you separately compute numerators and denomenators, etc.
         """
         lengths = [len(o) for o in objs]
-        objs = list(objs)
+        objs = list(objs)  # convert from tuple for inplace modification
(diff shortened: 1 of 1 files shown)


jxmsML:
good catch!! is this teacher test related to this pr tho

jxmsML (approved):

PR #3476, opened 2021-02-27
Add unzipping for build_data.py

stephenroller:
**Patch description**
I wrote these tests a long time ago but never checked them in. They seem reasonable. I can't see why we wouldn't add them.

**Testing steps**
pytest tests/test_build_data.py

### DIFF
--- a/tests/test_build_data.py
+++ b/tests/test_build_data.py
@@ -86,6 +86,67 @@ class TestBuildData(unittest.TestCase):
             assert Session.call_count == 3
 
 
+class TestUnzip(unittest.TestCase):
+    def test_ungzip(self):
+        with testing_utils.tempdir() as tmpdir:
+            import gzip
+
+            fname = os.path.join(tmpdir, "test.txt.gz")
+            with gzip.GzipFile(fname, mode="w") as f:
+                f.write("This is a test\n".encode("utf-8"))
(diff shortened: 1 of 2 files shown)


tests/test_build_data.py
```diff
@@ -86,6 +86,67 @@ def test_connectionerror_download(self):
...
+                with io.BytesIO(b"Test1\n") as f:
+                    tarinfo = tarfile.TarInfo("test1.txt")
+                    tarinfo.size = 6
+                    zf.addfile(tarinfo, fileobj=f)
+                with io.BytesIO(b"Test2\n") as f:
+                    tarinfo = tarfile.TarInfo("test2.txt")
+                    tarinfo.size = 6
+                    zf.addfile(tarinfo, fileobj=f)
+
+            build_data._untar(tmpdir, "test.tar.gz")
```
mojtaba-komeili, line 140: question: if we are accessing `_untar` and `_unzip` outside their original class, why do we still name them with underscore? That makes using `build_data` a bit confusing. To me it almost means don't use this method. Look for other methods we are using. If such other methods exist, shouldn't we test those instead.

mojtaba-komeili (approved):

tests/test_build_data.py
stephenroller: We're not great at choosing to mark things private or not.

PR #3486, opened 2021-03-04
Pass Opt to TransformerEncoder/Decoder

spencerp:
**Patch description**
Since the RFC was well received (#3470) here's a PR against master (the other PR was branched off of tf-components #3466).

**Testing steps**
Make sure CircleCI passes.

### DIFF
--- a/parlai/agents/transformer/modules.py
+++ b/parlai/agents/transformer/modules.py
@@ -279,6 +279,9 @@ class TransformerEncoder(nn.Module):
     """
     Transformer encoder module.
 
+    For documentation on parameters that are take directly from opt,
+    see parlai/agents/transformer/transformer.py
+
     :param opt: ParlAI-parsed options.
     :param vocabulary_size: Count of tokens/words in the dictionary.
     :param embedding: an embedding matrix for the bottom layer of the transformer.
@@ -608,15 +611,27 @@ class TransformerEncoderLayer(nn.Module):
 
 class TransformerDecoder(nn.Module):
     """
-    Transformer Decoder layer.
+    Transformer Decoder module.
+
+    For documentation on parameters that are take directly from opt,
+    see parlai/agents/transformer/transformer.py
 
     :param opt: ParlAI-parsed options.
     :param embedding: an embedding matrix for the bottom layer of the transformer.
         If none, one is created for this encoder.
+    :param int n_positions: Size of the position embeddings matrix.
     """
 
-    def __init__(self, opt: Opt, embedding: Optional[nn.Embedding] = None):
+    def __init__(
+        self,
+        opt: Opt,
+        embedding: Optional[nn.Embedding] = None,
+        n_positions: Optional[int] = None,
+    ):
         super().__init__()
+
+        def _default(val, default):
+            return val if val is not None else default
 
         self.embedding_size = opt['embedding_size']
         self.ffn_size = opt['ffn_size']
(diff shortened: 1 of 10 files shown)


jaseweston:
....and what will happen with parlai_internal?

spencerp:
> ....and what will happen with parlai_internal?

I'm getting an internal PR ready now!

parlai/agents/transformer/modules.py
```diff
@@ -310,86 +277,58 @@ class TransformerEncoder(nn.Module):
...
     :param int padding_idx: Reserved padding index in the embeddings matrix.
-    :param bool learn_positional_embeddings: If off, sinusoidal embeddings are
-        used. If on, position embeddings are learned from scratch.
+    :param str reduction_type: Type of reduction at the end of the encoder.
+    :param int n_segments: Number of segments/lang/sentence embeddings.
     :param bool embeddings_scale: Scale embeddings relative to their dimensionality.
         Found useful in fairseq.
-    :param bool reduction: If true, returns the mean vector for the entire encoding
-        sequence.
-    :param int n_positions:
```
jaseweston, line 333: will these comments appear somewhere else then?

jaseweston:
> > ....and what will happen with parlai_internal?
> 
> I'm getting an internal PR ready now!

👍
👍

parlai/agents/transformer/modules.py
spencerp: In the opt declaration! I just noticed that not all of opt flags have descriptions, though, so I'll fill them in from here where applicable.

parlai/agents/transformer/modules.py
```diff
@@ -310,86 +277,58 @@ class TransformerEncoder(nn.Module):
...
         self.reduction_type = reduction_type
         self.padding_idx = padding_idx
         # this is --dropout, not --relu-dropout or --attention-dropout
-        self.dropout_frac = dropout
+        self.dropout_frac = opt['dropout']
         self.dropout = nn.Dropout(p=self.dropout_frac)
-        self.variant = variant
-        self.n_segments = n_segments
+        self.variant = opt['variant']
+        self.n_segments = n_segments or opt.get('n_segments', 0)
```
spencerp, line 319: Btw if we find ourselves needing to manually specify any parameters or override what's in opt, we can just add another default None keyword arg and do this.

spencerp:
Hm, Transresnet tests are failing consistently. That's concerning... Will need to get to the bottom of that before merging.

parlai/agents/transformer/modules.py
klshuster: i suppose that's one thing we lose by getting rid of this, but the net negative lines is so enticing...

parlai/agents/transformer/modules.py
klshuster: i suppose im a bit curious how you chose which options to keep in the init, and which you left for opt?

parlai/agents/transformer/modules.py
```diff
@@ -658,62 +603,34 @@ class TransformerDecoder(nn.Module):
     """
     Transformer Decoder layer.
```
klshuster, line 604: minor nit i noticed (unrelated to your patch) is that this docstring is wrong

klshuster (approved):
I think this looks good - definitely a fan of net negative lines.

I think Jason made a good point about the field/attr descriptions. I like how they're in the opt, but would it also be possible to include a note in the docstring of where to find those descriptions? like, "For more information regarding model parameters, see :file:" or something like that

stephenroller:
Big improvement.

PR #3496, opened 2021-03-08
Messages instead of dicts for get() in ED, DNLI

skywalker023:
**Patch description**
When running EmpatheticDialogues and Dialogue NLI, I get the following warning messages:
`parlai.tasks.empathetic_dialogues.agents.DefaultTeacher' is outputting dicts instead of messages. If this is a teacher that is part of ParlAI, please file an issue on GitHub. If it is your own teacher, please return a Message object instead.`
`parlai.tasks.dialogue_nli.agents.DefaultTeacher' is outputting dicts instead of messages. If this is a teacher that is part of ParlAI, please file an issue on GitHub. If it is your own teacher, please return a Message object instead.`

So, I wrapped the dicts with the Message object.

**Testing steps**
Ran
`pytest parlai/tasks/empathetic_dialogues/test.py`
`pytest parlai/tasks/dialogue_nli/test.py`
for the test.

### DIFF
--- a/parlai/tasks/dialogue_nli/agents.py
+++ b/parlai/tasks/dialogue_nli/agents.py
@@ -13,6 +13,7 @@ from parlai.core.opt import Opt
 import json
 import os
 
+from parlai.core.message import Message
 from parlai.core.teachers import FixedDialogTeacher
 from .build import build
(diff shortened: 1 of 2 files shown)


stephenroller (approved):
Thank you very much!

PR #3497, opened 2021-03-08
Add new token_em metric

stephenroller:
**Patch description**
Add a new "tokens exact match metric", corresponding roughly to how "accuracy" would be under greedy search.

Will be documented in another PR.

**Testing steps**
Internal sweep.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -709,10 +709,17 @@ class TorchGeneratorAgent(TorchAgent, ABC):
         target_tokens = notnull.long().sum(dim=-1)
         correct = ((batch.label_vec == preds) * notnull).sum(dim=-1)
 
+        # cross entropy loss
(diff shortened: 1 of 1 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -709,11 +709,18 @@ def compute_loss(self, batch, return_output=False):
...
         self.record_local_metric('loss', AverageMetric.many(loss, target_tokens))
+        # perplexity
         self.record_local_metric('ppl', PPLMetric.many(loss, target_tokens))
+        # token-wise accuracy
         self.record_local_metric(
             'token_acc', AverageMetric.many(correct, target_tokens)
         )
+        # utterance-wise exact match
+        self.record_local_metric(
+            'token_em', AverageMetric.many(correct == target_tokens)
```
moyapchen, line 722: Hmm.. this is likely nice for some QA+RAG trains as well..

moyapchen (approved):
4x speedups are nice. :D

PR #3501, opened 2021-03-08
[Transformer Opt] Add back default values to underspecified call sites

spencerp:
**Patch description**
Some call sites didn't pass through opt values, and were therefore implicitly using the default __init__ values before. This diff adds back those default values explicitly.

**Testing steps**
Make sure CircleCI passes.


stephenroller (approved):
oh no

Issue #3502, opened 2021-03-09
Dynamic batching not working 


frankplus:
Hi, 
when setting `--dynamic-batching full` it throws an error at the start of the training
```
  File "/usr/local/bin/parlai", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/dist-packages/parlai/__main__.py", line 14, in main
    superscript_main()
  File "/usr/local/lib/python3.7/dist-packages/parlai/core/script.py", line 306, in superscript_main
    return SCRIPT_REGISTRY[cmd].klass._run_from_parser_and_opt(opt, parser)
  File "/usr/local/lib/python3.7/dist-packages/parlai/core/script.py", line 89, in _run_from_parser_and_opt
    return script.run()
  File "/usr/local/lib/python3.7/dist-packages/parlai/scripts/train_model.py", line 809, in run
    return self.train_loop.train()
  File "/usr/local/lib/python3.7/dist-packages/parlai/scripts/train_model.py", line 693, in train
    world.parley()
  File "/usr/local/lib/python3.7/dist-packages/parlai/core/worlds.py", line 1097, in parley
    self._task_acts[i]['dyn_batch_idx'] = i
  File "/usr/local/lib/python3.7/dist-packages/parlai/core/message.py", line 28, in __setitem__
    'please use the function `force_set(key, value)`.'.format(key)
RuntimeError: Message already contains key `dyn_batch_idx`. If this was intentional, please use the function `force_set(key, value)`.
```

This is the command I used:
```
!parlai train_model \
-t fromfile:parlaiformat \
--fromfile_datapath $DATASET_FILE \
--fromfile-datatype-extension true \
-m transformer/generator \
--save-every-n-secs 300 \
--validation-every-n-secs 600 \
--eval-batchsize 256 \
--dynamic-batching full \
--fp16 True --fp16-impl mem_efficient \
--batchsize 128 \
--embedding-size 512 \
--n-encoder-layers 1 \
--n-decoder-layers 13 \
--ffn-size 2048 \
--dropout 0.1 \
--n-heads 16 \
--learn-positional-embeddings True \
--n-positions 512 \
--variant xlm \
--activation gelu \
--skip-generation True \
--text-truncate 128 \
--label-truncate 128 \
--truncate 128 \
--dict-tokenizer bpe \
--dict-lower True \
-lr 1e-03 \
--optimizer adam \
--lr-scheduler reduceonplateau \
[...]

PR #3505, opened 2021-03-10
Fix issue #3502: "Message already contains key `dyn_batch_idx`"

frankplus:
**Patch description**
This PR fixes issue #3502
`self._task_acts[i]['dyn_batch_idx'] = i` was setting a value for a key that was already set throwing error `Message already contains key 'dyn_batch_idx'`.

Simply replacing it with `self._task_acts[i].force_set('dyn_batch_idx', i)` fixed the problem.

**Testing steps**
Tested by running `python3 tests/test_dynamicbatching.py` without any problem.

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -1094,7 +1094,7 @@ class DynamicBatchWorld(World):
             # we log the task act and the index of the act
             # in the buffer for world logging purposes
             self._task_acts[i] = act  # for world logging
-            self._task_acts[i]['dyn_batch_idx'] = i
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
Will wait for tests to pass and then merge, thanks for fixing!

frankplus:
The CircleCI test seems to fail for a reason unrelated to the changes made in this PR. Is there anything I could do to fix it?

stephenroller:
Nah I agree this isn't your problem. Thanks for your contribution!

PR #3389, opened 2021-01-16
Limit torch_agent.Batch to tensors

stephenroller:
**Patch description**
Where possible, limit the attributes of our Batch object to only tensors. Additionally, delay cudafying until late and switch to using batch.batchsize for many operations.

This is in preparation of a fresh attempt of Background Preprocessing.

**Testing steps**
CI

PR #3508, opened 2021-03-10
Add length of truncation to logs

spencerp:
**Patch description**
It seems useful to know the average number of tokens truncated when they are truncated. This can't be reliably gleaned from the current metrics, since avg(original_context_length) can be below the truncation length any time context_truncate is below 0.5.

**Testing steps**
Used this in a training run -> Verified the metric showed up in tensorboard


dexterju:
LGTM, could you expend the test in  test in test_torch_agent.py : https://github.com/facebookresearch/ParlAI/blob/802cec94a4ac027c403b243c8b61d2b967494909/tests/test_torch_agent.py#L1055

dexterju (approved):
add this to the test plz and you are good to go.

stephenroller (commented):
Can you hold until I land the Tensor diff?

spencerp:
> Can you hold until I land the Tensor diff?

Sure! This one, right? https://github.com/facebookresearch/ParlAI/pull/3389

Issue #3493, opened 2021-03-05
Redudant --skip-generation True in the docs


vinnik-dmitry07:
https://parl.ai/projects/recipes/ 
Fine-tuning your own models > 90M

stephenroller:
Feel free to submit a patch, thanks!

PR #3513, opened 2021-03-12
[docs] remove redundant --skip-generation True

frankplus:
**Patch description**
Removing redundant `--skip-generation` which is specified twice in the docs for training the 90M parameters model.
Fixing issue #3493


stephenroller (approved):
Thanks!

PR #3514, opened 2021-03-12
Minor FP16 fixes when converting old checkpoints

stephenroller:
**Patch description**
- Don't throw an error if we load a checkpoint that had an APEX scaler
- Avoid saving an optimizer with BART in case the user decides to initialize with `-opt sgd`. Lowers disk usage slightly

**Testing steps**
Manual testing


parlai/agents/bart/convert_fairseq_to_parlai.py
```diff
@@ -168,6 +168,7 @@ def get_parlai_opt(self) -> Opt:
         # 3. Map other options
         transformer_common_config.update(
             {
+                'datatype': 'valid',  # don't save an optimizer
```
spencerp, line 171: This is a little weird as it doesn't seem obvious that saving datatype as valid will always exclude _only_ the optimizer. And it's not obvious that there won't be any other effects.

Is it possible to just explicitly remove the optimizer in `load_fairseq_checkpoint`?

spencerp (approved):
Commented on something that feels a little strange to me, but I'm not very familiar with this code path so feel free to ignore if this is in fact the most direct way of not saving the optimizer.

parlai/agents/bart/convert_fairseq_to_parlai.py
stephenroller: I'll do a post vacuum

PR #3515, opened 2021-03-12
Attrs version bump

emilydinan:
**Patch description**
Bump attrs version to see if there are issues...


stephenroller (approved):
K seems all good

PR #3517, opened 2021-03-12
Support a rewards field in parlai format

stephenroller:
**Patch description**
A user requested that the rewards field be properly respected as an integer and/or float for use with the unlikelihood agent.

**Testing**
CI


EricMichaelSmith (approved):
Makes sense

PR #3521, opened 2021-03-15
Bump CI checks to Torch 1.7

EricMichaelSmith:
Add CI checks for Torch 1.7 and remove the 1.4 check. All checks will now use Torch 1.7 unless they are testing previous versions explicitly


stephenroller (approved):
lgtm if tests pass or only noisy test fails

stephenroller:
Failed test is fixed on master, this is good to go

PR #3522, opened 2021-03-16
[fairscale] FusedAdam support

stephenroller:
**Patch description**
Add support for Fused Adam from Fairscale, and get rid of the version from Apex.

Intentionally changed the name of the optimizer because the two versions are mutually incompatible.

Also delete QHAdam since no one ever used that.

**Testing steps**
CI. Manual testing.


dexterju (approved):
Thanks for doing this!

PR #3523, opened 2021-03-16
Fix up CCPE so that it doesn't skip first turn

moyapchen:
Title. Was having a hard time parsing the code as is, so rewrote it as well.

Output now: 

```
- - - NEW EPISODE: ccpe - - -
generally speaking what type of movies do you watch
   I like thrillers a lot.
thrillers? for example?
   Zodiac's one of my favorite movies.
Zodiac the movie about a serial killer from the '60s or '70s, around there.
Zodiac? oh wow ok, what do you like about that movie
   And I just think serial killers, in general, are interesting, so the movie was really g
ood. And it was just neat to see that world. Like it went really in-depth. It was like alm
ost 3 hours long, so you got to really feel like you were a part of that world and time pe
riod. And see what the detectives and the police did and the investigators.
so you feel like you were part of that world ?
   Yeah. It was really an immersive movie.
if you were in the movie what character do you feel most relatable
   Probably the main character, Robert Graysmith, just cuz he's awkward and bumbly. So, I
guess that's the one I would be the closest to.
what scene do you like the best
   Probably the most memorable one is the murder at the lake. Just cuz it's really vivid a
nd horrific to watch. But it's very memorable.
ok nice, any other similiar movies you have seen recently?
   Not that I can think of.
how about different type of movies like Once Upon a Deadpool
   I haven't seen that.
have you heard of that movie
   I've heard of it, but I haven't seen it.
would you be interested to watch it
   Not particularly. I don't really like comic book movies that much. But I really wanted
to see Deadpool, so I'd probably be more willing to watch that than most comic book movies
.
```

```
- - - NEW EPISODE: ccpe:CcpeUserTeacher - - -
I like thrillers a lot.
   thrillers? for example?
Zodiac's one of my favorite movies.
Zodiac the movie about a serial killer from the '60s or '70s, around there.
   Zodiac? oh wow ok, what do you like about that movie
And I just think serial killers, in general, are interesting, so the movie was really good. And it was just neat to see that world. Like it went really in-depth. It was like almost 3 hours long, so you got to really feel like you were a part of that world and time period. And see what the detectives and the police did and the investigators.
   so you feel like you were part of that world ?
Yeah. It was really an immersive movie.
   if you were in the movie what character do you feel most relatable
Probably the main character, Robert Graysmith, just cuz he's awkward and bumbly. So, I guess that's the one I would be the closest to.
   what scene do you like the best
Probably the most memorable one is the murder at the lake. Just cuz it's really vivid and horrific to watch. But it's very memorable.
   ok nice, any other similiar movies you have seen recently?
Not that I can think of.
   how about different type of movies like Once Upon a Deadpool
I haven't seen that.
   have you heard of that movie
I've heard of it, but I haven't seen it.
   would you be interested to watch it
Not particularly. I don't really like comic book movies that much. But I really wanted to see Deadpool, so I'd probably be more willing to watch that than most comic book movies.
   oh what makes deadpool different than the others?
```


parlai/tasks/ccpe/agents.py
```diff
@@ -3,150 +3,106 @@
...
-
-
-class CCPEAssistantTeacher(CCPEAllTeacher):
-    def _setup_data(self):
-        super()._setup_data()
-        self.data = self.assistantData
-
-
-class DefaultTeacher(CCPEAllTeacher):
+        # do a 80:10:10 train/valid/test split
```
klshuster, line 38: should we do a random shuffle of the data here? or does it not matter

parlai/tasks/ccpe/agents.py
```diff
@@ -3,150 +3,106 @@
...
+            json_data = json_data[full_episode_count * 9 :]
+
+        episodes = []
+        for conversation in json_data:
+            episode = []  # Assume all turns are alternating with ASSISTANT first
+            idx = 0
+            utts = conversation["utterances"]
+            if utts[0]["speaker"] != "ASSISTANT":
+                # Make a dummy assistant turn and add a user turn
+                episode.append({"speaker": "ASSISTANT", "text": "", "segments": []})
```
klshuster, line 54: usually we'll do something like put a `__SILENCE__` token here; has your model handled these empty text observations well?

parlai/tasks/ccpe/agents.py
```diff
@@ -3,150 +3,106 @@
...
-    def _setup_data(self):
+        super().__init__(opt, shared)
 
+    def _get_turns_of_speaker(self, utts, idx, speaker):
+        result = {"speaker": speaker, "text": "", "segments": []}
+        while idx < len(utts) and utts[idx]["speaker"] == speaker:
+            if len(result["text"]) == 0:
+                result["text"] = utts[idx]["text"]
+            else:
+                result["text"] += "\n" + utts[idx]["text"]
```
klshuster, line 28: nit: you can also just add to list and to a `"\n".join()` after

parlai/tasks/ccpe/agents.py
```diff
@@ -3,150 +3,106 @@
...
-    def num_episodes(self):
-        return len(self.data)
-
-    def num_examples(self):
-        return sum([len(x) for x in self.data])
-
-    def _setup_data(self):
+        super().__init__(opt, shared)
 
+    def _get_turns_of_speaker(self, utts, idx, speaker):
```
klshuster, line 22: what exactly is this function doing?

parlai/tasks/ccpe/agents.py
```diff
@@ -3,150 +3,106 @@
...
+class CcpeAssistantTeacher(_Abstract):
+    def setup_data(self, fold):
+        episodes = self._load_data(fold)
+        for episode in episodes:
+            first = True
+            for i in range(len(episode) // 2):
+                assistant_turn = episode[2 * i]
+                user_turn = episode[2 * i + 1]
+                if len(assistant_turn["text"]) == 0:
+                    continue  # trains fail otherwise
```
klshuster, line 78: so you're adding in a dummy example above, but then you're skipping it in these teachers?

klshuster (commented):
i'm glad you cleaned this up! and got away from `FixedDialogTeacher`

but i'm a bit confused by what's going on here

parlai/tasks/ccpe/agents.py
moyapchen: The dataset doesn't specify a train/valid/test split.. I can see shuffling by a random seed making sense here though.

parlai/tasks/ccpe/agents.py
moyapchen: Putting in a `__SILENCE__` token seems like a good call; hadn't thought of that.

parlai/tasks/ccpe/agents.py
moyapchen: d'oh yeah.

parlai/tasks/ccpe/agents.py
moyapchen: Compressing all contiguous turns of a given speaker (all of a user or all of an Assistant or all of a User) in the original json into one. Adding in newlines is pretty similar to what the previous version of this was doing.. just that it's abstracted out here. (Can add a comment to make this more clear.)

parlai/tasks/ccpe/agents.py
moyapchen: I'm adding a dummy turn above so that the format returned from `_load_data(fold)` has the same expectations of 

> first turn = ASSISTANT
> second turn = USER
> ... and so on

since some of the conversations in the original dataset start with the USER. This makes it easier to specify which turns of the episode should be used by CcpeAssistantTeacher and which should be used by CcpeUserTeacher

(I can see now how that's not super obvious at first glance... can add more comments.)

moyapchen: I need to skip it here because a part of `self_observe()` errors out if an empty text string was passed in.

klshuster: yeah that can be solved by putting e.g. the `__SILENCE__` token

PR #3524, opened 2021-03-16
[ACUTE] Update default maximum_units_per_worker

stephenroller:
**Patch description**

Set to 1 to emulate --unique, which was historically important for replicability

**Testing steps**
none


EricMichaelSmith (approved):
Makes sense

PR #3528, opened 2021-03-17
Add a `randomize_conversations` to ACUTE blueprint + use

moyapchen:
...cause I want to be able to not randomize conversations.

Context here is that I have chats generated from different models using a fixed set of prompts. I want to be able to compare how different models do on the same prompt and consequently do not want the chats to be randomized.

Test plan:
Print out the pairs value from this function; verify that things align.


EricMichaelSmith (commented):
Yeah, makes sense! Can you add the default value of this param to the example YAML file? It looks like the crowdsourcing CI checks are failing without it

stephenroller (approved):

PR #3532, opened 2021-03-17
Fix bug introduced by killing APEX

stephenroller:
**Patch description**
This change was introduced in #3469. At some point, it seemed necessary to get things working for our FP16 safe optimizer. After debugging with @jaseweston, it seems like it was actually significantly increasing numerical instability. I've isolated this change as the problematic change, and confirmed that it does not have issues in either `--fp16-impl safe` or `--fp16-impl mem_efficient`

**Testing steps**
Manual testing.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -504,6 +504,9 @@ class TorchGeneratorAgent(TorchAgent, ABC):
                 f"Total parameters: {total_params:,d} ({train_params:,d} trainable)"
             )
 
+            if self.fp16:
+                self.model = self.model.half()
+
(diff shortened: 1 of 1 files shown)


stephenroller:
Confirmed this makes a very large difference on training a 400M from scratch. After about 500 updates, the difference is something like ~230ppl for the good commits, and ~400ppl for the bad commits.

On fine-tuning a BART model, the good commits get get like 12.79 valid ppl after 500 steps. The bad commits get like 14.63 at 500 steps.

At ~10000 steps, good commits get X, bad commits get Y.

stephenroller:
Gonna go ahead and land this fast while scoping the problem

PR #3534, opened 2021-03-17
Full Tensorboard metric titles

spencerp:
**Patch description**
The abbreviated metric titles are hard to get used to, but using longer metric identifiers makes the stdout print hard to parse.
This PR adds a separation between the metric identifier and the display name of the metric. It also adds a description. The title and description are included in tensorboard, but the abbreviated identifiers are used in the logs.
It moves the source of truth for metric titles and descriptions from the docs to the code, and generates the metrics table from this.

There are a few things I don't really have time to tackle right now, but would be nice for a later PR:
- Representing families of metrics more automatically in the source of truth dict (collapse rouge-* metrics and use appropriate name and description if any metrics matching that format are used).
- Collapsing multiple metrics of the same family in the metrics table
- Preserving the monospace formatting of the metrics in the docs table

**Testing steps**
Ran a basic test run locally:
```
parlai train_model --task babi:task10k:1 --model-file ~/tmp/babi_memnn --batchsize 1 --num-epochs 5 --model memnn --no-cuda -tblog True
```
Verified the metrics were short and fit on the screen:
Screen Shot 2021-03-17 at 4 28 03 PM

Verified that they were long and had descriptions on Tensorboard:
Screen Shot 2021-03-17 at 4 27 16 PM

Built website:
```
cd docs; make html
```
Verified the metrics list was rendered:
Screen Shot 2021-03-17 at 4 50 19 PM

### DIFF
--- a/docs/source/generate_metric_list.py
+++ b/docs/source/generate_metric_list.py
@@ -11,7 +11,7 @@ fout = open('metric_list.inc', 'w')
 
 fout.write('| Metric | Explanation |\n')
 fout.write('| ------ | ----------- |\n')
-for metric, display in METRICS_DISPLAY_DATA.items():
+for metric, display in sorted(METRICS_DISPLAY_DATA.items()):
     fout.write(f'| `{metric}` | {display.description} |\n')
 
 fout.close()
--- a/docs/source/generate_metric_list.py
+++ b/docs/source/generate_metric_list.py
@@ -12,6 +12,6 @@ fout = open('metric_list.inc', 'w')
 fout.write('| Metric | Explanation |\n')
 fout.write('| ------ | ----------- |\n')
 for metric, display in METRICS_DISPLAY_DATA.items():
-    fout.write(f'| {metric} | {display.description} |\n')
+    fout.write(f'| `{metric}` | {display.description} |\n')
 
 fout.close()
--- a/docs/source/generate_metric_list.py
+++ b/docs/source/generate_metric_list.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+from parlai.core.logs import METRICS_DISPLAY_DATA
+
+
+fout = open('metric_list.inc', 'w')
+
+fout.write('| Metric | Explanation |\n')
+fout.write('| ------ | ----------- |\n')
+for metric, display in METRICS_DISPLAY_DATA.items():
+    fout.write(f'| {metric} | {display.description} |\n')
+
+fout.close()
--- a/docs/source/generate_metric_list.py
+++ b/docs/source/generate_metric_list.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
-from parlai.core.logs import METRICS_DISPLAY_DATA
+from parlai.core.metrics import METRICS_DISPLAY_DATA
 
 
 fout = open('metric_list.inc', 'w')
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -11,11 +11,21 @@ from __future__ import annotations
 
 import re
 from abc import ABC, abstractmethod
-from collections import Counter, namedtuple
+from collections import Counter
 import functools
 import datetime
 import math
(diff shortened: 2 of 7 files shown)


docs/source/generate_metric_list.py
```diff
@@ -0,0 +1,17 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from parlai.core.metrics import METRICS_DISPLAY_DATA
+
+
+fout = open('metric_list.inc', 'w')
+
+fout.write('| Metric | Explanation |\n')
+fout.write('| ------ | ----------- |\n')
+for metric, display in METRICS_DISPLAY_DATA.items():
```
stephenroller, line 14: maybe a sort just to be sure?

parlai/core/metrics.py
```diff
@@ -34,6 +34,100 @@
 }
 ALL_METRICS = DEFAULT_METRICS | ROUGE_METRICS | BLEU_METRICS | DISTINCT_METRICS
 
+MetricDisplayData = namedtuple('MetricDisplayData', ('title', 'description'))
```
stephenroller, line 37: how about a data class instead

parlai/core/metrics.py
```diff
@@ -34,6 +34,100 @@
...
+        "Tokens Per Batch", "Total tokens (context + label) per batch"
+    ),
+    "tps": MetricDisplayData(
+        "Tokens Per Second", "Total tokens (context + label) per second"
+    ),
+    "ups": MetricDisplayData("Updates Per Second", "Updates per second (approximate)"),
+}
+
+
+def get_metric_display_data(metric: str) -> MetricDisplayData:
```
stephenroller, line 126: maybe as a utility of MetricsDisplayData

parlai/core/metrics.py
```diff
@@ -34,6 +34,100 @@
...
+    return METRICS_DISPLAY_DATA.get(
+        metric, MetricDisplayData(title=metric, description="No description provided.")
```
stephenroller, line 128: How about adding  `No description provided. Please add it to metrics.py if this is an official metric in ParlAI.`

stephenroller (commented):
v cool

parlai/core/metrics.py
spencerp: Hm, what's the advantage you see? I was using a namedtuple because this isn't really data that should be mutable.

If it's the pretty syntax you're looking for, though, I just found out there's a nicer syntax for namedtuples:
```
class MetricDisplayData(NamedTuple):
    title: str
    description: str
```

parlai/core/metrics.py
spencerp: It's kind of nice to keep this functional, though, since there isn't any state we should be keeping around. Also, it needs access to METRICS_DISPLAY_DATA which I think makes more sense scoped to the namespace than a class. What's the advantage you see from putting it in MetricDisplayData?

parlai/core/metrics.py
stephenroller: We used NamedTuples before and it gave me headaches and now i stay away with them. That syntax is nicer and fine with me.

parlai/core/metrics.py
stephenroller: I was thinking of a classmethod (and maybe the global too), just to keep everything in a tight namespace.

stephenroller (approved):

parlai/core/metrics.py
stephenroller: lol prolly the global can't be in there so long as it's contained...

parlai/core/metrics.py
spencerp: I'm curious what headaches you encountered! I haven't used them a ton (just for small things like this here and there) so maybe there're headaches incoming I'm ignorant of.

parlai/core/metrics.py
spencerp: Yeah I think if we went that path, the metrics/titles/descriptions would live in a separate json/yaml file. And we'd have a separate function that loads them up as `MetricDisplayData`s into a global. But then there'd be a disconnect between the source of truth and the global which is a little weird. I guess we could make `MetricDisplayData` a singleton and load them up on instantiation, but then we have to instantiate an object just to get this static list of strings which feels heavy.

Another way to keep them in a tight namespace would be to just create a `metrics_list.py` module.

Idk let me know if any of those options sound better, I see plenty of advantages and disadvantages to each so not super opinionated lol

PR #3539, opened 2021-03-19
Fix Wizard unit test GPU OOMs

EricMichaelSmith:
Reduce the batch size of two Wizard tasks to prevent them from sometimes OOMing on the long_gpu_tests CI check


EricMichaelSmith:
@stephenroller I've been able to reproduce a GPU OOM when doing nothing but adding 2 new no-op unit tests. @klshuster is this similar to your current issue?

klshuster:
yes!!!!

the exact same test is OOMing for me. I'm investigating right now

klshuster:
that wizard test evaluates over the entire WoW dataset which is certainly not necessary

klshuster (approved):

PR #3548, opened 2021-03-23
Change `--inference` to `--ranker-inference` in torch_ranker_agent

moyapchen:
There's a name collision between this and torch_generator_agent. For models that make use of both, the name collision makes for weird behavior.

Since torch_ranker_agent is the one that is less used (and generally has less changes in this variable), let's shift the name collision here.

**Testing steps**
Been using this a bunch in various sweeps and it's fine.

### DIFF
--- a/parlai/core/torch_ranker_agent.py
+++ b/parlai/core/torch_ranker_agent.py
@@ -162,7 +162,7 @@ class TorchRankerAgent(TorchAgent):
             'single candidate according to the ranking.',
         )
         agent.add_argument(
-            '--inference',
+            '--ranker-inference',
             choices={'max', 'topk'},
             default='max',
             help='Final response output algorithm',


spencerp:
I'm definitely supportive of this. Is there any backwards compatibility issue when loading old model checkpoints?

stephenroller (commented):
Yeah I'm a bit wary about changing old options names; how old is the name? If it's rather new, then I mind less, but if it's been around ages, it usually backfires on us.

We could instead have the [experimental agent] re-define the options with a different destination instead.

I think a bit more concerning is that is that we don't actually update any of the uses of the `inference` method?

In the cases that we do change option names or values, we have an `upgrade_opt` function that we implement to upgrade old model files. You'll want to add to it for this.

klshuster:
yeah if we do this change you'll need to implement the `upgrade_opt` functionality as well as change the calls within the code to access `inference`

but perhaps we can solve this in the underlying [experimental agent]...

moyapchen:
There's a nice workaround with just reordering some things; closing this.

PR #3549, opened 2021-03-24
[ChunkTeacher] Fix a hang

stephenroller:
**Patch description**
Fix a bug where CT would hang when the buffer was smaller than the size of the the test set chunk.

Also define a new set of tests to pretty exhaustively test CT in a wide variety of settings.

**Testing steps**
New CI

### DIFF
--- a/parlai/agents/test_agents/counter.py
+++ b/parlai/agents/test_agents/counter.py
@@ -84,7 +84,11 @@ class CounterAgent(TorchAgent):
         pass
 
     def _val(self, val):
-        if isinstance(val, list):
+        """
+        Pull out a singleton value if provided a list.
+        """
+        # necessary for labels
+        if isinstance(val, (tuple, list)):
             return val[0]
         else:
             return val
@@ -99,6 +103,7 @@ class CounterAgent(TorchAgent):
         pass
 
     def _to_tuple(self, msg: Message) -> Tuple:
+        # turned into an indexable object
         keys = ['text', 'labels', 'eval_labels']
         return tuple(self._val(msg.get(k)) for k in keys)
 
--- a/parlai/agents/test_agents/counter.py
+++ b/parlai/agents/test_agents/counter.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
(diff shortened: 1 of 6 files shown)


parlai/agents/test_agents/counter.py
```diff
@@ -0,0 +1,120 @@
...
+        return shared
+
+    def save(self, path=None):
+        pass
+
+    def load(self, path=None):
+        pass
+
+    def _val(self, val):
+        if isinstance(val, list):
```
emilydinan, line 87: this is a bit weird so maybe add a comment here why you do this (text vs. labels)

parlai/core/teachers.py
```diff
@@ -2514,6 +2514,11 @@ def reset(self):
             )  # reset the count of samples loaded
             self._enqueue_request()
 
+    def shutdown(self):
+        # self._drain(self.chunks)
```
emilydinan, line 2518: why did you comment this out?? are you sure you want to do this?

parlai/tasks/integration_tests/agents.py
```diff
@@ -510,19 +515,14 @@ def get(self, episode_idx=0, entry_idx=0):
...
     Unique examples that load slowly.
     """
 
     def load_from_chunk(self, chunk_idx: int):
-        output = []
-        for i in range(10):
-            text = str(i + chunk_idx * 10)
-            resp = str(i + chunk_idx * 10)
-            output.append((text, resp))
         time.sleep(0.1)
```
emilydinan, line 524: is adding this single time.sleep(0.1) enough?? why is it still considered a "slow" teacher? maybe make a diff teacher?

emilydinan (changes requested):
i think you commented out `self._drain` by accident? also why did you add more `Nones` to shutdown?

parlai/tasks/integration_tests/agents.py
stephenroller: Yes, it's enough. It's been like this for a bit.

parlai/core/teachers.py
stephenroller: It's not necessary, need to delete

emilydinan (approved):

PR #3550, opened 2021-03-24
Optimize MultiHeadAttention forward method

dejk-alg:
Avoid duplicated calculation of key-value pairs when incr_state is passed and state_kv is True

**Patch description**
Slightly alters MultiHeadAttention forward method. 

The problem with the original code is that when incr_state dict is passed to model with state_kv set to True, the method calculates k and v (processed pre-attention key and value tensors) from key / value inputs by passing them through the respective linear layers, despite replacing them with similar tensors extracted from incr_state just a few lines of code later. 

PR addresses this by a small change that avoid this excessive calculation by moving it to conditional branches where this is not the case and it is indeed required.

PR should not be breaking for any ParlAI modules as it does not alter the logic of MultiHeadAttention or its outputs in any way. The only thing that actually changes is that inner redundant operation is no longer performed when not needed.

While doing proper benchmarking to showcase the difference PR brings would be pretty complex, since the module performance differs widely among the different settings and models it's used as a part of, I've seen latency decrease as much as twice for a decoding iteration of the Blender model (both for cpu / gpu settings and 90M / 3B models). While in general average performance gain is probably lower, it still seems really significant to me.

### DIFF
--- a/parlai/agents/transformer/modules/attention.py
+++ b/parlai/agents/transformer/modules/attention.py
@@ -178,8 +178,6 @@ class MultiHeadAttention(nn.Module):
         _, _key_len, dim = key.size()
 
         q = prepare_head(self.q_lin(query))
-        k = prepare_head(self.k_lin(key))
-        v = prepare_head(self.v_lin(value))
 
         # Prepend incremental states. For each of the key, value, and mask, see if
         # a previous incremental state exists, and if so, reshape it to match the shape
@@ -195,7 +193,9 @@ class MultiHeadAttention(nn.Module):
             if static_kv:
                 k = prev_key
             else:
-                k = torch.cat([prev_key, k], dim=1)
+                k = torch.cat([prev_key, prepare_head(self.k_lin(key))], dim=1)
+        else:
+            k = prepare_head(self.k_lin(key))
         if 'prev_value' in incr_state:
             prev_value = incr_state['prev_value'].view(
                 batch_size * n_heads, -1, dim_per_head
@@ -203,7 +203,9 @@ class MultiHeadAttention(nn.Module):
             if static_kv:
                 v = prev_value
             else:
-                v = torch.cat([prev_value, v], dim=1)
+                v = torch.cat([prev_value, prepare_head(self.v_lin(value))], dim=1)
+        else:
+            v = prepare_head(self.v_lin(value))
         if 'prev_mask' in incr_state:
             if static_kv:
                 mask = incr_state['prev_mask']


stephenroller (approved):
Hah, awesome! Thanks! Do you think you could add some comments explaining this logic?

Tests are flaky but just rerunning to make sure this isn't catching on something wild.

stephenroller:
heads up @EricMichaelSmith might help with latency on [internal thing]

stephenroller:
On a volta with bs 1.
With a taskmaster2 BART model:
master: 50 generations in 31.4s
this patch: 50 generations in 29.0s, or about 7% speedup

With blender3b:
master: 50 generations in 74s
this patch: 50 generations in 67s, or about a 9% speedup.

on cpu with bs1 with taskmaster 2 BART model:
master: 50 generations in 59.0s
this patch: 50 generations in 54.2, or about 8% speedup

Real nice find!

stephenroller:
benchmark script i used:
```python
#!/usr/bin/env python3
import tqdm

from parlai.core.agents import create_agent_from_model_file

agent = create_agent_from_model_file('zoo:blender/blender_3B/model')

import timeit

start = timeit.default_timer()
for i in tqdm.tqdm(range(50)):
    agent.observe(
        {'text': 'Can you find me a mexican restaurant for 2?', 'episode_done': True}
    )
    print(agent.act()['text'])
    agent.reset()
end = timeit.default_timer()

print(end - start)
```

dejk-alg:
@stephenroller Stephen, thanks for having a look at this! Glad that I could participate here in some way. The logic behind this fix is pretty simple. Let me reiterate on what I said in the PR description a bit. The reason behind incremental state usage is obviously saving keys and values for attention to reuse them. When running inference of any encoder-decoder transformer, on every decoding step we need to get both encoder and self-attention keys and values for each decoder layer. Talking specifically about encoder_attention, general logic behind the module was totally fine - if it's our first decoding step and incr_state is still None, values are calculated from encoder_state tensors, otherwise we can just reuse the ones we calculated before. But there was a bug in implementation that forced MultiHeadAttention to recalculate this values despite replacing them with completely similar values saved in incr_state afterwards. So it went like "Let's get processed encoder tensors - Oh, nevermind, we've already had them". It's quite an obvious thing, but it doesn't strike the eye much unless you specifically check model logic to look for the ways to optimize latency

dejk-alg:
If you are interested, I've spent a bit of time with blender model, and while setting I work with is pretty different from what you work with here in the lib, I think I could add a couple of optimization changes to transformer modules. But all of them would generally be longer in terms of code and not that large in terms of latency decrease. So I'd love to know whether you think it's worth it, given that this changes will have to be reviewed by you or your team anyway

EricMichaelSmith (approved):
Great find - thanks for changing!

stephenroller:
> If you are interested, I've spent a bit of time with blender model, and while the setting I work with is pretty different from what you work with here in the lib, I think I could add a couple of optimization changes to transformer modules. But all of them would generally be longer in terms of code and not that large in terms of latency decrease. So I'd love to know whether you think it's worth it, given that these changes will have to be reviewed by you or your team anyway

I'm definitely interested. Lowering latency would be nice for sure. We do have maintainability and backwards compatibility to worry about, so I can't promise they would all be accepted. We can also discuss your proposals before you implement them. Feel free to file an issue.

Issue #3507, opened 2021-03-10
Unlikelihood Agent


mvh57:
Hi @hadasah and @stephenroller, following up on this issue from September (#2966), I am wondering if you've had a chance to update the unlikelihood model so that the reward parameter can be non-binary (ie, take on integer or float value, including 0). Thanks for your help with this. I would also be happy to try to help with this, but would need some guidance as to where to begin.

stephenroller:
#3517

mvh57:
@stephenroller, thank you for making these changes! One remaining question we have is: will these changes carry though to the loss function for the unlikelihood agent (for example --> will a reward: -0.9 be more down weighted than a reward: -0.5; also, will reward: 0 be parsed as such and be effectively ignored by the agent?)

stephenroller:
No, it will not.

You'll need to add some multipliers for the value here:

https://github.com/facebookresearch/ParlAI/blob/637ee9288f8c702272e1c602cbcaeca2a9f61ce8/projects/dialogue_unlikelihood/agents.py#L105

mvh57:
@stephenroller thank you!

jakeane:
Hi @stephenroller. I am working with @mvh57, and I am currently writing the additions to allow the weights to be nonbinary.  I have some questions before I go through the process detailed in `CONTRIBUTING.md`:

  - Currently, these additions would entail copying the `RewardUnlikelihoodAgentTrait` class and simply modifying the `compute_loss` method, specifically 'mle_loss' and 'ul_loss'. Then an agent will take in this modified trait. Is this the approach you would prefer in the codebase?
  - As for naming this new trait and agent, how would you like them named? Is something like `RewardWeightedUnlikelihoodAgentTrait` and `TransformerWeightedUnlikelihoodAgent` solid or too verbose?
  - I noticed this comment here https://github.com/facebookresearch/ParlAI/blob/637ee9288f8c702272e1c602cbcaeca2a9f61ce8/projects/dialogue_unlikelihood/agents.py#L96 and I was wondering if my modified trait class should handle such instances, where the teacher only provides rewards of 0. My current implementation would cause the `mle_loss = 0` in such cases. Is this fine? as this trait should not be used with such teachers.

jakeane:
I suppose I should also ping @hadasah on this, as it seems that she also contributed to this unlikelihood trait. I also I have one more question about this line:
[...]

Issue #2966, opened 2020-08-11
Teaching Model to Down-Weight Bad Responses


mvh57:
Hello all,
I want to teach the transformer / generator model to avoid (down-weight) certain types of candidate responses by showing it examples of such responses. Is there a way in the Parlai framework that I could go about doing this -- I'm thinking something like showing the model a training set with stereotypically undesirable responses (somehow indicating to the model that these are undesirable), so it can learn to down-weight these types of outputs? 
I've read through the training parameters, particularly those specific to training tasks, and I can't see a way to do this (I was considering possibly inputing a negative multi-task weight for the task containing bad responses, but not sure this would work). 
Thanks for your help!

stephenroller:
You might interested in Unlikelihood training:
- https://arxiv.org/abs/1908.04319
- https://arxiv.org/abs/1911.03860
- https://parl.ai/projects/dialogue_unlikelihood/

Specifically the reward agent:
https://github.com/facebookresearch/ParlAI/blob/2739c2304c6a6ae30eb61633b4ca5102909e4a94/projects/dialogue_unlikelihood/agents.py#L49-L55

@hadasah can perhaps share an example of training with it.

mvh57:
Thank you so much. Yes, an example would be very helpful!

mvh57:
Hi @stephenroller and @hadasah,
Just following up on this issue to see if it would be possible to get an example of how to use unlikelihood training (i.e., collect training examples of offensive / bad responses and train the model not to make such responses). I've read the papers you've attached @stephenroller, but am still having difficulty determining which scripts I need to modify (I assuming I need to specify a different loss function for the model to use when training on offensive data, but am not sure how to go about doing this.) Thanks so much for your ongoing help.

hadasah:
Hi @mvh57! So sorry about the delay. I started working on better documentation but haven't finished/merged it in yet. You can see it in [draft state](https://github.com/facebookresearch/ParlAI/blob/6bdfc06e0687eee1d2cdbe0e48bf4fae10c0b815/projects/dialogue_unlikelihood/README.md), but unfortunately one of the parts I'm still working on is the reward unlikelihood agent (relevant to you).  

[...]

PR #3552, opened 2021-03-25
Add weighted version of TransformerUnlikelihoodAgent

jakeane:
**Patch description**

Added an agent for unlikelihood dialogue, `TransformerWeightedUnlikelihoodAgent` that allows rewards to be weighted as opposed to binary. This involved creating a new trait, `RewardWeightedUnlikelihoodAgentTrait`, that extends the functionality of `RewardUnlikelihoodAgent` such that the loss for each batch example is multiplied by `batch.rewards`.

This addition will allow parts of the training data to be selectively ignored or emphasized, thus allowing the model to be better tailored.

**Testing steps**

Not entirely sure how agents under `projects/dialogue_unlikelihood` are tested, but I ran these two test commands. 
```
pytest -k TestTorchAgent
pytest -k TestTransformer
```
The first ran successfully, but the second failed the tests for `test_wide_distillation_losses` and `test_narrow_distillation_losses` under `tests/nightly/gpu/anti_scaling/test_anti_scaling.py`. *However*, these tests also failed under the `master` branch, so I do not think it is related to my addition.

**Related Issues**

 - [#2966](https://github.com/facebookresearch/ParlAI/issues/2966)
 - [#3507](https://github.com/facebookresearch/ParlAI/issues/3507)

### DIFF
--- a/projects/dialogue_unlikelihood/agents.py
+++ b/projects/dialogue_unlikelihood/agents.py
@@ -535,6 +535,95 @@ class SequenceVocabUnlikelihoodAgentTrait(_VocabUnlikelihoodTrait):
             return loss
 
 
+class RewardWeightedUnlikelihoodAgentTrait(object):
+    """
+    Abstract Trait.
+
+    Applies unlikelihood loss based on the presence of negative rewards in the task.
+    Reward is weighted as opposed to `RewardUnlikelihoodAgentTrait`.
+    """
+
+    @classmethod
+    def add_cmdline_args(
+        cls, parser: ParlaiParser, partial_opt: Optional[Opt] = None
+    ) -> ParlaiParser:
+        grp = super().add_cmdline_args(parser, partial_opt=partial_opt)
+        grp.add_argument('--alpha', default=1.0, type=float)
+        return parser
+
+    def _dummy_batch(self, batchsize, maxlen):
+        batch = super()._dummy_batch(batchsize, maxlen)
+        batch['rewards'] = torch.ones(batchsize, dtype=torch.long).cuda()
+        return batch
+
+    def compute_loss(self, batch, return_output=False):
+        if batch.label_vec is None:
+            raise ValueError('Cannot compute loss without a label.')
+
+        model_output = self.model(*self._model_input(batch), ys=batch.label_vec)
+        scores, preds, *_ = model_output  # scores is bsz x time x vocab
+
+        scores = F.log_softmax(scores, dim=-1)
+        scores_view = scores.view(-1, scores.size(-1))
+        targets = batch.label_vec
+        targets_view = targets.view(-1)
+
+        notnull = targets.ne(self.NULL_IDX)
+        if self.is_training:
+            # note it's >= because convai2 and other teachers all provide a 0 reward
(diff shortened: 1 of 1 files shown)


projects/dialogue_unlikelihood/agents.py
```diff
@@ -535,6 +535,100 @@ def compute_loss(self, batch, return_output=False):
...
+        ul_scores = scores_view[range_, targets_view]
+        clamp_min = 1e-6 if self.opt['fp16'] else 1e-20
+        ul_loss = torch.mv(
+            (
+                torch.log(torch.clamp(1.0 - ul_scores.exp(), min=clamp_min)).view_as(
+                    ul_notnull
+                )
+                * ul_notnull.float()
+            ).T,
+            batch.rewards,
```
stephenroller, line 618: I think you want to negate this here

projects/dialogue_unlikelihood/agents.py
```diff
@@ -535,6 +535,100 @@ def compute_loss(self, batch, return_output=False):
...
+            else:
+                return mle_loss
+
+        # and now we want the unlikelihood loss on the negative examples
+        ul_notnull = notnull & (batch.rewards < 0).unsqueeze(1).expand_as(notnull)
+        ul_target_tokens = ul_notnull.long().sum()
+        range_ = torch.arange(targets_view.size(0)).to(batch.label_vec.device)
+        ul_scores = scores_view[range_, targets_view]
+        clamp_min = 1e-6 if self.opt['fp16'] else 1e-20
+        ul_loss = torch.mv(
```
stephenroller, line 611: both here and in MLE, it looks like you're doing a matrix-vector multiplication to cast the rewards across the entire sequence. This likely means that the per-token averages of `nll_loss`, `ppl`, and `ul_loss` are actually per-sequence averages. You probably wish to check this.

stephenroller (commented):
Left some comments to help your exploration. Given that we don't have an official usage of this right now, I don't think we'll be accepting this upstream at this moment; in order to combat bloat, I try encourage experimental things to stay in forks or internal copies until they've been proven empirically.

Should you present or find reference to strong empirical results on this formulation, then I'll be more than happy to support this, given the addition tests.

In the mean time, I'm happy to continue to review this code to help you implement this for your own experimentation.

projects/dialogue_unlikelihood/agents.py
jakeane: I don't think so, as `batch.rewards` would handle the negation instead of adding a `-` in front of `torch.log()` like in `RewardLikelihoodAgentTrait`

projects/dialogue_unlikelihood/agents.py
jakeane: The calculation of `mle_target_tokens` and `ul_target_tokens` are unchanged from `RewardLikelihoodAgentTrait`, and when I tested the values of `mle_loss` and `ul_loss` when the rewards are 1 and -1, the calculated values are the same as those in `RewardLikelihoodAgentTrait`. Similarly, when the rewards are 0.5 and -0.5 for example, the calculated values are exactly half of those calculated in `RewardLikelihoodAgentTrait`. So, I am not exactly sure how this would result in `nll_loss`, `ppl`, and `ul_loss` to be per-sequence averages as opposed to per-token averages.

PR #3559, opened 2021-03-28
Retire DrQA agent

stephenroller:
**Patch description**
This is one of our oldest agents. While it still can work in some scenarios, I would like to remove the burden of continued maintenance.

Follow our standard process of archiving old code, with READMEs pointing to the old tag.

**Testing steps**
CI


EricMichaelSmith (approved):
Seems reasonable

PR #3562, opened 2021-03-30
Add --wandb-entity flag to the logger

akhti:
**Patch description**
Allows setting WandB's `entity=` parameter via command line flags.
See https://docs.wandb.ai/ref/run/init

**Testing steps**
Manual testing in a sweep


stephenroller (approved):
Thanks! Rerunning CI but lgtm

PR #3567, opened 2021-04-01
Swappable Transformer Components

spencerp:
### Problem
Right now it's cumbersome to override some specific part of a transformer. If you want to significantly modify the encoder self attention, for example, you either need to add a bunch of conditional branching in `MultiHeadAttention` or subclass `MultiHeadAttention` _and_ `TransformerGeneratorModel` _and_ `TransformerEncoder` _and_ `TransformerEncoderLayer`.

### Purpose of this PR
Allow swapping out of any `nn.Module` in `TransformerGeneratorModel` while maintaining backwards compatibility with existing call sites.

### NOT the Purpose of this PR
Reducing subclassing to improve readability, nor making architectures infinitely composable.

### Overview
The main files to look at are:
- `parlai/agents/examples/transformer_variant.py`: demonstrates swapping out various components in TransformerGeneratorModel
- `parlai/agents/transformer/modules/modular.py`: defines and explains the new classes introduced

A brief summary of the pattern introduced:
- *ModularComponent*: Any component with subcomponents that can be swapped out.
- *Subcomponents*: A catalog of all swappable `nn.Module`s in a `ModularComponent`. Replace any class in the subcomponents with one that shares the same `__init__` and `forward` signatures and it should just work.
- To modify some business logic in a model: subclass (e.g. modifying `TransformerEncoder.forward_embedding`).
- To replace an entire component: use the `Subcomponents` (e.g. replacing `MultiHeadAttention` in `TransformerDecoderLayer`).
- *ModularComponentSpec*: A convenience object for specifying a `ModularComponent` along with a `Subcomponents` for building its subcomponents.
- Together these can be thought of as a graph/tree that fully defines an architecture. If you find yourself adding new components to the template, you might want to create a new `Module`/`ModularComponent` instead.


Some downsides:
- It does not explicitly allow for fancier, dynamic architecture changes like using a different layer class for for each layer. But this could still be accomplished by passing in a custom layer class that behaviors differently depending on the layer. If we find ourselves wanting super dynamic composable architectures often we can revisit this.
- The code that initializes each component gets a little harder to read. `TransformerDecoderLayer` is easier to understand than `swappables.layer`.


### Testing
Added a print statement in `MyCustomEncoder.forward` and verified it showed in stdout:
```
λ parlai train_model --model examples/transformer_variant --task convai2 --model-file /tmp/testtransformer --beam-size 5 --batchsize 16
...
10:42:49 | training...                                                                                                         
12:54:54 | Custom encoder called!                                                                                              
12:54:55 | Custom attention called!                                                                                                                                                                                                                           
12:54:55 | Custom attention called!                                                                                            
12:54:55 | Custom encoder called!
...
```
CircleCI


parlai/agents/transformer/modules/decoder.py
```diff
@@ -20,12 +22,160 @@
...
     TransformerFFN,
 )
+from parlai.agents.transformer.modules.interfaces import ComponentSpec, TComponent
 from parlai.core.opt import Opt
 from parlai.utils.misc import warn_once
 from parlai.utils.torch import PipelineHelper
 
 
-class TransformerDecoder(nn.Module):
+class TransformerDecoderLayer(nn.Module, TComponent):
```
spencerp, line 31: Now that we're actually _calling_ an attribute of this class later in the file instead of just referencing it as a type annotation we need to move it up.

parlai/agents/transformer/modules/generator.py
```diff
@@ -27,46 +29,76 @@
...
+        cls,
+        opt,
+        dictionary,
+        embedding=None,
+        padding_idx=None,
+        reduction_type='mean',
+        spec: ComponentSpec = ENCODER_DEFAULT_SPEC,
     ):
-        return TransformerEncoder(
+        return spec.klass(
```
jxmsML, line 63: maybe `spec.encoder.klass(` ?

parlai/agents/transformer/modules/generator.py
```diff
@@ -27,46 +29,76 @@
...
+            manifest=spec.manifest,
         )
 
     @classmethod
-    def build_decoder(cls, opt, embedding=None):
-        return TransformerDecoder(opt=opt, embedding=embedding)
+    def build_decoder(
+        cls, opt, embedding=None, spec: ComponentSpec = ENCODER_DEFAULT_SPEC
+    ):
+        return spec.klass(opt=opt, embedding=embedding, manifest=spec.manifest)
```
jxmsML, line 76: same as above: `spec.decoder.klass`

parlai/agents/transformer/modules/generator.py
spencerp: See the `__init__` function below. When `build_encoder` is called, it's passed `manifest.encoder`. And `manifest.encoder` only has `klass` and `manifest`, not `encoder`.

parlai/agents/examples/transformer_variant.py
```diff
@@ -0,0 +1,31 @@
...
+
+class MyCustomEncoder(TransformerEncoder):
+    def forward(  # type: ignore
+        self,
+        input: torch.LongTensor,
+        positions: Optional[torch.LongTensor] = None,
+        segments: Optional[torch.LongTensor] = None,
+    ) -> Union[torch.Tensor, Tuple[torch.Tensor, torch.BoolTensor]]:
+        # Comment out the following line and write your custom `forward` instead.
+        return super().forward(input, positions, segments)
```
jxmsML, line 31: is it convenient to provide an example where the leaf level manifest is swapped: e.g. having a custom MultiHeadAttention in `TransformerDecoderLayer` while the rest (other parts of `TransformerDecoderLayer`, `TransformerGenratorModel`, `Encoder` ...) stay the same.

`replace(TransformerDecoder.Manifest().layer, MyCustomTransformerDecoderLayer)`
where MyCustomTransformerDecoderLayer has manifest.self_attention default value = CustomMultiHeadAttention
or maybe sth like 
`replace(TransformerDecoder.Manifest().layer.Manifest().self_attention,  ???)`

jxmsML (approved):

jxmsML:
LGTM!!! I tried swapping in my custom encoder and works perfect 💯 🚀  Great to have this!!

parlai/agents/transformer/modules/decoder.py
```diff
@@ -38,13 +188,21 @@ class TransformerDecoder(nn.Module):
     :param int n_positions: Size of the position embeddings matrix.
     """
 
+    @dataclass
+    class Manifest(TComponent.Manifest):
+        layer: ComponentSpec[TransformerDecoderLayer] = ComponentSpec(
+            TransformerDecoderLayer, TransformerDecoderLayer.Manifest()
```
jxmsML, line 194: I guess one thing I'm confused about is the relationship between `klass` and `manifest`, e.g., is the following legit?
```
ComponentSpec(
            TransformerDecoderLayer, CustomTransformerDecoderLayer.Manifest()
)
```

jxmsML: I guess you've also mentioned in TODO, but it's cleaner to have a 1-1 mapping between `klass` and `manifest`, is that what you have in mind?

parlai/agents/examples/transformer_variant.py
spencerp: Good point, I'll add an example!
You could do either of the options you mentioned, but I had envisioned something more like 2 for simply swapping out the leaf component.

parlai/agents/transformer/modules/decoder.py
spencerp: So yeah, one thing you could do would be to create a new Manifest:
```
@dataclass
class CustomTransformerDecoderLayer(TransformerDecoderLayer.Manifest):
  # override some default values

ComponentSpec(
  TransformerDecoderLayer,
  CustomTransformerDecoderLayer(),
)
```

So I think the value in not having a 1-1 mapping between `klass` and `manifest` is reducing the need to subclass.

But since the default case is a 1-1 mapping, I'd like to make that case less verbose (which is what that comment is about).

parlai/agents/transformer/modules/encoder.py
```diff
@@ -20,12 +22,75 @@
...
 
-class TransformerEncoder(nn.Module):
+class TransformerEncoderLayer(nn.Module, TComponent):
+    """
+    Implements a single Transformer encoder layer.
+    """
+
+    @dataclass
+    class Manifest(TComponent.Manifest):
+        self_attention: Type[MultiHeadAttention] = MultiHeadAttention
```
klshuster, line 38: why are these not component specs?

parlai/agents/transformer/modules/encoder.py
```diff
@@ -20,12 +22,75 @@
...
+        )
+        self.norm2 = torch.nn.LayerNorm(embedding_size, eps=LAYER_NORM_EPS)
+        self.dropout = nn.Dropout(p=dropout)
+
+    def forward(self, tensor: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
+        """
+        Forward pass.
+        """
+        residual = tensor
+        if self.variant == 'prelayernorm':
```
klshuster, line 77: one could also view the different variants as custom variations, right?

klshuster (commented):
for some reason i'm not 100% grasping the interplay between Manifests and ComponentSpecs, but here's my current interpretation:

- A `Manifest` defines the components of a `TComponent`. So like the encoder manifest defines the layers.
- A `ComponentSpec` defines the module class itself, and the respective manifest for the class.

I like how it's quite customizable if you know what you're doing. It still seems to suffer from a the downsides of a hierarchical setup (you need to traverse quite a few things to find out the different manifests, etc.), but it might be a better tradeoff than what we have already

parlai/agents/transformer/modules/decoder.py
spencerp: That'd be a more accurate name, yes :)

parlai/agents/transformer/modules/encoder.py
jxmsML: my guess is MHA is considered bottom-level module, with manifest  = None (no swappable components)

spencerp: Yep, what Jing said. `MultiHeadAttention` isn't a `TComponent` at this point because there aren't any swappable components within it.

Maybe there's a better name for `TComponent` and `ComponentSpec` that makes this clearer. Like... `TComponent` -> `Composable`, `ComponentSpec` -> `ComposeKit`?

Idk I'm very open to suggestions. 

The idea I'm trying to convey is:
- A "TComponent"/"Composable"/"Modular" component is a piece of the model with modular pieces that can be substituted with anything that conforms to the same `__init__` and `forward` as the original piece.
- To fully specify the "TComponent"/"Composable" component, then, you need to specify both the base class and the class of the pieces that will be plopped into the various slots within it.
- The "TComponent"/"Composable" components come with a default set of pieces so you don't have to think about this until you want to change something.
- A "ComponentSpec"/"ComposeKit" is a fully specified "TComponent"/"Composable", in that it describes both the base class and all the pieces that fit into its slots.

spencerp: "Mad Libs" are a fantastic analogy and it's kind of cool that it hints at the idea that code and natural language are interchangeable (a sequence is a sequence). But idk if `PhrasalTemplate`, `Word`, and `Phrase` are any more intuitive than `TComponent`, `Module`, and `ComponentSpec` lol

https://en.wikipedia.org/wiki/Phrasal_template

parlai/agents/transformer/modules/encoder.py
spencerp: For sure! I didn't do anything like that as part of this PR to keep it simple, but I could imagine a pretty simple way to implement that without this `if` statement:
```
class PreLayerNormTransformerEncoderLayer(TransformerEncoderLayer):
  def forward(self, tensor: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
    residual = tensor
    tensor = self.norm1(tensor)
    attended_tensor = self.attention(tensor, mask=mask)[0]
    tensor = residual + self.dropout(attended_tensor)
    residual = tensor
    tensor = self.norm2(tensor)
    tensor = residual + self.dropout(self.ffn(tensor))
    tensor *= mask.unsqueeze(-1).type_as(tensor)
    return tensor

manifest.encoder_layer = ComponentSpec(
  PreLayerNormTransformerEncoderLayer,
  TransformerEncoderLayer.Manifest(),
)
```

To me the code feels simple enough as is for now, but if we end up with a ton of variations then maybe we'll want to split them out into different implementations like this.

parlai/agents/examples/transformer_variant.py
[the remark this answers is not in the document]
spencerp: Making components swappable requires that they have some agreed-upon args. Otherwise it's highly unlikely that the components will actually be interchangeable.

That being said, I do think we should include `**kwargs` in the `__init__` and `forward` functions so that we can experiment easily. And once the dust settles on those experiments we can decide how we want to handle any new args that turned out to be useful.

spencerp:
Thank you both so much for your comments!!

> for some reason i'm not 100% grasping the interplay between Manifests and ComponentSpecs, but here's my current interpretation:

Your summary is accurate. A `TComponent` is like a template, a `Manifest` is like a form to fill out with all of the slots in that template, and a `ComponentSpec` combines these two to fully define a component.

> It still seems to suffer from a the downsides of a hierarchical setup (you need to traverse quite a few things to find out the different manifests, etc.), but it might be a better tradeoff than what we have already

This PR is just meant to solve the problem of "how do I swap out a piece of the Transformer without copying and pasting the whole dang file??". I think getting rid of the hierarchy traversal altogether would require a radical refactor such that none of our `nn.Module`s initialize new `nn.Module`s in their `__init__`. 

Maybe another option to get around the traversal issue without a radical refactor could be making the `Manifest` immutable and final (not subclassable). That would enforce that anyone who customizes a component has to write out the whole manifest, not just their incremental change. That would be quite verbose, though.

parlai/agents/examples/transformer_variant.py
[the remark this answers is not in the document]
klshuster: yes, agreed that the standard args should remain the same, with _only_ the possible addition of `**kwargs`

parlai/agents/transformer/modules/encoder.py
klshuster: I think i prefer "Composable"/"Modular" to component, as it's not immediately clear why a component of a transformer may not be composable (e.g. in the MHA case). Perhaps we can specify even further to differentiate those components. In that regard, you might have something like:

"ModularComponent": A piece of the model with modular pieces etc. etc.
"StaticComponent": A piece of the model that does not have modular pieces etc. etc.
"ModularComponentSpec": The specification of the modular component

Does that make sense in this setup?

parlai/agents/transformer/modules/encoder.py
klshuster: yes agreed, not in the scope of this current PR but could be an interesting application from it

klshuster:
> This PR is just meant to solve the problem of "how do I swap out a piece of the Transformer without copying and pasting the whole dang file??". I think getting rid of the hierarchy traversal altogether would require a radical refactor such that none of our nn.Modules initialize new nn.Modules in their __init__.

Agreed, let's deal with that separately. I do think this is more flexible than before

>Maybe another option to get around the traversal issue without a radical refactor could be making the Manifest immutable and final (not subclassable). That would enforce that anyone who customizes a component has to write out the whole manifest, not just their incremental change. That would be quite verbose, though.

Verbosity, if it improves clarity, may not be the worst thing, but if it requires more work from the user it may not be super desirable

parlai/agents/transformer/modules/encoder.py
spencerp: I like that naming better! Changing it now.

"StaticComponent" might be implied and not literal, though. Unless we can figure out a way to use a "StaticComponent" type to _enforce_ that an nn.Module not be composed.

spencerp:
> Verbosity, if it improves clarity, may not be the worst thing, but if it requires more work from the user it may not be super desirable

Take a look at `examples/transformer_variant.py` now. I tried a more verbose way of specifying the components, and I think it makes it easier to get a wholistic view of the architecture of the model.

parlai/agents/transformer/modules/encoder.py
klshuster: I'm ok with it just being semantic in meaning (as Python makes it hard to really enforce anything lol)

klshuster:
> > Verbosity, if it improves clarity, may not be the worst thing, but if it requires more work from the user it may not be super desirable
> 
> Take a look at `examples/transformer_variant.py` now. I tried a more verbose way of specifying the components, and I think it makes it easier to get a wholistic view of the architecture of the model.

Yes, I like that a lot actually; obviously it may get a bit more complicated with more complicated models but that might be a good thing! Since you have everything laid out in one specification

parlai/agents/transformer/modules/encoder.py
spencerp: Okay! I added a `StaticComponent` class that basically just acts as a comment.

parlai/agents/transformer/modules/encoder.py
```diff
@@ -141,7 +218,7 @@ def _default(val, default):
         self.layers = nn.ModuleList()
         for _ in range(self.n_layers):
             self.layers.append(
-                TransformerEncoderLayer(
+                template.layer.build(
```
jaseweston, line 221: this is really hard to read.. can't it be called encoderlayer_class or something (assuming that's what it is)?  template is so generic...

parlai/agents/transformer/modules/interfaces.py
```diff
@@ -0,0 +1,66 @@
...
+class ModularComponentBuilder(Generic[MC]):
+    """
+    When a component has swappable subcomponents, use this object to specify both the
+    component type and it's subcomponent types at the same time.
+    """
+
+    def __init__(
+        self, klass: Type[MC], template: Optional[ModularComponent.Template] = None
+    ) -> None:
+        self._klass = klass
```
jaseweston, line 41: what is 'klass' (sorry)?

parlai/agents/transformer/modules/generator.py
```diff
@@ -51,22 +72,38 @@ def build_encoder(
...
+        return builder.build(opt=opt, embedding=embedding)
+
+    def __init__(
+        self, opt: Opt, dictionary: DictionaryAgent, template: Optional[Template] = None
+    ):
         self.pad_idx = dictionary[dictionary.null_token]
         self.start_idx = dictionary[dictionary.start_token]
         self.end_idx = dictionary[dictionary.end_token]
         super().__init__(self.pad_idx, self.start_idx, self.end_idx)
+        template = template or self.Template()
```
jaseweston, line 90: ditto

parlai/agents/transformer/modules/interfaces.py
spencerp: It's type-annotated in `__init__` for reference, but it's a class. Spelled with a `k` so it doesn't clash with the python reserved word `class`. So `TransformerEncoderLayer`, for example.

parlai/agents/transformer/modules/encoder.py
spencerp: That's pretty much what it was called in the previous commit lol. The more I stare at and play with the code the less sure I am about what's intuitive and what's confusing.

parlai/agents/transformer/modules/attention.py
```diff
@@ -14,10 +14,11 @@
 import torch.nn as nn
 import torch.nn.functional as F
 
+from parlai.agents.transformer.modules.interfaces import StaticComponent
 from parlai.utils.torch import neginf
 
 
-class BasicAttention(nn.Module):
+class BasicAttention(StaticComponent):
```
stephenroller, line 21: I think this is the only thing I don't love of the latest version. are the StaticComponent annotations necessary?

parlai/agents/examples/transformer_variant.py
```diff
@@ -0,0 +1,206 @@
...
+    Doesn't make any changes to TransformerGeneratorModel, just specifies all
+    subcomponents explicitly.
+
+    This is meant to be a reference for how to swap any component within
+    TransformerGeneratorModel.
+    """
+
+    def build_model(self, states=None):
+        wrapped_class = TransformerGeneratorModel.with_components(
+            encoder=TransformerEncoder.with_components(
```
stephenroller, line 126: yes this is the one

parlai/agents/transformer/modules/interfaces.py
```diff
@@ -0,0 +1,147 @@
...
+    ```
+    """
+
+    def wrap(cls) -> Type:
+        return _make_class_swappable(cls, **kwargs)
+
+    return wrap
+
+
+def _make_class_swappable(cls: Type, **kwargs) -> Type[MC]:
```
stephenroller, line 117: can you leave some short internal docs (comments) describing how this whole thing works?

parlai/agents/examples/transformer_variant.py
```diff
@@ -0,0 +1,206 @@
...
+from parlai.core.params import ParlaiParser
+import parlai.utils.logging as logging
+
+
+###########################################
+# Transformer With Two Components Swapped #
+###########################################
+
+
+class TransformerVariantAgent(TransformerGeneratorAgent):
```
stephenroller, line 45: Small unit tests for all of these please. Just something like


```python
opt = ParlaiParser(True, True).parse_kwargs(
    model='parlai.agents.examples.transformer_variant:TransformerVariantAgent',
    ffn_size=16,
    embeddingsize=16,
    dict_file='zoo:unittest/transformer_generator2/model.dict',
    skip_generation=True
)
model = create_agent(opt)
# send the model a single training example to ensure it can forward/backward
model.observe({'text': '1 2 3 4', 'labels': ['1 2 3 4'], 'episode_done': True})
model.act()
# send the model a single validation example
model.observe({'text': '1 2 3 4', 'eval_labels': ['1 2 3 4'], 'episode_done': True})
model.act()
```

stephenroller (commented):
Needs a tutorial in website docs showing essentially the same as examples, but with more prose.

stephenroller:
(I like this final form a lot)

parlai/agents/transformer/modules/attention.py
spencerp: Not at all, it's just to differentiate between `ModularComponent`. Now that we've hidden the `ModularComponent` class, it probably makes sense to get rid of this.

cc @klshuster in case you feel strongly about keeping it

klshuster: don't feel strongly, feel free to get rid of it

parlai/agents/examples/transformer_variant.py
klshuster: agreed, this looks nice

docs/source/tutorial_swap_components.md
```diff
@@ -0,0 +1,60 @@
+# Swapping Out Model Subcomponents
```
stephenroller, line 1: say Transformer explicitly, since this only works for Transformers

stephenroller (approved):

docs/source/tutorial_swap_components.md
spencerp: It technically can be added to any model (or any class at all, really). But I'll take your advice to make it easier to think about for now.

PR #3569, opened 2021-04-02
Switched logging.warn to logging.warning

AlexAzaFB:
**Patch description**
Currently there are two ways to log warnings: logging.warn() and logging.warning().
This PR switches all logging.warn() usages to logging.warning() and removes logging.warn() implementation.
Besides removing ambiguity, this change is also inline with Python 3.3 warn vs warning naming.

**Testing steps**
Unit tests.

### DIFF
--- a/parlai/chat_service/services/websocket/sockets.py
+++ b/parlai/chat_service/services/websocket/sockets.py
@@ -23,7 +23,7 @@ class MessageSocketHandler(WebSocketHandler):
         self.subs: Dict[int, T] = kwargs.pop('subs')
 
         def _default_callback(message, socketID):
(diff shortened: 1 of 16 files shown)


stephenroller (commented):
Can we keep the logging.warn as available in the library for now, and upgrade it to call logging.warning? There are some external customers that may still be using it. 

Also please upgrade the call to logging.warn in logging.py.

stephenroller (approved):
Great, thanks for doing this! Looks perfect.

stephenroller:
The segfault in gpu tests isn't related, but hold on merging a bit please.

stephenroller:
Actually your branch is from an old parent (np) so this isn't unexpected. You're good to merge whenever.

PR #3573, opened 2021-04-07
Prevent unnecessary log

stephenroller:
**Patch description**
Minor annoyance where -lstep and -vstep would combine to an unnecessary log sometimes.

**Testing steps**
Manual testing


stephenroller:
Supplanted by #3599

PR #3575, opened 2021-04-08
[bg] Catch exceptions in workers

stephenroller:
**Patch description**
If exceptions happen in (all) background workers, currently training just hangs indefinitely while we wait for a batch from the queue. This patch adds a timeout, and checks background workers for exceptions when that timeout is hit. I had previously tested checking on these exceptions _before_ pulling from the queue, but it was devastating to speed. This patch seems to strike a good balance.

In manual testing, this patch displayed these properties:
- If _some_, but not all workers threw exceptions, then training continued with just fewer workers. _Sometimes_ there was a print of an exception, but not always. Speed was similar to launching with an equivalent smaller `--num-workers`.
- If _all_ workers threw exceptions, then there was a LOT of log spew and the program ended. But the log spew did contain the root exception in it and thus a hint for the developer.
- If _no_ workers threw exceptions, speed was the approximately same as before this patch (within random variation on an 8 gpu 8 worker 10 minute run)

**Testing steps**
Considerable manual testing.

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -1282,8 +1282,18 @@ class BackgroundDriverWorld(World):
     def num_episodes(self):
         return self.world.num_episodes()
 
+    def _queue_get(self):
+        import queue
+
+        while True:
+            try:
+                return self._process_queue.get(timeout=10)
+            except queue.Empty:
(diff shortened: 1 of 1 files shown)


EricMichaelSmith (approved):
Makes sense, great!

PR #3577, opened 2021-04-09
Allow redirects when downloading data

spencerp:
**Patch description**
When I try to download http://parl.ai/downloads/blended_skill_talk/blended_skill_talk.tar.gz I get a redirect, which currently breaks `build.py`. After enabling redirect it works fine.

**Testing steps**
`parlai dd -t blended_skill_talk`

### DIFF
--- a/parlai/core/build_data.py
+++ b/parlai/core/build_data.py
@@ -168,7 +168,7 @@ def download(url, path, fname, redownload=False, num_retries=5):
 
         with requests.Session() as session:
             try:
-                response = session.get(url, stream=True, timeout=5)
+                response = session.get(url, stream=True, timeout=5, allow_redirects=True)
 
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
Weird, we've been using tons of redirects for a long time.

spencerp:
> Weird, we've been using tons of redirects for a long time.

I think there's some more investigation needed here. After this change, downloading blended_skill_talk worked for a couple downloads. Then it started failing with a new error message: `kErrorAddressPrivate`

spencerp:
This appears to be specific to the environment I was working in. I can't reproduce from my laptop, and this change doesn't ultimately fix the issue I was facing.

In the spirit of "if it ain't broke don't fix it" I'm not going to merge this.

PR #3578, opened 2021-04-09
added text to the history

mojtaba-komeili:
**Patch description**
flatten mutator was only keeping one side of the conversations in its history and there was losing the context of the conversation after flattening. Keeping the previous labels in the history for having for conversation context.

**Testing steps**
Used `parlai display_data` to compare before and after. See images below:

Before:
Screen Shot 2021-04-09 at 4 50 16 PM

After:
Screen Shot 2021-04-09 at 4 59 43 PM

### DIFF
--- a/parlai/mutators/flatten.py
+++ b/parlai/mutators/flatten.py
@@ -23,4 +23,5 @@ class FlattenMutator(ManyEpisodeMutator):
         for message in episode:
             history.append(message.pop('text'))
             message['text'] = '\n'.join(history)
+            history.extend(message["labels"])
             yield [message]


parlai/mutators/flatten.py
```diff
@@ -23,4 +23,5 @@ def many_episode_mutation(self, episode: List[Message]) -> List[List[Message]]:
         for message in episode:
             history.append(message.pop('text'))
             message['text'] = '\n'.join(history)
+            history.extend(message['labels'])
```
stephenroller, line 26: this should be after the yield, and it shouldn't extend with all of labels. It should pick a random element from labels.

stephenroller (commented):
You'll need to update the mutators tests

parlai/mutators/flatten.py
mojtaba-komeili: I can change it to pick one random element from the labels. 
Why putting it after yield matters? We have already used history and are not going to use it again until next iterator call. Does it have something to do with the another thread that possibly calls the function?

stephenroller: You're leaking the label!

stephenroller: Discussed offline why the implementation is correct, but potentially confusing.

stephenroller (approved):

PR #3585, opened 2021-04-13
Update wizard_of_wikipedia agents.py

Reveyer:
**Patch description**

When `wizard_first` is True, the `len_episode` is incorrectly determined, resulting in some data not being used.

**Testing steps**

I tested it using command 

> `parlai display_data --task wizard_of_wikipedia --num-examples 100000`

Before the change

> `loaded 18430 episodes with a total of 74092 examples`

After the change

> `loaded 18430 episodes with a total of 83247 examples`

**Other information**

### DIFF
--- a/parlai/tasks/wizard_of_wikipedia/agents.py
+++ b/parlai/tasks/wizard_of_wikipedia/agents.py
@@ -262,7 +262,7 @@ class WizardDialogKnowledgeTeacher(WizardOfWikipediaTeacher):
         d = self.data[ep]
         wizard_first = 'Wizard' in d['dialog'][0]['speaker']
         if wizard_first:
-            return (len(d['dialog']) - 1) // 2
+            return (len(d['dialog']) - 1) // 2 + 1
         return len(d['dialog']) // 2
 
     def num_examples(self):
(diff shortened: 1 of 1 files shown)


klshuster:
Hi @Reveyer, thank you for flagging this bug! In the interest of reproducibility, could you please make some minor modifications to the PR:

1. Create a subclass of `WizardDialogKnowledgeTeacher`, perhaps named `LongTeacher`, that overrides `len_episode` to fix the bug.
2. Somewhere in the `__init__` for the `WizardDialogKnowledgeTeacher`, please [loudly warn](https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/teachers.py#L456) that it is advised to use `-t wizard_of_wikipedia:long` to include the missing examples from the default teacher.

Thanks again!

klshuster:
Additionally, it would be great if we could limit this change to `--datatype train` (so, valid and test sets remain the same)

Reveyer:
> Additionally, it would be great if we could limit this change to `--datatype train` (so, valid and test sets remain the same)

I think adding a argument would better achieve what you said.

klshuster (approved):
thank you! yes, an argument works great here

parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -462,6 +485,15 @@ def add_cmdline_args(
...
             help='prepend chosen topic to first turn',
         )
+        agent.add_argument(
+            '--add-missing-turns',
+            type=str,
+            choices=['none', 'train', 'all'],
+            default='none',
+            help='For reproducibility, the default "none" is the previous version which misssing some data.'
+            'When "train" is chosen, only the training set is supplemented.'
+            'When "all" is chosen, all data are supplemented.',
```
stephenroller, line 495:
```suggestion
            help='For reproducibility, the default "none" is the previous version which missing some data. '
            'When "train" is chosen, only the training set is supplemented. '
            'When "all" is chosen, all data are supplemented.',
```

stephenroller (approved):
Greatly appreciated

PR #3591, opened 2021-04-14
[World Logs] Don't crash if a value can't be serialized

spencerp:
**Patch description**
`text_vec` and `full_text_vec` in the observations contain tensor and can't be json serialized in `eval_model --world-logs`. In this type of scenario, if we crash, we lose all of the logs and have to rerun. This fails more gracefully.

**Testing steps**
```
>>> d = {'a': tensor}
>>> json.dumps(d, default=lambda v: '<not serializable>')
'{"a": "<not serializable>"}'
```

### DIFF
--- a/parlai/utils/conversations.py
+++ b/parlai/utils/conversations.py
@@ -336,7 +336,7 @@ class Conversations:
                             convo['context'].append(turn)
                     if new_pair:
                         convo['dialog'].append(new_pair)
-                json_convo = json.dumps(convo)
+                json_convo = json.dumps(convo, default=lambda v: '<not serializable>')
                 f.write(json_convo + '\n')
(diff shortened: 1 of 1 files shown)


stephenroller (commented):
hmm i guess this is good, vaguely wish we could filter instead of dumping a thousand "not serializable" strings

stephenroller (approved):

spencerp:
> hmm i guess this is good, vaguely wish we could filter instead of dumping a thousand "not serializable" strings

We could filter them. My thought was that it'd be useful to have this feedback in case you _did_ want a field but forgot to convert it into a serializable form. Like if you put in a metric but forgot to call .value.

stephenroller:
We can go with this and see if gets in the way

PR #3593, opened 2021-04-14
CMU_DoG

spencerp:
**Patch description**
Adding CMU_DoG dataset. This is mostly a copypasta of code that wasn't yet checked in, and was originally written mostly by Moya and myself.

**Testing steps**
```
λ parlai dd -t cmu_dog
λ parlai dd -t cmu_dog --cmu-dog-split-type original
λ parlai dd -t cmu_dog --cmu-dog-split-type seen --datatype test
λ parlai dd -t cmu_dog --cmu-dog-split-type unseen --datatype test
λ parlai dd -t cmu_dog --cmu-dog-include-knowledge-keys movieName
λ parlai dd -t cmu_dog --cmu-dog-provide-movie-context False --mutators knowledge_only_when_updated,prepend_knowledge_to_message
λ parlai dd -t cmu_dog --cmu-dog-only-with-knowledge False --cmu-dog-rating 1 --cmu-dog-multi-msg-delimiter "__" --cmu-dog-fact-delimiter "/"
```
Unit tests
```
λ pytest -k test_zootasks
λ pytest --force-regen parlai/tasks/cmu_dog/test.py  # fails initially, as expected
λ pytest parlai/tasks/cmu_dog/test.py  # succeeds
```

### DIFF
--- a/parlai/tasks/cmu_dog/agents.py
+++ b/parlai/tasks/cmu_dog/agents.py
@@ -162,9 +162,15 @@ class CMUDocumentGroundedConversationsTeacher(DialogTeacher):
     def __init__(self, opt: Opt, shared: TShared = None):
         opt = copy.deepcopy(opt)
         self.delimiter = opt.get('delimiter', '\n')
+        split_type = SplitType(opt.get("cmu_dog_split_type"))
+        if split_type == SplitType.ORIGINAL:
+            logger.warning(
+                "`original` split type contains duplicate conversations across train, "
+                "valid, and test. See https://github.com/festvox/datasets-CMU_DoG/issues/2 "
+                "for more detail."
+            )
         opt['datafile'] = _datafile(
-            split=DatatypeHelper.fold(opt['datatype']),
-            split_type=SplitType(opt.get("cmu_dog_split_type")),
+            split=DatatypeHelper.fold(opt['datatype']), split_type=split_type
         )
         super().__init__(opt, shared)
         if shared:
--- a/parlai/tasks/cmu_dog/agents.py
+++ b/parlai/tasks/cmu_dog/agents.py
@@ -41,11 +41,10 @@ def _datapath(opt: Opt) -> str:
     return os.path.join(opt['datapath'], 'cmu_dog')
 
 
-def _datafile(split: str, opt: Opt) -> str:
+def _datafile(split: str, split_type: SplitType) -> str:
     """
     Returns the filename, e.g. train.json.
     """
-    split_type = opt.get("cmu_dog_split_type")
     if split_type == SplitType.ORIGINAL:
         return f"{split}.json"
     if split_type == SplitType.SEEN:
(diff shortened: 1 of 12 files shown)


parlai/tasks/cmu_dog/agents.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def _all_split_datafiles(opt: Opt) -> List[str]:
+    datafiles = []
+    split_type = SplitType(opt.get("cmu_dog_split_type"))
+    if split_type in {SplitType.SEEN, SplitType.UNSEEN}:
+        # For seen/unseen split, the full set of dialogs is split
+        # across train, valid, test seen, and test unseen
+        for split in ['train', 'valid', 'test']:
+            datafiles.append(_datafile(split, SplitType.SEEN))
```
moyapchen, line 73: [No-op, just a comment] As a sidenote, 

> list(train_seen + valid_seen + test_seen + test_unseen) = list(train_original_deduped + valid_original_deduped + test_original_deduped)

so we could do that... but the code is probably clearer this way. 

(On that note, this does mean that we will be duplicate counting conversations when we do F1 on the original split, but it seems like we're deciding to be okay with that.)

moyapchen (approved):
Thanks for putting this up!

parlai/tasks/cmu_dog/agents.py
spencerp: > list(train_seen + valid_seen + test_seen + test_unseen) = list(train_original_deduped + valid_original_deduped + test_original_deduped)

I'm leaning toward leaving these separate for code clarity. But it might be good to add this relationship as a unit test so that it's documented.

> we will be duplicate counting conversations when we do F1 on the original split

When you say "original split" you mean the one that isn't deduplicated, right? I guess I'm kind of assuming that if someone is intentionally using the split with duplicate conversations, it's probably desirable that the conversations be duplicated in the metric calculations as well.

Maybe I should add a warning print if people use the `original` split type, and point them to your issue (https://github.com/festvox/datasets-CMU_DoG/issues/2) so they know what they're getting into.

parlai/tasks/cmu_dog/agents.py
```diff
@@ -0,0 +1,370 @@
...
+            choices=[1, 2, 3],
+            nargs="+",
+            help='The higher the number, the better quality the conversation. '
+            'For each rating, the number of conversations is as follows: 1-1443, 2-2142, 3-527',
+        )
+        cmu_dog.add_argument(
+            "--cmu-dog-only-with-knowledge",
+            type=bool,
+            default=True,
+            help="Optionally train only the sides of the conversation that have access to knowledge.",
```
klshuster, line 200: nit: as this defaults to True, and is presumably the desired arg, might be good to update the help comment here, like "Set False to optionally train both sides of conversation"

parlai/tasks/cmu_dog/agents.py
```diff
@@ -0,0 +1,370 @@
...
+        cmu_dog.add_argument(
+            "--cmu-dog-fact-delimiter",
+            type=str,
+            default=";",
+            help="When a section of the knowledge contains multiple facts, they will be concatenated with this delimiter.",
+        )
+        cmu_dog.add_argument(
+            "--cmu-dog-include-knowledge-keys",
+            type=str,
+            default='cast,critical_response,director,genre,introduction,movieName,rating,year',
```
klshuster, line 217: should we make `movieName` be the default here? or at least the `recommended`? as at least for our future experimentation we'll like to use that split.

alternatively, maybe we can provide a teacher, like `MovieNameTeacher`, which sets this arg for you (so you can do something like `parlai dd -t cmu_dog:movie_name` and it'll work appropriately)

parlai/tasks/cmu_dog/build.py
```diff
@@ -0,0 +1,33 @@
...
+# LICENSE file in the root directory of this source tree.
+import parlai.core.build_data as build_data
+import os
+from parlai.core.build_data import DownloadableFile
+
+RESOURCES = [
+    DownloadableFile(
+        'http://parl.ai/downloads/cmu_dog/cmu_dog.tar.gz',
+        'cmu_dog.tar.gz',
+        '30d2bac0dae6b4e4c0b94ba581fffaf1acb46838480f7ad6736ad03d9312ae9d',
```
klshuster, line 14: messaged offline about this but we should be downloading the dataset from the source, and building the splits ourselves (in the `build` function below)

klshuster (commented):
thanks! see comments, though as this is merged i think the build comment is the only one necessary to address (others are mostly nits)

PR #3597, opened 2021-04-16
[tests] Kill test_eval_fixed

stephenroller:
**Patch description**
This test sucks. It's our most common failure. Kill it with fire, it's not even worth saving.

**Testing steps**
```
$ pytest --seconds 600 tests/test_tra.py
418 passed, 133 skipped, 912 warnings in 636.21s (0:10:36)
```


EricMichaelSmith (approved):
Yeah makes sense

PR #3598, opened 2021-04-17
[params] Fix parser in python 3.8

stephenroller:
**Patch description**
ParlAI likes to use short options like `-hs 1024`. Since Python 3.8, this has been broken, and gets parsed as `-h s 1024`.

Although we've patched a few places previously ([example](https://github.com/facebookresearch/ParlAI/pull/3284)), there continue to be little bits and places left. This patch should fix all of them by changing everything to double dashes under the hood.

**Testing steps**
CI, manual testing.

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -13,7 +13,6 @@ import os
 import sys as _sys
 import datetime
 import parlai
-from packaging import version
 
 try:
     import git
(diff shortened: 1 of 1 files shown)


parlai/core/params.py
```diff
@@ -971,11 +971,29 @@ def parse_known_args(self, args=None, namespace=None, nohelp=False):
         if args is None:
             # args default to the system args
             args = _sys.argv[1:]
+
         args = fix_underscores(args)
+        # handle the single dash stuff. See _handle_single_dash_addarg for info
+        actions = set()
+        for action in self._actions:
+            actions = actions | set(action.option_strings)
```
mojtaba-komeili, line 979: Why not just `add(action)`?

parlai/core/params.py
```diff
@@ -1285,12 +1303,38 @@ def _handle_custom_options(self, kwargs):
...
+        `-e mb`.
+
+        Here we rewrite them into long args to get around the nonsense.
+        """
+        if _sys.version_info < (3, 8, 0):
+            # older python works fine
+            return args
+
+        out_long = []
+        out_short = []
```
mojtaba-komeili, line 1322: nit: why can't  they both use the same list? You returned their aggregated version only anyway.

mojtaba-komeili (approved):

parlai/core/params.py
stephenroller: bc i'm being cute and swapping the ordering

stephenroller: (we tend to define options like `add_argument('-hs', '--hiddensize')` but we need it to `add_argument('--hiddensize', `--hs`) or otherwise it will try to store the value to `opt['hs']`

parlai/core/params.py
stephenroller: I could use `set.update(action.option_strings)` but not `set.add`. option_strings is a list

parlai/core/params.py
mojtaba-komeili: Oh, I see. That makes sense.

parlai/core/params.py
stephenroller: Quick timer check shows that's much faster, so I'll update

mojtaba-komeili: Ok, I didn't know it is list. It makes sense like this then. Thanks for clarifying.

PR #3599, opened 2021-04-17
[TA] Fix preemption of cosine scheduler

stephenroller:
**Patch description**
Cosine schedulers would handle preemption poorly. Rather than resuming, they would reset, iff we were past warmup. Much to my surprise, only cosine had this issue.

**Testing steps**
New CI.


klshuster (approved):
thanks, sorry you got stuck with doing this. all looks good to me

PR #3600, opened 2021-04-17
Constrained Beam Search

PolKul:
**Patch description**

Added Constrained Beam Search algorithm for the torch_generator model. The constraints are usually required in order to guide decoding process in the right direction. The algorithm was ported from the Fairseq library. You can find more about the algorithm and the supporting papers [here](https://github.com/pytorch/fairseq/tree/master/examples/constrained_decoding)   

**Testing steps**

The demo code can be run from tests/run_constrained_beam_search.py. The constraints are defined in the "constraints" parameter, which is a list of all the constraints for the decoder. Each constraint can be either one more multiple words.

The result of the decoder should contain all the constraints from the list. 

**Other information**


parlai/core/search.py
```diff
@@ -0,0 +1,817 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
stephenroller, line 1: This whole file looks like copypasta from another project (fairseq? hugging face?)

Is it really necessary? It's extremely large and seems to have a lot of redundancy with our other abstractions

PolKul: well, I've just taken this whole file from Parlai library because it contains LexicallyConstrainedBeamSearch. Basically we just need that class (and its parent Search class of course) but there are some more interesting Search sibling classes there like DiverseBeamSearch for example, which can be explored and implemented in Parlai later on as well. Also for the ease of updates. we can update the whole search.py file when required. But please let me know if you prefer to extract just the LexicallyConstrainedBeamSearch from it, we can do it as well.

stephenroller: I definitely want to extract the very minimum. I have too much code to maintain as it is.

PolKul: I've trimmed search.py Please check now

PR #3601, opened 2021-04-17
fix update_freq when initializing CUDA buffer

Reveyer:
**Patch description**

This fixes an issue of initializing  CUDA buffer when using  the accumulated gradients·

**Testing steps**

Tested with [projects.wizard_of_wikipedia.generator.agents:EndToEndAgent](https://github.com/facebookresearch/ParlAI/blob/master/projects/wizard_of_wikipedia/generator/agents.py) agent with train_model.

Before the change
```
20:14:55 | time:50s total_exs:6400 total_steps:100 epochs:0.09
    clen  clip  ctpb  ctps  ctrunc  ctrunclen  exps  exs  gpu_mem  know_acc  know_chance  llen        lr  ltpb  ltps   ltrunc  ltrunclen  token_acc  token_em  total_train_updates  tpb   tps   ups
   75.89     0  2146  9167   .1884      8.831 136.7 3200    .3320         1       .03965 23.46 1.005e-05 746.9  3191 .0003125      .1209          0         0                  100 2893 12357 2.148
```

After the change
```
20:23:45 | time:92s total_exs:12800 total_steps:100 epochs:0.17
    clen  clip  ctpb  ctps  ctrunc  ctrunclen  exps  exs  gnorm  gpu_mem  know_acc  know_chance  know_loss  llen  loss        lr  ltpb  ltps  ltrunc  ltrunclen   ppl  token_acc  token_em  \
   75.48 .1000  2151 10052   .1866      8.264 149.5 6400 .08706    .2992    .04063       .03927       6.19 23.02 10.21 1.005e-05 736.5  3442       0          0 27192  1.358e-05         0
    total_train_updates  tpb   tps   ups
                    100 2888 13493 1.168
```

**Other information**

The reason for this issue is that initializing the buffer also updates the parameter `self._number_grad_accum`, causing the loss to become `nan`.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -2255,7 +2255,7 @@ class TorchAgent(ABC, Agent):
             # Only print in the non-shared version.
             logging.info(f'{self.id}: full interactive mode on.')
 
-    def backward(self, loss):
+    def backward(self, loss, is_init_buffer=False):
         """
         Perform a backward pass.
 
@@ -2264,7 +2264,7 @@ class TorchAgent(ABC, Agent):
         """
         update_freq = self.opt.get('update_freq', 1)
 
-        if update_freq > 1:
+        if update_freq > 1 and not is_init_buffer:
             # gradient accumulation, but still need to average across the minibatches
             loss = loss / update_freq
             self._number_grad_accum = (self._number_grad_accum + 1) % update_freq
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -838,7 +838,7 @@ class TorchAgent(ABC, Agent):
         # set interactive mode or not according to options.
         self.set_interactive_mode(opt.get('interactive_mode', False), shared)
 
-        self.buffer_initialized = True
+        self.buffer_initialized = False
 
     def build_history(self):
         """
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -840,7 +840,6 @@ class TorchAgent(ABC, Agent):
 
         self.buffer_initialized = True
 
-
     def build_history(self):
         """
         Return the constructed history object.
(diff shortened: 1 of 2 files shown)


stephenroller:
Sorry can you explain a bit more? I don't see where the nan is coming from. It is problematic if nodes are having their `update_freq` get out of sync, but we should really only see that if we oom'd on forward or backward, in which case, that worker actually DOES need its `update_freq` moved forward.

Reveyer:
You are right! I just re-debugged the code again and found that the source of the loss is generated through 

https://github.com/facebookresearch/ParlAI/blob/36bc4ae4bb7523f6089a7fbe7b9c0fa0260994de/projects/wizard_of_wikipedia/generator/modules.py#L34-L37

When using  the dummy batch, nan is generated here. Then the loss will become nan.

stephenroller:
Seems like the right solution is to make it `sentence_sums /= max(divisor, 1)`

Reveyer:
@stephenroller Since this may also happen in other agents, I have re-changed the judgment condition so that it will only ignore the use of dummy batch. What do you think about this?

stephenroller:
I don't think we want to change the logic around update/cuda buffer initialization. It's meant to be a (very poorly named) OOM-recovery mechanism. During distributed training, all workers must perform a gradient sync at _exactly_ the same time, or there will be a system hang. So if one worker OOMs, we do a dummy batch with zero loss, so that it has something to synchronize (all zeros, but still):

https://github.com/facebookresearch/ParlAI/blob/36bc4ae4bb7523f6089a7fbe7b9c0fa0260994de/parlai/core/torch_generator_agent.py#L617

That update_freq is incremented during that original first initialization isn't really problematic. Yeah okay, the gradients are slightly smaller than the next batch, but that's minor. The real problem is that `compute_loss` is returning a NaN and propagating, and that's because of a bug in just this agent.

I am planning on, sometime in the next month, removing this _init_cuda_buffer logic and dummy batches entirely, and coding the recovery mechanism as an explicit zero-grad sync. Writing dummy batches is annoying.

Reveyer:
Thank you for your reply and I look forward to the subsequent releases.😆

stephenroller:
Just to be clear, I would very much welcome a fix for the wizard agent that protects that divide-by-zero!

Reveyer:
Haha, ok, I'll resubmit a new PR!

PR #3612, opened 2021-04-21
[Safety] Fix a Static Task bug and Safety README

jxmsML:
**Patch description**
Patch on safety human safety
- Just notice that the UI always enable response field in static task (if `responseField !==null`, which is always `true` given the default value for `responseField` is `False`), add a fix to it.
- More explanation on how to run human safety evaluation, with new script parsing world logs to human eval ready format.
- change logging info -> warn once, otherwise the loggings output is nasty for `bs > 1`

Patch on world logging (moved to https://github.com/facebookresearch/ParlAI/pull/3674), this branch is rebased on the feature branch `convo_log_pad` in https://github.com/facebookresearch/ParlAI/pull/3674
- filter by `message.is_padding()` when write episodes to `self._current_episodes`.
- modify the test `tests/test_eval_model.py` on `test_save_report`.

**Testing steps**

**Other information**

### DIFF
--- a/parlai/utils/world_logging.py
+++ b/parlai/utils/world_logging.py
@@ -76,7 +76,9 @@ class WorldLogger:
         msgs = []
         for act in acts:
             # padding examples in the episode[0]
-            if isinstance(act, Message) and act.is_padding():
+            if not isinstance(act, Message):
+                act = Message(act)
+            if act.is_padding():
                 break
             if not self.keep_all:
                 msg = {f: act[f] for f in self.keep_fields if f in act}
--- a/parlai/utils/world_logging.py
+++ b/parlai/utils/world_logging.py
@@ -16,6 +16,7 @@ from parlai.utils.misc import msg_to_str
 from parlai.utils.conversations import Conversations
 from parlai.utils.io import PathManager
 import parlai.utils.logging as logging
+from parlai.core.message import Message
 
 import copy
 from tqdm import tqdm
@@ -74,12 +75,17 @@ class WorldLogger:
         """
         msgs = []
         for act in acts:
+            # padding examples in the episode[0]
+            if isinstance(act, Message) and act.is_padding():
+                break
             if not self.keep_all:
                 msg = {f: act[f] for f in self.keep_fields if f in act}
             else:
                 msg = act
             msgs.append(msg)
 
+        if len(msgs) == 0:
+            return
         self._current_episodes.setdefault(idx, [])
         self._current_episodes[idx].append(msgs)
 
--- a/projects/safety_recipes/README.md
+++ b/projects/safety_recipes/README.md
@@ -81,13 +81,13 @@ To enable onboarding task, please edit the [onboarding task](https://github.com/
 
 To create your own chat logs[here](https://github.com/facebookresearch/ParlAI/blob/master/projects/safety_recipes/human_safety_evaluation/task_config/task_data.jsonl)
 
-1) The following command will generate original chat logs of Blender 90M model responding to Bot Adversarial Task testset for human safety evaluation and write the 180 examples to `tmp/world_logs.jsonl` in ParlAI format:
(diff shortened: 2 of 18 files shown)


projects/safety_recipes/README.md
```diff
@@ -75,7 +75,24 @@ parlai eval_model -mf zoo:sensitive_topics_classifier/model -t sensitive_topics_
...
+The following command will generate chat logs of Blender 90M model responding to Bot Adversarial Task testset for human safety evaluation and write the 180 examples to `tmp/world_logs.jsonl` 
+```
+parlai em -t bot_adversarial_dialogue:HumanSafetyEvaluation --flatten-dialogue False -mf zoo:blender/blender_90M/model -bs 64 --world-logs tmp/world_logs.jsonl
+```
+and to automatically format the generated logs in `world_logs.jsonl` to human safety evaluation ready format:
+```
+python projects/safety_recipes/human_safety_evaluation/format_safety_ready.py --world-logs-path tmp/world_logs.jsonl --eval-logs-dir tmp/human_safety_evaluation
+```
+
+2) Specify turn indices per conversation to annotate [here](https://github.com/facebookresearch/ParlAI/blob/master/projects/safety_recipes/human_safety_evaluation/task_config/annotation_indices.jsonl): each line represents the list of utterance indices to be annotated for safety for the corresponding conversation in the chat logs. For bot adversarial test set consisting of 180 examples, we only evaluate the last reply of each conversation.
```
emilydinan, line 94: is this part necessary if we use the command above?

emilydinan: if yes, can we make it automated? if no, can we make it clear that it's not necessary?

projects/safety_recipes/human_safety_evaluation/format_safety_ready.py
```diff
@@ -0,0 +1,79 @@
...
+        help="dir to the store annotate_indices.jsonl and task_data.jsonl ready for direct use in human safety evaluation format",
+    )
+
+    args = parser.parse_args()
+    world_logs_path = args.world_logs_path
+    chatlogs = []
+    with PathManager.open(world_logs_path) as data_file:
+        for l in data_file.readlines():
+            episode = json.loads(l.strip())
+            # TODO: when conversation format is finished please remove this line;
```
emilydinan, line 37: what does this mean?

projects/safety_recipes/README.md
jxmsML: ah yes. I just edit this ( running `format_safety_ready.py` should automatically generate the `annotation_indices.jsonl` as well as the` task_data.jsonl`)

projects/safety_recipes/human_safety_evaluation/format_safety_ready.py
jxmsML: It seems like there is a bug in conversation format that would generate lines like:
```
{"dialog": [[{"batch_padding": true, "episode_done": true, "id": "bot_adversarial_dialogue:HumanSafetyEvaluation.persona_False_flatten_False"}, {"id": "TransformerGenerator", "episode_done": false}]], "context": [], "metadata_path": "tmp/world_logs.metadata"}
```
I added a hack to skip those when parsing but, there is room for removing that hack after the bug above is fixed.

emilydinan (approved):
seems ready to go, but let's wait to merge until the aforementioned bug is fix so we can get rid of that comment/hack and rebase on top

parlai/utils/world_logging.py
```diff
@@ -74,12 +75,17 @@ def _add_msgs(self, acts, idx=0):
         """
         msgs = []
         for act in acts:
+            # padding examples in the episode[0]
+            if isinstance(act, Message) and act.is_padding():
```
jxmsML, line 79: only filter out if act is `Message` otherwise it'll break the unittests for `act` is `dict`

jxmsML:
This pr is rebased on https://github.com/facebookresearch/ParlAI/pull/3674. (separate the changes on safety test and world log saving)

PR #3613, opened 2021-04-21
Modify dictionary load to only rstrip to handle space prefixed tokens

ankitade:
**Patch description**
If special tokens are prefixed by space, while loading them from the dict for eval, the space was being stripped out. Modifying the strip to be rstrip and strip only line breaks

**Testing steps**
Will wait for CI tests to pass


stephenroller:
Whyyyy didn't this run any tests?

stephenroller (approved):
Great

PR #3616, opened 2021-04-22
bpe dropout modification

melanierk:
**Patch description**
Trying a different variant of BPE dropout (jointly applying the same dropout rules to both context and label, as opposed to independently applying the dropout). 
**Testing steps**
ipdb for a bart model to see that tokenizer rule resampling is applied when we want it to be, plus new unit test in test_dict.py. I tried to add a test in test_torch_agent.py, but got a little stuck figuring out a good way to mock the dictionary attribute and its fn, set_tokenization_mode. If anyone has ideas how to do that I can add one here. 

**Other information**
Will not land until we see that this is an improvement.

### DIFF
--- a/parlai/core/dict.py
+++ b/parlai/core/dict.py
@@ -248,9 +248,6 @@ class DictionaryAgent(Agent):
             'dict_textfields', DictionaryAgent.default_textfields
         ).split(",")
 
-        # used to signal whether we should use training time tricks, like bpe droput
-        self._tokenization_mode = TokenizationMode.TEST_TIME_LABEL
-
         try:
             self.tokenizer_fun = getattr(self, self.tokenizer + '_tokenize')
         except AttributeError:
@@ -827,7 +824,7 @@ class DictionaryAgent(Agent):
         """
         return str(self.freq)
 
-    def set_tokenization_mode(self, mode: TokenizationMode):
+    def set_tokenization_mode(self, mode: TokenizationMode, resample_bpe: bool = False):
         """
         Indicate what "kind" of tokenization is being done.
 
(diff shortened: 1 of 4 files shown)


parlai/core/dict.py
```diff
@@ -248,9 +248,6 @@ def __init__(self, opt: Opt, shared=None):
             'dict_textfields', DictionaryAgent.default_textfields
         ).split(",")
 
-        # used to signal whether we should use training time tricks, like bpe droput
-        self._tokenization_mode = TokenizationMode.TEST_TIME_LABEL
```
melanierk, line 252: I believe this was dead code.

parlai/core/dict.py
```diff
@@ -827,7 +824,7 @@ def __str__(self):
         """
         return str(self.freq)
 
-    def set_tokenization_mode(self, mode: TokenizationMode):
+    def set_tokenization_mode(self, mode: TokenizationMode, resample_bpe: bool = False):
```
melanierk, line 827: This may not be the best way to abstract ---  it feels a bit weird to put a bpe related thing in the dict/agent level but I'm not sure if there's a better way.

parlai/core/dict.py
stephenroller: not sure? Dictionary should be initialized into evaluation mode by default. Maybe it should be TEST_TIME_TEXT?

Anyway if it's initializing to never dropping out, it's ok

parlai/core/dict.py
stephenroller: Yeah I didn't like the abstraction leak either. That's why I was trying to signal it on the TokenizationMode

parlai/core/dict.py
```diff
@@ -838,10 +835,17 @@ def set_tokenization_mode(self, mode: TokenizationMode):
...
 
-        Use True for training time, False for not.
+        mode: use True for training time, False for not.
+        resample_bpe: if True, reset the dropout of the merge rules for BPE
         """
         self._context_mode = mode
         if hasattr(self, 'bpe'):
-            # enable bpe dropout only in texts at training time. disable all
-            # other times
-            self.bpe.enable_bpe_dropout(mode == TokenizationMode.TRAIN_TIME_TEXT)
```
stephenroller, line 847: why can't `enable_bpe_dropout` internally call `dropout_bpe_ranks`?

stephenroller (commented):
implementation looks correct from my quick glance. Some nits about structure but we can handle these after experimentation to figure out if it's even worth doing

PR #3623, opened 2021-04-27
Mutators inside `parlai_internal` and `parlai_fb`

stephenroller:
**Patch description**
Allow ParlAI to pick up mutators from internal-only code for internal development.

**Testing steps**
Manual testing.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -611,6 +611,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
 
         This is also used in distributed mode to force a worker to sync with others.
         """
+        return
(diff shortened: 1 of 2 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -611,6 +611,7 @@ def _init_cuda_buffer(self, batchsize, maxlen, force=False):
 
         This is also used in distributed mode to force a worker to sync with others.
         """
+        return
```
stephenroller, line 614: should not be a part of this

stephenroller, line 614:
```suggestion
```

EricMichaelSmith (approved):
Parallelism is clear, makes sense

PR #3624, opened 2021-04-27
Add glue hf datasets

meganung:
Adding all glue datasets (https://huggingface.co/datasets/glue) and super glue datasets (https://huggingface.co/datasets/super_glue).

Added test.py for both glue and superglue and all pass.

### DIFF
--- a/parlai/tasks/glue/agents.py
+++ b/parlai/tasks/glue/agents.py
@@ -42,6 +42,30 @@ class MnliTeacher(AbstractHuggingFaceTeacher):
     }
 
 
+class MnliMatchedTeacher(AbstractHuggingFaceTeacher):
+    """
+    Note: this is an evaluation dataset so it only has valid and test splits
+    """
+
+    hf_path = 'glue'
+    hf_name = 'mnli_matched'
+    hf_text_fields = ['premise', 'hypothesis']
+    hf_label_field = 'label'
+    hf_splits_mapping = {'train': None, 'valid': 'validation', 'test': 'test'}
+
+
+class MnliMismatchedTeacher(AbstractHuggingFaceTeacher):
+    """
+    Note: this is an evaluation dataset so it only has valid and test splits
+    """
+
+    hf_path = 'glue'
+    hf_name = 'mnli_mismatched'
+    hf_text_fields = ['premise', 'hypothesis']
+    hf_label_field = 'label'
+    hf_splits_mapping = {'train': None, 'valid': 'validation', 'test': 'test'}
+
+
 class MrpcTeacher(AbstractHuggingFaceTeacher):
     hf_path = 'glue'
     hf_name = 'mrpc'
--- a/parlai/tasks/glue/agents.py
+++ b/parlai/tasks/glue/agents.py
@@ -4,15 +4,22 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+from parlai.core.teachers import MultiTaskTeacher
 from parlai.tasks.huggingface.agents import AbstractHuggingFaceTeacher
+from copy import deepcopy
 
 
 class AxTeacher(AbstractHuggingFaceTeacher):
+    """
+    Note: this is an evaluation dataset so it only has a test split
+    Use a model trained on MulitNLI to produce predictions for this dataset.
+    """
+
     hf_path = 'glue'
     hf_name = 'ax'
     hf_text_fields = ['premise', 'hypothesis']
     hf_label_field = 'label'
-    hf_splits_mapping = {'train': 'test', 'valid': 'test', 'test': 'test'}
+    hf_splits_mapping = {'train': None, 'valid': None, 'test': 'test'}
 
 
 class ColaTeacher(AbstractHuggingFaceTeacher):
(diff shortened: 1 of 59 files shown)


stephenroller (commented):
Wow move fast! Great to see

- Add tests for all of them, please
- Add a MultiTeacher as the new default for glue and superglue. I'll show you an example shortly

parlai/tasks/glue/agents.py
```diff
@@ -7,6 +7,14 @@
 from parlai.tasks.huggingface.agents import AbstractHuggingFaceTeacher
 
 
+class AxTeacher(AbstractHuggingFaceTeacher):
+    hf_path = 'glue'
+    hf_name = 'ax'
+    hf_text_fields = ['premise', 'hypothesis']
+    hf_label_field = 'label'
+    hf_splits_mapping = {'train': 'test', 'valid': 'test', 'test': 'test'}
```
meganung, line 15: This is an evaluation dataset (see https://huggingface.co/datasets/glue) so it only has a test split. 
An alternative is to have train and valid map to None.

parlai/tasks/glue/agents.py
```diff
@@ -15,5 +23,73 @@ class ColaTeacher(AbstractHuggingFaceTeacher):
...
+class MnliTeacher(AbstractHuggingFaceTeacher):
+    hf_path = 'glue'
+    hf_name = 'mnli'
+    hf_text_fields = ['premise', 'hypothesis']
+    hf_label_field = 'label'
+    hf_splits_mapping = {
+        'train': 'train',
+        'valid': 'validation_matched',
+        'test': 'test_matched',
+    }
```
meganung, line 35: Mnli dataset also has `validation_mismatched` and `test_mismatched` splits, I just picked one for the teacher.

parlai/tasks/huggingface/agents.py
```diff
@@ -39,6 +40,7 @@ def __init__(self, opt, shared=None):
         self.hf_split = self.hf_splits_mapping[self.fold]
         self.data_path = self._path(opt)
         opt['datafile'] = self.data_path
+        make_dir(opt['datafile'])
```
meganung, line 43: Added this so the .lengths files would be saved properly- can leave this here or for consistency do we want to move this into a new build.py file?

parlai/tasks/glue/agents.py
stephenroller: Hm, can you go read the superglue and mnli paper and try to figure out which is right? We shouldn't guess.

parlai/tasks/glue/agents.py
stephenroller: Good call out. Not sure what's going to happen if we make an empty teacher but we can try

parlai/tasks/huggingface/agents.py
stephenroller: Huh wouldn't think it was necessary... okay this is fine.

stephenroller:
https://github.com/facebookresearch/ParlAI/blob/2f0e38754315310818d79c41019b64b1bba5587a/parlai/tasks/decanlp/agents.py#L67-L83

Example of a MultiTeacher

parlai/tasks/glue/agents.py
meganung: "For MNLI, we report accuracy averaged over the matched and mismatched test sets" ... they use both.

parlai/tasks/glue/agents.py
```diff
@@ -15,5 +30,116 @@ class ColaTeacher(AbstractHuggingFaceTeacher):
...
+class MnliMismatchedTeacher(AbstractHuggingFaceTeacher):
+    """
+    Note: this is an evaluation dataset so it only has valid and test splits
+    """
+
+    hf_path = 'glue'
+    hf_name = 'mnli_mismatched'
+    hf_text_fields = ['premise', 'hypothesis']
+    hf_label_field = 'label'
+    hf_splits_mapping = {'train': None, 'valid': 'validation', 'test': 'test'}
```
meganung, line 66: added these 2 as consistency since on hugging face they also have them as separate datasets but technically you can access the data from the mnli dataset splits ('validation_mismatched', 'validation_matched', and the same for test).

parlai/tasks/glue/agents.py
```diff
@@ -4,7 +4,22 @@
...
+    """
+    Note: this is an evaluation dataset so it only has a test split
+    Use a model trained on MulitNLI to produce predictions for this dataset.
+    """
+
+    hf_path = 'glue'
+    hf_name = 'ax'
+    hf_text_fields = ['premise', 'hypothesis']
+    hf_label_field = 'label'
+    hf_splits_mapping = {'train': None, 'valid': None, 'test': 'test'}
```
meganung, line 22: for evaluation datasets, changed mapping of train and valid to None e.g. `hf_splits_mapping = {'train': None, 'valid': None, 'test': 'test'}` and removed the task from the tests and the multitaskteacher, otherwise will produce errors. e.g. `display_data -t glue:ax` will work only if you set the datatype to test now.

parlai/tasks/glue/agents.py
stephenroller: Great, just checked the leaderboard and those are separate numbers, this seems good

stephenroller (approved):
Awesome!

PR #3625, opened 2021-04-28
Adding parlai_fb and parlai_interal as alternative scripts directories

skiingpacman:
**Patch description**
Extending where the parlai command line looks for scripts. This is only relevant when working internally at FB.

**Testing steps**
See internal diff D28044715.


stephenroller (approved):
Thanks!

PR #3621, opened 2021-04-27
[fbcode] Lint.

stephenroller:
**Patch description**
FBCode consistently complains about a number of small whitespace problems in READMEs in our repo. Normally I just run `arc lint --accept` to make sure these changes are included on a sync, but it's getting annoying, so pull them into GitHub.

**Testing steps**
CI

PR #3628, opened 2021-04-28
[tests] Make crowdsource tests whitespace-agnostic

stephenroller:
**Patch description**
#3621 broke crowdsourcing tests by stripping a bunch of whitespace from our regression files. Annoying, but just correct for this in the test and strip the comparison point as well.

**Testing steps**
CI.

### DIFF
--- a/parlai/crowdsourcing/tasks/acute_eval/util.py
+++ b/parlai/crowdsourcing/tasks/acute_eval/util.py
@@ -191,7 +191,7 @@ class AbstractFastAcuteTest(AbstractOneTurnCrowdsourcingTest):
             results_folder=results_folder, file_suffix=file_suffix
         )
         with open(file_path) as f:
-            contents = f.read()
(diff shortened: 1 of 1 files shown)


parlai/crowdsourcing/tasks/acute_eval/util.py
```diff
@@ -191,7 +191,7 @@ def _check_file_contents(
             results_folder=results_folder, file_suffix=file_suffix
         )
         with open(file_path) as f:
-            contents = f.read()
+            contents = f.read().rstrip('\n') + '\n'
```
EricMichaelSmith, line 194: Okay, this works - maybe there can be a comment saying why this is needed (i.e. to be insensitive to whitespace linting issues?)

EricMichaelSmith (approved):
Yeah, seems reasonable, thanks!

PR #3634, opened 2021-04-29
skipping the padding message

mojtaba-komeili:
**Patch description**
The rag agent may crash if there is batch padding message. These message do not contain the `text` data that it want to encode.

**Test**
It used to crash if it was receiving a batch padding message. It successfully finishes with this.

### DIFF
--- a/parlai/agents/rag/rag.py
+++ b/parlai/agents/rag/rag.py
@@ -251,9 +251,9 @@ class RagAgent(TransformerGeneratorRagAgent, BartRagAgent, T5RagAgent):
         """
         Overrides TA.observe to tokenize the retriever query.
         """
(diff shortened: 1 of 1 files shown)


stephenroller:
Use obs.is_padding_example()

stephenroller (approved):
Is observation not a message already?! This is doing a shallow copy which is fine, but we should have gotten those dicts out of there by now :(

klshuster:
observation can be a dict but `TorchAgent.observe` takes care of that so this copy is probably not necessary but alas (`self._generation_agent.observe()` is guaranteed to return a `Message`)

stephenroller:
I tried to make the teachers also forcefully output Messages by now

klshuster:
interestingly the WoW teachers do not

PR #3637, opened 2021-04-30
[fbcode] Fix bug with internal tensorboard

stephenroller:
**Patch description**
Internal fbcode tensorboard is a few versions behind, and doesn't support these advanced features. Just silently omit them if an error is thrown, and fall back to The Old Ways(TM).

**Testing steps**
Manual testing in fbcode.


spencerp (approved):
Thanks for this fix!! Sorry I didn't check the version of internal tensorboard

stephenroller:
Confirmed internal testing, gonna merge.

PR #3643, opened 2021-05-04
[fbcode] Bring in json_safe_payload method

stephenroller:
**Patch description**
Internal discussion for the need for this patch.

**Testing steps**
CI, internal CI.


spencerp (approved):
Thank you!!

PR #3644, opened 2021-05-04
[torchscript] Support special tokens in torchscript module.

stephenroller:
**Patch description**
Add special tokens support for torchscript. The implementation doesn't reflect the original, because the original just used recursion, which torchscript doesn't support.

**Testing steps**
New CI. Internal testing

### DIFF
--- a/parlai/torchscript/modules.py
+++ b/parlai/torchscript/modules.py
@@ -70,6 +70,7 @@ class TorchScriptGreedySearch(nn.Module):
             bpe_encoder=orig_bpe.encoder,
             bpe_byte_encoder=orig_bpe.byte_encoder,
             fused_key_bpe_ranks=fused_key_bpe_ranks,
+            special_tokens=agent._get_special_tokens(),
         )
 
         # History tracking and start/end tokens
@@ -451,6 +452,7 @@ class ScriptableGpt2BpeHelper(object):
         encoder: Dict[str, str],
         byte_encoder: Dict[int, str],
         fused_key_bpe_ranks: Dict[str, float],
+        special_tokens: List[str],
     ):
 
         self.add_prefix_space = add_prefix_space
@@ -467,6 +469,11 @@ class ScriptableGpt2BpeHelper(object):
 
         self.bpe_ranks = fused_key_bpe_ranks
 
+        # special tokens
+        self._special_tokens: Dict[str, int] = {}
+        for st in special_tokens:
+            self._special_tokens[st] = 1
+
     def encode(self, text: str) -> List[str]:
         """
         Tokenize text.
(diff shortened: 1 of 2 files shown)


parlai/torchscript/modules.py
```diff
@@ -481,7 +488,41 @@ def encode(self, text: str) -> List[str]:
         """
         if self.add_prefix_space:
             text = f' {text}'
-        return self.helper_encode(text)
+
+        # constants for readability
```
EricMichaelSmith, line 492: hmm perhaps it'd help to have a 1-sentence comment about how the special tokens code works at a high level, and maybe also what FINAL and SPLITABLE are?

tests/nightly/gpu/test_torchscript.py
```diff
@@ -62,6 +62,72 @@ def test_token_splitter(self):
                     if idx + 1 == num_examples:
                         break
 
+    def test_special_tokenization(self):
+        from parlai.core.dict import DictionaryAgent
```
EricMichaelSmith, line 66: nit: these imports could go to the top, right? Or at least the first 2?

tests/nightly/gpu/test_torchscript.py
```diff
@@ -62,6 +62,72 @@ def test_token_splitter(self):
...
+        from parlai.core.params import ParlaiParser
+        from parlai.torchscript.modules import ScriptableDictionaryAgent
+
+        SPECIAL = ['Q00', 'Q01']
+        text = "Don't have a Q00, man! Have a Q01 instead."
```
EricMichaelSmith, line 71: 😆

tests/nightly/gpu/test_torchscript.py
```diff
@@ -62,6 +62,72 @@ def test_token_splitter(self):
...
+                bpe_byte_encoder=orig_bpe.byte_encoder,
+                fused_key_bpe_ranks=fused_key_bpe_ranks,
+                special_tokens=[],
+            )
+
+            tokenized = sda.txt2vec(text)
+            assert len(tokenized) == 15
+            assert sda.vec2txt(tokenized) == text
+            nice_tok = [sda.ind2tok[i] for i in tokenized]
+
```
EricMichaelSmith, line 106: Nit: a few variables that are assigned to but never used, given linting messages

tests/nightly/gpu/test_torchscript.py
```diff
@@ -62,6 +62,72 @@ def test_token_splitter(self):
...
+                bpe_encoder=orig_bpe.encoder,
+                bpe_byte_encoder=orig_bpe.byte_encoder,
+                fused_key_bpe_ranks=fused_key_bpe_ranks,
+                special_tokens=SPECIAL,
+            )
+
+            special_tokenized = sda.txt2vec(text)
+            assert len(special_tokenized) == 15
+            assert sda.vec2txt(special_tokenized) == text
+            assert special_tokenized != tokenized
```
EricMichaelSmith, line 128: Thought: it could be even more explicit to check the actual strings of the output tokens, instead of just their length and whether they match with/without special tokens. No strong opinion on this either way, though

EricMichaelSmith (approved):
Makes sense, minor nits

stephenroller:
Sorry, landing in expediency.

PR #3657, opened 2021-05-19
Fix RAG generation for T5 FiD

moyapchen:
I noticed a big discrepency between exact_match and token_em after I switched to using T5 FiD in public ParlAI.
Per discussion with Kurt, it's cause the generate for T5 is supposed to call TorchGeneratorAgent, rather than using the T5 generation via Hugging Face. This was in the original itnernal implementation, but got lost at some point while open sourcing.

Verified by printing out forced decoding outputs and the output from eval_model. Verified that when `token_em == 1` that the output of eval_model was also the same. Models trained with this also doing a lot better.


klshuster (approved):
thanks for fixing this!

PR #3658, opened 2021-05-19
[logging] Fix suppression of debug logs

stephenroller:
**Patch description**
Fix an issue where usages of `create_agent_from_model_file` caused logging to bump back to INFO.

The reason is that we eventually call this line to try to fill in missing args with defaults:
https://github.com/facebookresearch/ParlAI/blob/91e883baf3352758fce6a48a075ce48e67d5caeb/parlai/core/agents.py#L368

Which would get the default args and accidentally call `set_log_level` here:
https://github.com/facebookresearch/ParlAI/blob/91e883baf3352758fce6a48a075ce48e67d5caeb/parlai/core/params.py#L1158

As a solution, I moved setting the log level to happen in the abstract ParlaiScript class.

**Testing steps**
Manual testing.

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -1155,7 +1155,6 @@ class ParlaiParser(argparse.ArgumentParser):
 
         self._process_args_to_opts(args)
         print_announcements(self.opt)
-        logging.set_log_level(self.opt.get('loglevel', 'info').upper())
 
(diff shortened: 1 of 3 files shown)


mojtaba-komeili (approved):
Thanks!

klshuster:
thanks for the fix!!

PR #3668, opened 2021-05-25
[RAG] Fix Generate Dense Embeddings Script

klshuster:
**Patch description**
At some point, this script broke, in that the default flag for `--share-encoders` was forcing the document and query encoders to be shared; as we used the document encoder to generate embeddings, this is no good.

**Testing steps**
Confirmed that the total number of parameters printed when running the generation script is correct.


klshuster:
Note that I have verified in old logs that the zoo indices are built with the correct embeddings

mojtaba-komeili (approved):

PR #3669, opened 2021-05-25
[TGA] Beam context block bug

emilydinan:
**Patch description**
Hit a bug with beam blocking in TGA. TGA has beam_block_full_context set to True by default, which requires the batch to have a `full_text_vec`. I have beam context blocking off, and am passing a cached text vec in the observation to avoid re-tokenization (which means `full_text_vec` never gets set) and hit this error. To remedy, I return an empty tensor when beam_context_blocking is turned off.


stephenroller (approved):
Let all tests pass but then okay

PR #3672, opened 2021-05-25
tqdm version changed to 4.38.0 due to spacey req 

liliarose:
tqdm version changed to 4.38.0 due to spacey req


stephenroller (approved):
Lgtm. Don't worry about mypy

PR #3674, opened 2021-05-26
Patch on world logging

jxmsML:
Patch on world logging
- filter by `message.is_padding()` when writing episodes to `self._current_episodes`.
- modify the test tests/test_eval_model.py on test_save_report to verify the no dummy world logs are stored (those with `batch_padding`  `True`)

before this patch:
the world logs saved will contain the following dummy episode:
```
{"dialog": [[{"batch_padding": true, "episode_done": true, "id": "bot_adversarial_dialogue:HumanSafetyEvaluation.persona_False_flatten_False"}, {"id": "TransformerGenerator", "episode_done": false}]], "context": [], "metadata_path": "random/world_logs.metadata"}
```

After this patch, no such padding examples will be saved to the world logs

**Testing steps**

**Other information**

### DIFF
--- a/parlai/utils/world_logging.py
+++ b/parlai/utils/world_logging.py
@@ -76,7 +76,9 @@ class WorldLogger:
         msgs = []
         for act in acts:
             # padding examples in the episode[0]
-            if isinstance(act, Message) and act.is_padding():
+            if not isinstance(act, Message):
+                act = Message(act)
+            if act.is_padding():
                 break
             if not self.keep_all:
                 msg = {f: act[f] for f in self.keep_fields if f in act}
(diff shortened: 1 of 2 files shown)


parlai/utils/world_logging.py
```diff
@@ -74,12 +75,17 @@ def _add_msgs(self, acts, idx=0):
         """
         msgs = []
         for act in acts:
+            # padding examples in the episode[0]
+            if isinstance(act, Message) and act.is_padding():
```
stephenroller, line 79: We might want to cast it as a message in case it happens to be a dict

stephenroller (approved):

stephenroller (approved):

PR #3676, opened 2021-05-27
Split folds when multiple domains helper

moyapchen:
We've got some use cases where we've got a single dataset that has different sub-components that can be swapped in + out. It's a little subtle how we make sure we're getting the same samples out of every domain (namely, we can't concatinate them all together), so make a helper function for it.

Test Plan:
Print out outputs in a dataset that use this, with a varying # of domains but one fixed domain. Verify that the same lines show up for the fixed domain.

### DIFF
--- a/parlai/utils/data.py
+++ b/parlai/utils/data.py
@@ -102,6 +102,7 @@ class DatatypeHelper:
         train_frac: float,
         valid_frac: float,
         test_frac: float,
+        seed: int = 42,
     ):
         """
         Need to be careful about how we setup random to not leak examples between trains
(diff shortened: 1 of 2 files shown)


parlai/utils/data.py
```diff
@@ -91,3 +93,35 @@ def is_streaming(cls, datatype: str) -> bool:
...
+        elif "valid" in fold:
+            start = train_frac
+            end = train_frac + valid_frac
+        else:
+            start = train_frac + valid_frac
+            end = 1.0
+
+        result = []
+        for domain in domains:
+            random.Random(42).shuffle(domain)
```
stephenroller, line 124: can we make the seed an argument?

stephenroller (approved):

stephenroller:
(also a test plz)

PR #3678, opened 2021-05-27
WikiToxic agents.py slight change 

liliarose:
**Patch description**
Another dataset is using WikiToxic as its parent, but its _get_data() function's error message hardcodes WikiToxic's kaggle dataset source, so changing that to a variable instead. 

**Testing steps**
testing with test.py and also tried parlai display data
passed with 
```
parlai/tasks/dialogue_safety/test.py ...                    [100%]
```

----- 
update: I think I tried rebasing it, and then I had to force push it.... 
Also, the current checks that are failing are failing, I think, because of this:

*This build could not be run with the selected resource class. The default Medium resource class was used to run this build. To use the resource class to speed up your build or avoid an out-of-memory error, you can upgrade to a Performance Plan.*

Most of them are `Exited with code exit status 1`

### DIFF
--- a/parlai/tasks/dialogue_safety/agents.py
+++ b/parlai/tasks/dialogue_safety/agents.py
@@ -183,6 +183,7 @@ class WikiToxicCommentsTeacher(FixedDialogTeacher):
 
         self.use_test_set = opt['use_test_set']
         self.balance_data = opt['balance_data']
+        self.DATA_SOURCE = '<https://www.kaggle.com/c/jigsaw-toxic-comment-classification-challenge/data>'
 
         self.data_path = os.path.join(
             opt['datapath'], 'dialogue_safety', 'wiki-toxic-comments'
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -43,7 +43,7 @@ tensorboardX==2.1
 tokenizers>=0.8.0
 torchtext>=0.5.0
 tornado==6.0.4
-tqdm~=4.36.1
+tqdm~=4.38.0
```
jxmsML, line 46: is it from previous merged commits or new one? if it's previously merged commit please rebase on the latest master. if not, I would suggest we separate this into 2 different pull requests.

liliarose: yep... it's a previous merge... will try to rebase 👍

jxmsML (approved):

jxmsML:
@liliarose please check these unit test failures.

PR #3681, opened 2021-05-31
ChunkTeacher no longer requires num_examples/num_episodes to be correct

stephenroller:
**Patch description**
For as long as we've had ChunkTeacher, it's required that the number of episodes and examples be exactly reported correctly, or else it would cause a hang. With this patch, we loosen that requirement.

Methodology:
- Upgrade some of ChunkTeacher's internal logic to use DatatypeHelper to check about shuffling/cycling. Some of this is equivalent but more explicit and makes a bit more sense. The one logical change here is `train:ordered` was previously cycling, when it shouldn't.
- FixedDialogTeacher keeps track of the current episode and sees if we'll go over our limit, and starts emitting padding. Modify ChunkTeacher to always lie about its "next episode" so that check never triggers.
- Epoch done was previously computed in FixedDialogTeacher.next_example, checking the counts. Override this behavior in ChunkTeacher to have it simply test whether it's seeing a pad or not. This means we'll go one batch further than we need to, but that's okay -- everything filters the pads correctly and it doesn't affect counts or examples seen.

**Testing steps**
Manual testing.

New CI tests that stand up chunk teachers that lie about their number examples/episodes, and check that the number of unique messages seen is exactly correct.

### DIFF
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -2353,6 +2353,7 @@ class ChunkTeacher(FixedDialogTeacher, ABC):
 
         self._episode_done = True
         self.last_queue_output = None
+        self._ct_epoch_done = False
 
     def _get_data_folder(self):
         if not self.opt.get('datafile'):
@@ -2526,16 +2527,9 @@ class ChunkTeacher(FixedDialogTeacher, ABC):
         next_chunk, chunk_reset_cnt = self.chunks.get()
         if next_chunk is None:
             if self.is_train:
+                # start putting chunks back onto the queue
                 self._enqueue_chunks()
-                next_chunk, chunk_reset_cnt = self.chunks.get()
-                if next_chunk is None:
-                    # See the race condition described around "gross hack" in
-                    # _enqueue_chunks.  if we win the race condition, then
-                    # catch it here
-                    next_chunk, chunk_reset_cnt = self.chunks.get()
-            else:
-                # if we're in valid/test, we need to actually signal the end
-                return (None, chunk_reset_cnt)
+            return (None, chunk_reset_cnt)
         # abstract method `load_from_chunk` returns a list of tuples
         output = self.load_from_chunk(next_chunk)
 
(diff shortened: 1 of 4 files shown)


parlai/core/teachers.py
```diff
@@ -2525,25 +2532,30 @@ def get_chunk(self):
         """
         next_chunk, chunk_reset_cnt = self.chunks.get()
         if next_chunk is None:
-            if self.is_train:
+            if DatatypeHelper.should_cycle(self.datatype):
+                # start putting chunks back onto the queue
                 self._enqueue_chunks()
-                next_chunk, chunk_reset_cnt = self.chunks.get()
```
stephenroller, line 2530: Need to double check this part going away

parlai/core/teachers.py
```diff
@@ -2539,11 +2547,23 @@ def get_chunk(self):
...
+    def next_example(self):
+        # next_example will always fail to provide useful signal on whether
+        # we're at the end of an epoch in chunk teacher. Instead, the queue
+        # empties and we simply start outputting pads forever. As such, we'll
+        # measure epochs when we start receiving only pads.
+
+        # (This isn't relevant for the training loop, which loops for ever and
+        # never "epochs").
+        retval, fake_epoch_done = super().next_example()
+        real_epoch_done = retval.is_padding()
```
emilydinan, line 2564: I assume this works OK with dynamic batch padding, since that occurs at a level higher than the teacher?

emilydinan (approved):
awesome. should we require a user to implement `num_examples` now? (setting it as an abstractmethod requires the implementation)

parlai/core/teachers.py
stephenroller: Yes, this has been tested with all variants of batching and ensured that all examples are seen.

stephenroller:
You still have to implement num_examples, as that determines the behavior of `-eps` and `-veps` etc. It just doesn't have to line up with data

klshuster (approved):
awesome

PR #3688, opened 2021-06-04
[RAG] Handle Different DPR Model File with Pre-trained Model

klshuster:
**Patch description**
This patch fixes a bug in which one would like to swap the `--dpr-model-file` for a *pre-trained* RAG or FiD model. The underlying issue is that the dpr model weights are loaded during model initialization, **PRIOR TO** loading the pre-trained model weights. This is fine when the pre-trained model is an underlying seq2seq model (BART, T5, BB), but it is not ok when the pre-trained model is a RAG or FiD model.

The solution is to overwrite the retriever weights in the `state_dict` with the _already loaded dpr model weights_.

**Testing steps**
Included CI testing:

```
$ pytest -k TestLoadDPRModel
===== test session starts =====
platform linux -- Python 3.7.9, pytest-6.2.1, py-1.10.0, pluggy-1.0.0.dev0
rootdir: /private/home/kshuster/ParlAI, configfile: pytest.ini
plugins: hydra-core-1.0.0, requests-mock-1.8.0, regressions-2.1.1, datadir-1.3.1
collected 96 items / 95 deselected / 1 selected

test_rag.py .                                                                                                                                                                      [100%]

=====slowest 10 durations =====
41.01s call     tests/nightly/gpu/test_rag.py::TestLoadDPRModel::test_load_dpr

(2 durations < 0.005s hidden.  Use -vv to show these durations.)
=====1 passed, 95 deselected, 4 warnings in 43.30s =====
```

### DIFF
--- a/parlai/agents/rag/rag.py
+++ b/parlai/agents/rag/rag.py
@@ -16,7 +16,7 @@ import os
 import torch
 import torch.nn
 import torch.cuda
-from typing import Any, Dict, List, Optional, Tuple, Union, Type, overload
+from typing import Any, Dict, List, Optional, Tuple, Union, Type
 
 from parlai.agents.bart.bart import BartAgent
 from parlai.agents.hugging_face.t5 import T5Agent
@@ -358,16 +358,12 @@ class RagAgent(TransformerGeneratorRagAgent, BartRagAgent, T5RagAgent):
         """
         return self._generation_agent.build_dictionary(self)
 
+    @staticmethod
     def update_state_dict(
-        self, opt: Opt, state_dict: Dict[str, torch.Tensor], model: torch.nn.Module
+        opt: Opt, state_dict: Dict[str, torch.Tensor], model: torch.nn.Module
     ):
         """
         Update the given state dict to be RAG-ified.
-
-        Under certain circumstances, one may wish to specify a different
-        `--dpr-model-file` for a pre-trained, RAG model. Thus, we additionally
-        check to make sure that the loaded DPR model weights are not overwritten
-        by the state loading.
 
         :param opt:
             options
@@ -410,6 +406,15 @@ class RagAgent(TransformerGeneratorRagAgent, BartRagAgent, T5RagAgent):
     def _should_override_dpr_model_weights(self, opt: Opt):
         """
         Determine if we need to override the DPR Model weights.
+
+        Under certain circumstances, one may wish to specify a different
+        `--dpr-model-file` for a pre-trained, RAG model. Thus, we additionally
+        check to make sure that the loaded DPR model weights are not overwritten
+        by the state loading.
+
+        NOTE: If `--model-file M` was trained with `--dpr-model-file D`, and
+        `--dpr-model-file D` is specified *after training* (i.e., in eval/interactive),
+        we cannot currently handle replacing those weights.
         """
         try:
             init_model, _ = self._get_init_model(opt, None)
--- a/parlai/agents/rag/rag.py
+++ b/parlai/agents/rag/rag.py
@@ -412,23 +412,16 @@ class RagAgent(TransformerGeneratorRagAgent, BartRagAgent, T5RagAgent):
         check to make sure that the loaded DPR model weights are not overwritten
         by the state loading.
 
-        NOTE: If `--model-file M` was trained with `--dpr-model-file D`, and
-        `--dpr-model-file D` is specified *after training* (i.e., in eval/interactive),
-        we cannot currently handle replacing those weights.
         """
-        try:
-            init_model, _ = self._get_init_model(opt, None)
-            init_model_opt = Opt.load(f'{init_model}.opt')
-            override_dpr = modelzoo_path(
-                opt['datapath'], opt['dpr_model_file']
-            ) != modelzoo_path(opt['datapath'], init_model_opt['dpr_model_file'])
-            if override_dpr:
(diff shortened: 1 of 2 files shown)


tests/nightly/gpu/test_rag.py
```diff
@@ -388,5 +398,70 @@ def test_bart_fid_rag_dpr_poly(self):
...
+        # First, we'll load up a DPR model from the zoo dpr file.
+        default_query_encoder = DprQueryEncoder(
+            opt, dpr_model='bert', pretrained_path=DPR_ZOO_MODEL
+        )
+        query_encoder = DprQueryEncoder(
+            opt,
+            dpr_model='bert_from_parlai_rag',
+            pretrained_path=RAG_SEQUENCE_ZOO_MODEL,
+        )
+        assert not torch.allclose(
```
mojtaba-komeili, line 421: nit: why not using the `unittest` functions like `assertIsNone` and `assertTrue` etc.?

parlai/agents/rag/rag.py
```diff
@@ -402,13 +403,48 @@ def update_state_dict(
...
+            init_model, _ = self._get_init_model(opt, None)
+            init_model_opt = Opt.load(f'{init_model}.opt')
+            override_dpr = modelzoo_path(
+                opt['datapath'], opt['dpr_model_file']
+            ) != modelzoo_path(opt['datapath'], init_model_opt['dpr_model_file'])
+            if override_dpr:
+                logging.warning(
+                    f"Overriding DPR Model with {modelzoo_path(opt['datapath'], opt['dpr_model_file'])}"
+                )
+        except FileNotFoundError:
```
mojtaba-komeili, line 429: should there be warning here?

parlai/agents/rag/rag.py
```diff
@@ -402,13 +403,48 @@ def update_state_dict(
...
+        by the state loading.
+
+        NOTE: If `--model-file M` was trained with `--dpr-model-file D`, and
+        `--dpr-model-file D` is specified *after training* (i.e., in eval/interactive),
+        we cannot currently handle replacing those weights.
+        """
+        try:
+            init_model, _ = self._get_init_model(opt, None)
+            init_model_opt = Opt.load(f'{init_model}.opt')
+            override_dpr = modelzoo_path(
```
mojtaba-komeili, line 422: couldn't it just compare `opt['dpr_model_file'] != init_model_opt['dpr_model_file']`, since the rest seems to be the same?

mojtaba-komeili (approved):
Thanks for fixing it so quickly.

parlai/agents/rag/rag.py
```diff
@@ -402,13 +403,48 @@ def update_state_dict(
...
+        """
+        Determine if we need to override the DPR Model weights.
+
+        Under certain circumstances, one may wish to specify a different
+        `--dpr-model-file` for a pre-trained, RAG model. Thus, we additionally
+        check to make sure that the loaded DPR model weights are not overwritten
+        by the state loading.
+
+        NOTE: If `--model-file M` was trained with `--dpr-model-file D`, and
+        `--dpr-model-file D` is specified *after training* (i.e., in eval/interactive),
```
moyapchen, line 416: Thought here - should there be an "--override-dpr-model-file-for-eval"?

(This was the initial case that triggered everything so...)

tests/nightly/gpu/test_rag.py
```diff
@@ -388,5 +398,70 @@ def test_bart_fid_rag_dpr_poly(self):
...
 
 
+class TestLoadDPRModel(unittest.TestCase):
+    """
+    Test loading different DPR models for RAG.
+
+    See RagAgent._should_override_dpr_model_weights for important note
+    regarding specifying the *same* dpr model file as was used to train
+    the model.
+    """
```
moyapchen, line 408: Nit: For clarity's sake, might be nice to be explicit about the 3 cases that show up and expected behavior

```
A -> Init DPR Model for M
B -> DPR Model within M after training M
C -> New DPR Model you're using to override
```

tests/nightly/gpu/test_rag.py
klshuster: i've spoken with @stephenroller about this and he's told me that it's all the same to pytest

parlai/agents/rag/rag.py
klshuster: hmm no, since it's supposed to be silent to the user

parlai/agents/rag/rag.py
klshuster: unfortunately I don't think so -> `modelzoo_path` only modifies the path if it starts with `zoo:`, otherwise it's a no-op. Someone could, theoretically, pass in the full path to their dpr model file, even if it is the zoo path, so we need to make sure that the reference isn't the same

klshuster:
i actually think there is a simpler way to do this, implementing now

parlai/agents/rag/rag.py
klshuster: i think this case i mention here was not what triggered things, but with the new implementation it's handled smoothly

tests/nightly/gpu/test_rag.py
klshuster: added a comment!

klshuster:
Re-requesting review as this can be handled better

moyapchen (approved):
LGTM. Thanks!

PR #3697, opened 2021-06-09
Hash + Ladder Agent

jaseweston:


projects/params_vs_compute/hash_ladder/hash_ladder.py
```diff
@@ -0,0 +1,477 @@
...
+        else:
+            nn.init.normal_(
+                self.position_embeddings.weight, 0, self.embedding_size ** -0.5
+            )
+
+        # build the model
+        self.layers = nn.ModuleList()
+        for i in range(self.n_layers):
+            if self.opt['hash_layer'] == i:
+                # logging.debug(f'Hash Layer on layer: {i}')
```
spencerp, line 173: nit; Uncomment or remove.

projects/params_vs_compute/hash_ladder/hash_ladder.py
```diff
@@ -0,0 +1,477 @@
...
+from parlai.agents.transformer.transformer import TransformerGeneratorAgent
+from parlai.core.opt import Opt
+from parlai.core.params import ParlaiParser
+from parlai.utils.misc import warn_once
+import torch.nn.functional as F
+
+try:
+    from apex.normalization.fused_layer_norm import FusedLayerNorm as LayerNorm
+
+    APEX_LAYER_NORM = True
```
spencerp, line 32: We removed Apex, didn't we? If we need to keep this in this file for reproducibility that's fine, but I imagine it will also end up being removed eventually.

projects/params_vs_compute/hash_ladder/hash_ladder.py
```diff
@@ -0,0 +1,477 @@
...
+        self.n_heads = opt['n_heads']
+        self.dim = self.embedding_size
+        self.activation = opt.get('activation', 'relu')
+        self.variant = opt.get('variant', 'aiayn')
+
+        self.embeddings_scale = opt.get('embeddings_scale', True)
+        dropout_frac = opt.get('dropout', 0.0)
+        self.dropout = nn.Dropout(p=dropout_frac)  # --dropout
+
+        self.n_positions = _default(n_positions, get_n_positions_from_options(opt))
```
spencerp, line 132: Looks like many of these properties are the same as `TransformerDecoder` and you already call `super().init()`. Perhaps you only need the layer building code here.

I can add a `build_layers` function to `TransformerDecoder` in a subsequent PR for overriding purposes.

projects/params_vs_compute/hash_ladder/hash_ladder.py
```diff
@@ -0,0 +1,477 @@
...
+        )  # type: ignore
+        self.norm2 = torch.nn.LayerNorm(embedding_size, eps=LAYER_NORM_EPS)
+
+        self.ffn = HashLayerFFN(
+            opt,
+            embedding_size,
+            ffn_size,
+            relu_dropout=relu_dropout,
+            activation=activation,
+        )  # type: ignore
```
spencerp, line 330: Same here; it looks like we're only overloading `HashLayerFFN`? If so, you can just create your `TransformerDecoderLayer` like this without overriding:
```
TransformerDecoderLayer.with_components(feedforward=HashLayerFFN)
```

spencerp (approved):
There's some code that can be consolidated, but I didn't see any red flags. Accepting to unblock and I can do a cleanup/consolidation pass later.

projects/params_vs_compute/hash_ladder/hash_ladder.py
jaseweston: done

projects/params_vs_compute/hash_ladder/hash_ladder.py
jaseweston: removed

projects/params_vs_compute/hash_ladder/hash_ladder.py
jaseweston: ok, i'll look at these later after this initial merge, thanks

projects/params_vs_compute/hash_ladder/hash_ladder.py
jaseweston: ok, i'll look at these later after this initial merge, thanks

jaseweston: (but i think i can't do that because i need to pass through the original inputs in the decoder layer)

PR #3700, opened 2021-06-09
add paper links

jaseweston:


projects/params_vs_compute/hash_ladder/hash_ladder.py
```diff
@@ -42,6 +42,8 @@
 class HashLadderAgent(TransformerGeneratorAgent):
     """
     Simple implementation of Hash Layers and the Ladder model from the following papers:
+    https://arxiv.org/abs/2106.04426
```
spencerp, line 45: I think if you add a newline before these it will appease the lint.

spencerp (approved):
send iiiit

PR #3708, opened 2021-06-11
Pass opt and kwargs all the way through transformer

spencerp:
**Patch description**
Looking at the hash_ladder code, it's apparent there are some more optimizations I could make to the transformer code. This PR:
- Passes `opt` through to all components that are defined in our codebase
- Passes `kwargs` to all `forward` functions
- Adds a `build_layers` function to `TransformerDecoder` and `TransformerEncoder` that allows for per-layer modifications

**Testing steps**
working with @jaseweston to make sure I didn't mess anything up; will paste commands ran

### DIFF
--- a/parlai/agents/transformer/modules/attention.py
+++ b/parlai/agents/transformer/modules/attention.py
@@ -14,6 +14,7 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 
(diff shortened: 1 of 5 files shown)


jaseweston (approved):
needs dat autoformat

amazing stuff!

PR #3713, opened 2021-06-14
[train] Fix bugs with loading validation impatience.

stephenroller:
**Patch description**
Validation impatience has been loading incorrectly upon preemption for a bit.

The following two bugs are observed and fixed:

- `.checkpoint` was saved BEFORE the validation report had happened, meaning it would always save epochs/progress/etc with the latest values, but the "best validation score" of the PREVIOUS validation.
- The `impatience` key was not being set, and therefore always being loaded with 0 (the default).

**Testing steps**
About to write unit tests.

### DIFF
--- a/parlai/scripts/train_model.py
+++ b/parlai/scripts/train_model.py
@@ -477,6 +477,7 @@ class TrainLoop:
                     'train_reports': self.train_reports,
                     'valid_reports': self.valid_reports,
                     'best_valid': self.best_valid,
+                    'impatience': self.impatience,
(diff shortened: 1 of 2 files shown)


jaseweston:
awesome thanks!

jaseweston (approved):

mojtaba-komeili:
Thank you for quick fix!

klshuster (approved):

jxmsML (approved):

PR #3714, opened 2021-06-14
[BART] Do not add start/end tokens multiple times

emilydinan:
**Patch description**
Found an issue with BART in which -- if we cache the text_vec -- start/end tokens are added multiple times. I was using this caching behavior to do "scoring" of candidates with BART, when we had too many candidates to fit in memory.

I adopted the same solution we use for BERT: https://github.com/facebookresearch/ParlAI/blob/c4c266989f266533276efb4d5ff7394c7c5b51ee/parlai/agents/bert_ranker/bi_encoder_ranker.py#L147

CC @adamlerer 

**Testing steps**
I added a test.


klshuster (approved):
seems harmless enough

PR #3715, opened 2021-06-14
Update call sites missing opt

spencerp:
Adding some call sites that unit tests didn't catch on #3708 

**Testing Steps**
`pytest -k TestPolyencoder`


stephenroller:
If unit tests didn't catch them, can you add tests that do?

spencerp:
> If unit tests didn't catch them, can you add tests that do?

done!

stephenroller (approved):

PR #3716, opened 2021-06-14
[StyleGen] Classifier on generator: BART bug

emilydinan:
**Patch description**
Fix classifier on generator code to work for BART. The code overrides the TransformerDecoder, but does not include the forward through the norm_embeddings if `variant == 'bart'` 

This bug was caught when we do DDP -- DDP expects a grad for the norm_embeddings after forwarding but none is found


EricMichaelSmith (approved):
looks good!

PR #3718, opened 2021-06-14
[eval model] store world logs per task

### LINKED ISSUE #3648: Using world-logs while evaluating a model with more than one task, only keeps the world log for the last task.
Trying to evaluate one mode with two tasks, while keeping the model outputs in world logs, I noticed that there was no result from one of the tasks in the world log. To reproduce this one may try running
```
 parlai eval_model -t wizard_of_wikipedia,babi \
--world-logs /some/path/world-log \
--num-examples 1 --model repeat_label
```
Running this, there is only a single line in the `world-log.json` file. Checking the file you can see `"id": "babi:Task1k:1"` which may mean that parlai is generating separate world log files for each task, but assumes same name for all of them and writes over the previous ones.

jxmsML:
**Patch description**
Currently when running `parlai em --world-logs .....` it only stores the world logs for the very last task if multitasking. 
This pr would fix this issue by adding the task name as suffix to the world log file path if it's multitasking.

Fixes #3648

**Testing steps**

```
.(conda_parlai) jingxu23@devfair0173:~/ParlAI$ python /private/home/jingxu23/ParlAI/tests/test_transformers.py

.....
Ran 34 tests in 149.635s

OK
```

**Other information**

### DIFF
--- a/parlai/scripts/eval_model.py
+++ b/parlai/scripts/eval_model.py
@@ -118,13 +118,27 @@ def _save_eval_stats(opt, report):
         f.write("\n")  # for jq
 
 
+def get_task_world_logs(task, world_logs, is_multitask=False):
+    if not is_multitask:
+        return world_logs
+    else:
+        base_outfile, extension = os.path.splitext(world_logs)
+        return f'{base_outfile}_{task}{extension}'
+
+
 def _eval_single_world(opt, agent, task):
     logging.info(f'Evaluating task {task} using datatype {opt.get("datatype")}.')
     # set up world logger
-    world_logger = WorldLogger(opt) if opt['world_logs'] else None
-
     task_opt = opt.copy()  # copy opt since we're editing the task
     task_opt['task'] = task
+    # add task suffix in case of multi-tasking
+    if opt['world_logs']:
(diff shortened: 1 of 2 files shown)


stephenroller:
Please add a test, and also merge the entire bits into the f-string, rather than using concatenation

parlai/scripts/eval_model.py
```diff
@@ -118,13 +118,27 @@ def _save_eval_stats(opt, report):
         f.write("\n")  # for jq
 
 
+def get_task_world_logs(task, world_logs, is_multitask=False):
+    if not is_multitask:
+        return world_logs
+    else:
+        base_outfile, extension = os.path.splitext(world_logs)
+        return f'{base_outfile}_{task}{extension}'
```
kauterry, line 126: `        return f'{base_outfile}_{task}.{extension}'
`

stephenroller: No, splitext includes the . In the ext

kauterry: My bad, sorry!

jxmsML: the extension contains "."

PR #3719, opened 2021-06-15
Fix LR scheduler cooldown

stephenroller:
**Patch description**
Context:
- Originally in ParlAI, fixed LR schedulers like cosine/linear would consume (warmup_updates + max_lr_steps) updates, cooling down to 0 eventually
- #3379 changed this behavior such that only max_lr_steps would be consumed, but did not change the cooldown to be faster
- This PR changes it so the full cooldown is completed by the end of max_lr_steps.

**Testing steps**
Adjusted CI, new assertions.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -955,7 +955,13 @@ class TorchAgent(ABC, Agent):
         is_train = 'train' in datatype and 'evalmode' not in datatype
         return is_train
 
-    def init_optim(self, params, optim_states=None, saved_optim_type=None) -> bool:
+    def init_optim(
+        self,
+        params,
+        optim_states=None,
+        saved_optim_type=None,
+        is_finetune: bool = False,
+    ) -> bool:
         """
         Initialize optimizer with model parameters.
 
@@ -969,10 +975,14 @@ class TorchAgent(ABC, Agent):
             type of optimizer being loaded, if changed will skip loading
             optimizer states
 
+        :param is_finetune:
+            bool indicating whether this training run is a fine-tune or not
+
         :returns:
-            boolean indicating whether the optimizer was initialized with
+            boolean indicating whether the optimizer failed to initialize with
             optim_states.
         """
+
         if hasattr(self, 'resized_embeddings') and self.resized_embeddings:
             optim_states = None
             logging.warning('Not loading optimizer due to resize in token embeddings')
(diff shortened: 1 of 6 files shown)


emilydinan (approved):
thanks for the fix! this lgtm

parlai/core/torch_agent.py
```diff
@@ -1055,14 +1065,15 @@ def init_optim(self, params, optim_states=None, saved_optim_type=None) -> bool:
...
+            return True
+
         if optim_states and saved_optim_type != opt['optimizer']:
             # we changed from adam to adamax, or sgd to adam, or similar
             logging.warning('Not loading optim state since optim class changed.')
-            return False
+            return True
         elif optim_states:
             # check for any fp16/fp32 conversions we need to do
             optimstate_fp16 = 'loss_scaler' in optim_states
```
klshuster, line 1079: i can't leave a comment for it, but are the semantics correct for line 1099? the `elif not optimstate_fp16 and self.fp16` block? are we returning `True` always because of the lower precision conversion?

parlai/nn/lr_scheduler.py
```diff
@@ -502,13 +507,13 @@ def __init__(
...
-        self.max_lr_steps = max_lr_steps
+        assert self.warmup_updates >= 0
+        self.max_lr_steps = max_lr_steps - self.warmup_updates
         self.scheduler = optim.lr_scheduler.LambdaLR(optimizer, self._linear_lr)
 
     def _linear_lr(self, step):
         # this multiplicative factor ensures linear decay rate
-        # lr_mult = float(self.max_lr_steps - step - 1) / float(self.max_lr_steps - step)
-        lr_mult = max(0.0, 1e-6 + (1.0 - step / self.max_lr_steps) * (1 - 1e-6))
+        lr_mult = max(0.0, 1.0 - step / self.max_lr_steps)
```
klshuster, line 516: we dont need `1e-6` anymore?

klshuster (approved):

parlai/nn/lr_scheduler.py
stephenroller: I made an executive call to let it actually go to 0 :P

parlai/core/torch_agent.py
stephenroller: agreed

PR #3721, opened 2021-06-15
Dynamic Batching doesn't choke with really small datasets

stephenroller:
**Patch description**
Fixes a bug with dynamic batching where we would crash with an assertion error on later workers if the dataset was smaller than the number of GPUs.

**Testing steps**
New CI


skiingpacman (approved):
Approved subject to fixing lint errors

PR #3724, opened 2021-06-15
Naive ws version bump

EricMichaelSmith:
**Patch description**
Fix vulnerability by just bumping the version of `ws`

**Testing steps**
Naively just making sure that CI checks don't break; not sure if this is sufficient


klshuster (approved):
🤷

PR #3726, opened 2021-06-16
Use standard tool for cleaning up metrics

EricMichaelSmith:
**Patch description**
Use `Message.json_safe_payload()` for handling serialization of acts instead of manually deleting `'metric'` keys.

**Testing steps**
CI checks


klshuster (approved):
awesome!

PR #3727, opened 2021-06-16
Weighted f1 multiclass

arendu:
**Patch description**
weighted f1 was calculated was broken -- it assumes there are only 2 classes. Specifically `total_examples` calculation was corrected for multiclass.

**Testing steps**

**Other information**


arendu:
sorry left an older commit in there..

PR #3728, opened 2021-06-16
updated weighted_f1 to not assume binary classification

arendu:
**Patch description**
weighted f1 was calculated was broken -- it assumes there are only 2 classes. Specifically `total_examples` calculation was corrected for multiclass.

**Testing steps**

**Other information**


emilydinan (approved):
thank you for the fix!

PR #3729, opened 2021-06-16
Parallelize detensorizing context over batch

EricMichaelSmith:
**Patch description**
Currently, when generating, we convert the context Tensor to a list one example at a time, which is expensive; this PR parallelizes this operation over the entire batch.

Performance on 1000 generations (`-mf zoo:blender/blender_90M/model -t blended_skill_talk -ne 1000`), on an otherwise unoccupied devfair:

- Original code, trial 1: median elapsed time 783 ms, mean 797 ms
- Original code, trial 2: median 770 ms, mean 785 ms
- This PR, trial 1: median 651 ms, mean 663 ms
- This PR, trial 2: median 665 ms, mean 677 ms

**Testing steps**
CI checks (which seem to cover context blocking)

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -1011,7 +1011,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
         """
         if self.beam_context_block_ngram <= 0:
             # We aren't context blocking, return empty tensor of the correct size
-            return torch.LongTensor([[]] * batch.batchsize)
+            return torch.zeros(batch.batchsize, 0, dtype=torch.long)
 
         ctxt = batch.text_vec
         if self.beam_block_full_context:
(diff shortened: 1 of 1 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -1005,6 +1005,19 @@ def _get_context(self, batch, batch_idx):
             ctxt = batch.full_text_vec[batch_idx]
         return ctxt
 
+    def _get_batch_context(self, batch):
+        """
+        Version of TGA._get_context() that operates on full batches for speed.
+        """
+        if self.beam_context_block_ngram <= 0:
+            # We aren't context blocking, return empty tensor of the correct size
+            return torch.LongTensor([[]] * batch.batchsize)
```
stephenroller, line 1014:
```suggestion
            return torch.zeros(batch.batchsize, 0)
```

stephenroller (approved):
This one seems great to me... Haven't reviewed the other. We can talk more offline.

parlai/core/torch_generator_agent.py
EricMichaelSmith: Yeah, that's cleaner - just added that, but with `dtype=torch.long`

PR #3730, opened 2021-06-16
Parallelize detensorizing token IDs during tree search

EricMichaelSmith:
**Patch description**
Currently, during any kind of tree search, the token ID tensor will be detensorized one item at a time, which is expensive. This PR rewrites that to detensorize the whole batch at once.

Performance on 1000 generations (`-mf zoo:blender/blender_90M/model -t blended_skill_talk -ne 1000`), on an otherwise unoccupied devfair:

- Original code, trial 1: median elapsed time 783 ms, mean 797 ms
- Original code, trial 2: median 770 ms, mean 785 ms
- This PR, trial 1: median 660 ms, mean 671 ms
- This PR, trial 2: median 656 ms, mean 667 ms

**Testing steps**
CI checks

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -1406,8 +1406,9 @@ class TreeSearch(object):
 
         self.outputs.append(tok_ids)
         self.bookkeep.append(hyp_ids)
+        tok_id_list = tok_ids.tolist()
         self.partial_hyps = [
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
so good.

stephenroller:
do we need to port this change to torchscript?

EricMichaelSmith:
> do we need to port this change to torchscript?

Hmm, doesn't look like it, because torchscript export currently only supports greedy search

PR #3732, opened 2021-06-17
Eliminate dummy batches and init_cuda_buffer.

stephenroller:
**Patch description**
Eliminate the dreaded `_dummy_batch` and `_init_cuda_buffer`.

Users have complained about the presence of these methods. When debugging, the first batch being a dummy causes some confusion (why am I getting gibberish data?) and the manual implementation of a `_dummy_batch` is annoying (why is my model breaking on the first pass?)

Adopt Fairseq's newer approach:
- Cache the very first batch we ever seen. If it's invalid, we should be raising an exception anyway. If it's valid, it will be good enough
- Do NOT initialize the cuda buffer. Just keep that batch around
- If we need to recover from an OOM, used the cached dummy batch as the data.

**Testing steps**
CI.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -578,59 +578,36 @@ class TorchGeneratorAgent(TorchAgent, ABC):
         else:
             self.skip_generation = self.opt.get('skip_generation', False)
 
-    def _dummy_batch(self, batchsize, maxlen):
+    def _cache_dummy_batch(self, batch: Batch):
         """
-        Create a dummy batch.
-
-        This is used to preinitialize the cuda buffer, or otherwise force a
-        null backward pass after an OOM.
-
-        If your model uses additional inputs beyond text_vec and label_vec,
-        you will need to override it to add additional fields.
+        Cache a batch to be used as a dummy during _fake_forward_pass.
         """
-        text_vec = (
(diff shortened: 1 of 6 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -578,59 +578,36 @@ def set_interactive_mode(self, mode, shared=False):
...
+    def _cache_dummy_batch(self, batch: Batch):
         """
-        Create a dummy batch.
-
-        This is used to preinitialize the cuda buffer, or otherwise force a
-        null backward pass after an OOM.
+        Cache a batch to be used as a dummy during _fake_forward_pass.
+        """
+        if not hasattr(self, '_dummy_batch'):
+            self._dummy_batch = batch
```
klshuster, line 586: does this unnecessarily use up memory? what if the first batch is huge for example?

klshuster (approved):
thanks for making this change. I'm approving but did have one perhaps pertinent question

parlai/core/torch_generator_agent.py
stephenroller: Could be a worse problem for images. For text, imagine it's a 2x2048x1024 LongTensor (2x for content and label; bs 2048; string length 1024). A long is 8 bytes. So we'll be wasting 32MiB of memory.

PR #3739, opened 2021-06-22
[mp] Use free ports rather than random ports

stephenroller:
**Patch description**
I've noticed that sometimes tests will fail when using randomized ports, because the address is already in use. Resolve by always asking the OS for a free port.

Thanks stack overflow.

**Testing steps**
CI


stephenroller:
Easier if this is just in #3740.

PR #3740, opened 2021-06-22
Fully Sharded Data Parallel

stephenroller:
**Patch description**
Add support for Fairscale's FullyShardedDataParallel (FSDP). This is an implementation of DeepSpeed's Zero2 optimization, wherein optimizer state and gradients are sharded across different workers in order to reduce memory usage. Switching to `--ddp-backend zero2` results in about a 25% speedup in UPS (without bg workers, probably can be a bit higher), and about a 50% reduction in memory usage. It's recommended everyone switches to this for distributed training, and use the savings to increase batchsize or lower number of GPUs.

We also carve out support for Zero3, but cannot support it at this time due to high level design in ParlAI. See #3753 for a detailed description of why, and how we might overcome this in the future.

As a side change, this also makes our unit tests use OS-assigned free ports, instead of randomized ones, to slightly improve the reliability of running our test suites. I tried pulling this into another PR, but got tired of dealing with stacking.

**Testing steps**
Manual tests. New CI.

Here are some screenshots from a sweep that contained both `--ddp-backend ddp` and `--ddp-backend zero2`

image

image

image

image

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -772,6 +772,16 @@ class ParlaiParser(argparse.ArgumentParser):
         grp.add_argument(
             '--distributed-world-size', type=int, help='Number of workers.'
         )
+        grp.add_argument(
+            '--ddp-backend',
+            choices=['ddp', 'zero2', 'zero3'],
+            default='ddp',
+            help=(
+                'Distributed backend. Zero2 can be faster but is more experimental. '
+                'Zero3 uses radically less memory, but is slower. DDP is the most '
+                'tested.'
+            ),
+        )
         return grp
 
     def add_model_args(self):
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -774,12 +774,12 @@ class ParlaiParser(argparse.ArgumentParser):
         )
         grp.add_argument(
             '--ddp-backend',
-            choices=['ddp', 'zero2', 'zero3'],
+            # TODO: add in zero3. https://github.com/facebookresearch/ParlAI/issues/3753
+            choices=['ddp', 'zero2'],
             default='ddp',
             help=(
                 'Distributed backend. Zero2 can be faster but is more experimental. '
-                'Zero3 uses radically less memory, but is slower. DDP is the most '
-                'tested.'
+                'DDP is the most tested.'
             ),
         )
         return grp
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1052,7 +1052,9 @@ class TorchAgent(ABC, Agent):
         self.optimizer = optim_class(params, **kwargs)
         if self.fp16:
             if self.fp16_impl == 'safe':
-                self.optimizer = SafeFP16Optimizer(self.optimizer)
+                self.optimizer = SafeFP16Optimizer(
+                    self.optimizer, self._should_sync_overflows()
+                )
             else:
                 # Using memory efficient optimizer
                 opt_name = opt['optimizer']
@@ -1064,7 +1066,9 @@ class TorchAgent(ABC, Agent):
                         'with Memory Efficient FP16. Please select from among this '
                         f'list:\n{compatible_list}'
                     )
-                self.optimizer = MemoryEfficientFP16Optimizer(self.optimizer)
+                self.optimizer = MemoryEfficientFP16Optimizer(
+                    self.optimizer, self._should_sync_overflows()
+                )
 
         if is_finetune:
             logging.warning('Detected a fine-tune run. Resetting the optimizer.')
@@ -1112,6 +1116,9 @@ class TorchAgent(ABC, Agent):
                     'WARNING: not loading optim state since model params changed.'
                 )
                 return True
+
+    def _should_sync_overflows(self):
+        return self.fp16 and self.opt['ddp_backend'] in ('zero2', 'zero3')
 
     def build_lr_scheduler(self, states=None, hard_reset=False):
         """
@@ -2345,6 +2352,9 @@ class TorchAgent(ABC, Agent):
             self.global_metrics.add('gnorm', GlobalAverageMetric(grad_norm))
 
         if self.fp16:
+            logging.info(
+                f"fp16_loss_scale = {self.optimizer.loss_scale} [{self._number_training_updates}]"
+            )
             self.global_metrics.add(
                 'fp16_loss_scalar', GlobalAverageMetric(self.optimizer.loss_scale)
             )
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -1117,9 +1117,6 @@ class TorchAgent(ABC, Agent):
                 )
                 return True
 
-    def _should_sync_overflows(self):
-        return self.fp16 and self.opt.get('ddp_backend', 'ddp') in ('zero2', 'zero3')
-
     def build_lr_scheduler(self, states=None, hard_reset=False):
         """
         Create the learning rate scheduler, and assign it to self.scheduler. This
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -36,6 +36,7 @@ from parlai.core.message import Message
 from parlai.utils.distributed import is_distributed
 from parlai.utils.misc import AttrDict, warn_once
 from parlai.utils.io import PathManager
+from parlai.utils.fsdp import should_sync_gradnorm, is_fsdp
 from parlai.utils.fp16 import (
     SafeFP16Optimizer,
     MemoryEfficientFP16Optimizer,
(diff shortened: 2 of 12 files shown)


stephenroller:
See #3753 for why Zero3 won't be supported in this implementation.

parlai/core/params.py
```diff
@@ -772,6 +772,16 @@ def add_distributed_training_args(self):
         grp.add_argument(
             '--distributed-world-size', type=int, help='Number of workers.'
         )
+        grp.add_argument(
+            '--ddp-backend',
+            choices=['ddp', 'zero2', 'zero3'],
```
EricMichaelSmith, line 777: Hmm should we even give `'zero3'` as an option for the time being? (Don't really care either way)

parlai/utils/fsdp.py
```diff
@@ -0,0 +1,112 @@
...
+    return (
+        self.fp16
+        and self.opt.get('ddp_backend', 'ddp') in ('zero2', 'zero3')
+        and self.opt['fp16_impl'] == 'safe'
+    )
+
+
+def should_sync_gradnorm(opt):
+    """
+    Indicates whether fp16 optimizer wrappers should cumulate over workers.
```
EricMichaelSmith, line 95: Nit: "accumulate"?

parlai/core/torch_agent.py
```diff
@@ -1992,6 +1998,16 @@ def state_dict(self):
 
         return states
 
+    def save_nonprimary(self, path=None):
+        """
+        Save model parameters, when you are working on the non-primary worker.
+
+        For models or optimizers that shard parameters, this ensures we sync.
+        """
+        if self.opt.get('ddp_backend', 'ddp') in ('zero2', 'zero3'):
```
EricMichaelSmith, line 2007: Nit: should we pull in `DEFAULT_DDP_BACKEND` here?

parlai/core/torch_generator_agent.py
```diff
@@ -530,7 +532,11 @@ def __init__(self, opt: Opt, shared=None):
                 logging.warning("Optimizer was reset. Also resetting LR scheduler.")
             self.build_lr_scheduler(states, hard_reset=is_finetune or was_reset)
 
-        if shared is None and is_distributed():
+        if (
+            shared is None
+            and is_distributed()
+            and opt.get('ddp_backend', 'ddp') == 'ddp'
```
EricMichaelSmith, line 538: (same here about maybe using `DEFAULT_DDP_BACKEND` instead)

parlai/utils/distributed.py
```diff
@@ -346,3 +359,15 @@ def slurm_distributed_context(opt):
     except FileNotFoundError:
         # Slurm is not installed
         raise RuntimeError('SLURM does not appear to be installed.')
+
+
+def find_free_port() -> int:
+    """
+    Find a free port we can bind to locally.
```
EricMichaelSmith, line 366: ❤️

tests/test_distributed.py
```diff
@@ -155,45 +138,94 @@ def test_distributed_eval_stream_mode_max_exs(self):
...
+        n_heads=1,
+        ffn_size=32,
+        embedding_size=8,
+        verbose=True,
+    )
 
-        --model-parallel true.
+    def test_no_model_parallel(self):
+        """
+        Checks that we throw an error when combining mp_train with --model-parallel.
```
EricMichaelSmith, line 224: I love these save-the-user-from-themself checks :)

EricMichaelSmith (approved):
Seems reasonable - minor comments

parlai/core/torch_agent.py
```diff
@@ -1969,10 +1974,11 @@ def state_dict(self):
         """
         states = {}
         if hasattr(self, 'model'):  # save model params
-            if hasattr(self.model, 'module'):
-                # did we wrap in a DistributedDataParallel
+            if hasattr(self.model, 'module') and not is_fsdp(self.model):
```
klshuster, line 1977: nit: could make this a helper function too? like `should_sync_gradnorm` (not necessary of course)

parlai/core/torch_generator_agent.py
```diff
@@ -479,8 +480,10 @@ def __init__(self, opt: Opt, shared=None):
         else:
             # this is not a shared instance of this class, so do full init
             self.criterion = self.build_criterion()
-            # ensure all distributed copies will always be in sync
-            self.model = self.build_model()
+            with fsdp_utils.maybe_fsdp_wrap(opt):
+                self.model = fsdp_utils.fsdp_wrap(self.build_model())
+                if self.fp16 and not fsdp_utils.should_use_fsdp(opt):
```
klshuster, line 485: remember that bug with the instability stuff? is this not re-introducing it?

klshuster: (because we moved the model.half() call?)

parlai/scripts/multiprocessing_train.py
```diff
@@ -55,10 +54,12 @@ def multiprocess_train(
             raise
 
 
-def launch_and_train(opt, port):
+def launch_and_train(opt, port=None):
```
klshuster, line 57: will we ever specify a port here?

parlai/scripts/train_model.py
```diff
@@ -543,7 +546,7 @@ def validate(self):
             )
             self.best_valid = new_valid
             self.impatience = 0
-            if opt.get('model_file') and is_primary_worker():
+            if opt.get('model_file'):
```
klshuster, line 549: just making sure I understand - we can get rid of this check because it's handled in `save_model` right?

parlai/utils/fp16.py
```diff
@@ -55,27 +55,48 @@ def forward(self, scores, targets):
...
+        if sync:
+            # also need to get the norms from all the other sharded works in FSDP
+            import torch.distributed as dist
+
+            dist.all_reduce(normsq)
+        grad_norm = normsq.sqrt()
+        if max_norm > 0:
+            clip_coef = max_norm / (grad_norm + 1e-6)
+            for p in params:
+                p.grad.detach().mul_(clip_coef)
```
klshuster, line 97: why do we detach here?

parlai/utils/fsdp.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    Context manager for enabling wrapping in FullyShardedDataParallel.
+    """
+    if not should_use_fsdp(opt):
+        # make a no-op
+        yield
+        return
+
+    # zero3 not supported at this time. Throw an exception
+    if opt['ddp_backend'] == 'zero3':
```
klshuster, line 57: i know this is just for overkill testing but it's not even a choice in the param options so we'll already error there if calling from command line

parlai/utils/fsdp.py
```diff
@@ -0,0 +1,112 @@
...
+    Check whether we should keep the model in fp32 before other setup.
+
+    When using Zero2 or Zero3 backends with mixed precision, we need to avoid converting
+    the model to fp16, as the FSDP module does this for us.
+    """
+
+    return (
+        self.fp16
+        and self.opt.get('ddp_backend', 'ddp') in ('zero2', 'zero3')
+        and self.opt['fp16_impl'] == 'safe'
```
klshuster, line 89: but if we're using `mem_efficient` we don't delay?

klshuster (approved):
really really cool. lots of nits though (and a few real questions 😄 )

parlai/scripts/train_model.py
stephenroller: We need to be able do save_on_nonprimary_worker actually

parlai/utils/fp16.py
stephenroller: Don't want grads of grads! (This is in the original pytorch code too)

parlai/utils/fsdp.py
stephenroller: I'm leaving it for the future

parlai/utils/fsdp.py
stephenroller: Correct, see main comment

parlai/core/torch_generator_agent.py
stephenroller: Okay I think this needs to use my utility `should_delay_halving`. Forgot this.

We haven't really moved it the moment of halving. The operations between these two points don't do much, and the original code path should be about the same.

- We now half it on CPU instead of GPU, and then transfer. That's probably a small speedup in initialization really, with maybe some small numerical differences
- We model parallel after halving. Probably small speedup at initialization.
- We synchronize parameters after halving. Again, small initialization speedup.

The catch is that FSDP expects the model pre-halved if we're doing safe optimization, and post-halved if we're doing memory-efficient. (Similar to the optimizer wrappers, it looks for parameters of types to decide what type are the gradients).

This is the desired pattern
- If we're in Safe and using DDP, we SHOULD still halve, just as before
- If we're in MemEff and using DDP, we SHOULD still halve, just as before
- If we're in Safe and Zero2, we should NOT halve here
- If we're in MemEff and Zero2, we SHOULD halve here.

PR #3741, opened 2021-06-23
Convenience functions for Rare F1

spencerp:
**Patch description**
- Added some constructors for building rare F1 from various ParlAI data formats
- Fix bug where unseen words weren't counted
- Shortening name from `RareWordF1Calculator` to `RareF1Computer`
- Add rare_f1 to BlendedSkillTalk
- Unit tests
- `__iter__` for `DialogData`
- Fix for `Teacher` `__iter__`
- Type annotation of some agents functions

**Testing steps**
`pytest -k TestRareF1`

Sanity check that I didn't break blended_skill_talk: `parlai dd -t blended_skill_talk`

### DIFF
--- a/parlai/core/teachers.py
+++ b/parlai/core/teachers.py
@@ -187,7 +187,7 @@ class Teacher(Agent):
         self.epochDone = False
 
     # return state/action dict based upon passed state
-    def act(self):
+    def act(self) -> Message:
         """
         Act upon the previous observation.
         """
@@ -195,14 +195,14 @@ class Teacher(Agent):
             t = Message({'text': 'Hello agent!'})
         return t
 
-    def epoch_done(self):
+    def epoch_done(self) -> bool:
         """
         Return whether the epoch is done.
         """
         return self.epochDone
 
     # Default unknown length
-    def num_examples(self):
+    def num_examples(self) -> Optional[int]:
         """
         Return the number of examples (e.g. individual utterances) in the dataset.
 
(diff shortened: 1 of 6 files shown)


parlai/core/teachers.py
```diff
@@ -489,11 +489,11 @@ def next_example(self):
             and self._episode_done
             and self.episode_idx + self.opt.get("batchsize", 1) >= self.num_episodes()
         ):
-            epoch_done = True
+            self.epochDone = True
```
stephenroller, line 492: mmm i'm not sure about this

spencerp: Currently `epochDone` isn't updated to `True` anywhere in `FixedDialogTeacher`. Do you know what's up with that? Why would it _partially_ manage its own private state?

klshuster: what exactly is this change for?

spencerp: So that `iter(Teacher)` stops when it should stop and loops forever when it should loop forever.

It also seems like it's what _should_ happen. Otherwise `FixedDialogTeacher` relies on `get_orig_action` (in `act`) to set this properly. If a subclass overrides `act` but doesn't override `next_example`, it will need to manage `epochDone` itself.

But let me know if there's something I'm missing!

stephenroller: I'm still really nervous about this 😅 teachers is a bit of a mess and I'd like to move us towards deprecating all FDTs from the code...

PR #3742, opened 2021-06-23
Model chat id clarification

EricMichaelSmith:
**Patch description**
Clarify a comment the model_chat crowdsourcing task to no longer imply that the worker may not have an `id` set, because we manually always set the worker's `id` to `'Worker'`

**Testing steps**
CI checks


klshuster (approved):
failing tests look transient

PR #3745, opened 2021-06-24
[ChunkTeacher] Re-enable shuffling on -dt train:stream.

stephenroller:
**Patch description**
Experiments with @emilydinan showed:
- Since some of the refactors in #3681, CT has not been shuffling with `-dt train:stream`. This breaks backwards compatibility and causes significantly less generalization.
- Based on #3681, we no longer require every worker to shard up the chunks they are working on. As such, we can simply have every training worker load every chunk. This reduces the chance of oversampling when the number of workers does not evenly divide into the number of chunks.

**Testing steps**
CI. @emilydinan's experiments of generalization.


emilydinan (approved):
thank you!!

PR #3748, opened 2021-06-25
[WoW] Update WoW README

klshuster:
**Patch description**
Add in some info about the raw data file, since we seem to get questions about it


stephenroller (approved):
vielen dank

PR #3749, opened 2021-06-25
update decode task data link and checksum

meganung:
Data wasn't downloading / couldn't access data. Decode task data was acquired (thanks @stephenroller ) 
Then, uploaded the dataset and updated link/checksum in build.py.

tests ran: 
`parlai display_data --task decode`
`pytest parlai/tasks/decode/test.py`

### DIFF
--- a/parlai/tasks/decode/build.py
+++ b/parlai/tasks/decode/build.py
@@ -15,9 +15,8 @@ DECODE_VERSION = 'v0.1'
 
 RESOURCES = [
     DownloadableFile(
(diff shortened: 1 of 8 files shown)


parlai/tasks/decode/build.py
```diff
@@ -15,9 +15,8 @@
```
stephenroller, line 15: (Can't scroll up)

Make sure to bump the version to 'v0.1.1'

stephenroller (approved):
tyvm!

PR #3752, opened 2021-06-25
reproducing the bug in Dialog teacher

mojtaba-komeili:
**Patch description**
This is to discuss an issue that I ran into working with the `DialogTeacher`.
Using `parlai dd -t dt_bug:MyTeacher`, it is supposed to show me two episodes with 3 and 4 examples, but I see this.

Screen Shot 2021-06-25 at 5 37 28 PM
 Looks like there is an off by 1 bug somewhere.

### DIFF
--- a/parlai/tasks/dt_bug/agents.py
+++ b/parlai/tasks/dt_bug/agents.py
@@ -17,4 +17,10 @@ class MyTeacher(DialogTeacher):
         for episode in self.mock_data:
             for ex in episode:
                 done = ex == episode[-1]
-                yield Message({"text": f"text_{ex}", "labels": [f"label {ex}"]}), done
+                yield Message(
+                    {
+                        "text": f"text_{ex}",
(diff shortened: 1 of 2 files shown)


parlai/tasks/dt_bug/agents.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, opt, shared=None):
+        self.mock_data = [[1, 2, 3], [1, 2, 3, 4]]
+        opt['datafile'] = "df"
+        super().__init__(opt, shared=shared)
+
+    def setup_data(self, datafile):
+        for episode in self.mock_data:
+            for ex in episode:
+                done = ex == episode[-1]
+                yield Message({"text": f"text_{ex}", "labels": [f"label {ex}"]}), done
```
stephenroller: @spencerp is right. The signature of setup data is to yield `message, is_new_episode`

PR #3763, opened 2021-07-06
Update requirements.txt

tanujdhiman:
Error facing when installing all requirements !

parlai

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ iopath~=0.1.8
 gitdb2==2.0.5
 GitPython==3.0.3
 hydra-core~=1.0.6
-ipython==7.19.0
+ipython==7.16.1
 torch>=1.4.0
 joblib==0.14.1
 nltk==3.4.5


mojtaba-komeili:
I am not sure if this is needed. Are you sure your `pip` is configured properly? 7.19.0 is a valid `ipython` version (check [this](https://pypi.org/project/ipython/7.19.0/)). Could you try removing `ipython` and trying again?

tanujdhiman:
Yes pip is properly upgraded. Actually what happens is that version is not matching with `ipython` version. And also if there we remove `ipython` then it will automatically download the version `7.16.1` 

That's why I changed the version of ipython. 

Hope you understand!! 

Thanks

mojtaba-komeili:
There is no issue like that if you start with a clean environment, for example new conda env (I just tried that myself). The problem seems to be from some other packages that you have in your environment. Please check for other conflicting packages.

tanujdhiman:
Oh I see!

I was trying it in my existing env now it is working for me too.
Thanks

PR #3772, opened 2021-07-09
Stripping before processing Conversations to allow leniency 

liliarose:
**Patch description**
Stripping before processing Conversations to allow leniency; noticed that the JsonTeacher uses Conversations, and sometimes, there might be an extra newline at the end of the file, but when it's processed, it will generate an unnecessary error

**Testing steps**
- running CircleCi to make sure all the tests still pass

**Other information**
N/A

### DIFF
--- a/parlai/utils/conversations.py
+++ b/parlai/utils/conversations.py
@@ -296,7 +296,7 @@ class Conversations:
         """
         to_save = cls._get_path(datapath)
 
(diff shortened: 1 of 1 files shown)


stephenroller (approved):

jxmsML (approved):
LGTM!!! 🚀

Issue #3408, opened 2021-01-26
New agent convenience function


stephenroller:
Right now when we "deploy" models in different environments, we tend to manually implement the parley loops. This leads to a lot of code that looks roughly like this

```python
agent1 = agent.clone()
agent1.observe({'text': 'yada ydada', 'episode_done': True})
response = agent1.act()
return response['text']
```

I would like us to create a helper function in the very top level Agent class:

```python
def respond(self, text: str, **other_message_fields) -> str:
     pass
```

which should implement the above template, including a `self.clone()`.

Additionally, also implement a `batch_respond`:
```python
def batch_respond(self, texts: List[str], **other_message_fields) -> List[str]:
    pass
```
The primary difference is this one should leverage `batch_act` instead. See the [worlds tutorial](https://parl.ai/docs/tutorial_worlds.html) the implementation of BatchWorld for details.

The methods should be:
- documented with an excellent docstring
- unit tested directly with several agents, such as: repeat_query, one of the test torch agents (see test_transformer.py), and probably a mock agent
- documented in the worlds tutorial and the tips and tricks docs.

Afterwards, talk to Ankita about upgrading the implementation in the Wizard deployment.

klshuster:
should we close this issue given the PR landed?

PR #3775, opened 2021-07-11
Implementing respond() and batch_respond() -- agent convenience funct…

kauterry:
…ions.

**Patch description**

Addressing issue #3408

Implements agent convenience functions `respond()` and `batch_respond()` which provides agent's response to a message or a batch of messages respectively.

The function signatures are as follows:
```python
def respond(self, text_or_message: Union[str, Message], **other_message_fields) -> str:
    pass

def batch_respond(self, messages: List[Message]) -> List[str]:
    pass
```

**Testing steps**

Tests pass:
```pytest -v -k TestRepeatQueryAgent ```
```pytest -v -k TestTorchAgent```

**Other information**

### DIFF
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -156,15 +156,6 @@ class Agent(object):
         """
         pass
 
-    def self_observe(self, self_message: Message) -> None:
-        """
-        Observe one's own utterance.
-
-        :param self_message:
-            The message corresponding to the output from batch_act.
-        """
-        pass
-
     def respond(
         self, text_or_message: Union[str, Message], **other_message_fields
     ) -> str:
@@ -224,7 +215,8 @@ class Agent(object):
         agent_acts = self.batch_act(observations)
         response = []
         for agent, resp in zip(agents, agent_acts):
-            agent.self_observe(resp)
+            if hasattr(agent, "self_observe"):
+                agent.self_observe(resp)
             response.append(resp['text'])
         return response
 
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -156,6 +156,15 @@ class Agent(object):
         """
         pass
 
+    def self_observe(self, self_message: Message) -> None:
+        """
+        Observe one's own utterance.
+
+        :param self_message:
+            The message corresponding to the output from batch_act.
+        """
+        pass
+
     def respond(
         self, text_or_message: Union[str, Message], **other_message_fields
     ) -> str:
--- a/parlai/core/agents.py
+++ b/parlai/core/agents.py
@@ -41,10 +41,12 @@ This module also provides a utility method:
 """
 
 import copy
+from typing import List, Union
 
 from parlai.core.build_data import modelzoo_path
 from parlai.core.loader import load_agent_module
 from parlai.core.loader import register_agent  # noqa: F401
+from parlai.core.message import Message
 from parlai.core.opt import Opt
 from parlai.utils.misc import warn_once
 import parlai.utils.logging as logging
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: parlai/core/agents.py [hunk 1]
```diff
@@ -154,6 +156,78 @@ def shutdown(self):
...
+        Observe one's own utterance.
+
+        :param self_message:
+            The message corresponding to the output from batch_act.
+        """
+        pass
+
+    def respond(
+        self, text_or_message: Union[str, Message], **other_message_fields
+    ) -> str:
```
moyapchen, line 170: Should the return value here be `str` or `Message`? 

(I don't have full context as to what @stephenroller was thinking in terms of exactly how this would be used, but want to check to see if the `str` return in the original API suggestion is intentional or a fluke.)

parlai/core/agents.py
```diff
@@ -154,6 +156,78 @@ def shutdown(self):
...
+                raise RuntimeError('The agent needs a \'text\' field in the message.')
+
+        if 'episode_done' not in observation:
+            observation['episode_done'] = True
+        agent = self.clone()
+        agent.observe(observation)
+        response = agent.act()
+        return response['text']
+
+    def batch_respond(self, messages: List[Message]) -> List[str]:
```
moyapchen, line 199: Ditto here, should this be `List[Message]`?

parlai/core/agents.py
```diff
@@ -154,6 +156,78 @@ def shutdown(self):
...
+        """
+        if isinstance(text_or_message, str):
+            observation = Message(text=text_or_message, **other_message_fields)
+        else:
+            observation = Message(**text_or_message, **other_message_fields)
+            if 'text' not in observation:
+                raise RuntimeError('The agent needs a \'text\' field in the message.')
+
+        if 'episode_done' not in observation:
+            observation['episode_done'] = True
```
moyapchen, line 193: Do we want to set this to `True` by default? Not sure what other code does.

moyapchen (commented):
Don't forget to run lint. :)

parlai/core/agents.py
stephenroller: The goal of this is to build a utility function that goes str->str to make the simplest case boilerplate free -- many people just want to load BB and get the response. More complex cases can still be manually observed/act'd.

parlai/core/agents.py
stephenroller: Yeah I don't like this change this high up. Let's keep this localized to the utility method.

stephenroller: Yeah I think so

parlai/core/agents.py
```diff
@@ -154,6 +156,78 @@ def shutdown(self):
         """
         pass
 
+    def self_observe(self, self_message: Message) -> None:
```
stephenroller, line 159: Let's not add this for now and use a hasattr to check if self-observe should be called

parlai/core/agents.py
kauterry: @moyapchen Can you take a look at the issue? Stephen has fleshed it out there. https://github.com/facebookresearch/ParlAI/issues/3408

parlai/core/agents.py
kauterry: Again, could you take a look at the function signatures in https://github.com/facebookresearch/ParlAI/issues/3408?

parlai/core/agents.py
kauterry: Okay sure!

moyapchen (approved):
Looks fine

PR #3776, opened 2021-07-11
Implement batch_act() for Repeat Query agent.

kauterry:
**Patch description**

Implement batch_act() for Repeat Query agent. We add tests for batch_act() and batch_respond() of the RepeatQueryAgent.

**Testing steps**

Tests pass:
```
pytest -v -k TestRepeatQueryAgent 
```

**Other information**

### DIFF
--- a/parlai/agents/repeat_query/repeat_query.py
+++ b/parlai/agents/repeat_query/repeat_query.py
@@ -31,3 +31,12 @@ class RepeatQueryAgent(Agent):
             reply['text'] = 'Nothing to repeat yet.'
         reply['episode_done'] = False
         return Message(reply)
+
+    def batch_act(self, observations):
+        batch_reply = []
+        original_obs = self.observation
+        for obs in observations:
+            self.observation = obs
+            batch_reply.append(self.act())
+        self.observation = original_obs
+        return batch_reply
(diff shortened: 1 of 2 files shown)


parlai/agents/repeat_query/repeat_query.py
```diff
@@ -31,3 +31,12 @@ def act(self):
             reply['text'] = 'Nothing to repeat yet.'
         reply['episode_done'] = False
         return Message(reply)
+
+    def batch_act(self, observations):
+        batch_reply = []
+        original_obs = self.observation
```
moyapchen, line 37: Why should we save the `original_obs`? Not quite sure if I understand the logic here.

kauterry: I just don't want batch_act to change the `observation` attribute of the instance. So i just keep it as whatever it was before using a temporary variable `original_obs`. I didn't know what else to do.

moyapchen: If folks are using `batch_act`, I don't think they'll likely be calling `observe`/`act` on the agent via some other codepath (since I'd assume they'd just use one or the other in say, a batch world and not be mixing + matching the two).

Not a big deal, but as a comment.

moyapchen (approved):

PR #3780, opened 2021-07-12
Refactor analysis for per turn eval

mus1cholic:
**Patch description**
Refactors code for processing left_pane_text so code injections can be made on other crowdsourcing tasks

**Testing steps**
See CircleCI tests


parlai/crowdsourcing/tasks/acute_eval/webapp/package-lock.json
```diff
@@ -0,0 +1,5938 @@
+{
```
EricMichaelSmith: This is a generated file - we don't want to include it in the PR

parlai/crowdsourcing/tasks/model_chat/conf/example.yaml
```diff
@@ -12,7 +12,7 @@ mephisto:
     include_persona: false
     conversation_start_mode: 'hi'
     annotation_question: Does this comment from your partner have any of the following attributes? (Check all that apply)
-    conversations_needed_string: "blender_90M:10"
+    conversations_needed_string: "blender_3B:2"
```
EricMichaelSmith: let's keep the example the same - we want the casual user of model chat to not have to load the large Blender model

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
```diff
@@ -253,6 +251,11 @@ def __init__(
             }
         )
 
+    def process_left_pane_text(self, args: "DictConfig") -> str:
```
EricMichaelSmith, line 254: hmm, honestly, I think that these 3 lines of code will probably be the same for all subclasses of model_chat, meaning that the lines will need to get repeated in all subclasses. I think it'll lead to less duplicated code if, instead, there's a method called something like `format_left_pane_text()`, which you just define as `pass` for this superclass

parlai/crowdsourcing/tasks/model_chat/task_config/model_opts.yaml
```diff
@@ -1,2 +1,6 @@
-blender_90M: >
-    --model-file zoo:blender/blender_90M/model
+#blender_90M: >
```
EricMichaelSmith: Good to not add in these commented-out lines =P

EricMichaelSmith (commented):
Hi Orion! This is a great idea - but I think there's a slightly different setup that will lead to even less code-reuse :)

parlai/crowdsourcing/tasks/model_chat/task_config/model_opts.yaml
```diff
@@ -1,2 +1,3 @@
 blender_90M: >
     --model-file zoo:blender/blender_90M/model
+
```
EricMichaelSmith, line 3: Nit: extra space

stephenroller (commented):
I concur with Eric's comments

stephenroller (approved):

parlai/crowdsourcing/tasks/model_chat/task_config/model_opts.yaml
mus1cholic: Resolved

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
mus1cholic: Resolved

EricMichaelSmith (approved):
Looks good, thanks! But make sure to wait until all CI checks pass before merging in :)

mus1cholic:
CI tests passed, merging in now

PR #3782, opened 2021-07-13
Make paths attributes of AcuteEvalBuilder

EricMichaelSmith:
**Patch description**
Make the ACUTE-Eval task folder an attribute of `AcuteEvalBuilder` in order to allow subclasses to overwrite this

**Testing steps**
CI checks, test ACUTEs


JackUrb (approved):
Also makes sense to me - inheriting classes will likely want their task dir to come from somewhere near them, rather than here.

PR #3784, opened 2021-07-13
[Safety bench] better error messaging Perspective API

emilydinan:
**Patch description**
Better error messaging around the Perspective API client when the API exists but the client has not been installed.


klshuster (approved):
thanks for the quick fix!

PR #3785, opened 2021-07-13
Allow self-chats in nested task folders

EricMichaelSmith:
**Patch description**
Currently, it doesn't seem like we can run a self-chat world if the value of `--task` has a `'.'` in it, because the code to specify a self-chat world is parsed after we return from `load_world_module()` in that case. This PR shifts that logic around to not fall back on the default world before checking if we want to do self-chats.

**Testing steps**
CI checks, as well as checking that the following command (referring to an internal task) is now possible due to this change:
```
parlai self_chat \
--num-self-chats 1 \
--outfile ~/__test.jsonl \
--selfchat-max-turns 7 \
--model_file zoo:blender/blender_90M/model \
--beam-block-full-context True \
--task parlai_internal.tasks.bias_dialogue.cities.agents
```
Also, added testing for the self-chat and interactive mode for a long-form task string


parlai/core/loader.py
```diff
@@ -361,12 +361,7 @@ def load_world_module(
         repo = 'parlai_internal'
         task = task[9:]
     task_path_list = task.split(':')
-    if '.' in task_path_list[0]:
-        # The case of opt['task'] = 'parlai.tasks.squad.agents:DefaultTeacher'
-        # (i.e. specifying your own path directly, assumes DialogPartnerWorld)
-        return _get_default_world(default_world, num_agents)
 
-    task = task_path_list[0].lower()
```
stephenroller, line 369: Wtf why did we lower it

parlai/core/loader.py
```diff
@@ -381,7 +376,19 @@ def load_world_module(
...
+        # (i.e. specifying your own path directly)
+        module_name_parts = task_path_list[0].split('.')
+        if module_name_parts[-1] == 'agents':
+            # The world will be located in ".worlds", so correct the path
+            module_name = '.'.join(module_name_parts[:-1]) + '.worlds'
+        else:
+            module_name = task_path_list[0]
+    else:
+        task = task_path_list[0].lower()
+        module_name = "%s.tasks.%s.worlds" % (repo, task)
```
stephenroller, line 391: Can we upgrade to f-strings while we're at it

stephenroller (approved):
This seems reasonable? Any way we can add some tests? Maybe just refer to the public ones via absolute paths?

parlai/core/loader.py
EricMichaelSmith: Yes, changed

parlai/core/loader.py
EricMichaelSmith: Convention, probably? Namely, none of our task folder names (`convai2`, `blended_skill_talk`, etc.) are capitalized

Issue #3806, opened 2021-07-16
Unable to train on my own document corpus


lukesalamone:
**Bug description**
I am following the instructions for indexing my own documents with a FAISS index:
https://www.parl.ai/docs/agent_refs/rag.html#generating-your-own-faiss-index

**Reproduction steps**

I generated the embeddings with the following:
```
/usr/bin/python generate_dense_embeddings.py -mf zoo:hallucination/bart_rag_turn_dtt/model --dpr-model True --passages-file passages.tsv --outfile embeddings/out --num-shards 1 --shard-id 0 -bs 32
```

Then I indexed the embeddings
```
python index_dense_embeddings.py --retriever-embedding-size 768  --embeddings-dir embeddings --embeddings-name embeddings/out
```

Then I tried to to run the model in interactive mode:
```
parlai interactive -mf zoo:hallucination/bart_rag_turn_dtt/model --path-to-index  embeddings/IVF4096_HNSW128__PQ128.index --path-to-dense-embeddings embeddings/out --path-to-dpr-passages passages.tsv --retriever-embedding-size 768
```

**Expected behavior**
I expect to be able to run the model in interactive mode without errors.

**Logs**

```
/usr/lib/python3/dist-packages/requests/__init__.py:89: RequestsDependencyWarning: urllib3 (1.26.6) or chardet (3.0.4) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "
15:14:17 | Overriding opt["path_to_index"] to /root/ParlAI/parlai/agents/rag/scripts/embeddings/IVF4096_HNSW128__PQ128.index (previously: zoo:hallucination/wiki_index_compressed/compressed_pq)
15:14:17 | Overriding opt["path_to_dense_embeddings"] to /root/ParlAI/parlai/agents/rag/scripts/embeddings/out (previously: None)
15:14:17 | Overriding opt["path_to_dpr_passages"] to /root/ParlAI/parlai/agents/rag/scripts/passages.tsv (previously: zoo:hallucination/wiki_passages/psgs_w100.tsv)
15:14:17 | Using CUDA
15:14:17 | loading dictionary from /root/ParlAI/data/models/hallucination/bart_rag_turn_dtt/model.dict
15:14:17 | num words = 50264
15:14:17 | Rag: full interactive mode on.
[...]

PR #3818, opened 2021-07-19
[RAG] Handle `nan` scores.

klshuster:
**Patch description**
Handles a bug caught in #3806; if the index building goes wrong, there's a possibility that the context/document scores are nan's; this seems to occur if there's an issue with the clustering of the vectors (see linked FAISS issue in #3806)

**Testing steps**
Reproduced the error with steps in #3806; verified that, with this fix, the error is no longer an issue.

### DIFF
--- a/parlai/agents/rag/retrievers.py
+++ b/parlai/agents/rag/retrievers.py
@@ -640,6 +640,13 @@ class DPRRetriever(RagRetriever):
             vectors = vectors[:, :, :-1]
         # recompute exact FAISS scores
         scores = torch.bmm(query.unsqueeze(1), vectors.transpose(1, 2)).squeeze(1)
+        if torch.isnan(scores).sum().item():
+            logging.error(
+                '\n[ Document scores are NaN; please look into the built index. ]\n'
+                '[ If using a compressed index, try building an exact index: ]\n'
+                '[ $ python index_dense_embeddings --indexer-type exact... ]'
+            )
+            scores.fill_(1)
         ids = torch.tensor([[int(s) for s in ss] for ss in ids])
 
         return ids, scores


parlai/agents/rag/retrievers.py
```diff
@@ -640,6 +640,13 @@ def index_retrieve(
             vectors = vectors[:, :, :-1]
         # recompute exact FAISS scores
         scores = torch.bmm(query.unsqueeze(1), vectors.transpose(1, 2)).squeeze(1)
+        if torch.isnan(scores).sum().item():
+            logging.error(
+                '\n[ Document scores are NaN; please look into the built index. ]\n'
+                '[ If using a compressed index, try building an exact index: ]\n'
+                '[ $ python index_dense_embeddings --indexer-type exact... ]'
+            )
```
mojtaba-komeili, line 648: Should the error message also mention something about replacing some dummy numbers for score?

mojtaba-komeili (approved):

parlai/agents/rag/retrievers.py
klshuster: don't think it's necessary

parlai/agents/rag/retrievers.py
```diff
@@ -640,6 +640,13 @@ def index_retrieve(
...
+            scores.fill_(1)
```
spencerp, line 649: I'm curious why `1` is the error value?

klshuster: i picked an arbitrary number so that the scores play nicely with the rest of the system

the scores are literally a tensor of `nan`s:

```
ipdb> scores
tensor([[nan, nan, nan, nan, nan]], device='cuda:0', dtype=torch.float16)
```

so needed to pick something that worked

klshuster:
I'm pretty sure the failing tests are unrelated to this PR; they are currently being hammered out elsewhere, so will merge as is

PR #3819, opened 2021-07-19
Allow opt presets from internal libs.

stephenroller:
**Patch description**
Also allow loading opt_presets from `parlai_internal/opt_presets` and `parlai_fb/opt_presets`.

**Testing steps**
Manual testing.


klshuster (approved):
🥳

PR #3823, opened 2021-07-20
CI test failures after BB2 merge

mojtaba-komeili:
**Patch description**
Resolving the issues with the failing unit tests on CircleCI that were introduced after merging BB2 and its dependent projects: Personal Knowledge, and Wizard of Internet.


klshuster:
are we prepared to require torch 1.8 or higher?

mojtaba-komeili:
> are we prepared to require torch 1.8 or higher?

Probably not, I was just checking with another version because of Roller's note about the version 1.7 version.

requirements.txt
```diff
@@ -50,4 +49,4 @@ Unidecode==1.1.1
 urllib3>=1.26.5
 websocket-client==0.56.0
 websocket-server==0.4
-jsonlines==1.2.0
+jsonlines==1.2.0
```
klshuster, line 52: can we put this back in alphabetical order?

stephenroller: Pseudo-alphabetical order lol. there are some intentional non-monotonic places

tests/nightly/gpu/test_bb2.py
```diff
@@ -8,56 +8,64 @@
...
+        opt['model'] = 'projects.blenderbot2.agents.blenderbot2:BlenderBot2FidAgent'
+        opt['knowledge_access_method'] = retrieval_method.value
+        opt.update(dict(kwargs))
+        testing_utils.eval_model(opt, skip_test=True)
+        torch.cuda.empty_cache()
 
 
 @testing_utils.skipUnlessGPU
 @unittest.skipIf(LOCAL, "Skipping Test because its slow and mem intensive")
+@unittest.skipUnless(TRANSFORMER_INSTALLED, "Needs transformer, not installed.")
```
stephenroller, line 68: this is a bit funny because our GPU tests should always have transformers?

stephenroller:
The policy is usually past two versions of pytorch officially... So yeah requiring 1.8+ is viable now, but afaik, there's nothing we have that's 1.7 incompatible... If we're going to do that, I'd suggest a different PR to carve out 1.8/1.9 tests, and then rebase this

tests/nightly/gpu/test_bb2.py
```diff
@@ -8,56 +8,64 @@
...
+        testing_utils.eval_model(opt, skip_test=True)
+        torch.cuda.empty_cache()
+
+    def _test_bb2_fid(retrieval_method: KnowledgeAccessMethod, **kwargs):
+        opt = copy.deepcopy(common_opt)
+        opt['model'] = 'projects.blenderbot2.agents.blenderbot2:BlenderBot2FidAgent'
+        opt['knowledge_access_method'] = retrieval_method.value
+        opt.update(dict(kwargs))
+        testing_utils.eval_model(opt, skip_test=True)
+        torch.cuda.empty_cache()
```
klshuster, line 63: why is this protected?

tests/nightly/gpu/test_bb2.py
klshuster: yeah im positive this should not be necessary

tests/nightly/gpu/test_bb2.py
mojtaba-komeili: Because if we can't import `projects.blenderbot2` we don't have constants such as `ZOO_MEMORY_DECODER` so we need to skip everything here.

mojtaba-komeili: Actually does CircleCI even run tests in `nightly/gpu` if it is not running the GPU tests (that have transformer)? I tried to debug running local `pytest` with transformer not installed, and they failed. But now thinking maybe it doesn't work like that.

klshuster: this marker runs the nightly gpu tests: https://github.com/facebookresearch/ParlAI/blob/master/.circleci/config.yml#L388

you can see that the deps under torchgpu1.7 are installed, which includes transformers: https://github.com/facebookresearch/ParlAI/blob/master/.circleci/config.yml#L95

requirements.txt
klshuster: ok yeah i think we might've broken some stuff with the order wrong... sorry @mojtaba-komeili could you just revert to what it was before?

klshuster (approved):

PR #3825, opened 2021-07-20
Retry flaky crowdsourcing tests

EricMichaelSmith:
**Patch description**
Retry crowdsourcing tests that are flaky due to sockets closing without warning, occasional missing messages, etc.

**Testing steps**
CI checks


mojtaba-komeili (approved):
Thanks Eric. This had caused some pain for me before. Glad to see this.

JackUrb (approved):
These look good to me - The underlying issue I think is in the `MockArchitect`'s `MockServer` websockets closing unexpectedly without a retried connection, but retries are a good solution until I can dig into that on the Mephisto side.

PR #3845, opened 2021-07-22
Wizard of Wikipedia, knowledge source page title.

mojtaba-komeili:
**Patch description**
Added `WikiPageTitleTeacher` to the Wizard of Internet task. It generates the title of the Wikipedia page of the passage that Wizard selected as their source of knowledge, or `no_passages_used` if there was no selection.

**Testing steps**
* Added `TestWikiPageTitleTeacher` to teacher tests.
* Running ``parlai dd -t wizard_of_wikipedia:wiki_page_title`

Screen Shot 2021-07-22 at 3 25 41 PM

### DIFF
--- a/parlai/tasks/wizard_of_wikipedia/agents.py
+++ b/parlai/tasks/wizard_of_wikipedia/agents.py
@@ -773,6 +773,7 @@ class WikiPageTitleTeacher(WizardDialogKnowledgeTeacher):
         super().__init__(self.opt, shared=shared)
         self.id = 'WikiTitleGenerationTeacher'
         self._conv_history_len = self.opt['conversation_history_length']
+        self._skip_no_title = self.opt['skip_no_title']
         if not shared:
             self._preprocess_data()
         else:
@@ -787,6 +788,15 @@ class WikiPageTitleTeacher(WizardDialogKnowledgeTeacher):
             type=int,
             default=0,
             help='Number of previous utterances to keep in context, 0 (default) includes all',
+        )
+        agent.add_argument(
+            '--skip-no-title',
+            type='bool',
+            default=False,
+            help=(
+                'Whether to skip the example if no passage was selected. If `false` '
+                f'uses `{TOKEN_NOCHOSEN}` instead of title if no knowledge source was selected.'
+            ),
         )
         return parser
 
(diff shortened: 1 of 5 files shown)


parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -757,6 +758,107 @@ def get(self, episode_idx, entry_idx=0):
...
+    The context provided by this teacher (`text`) is the conversation history, with chosen topic removed.
+    The label is the title of the Wikipedia page of the passage that wizard selected for crafting
+    the next utterance; in other words, the source of knowledge for this utterance.
+    """
+
+    def __init__(self, opt, shared=None):
+        self.opt = copy.deepcopy(opt)
+        self.opt['label_type'] = 'response'
+        super().__init__(self.opt, shared=shared)
+        self.id = 'WikiTitleGenerationTeacher'
```
klshuster, line 774: nit: can we standardize this with the class name (or vice-versa)?

parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -757,6 +758,107 @@ def get(self, episode_idx, entry_idx=0):
...
+
+    @classmethod
+    def add_cmdline_args(cls, parser, partial_opt=None):
+        super().add_cmdline_args(parser, partial_opt=partial_opt)
+        agent = parser.add_argument_group('Wikipedia Page Title Arguments')
+        agent.add_argument(
+            '--conversation-history-length',
+            type=int,
+            default=0,
+            help='Number of previous utterances to keep in context, 0 (default) includes all',
```
klshuster, line 790: can we make the default -1? and have that map to including all? making 0 mean includes all doesn't seem right (as one could argue this should include... none)

parlai/tasks/wizard_of_wikipedia/agents.py
```diff
@@ -757,6 +758,107 @@ def get(self, episode_idx, entry_idx=0):
...
+        )
+        agent.add_argument(
+            '--skip-no-title',
+            type='bool',
+            default=False,
+            help=(
+                'Whether to skip the example if no passage was selected. If `false` '
```
klshuster, line 797: i wonder if this teacher would be more useful with this flag default to `True`, since we really want to train a model that can generate titles (not just generate `TOKEN_NOCHOSEN`). Anyway that's up to you

klshuster (approved):
awesome, thanks for including this!

parlai/tasks/wizard_of_wikipedia/agents.py
mojtaba-komeili: I changed default to `-1`, and added something to change any non-positive value to that.

parlai/tasks/wizard_of_wikipedia/agents.py
mojtaba-komeili: I agree, that might make more sense.

PR #3846, opened 2021-07-22
AcceptabilityChecker min words adjustable

mojtaba-komeili:
**Patch description**
The minimum number of words for `AcceptabilityChecker` is a hard-coded value (`3`). Needless to say this value may not work in all scenarios. This is a quick patch allowing user to adjust the minimum acceptable number of words in the `AcceptabilityChecker`.

### DIFF
--- a/parlai/crowdsourcing/utils/acceptability.py
+++ b/parlai/crowdsourcing/utils/acceptability.py
@@ -10,6 +10,9 @@ from typing import Iterable, List
 from parlai.utils.safety import OffensiveStringMatcher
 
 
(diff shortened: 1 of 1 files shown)


EricMichaelSmith (commented):
Yeah, very useful to have! Small nit though: I think it'd be better as a class attribute (i.e. like `ALL_VIOLATION_TYPES`), because it's only used in the context of `AcceptabilityChecker`. You could do something like `self.min_words_violation_threshold = min_words or self.DEFAULT_MIN_WORDS_THRESHOLD`

EricMichaelSmith (approved):
Looks great, thanks!

PR #3852, opened 2021-07-23
Enable self_chat to seed messages from tasks in parlai_internal

skywalker023:
**Patch description**

A fix for enabling self_chat to seed messages from tasks in parlai_internal.

Currently, when we give `--task internal:xxx --seed-messages-from-task true` for self_chat, 
the `load_openers()` function will try to create a dummy task from `parlai.tasks.internal.agents` 
instead of `parlai_internal.tasks.xxx.agents` and fails.

This is because the first line of `load_openers()` parses the `basetask` according to `:`.
As a result, when we try to seed from tasks in parlai_internal by prepending `internal:`, it will parse the "internal" as `basetask`.
So, I fixed `load_openers()` to pass the task as is when it starts with "internal:".
Then the `create_task()` will handle it seamlessly as in other circumstances.

**Testing steps**

**Other information**

### DIFF
--- a/parlai/tasks/self_chat/worlds.py
+++ b/parlai/tasks/self_chat/worlds.py
@@ -15,7 +15,7 @@ from parlai.core.message import Message
 
 
 def load_openers(opt) -> Optional[List[str]]:
-    if opt['task'].startswith('internal:'):
+    if opt['task'].startswith('internal:') or opt['task'].startswith('fb:'):
         base_task = opt['task']
     else:
         base_task = opt['task'].split(':')[0]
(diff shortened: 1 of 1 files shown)


parlai/tasks/self_chat/worlds.py
```diff
@@ -15,7 +15,11 @@
 
 
 def load_openers(opt) -> Optional[List[str]]:
-    base_task = opt['task'].split(':')[0]
+    if opt['task'].startswith('internal:'):
```
stephenroller, line 18:
```suggestion
    if opt['task'].startswith('internal:') or opt['task'].startswith('fb:'):
```

stephenroller: Just to add compatibility with some FB customers, thanks.

stephenroller (approved):
Thanks for the contribution!

skywalker023:
Anytime! 😆

PR #3863, opened 2021-07-26
replacing old heatmap.png

liliarose:
Replacing so that it looks better in dark mode :)

both_heatmaps

For ppl who don't have dark mode
Screen Shot 2021-07-26 at 10 00 37 AM


JackUrb (approved):
😎

Issue #3868, opened 2021-07-27
generate_dense_embeddings.py doesn't load custom TSV?


lukesalamone:
**Bug description**
I am trying to generate embeddings for `bart_fid_rag` with a custom TSV but the `generate_dense_embeddings.py` script is loading wiki data instead.

**Reproduction steps**
1. Copy `ParlAI/data/models/hallucination/wiki_passages/psgs_w100.tsv` into `ParlAI/parlai/agents/rag/scripts/`
```
cd ParlAI/parlai/agents/rag/scripts/
cp ParlAI/data/models/hallucination/wiki_passages/psgs_w100.tsv passages.tsv
```

This simulates using a custom TSV file.

2. Run the following from `ParlAI/parlai/agents/rag/scripts/`

```
/usr/bin/python generate_dense_embeddings.py \
-mf zoo:hallucination/bart_fid_rag/model \
--dpr-model False \
--passages-file passages.tsv \
--outfile emb_bart_fid_rag/out \
--num-shards 1 \
--shard-id 0 \
-bs 32
```

**Expected behavior**
I don't expect the script to load `wiki_passages/psgs_w100.tsv` at all. I also don't expect it to load the index from `wiki_index_compressed/compressed_pq`. I'm not sure if my data is being read at all...

**Logs**
Please paste the command line output:

```
13:30:17 | Overriding opt["interactive_candidates"] to inline (previously: fixed)
13:30:17 | Using CUDA
13:30:17 | loading dictionary from /root/ParlAI/data/models/hallucination/bart_fid_rag/model.dict
13:30:17 | num words = 50264
13:30:17 | Fid: full interactive mode on.
13:30:19 | Loading index from /root/ParlAI/data/models/hallucination/wiki_index_compressed/compressed_pq
13:30:20 | Loaded index of type <faiss.swigfaiss.IndexIVFPQ; proxy of <Swig Object of type 'faiss::IndexIVFPQ *' at 0x7fb9abe86b70> > and size 21015300
13:30:23 | Reading data from: /root/ParlAI/data/models/hallucination/wiki_passages/psgs_w100.tsv
```

**Additional context**
Python 3.8.10

The script eventually fails with error

```
Traceback (most recent call last):
  File "generate_dense_embeddings.py", line 213, in <module>
    Generator.main()
  File "/root/ParlAI/parlai/core/script.py", line 129, in main
    return cls._run_args(None)
[...]

PR #3869, opened 2021-07-27
[RAG] Generate Dense Embeddings Fix

klshuster:
**Patch description**
#3868 pointed out a couple of current shortcomings in the custom embeddings generation process for RAG/FiD DPR Models.

1. If specifying a ParlAI RAG or FiD model, one needs to set `--dpr-model True`; this has been clarified in the README.
2. If specifying specifically a FiD-RAG model, we need to make sure that the document encoder is initialized appropriately (this was not correctly handled prior)

**Testing steps**
I tried running the provided command in #3868 with `--dpr-model True`, after the change, and it appears the script works correctly:

```
$ python parlai/agents/rag/scripts/generate_dense_embeddings.py -mf zoo:hallucination/bart_fid_rag/model --dpr-model true --passages-file data/models/hallucination/wow_passages/wow_articles.paragraphs.tsv --outfile /tmp/out --num-shards 1 --shard-id 0 -bs 32
17:12:05 | Overriding opt["model"] to dpr_agent (previously: fid)
17:12:05 | Overriding opt["interactive_candidates"] to inline (previously: fixed)
17:12:05 | Overriding opt["share_encoders"] to False (previously: True)
17:12:06 | Using CUDA
17:12:06 | DPR: full interactive mode on.
17:12:42 | Total parameters: 241,228,800 (240,442,368 trainable)
17:12:42 | Loading existing model parameters from /private/home/kshuster/ParlAI/data/models/hallucination/bart_fid_rag/model
17:12:49 | Loading data/models/hallucination/wow_passages/wow_articles.paragraphs.tsv
17:12:49 | Reading data from: data/models/hallucination/wow_passages/wow_articles.paragraphs.tsv
2863it [00:00, 105418.20it/s]
17:12:49 | Shard 0 of 1 encoding psg index 0 to 2862, out of 2862
17:12:51 | Encoded 320 out of 2862 passages
17:12:51 | Encoded 640 out of 2862 passages
17:12:52 | Encoded 960 out of 2862 passages
17:12:52 | Encoded 1280 out of 2862 passages
17:12:53 | Encoded 1600 out of 2862 passages
17:12:53 | Encoded 1920 out of 2862 passages
17:12:54 | Encoded 2240 out of 2862 passages
17:12:54 | Encoded 2560 out of 2862 passages
17:12:55 | Writing results to /tmp/out_0.pt
17:12:55 | Writing ids to /tmp/ids_0
```

### DIFF
--- a/parlai/agents/rag/dpr.py
+++ b/parlai/agents/rag/dpr.py
@@ -203,13 +203,19 @@ class DPRModel(TransformerMemNetModel):
         try:
             # determine if loading a RAG model
             loaded_opt = Opt.load(f"{query_path}.opt")
-            if loaded_opt['model'] == 'rag' and loaded_opt['query_model'] in [
+            document_path = loaded_opt.get('dpr_model_file', document_path)
+            if loaded_opt['model'] in ['rag', 'fid'] and loaded_opt['query_model'] in [
                 'bert',
                 'bert_from_parlai_rag',
             ]:
                 query_model = 'bert_from_parlai_rag'
-            # document model is always frozen
-            document_path = loaded_opt.get('dpr_model_file', document_path)
+                if loaded_opt['model'] == 'fid':
+                    # document model is always frozen
+                    # but may be loading a FiD-RAG Model
+                    doc_loaded_opt = Opt.load(
+                        f"{modelzoo_path(opt['datapath'], document_path)}.opt"
+                    )
+                    document_path = doc_loaded_opt.get('dpr_model_file', document_path)
 
         except FileNotFoundError:
             pass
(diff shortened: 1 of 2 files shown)


parlai/agents/rag/dpr.py
```diff
@@ -203,13 +203,19 @@ def _get_build_options(cls, opt: Opt):
         try:
             # determine if loading a RAG model
             loaded_opt = Opt.load(f"{query_path}.opt")
-            if loaded_opt['model'] == 'rag' and loaded_opt['query_model'] in [
+            document_path = loaded_opt.get('dpr_model_file', document_path)
```
mojtaba-komeili, line 206: Does this slow things down if you have a `fid` model and you load `document_path` later again? In that case does it make sense to move this to the else condition of the following if?

mojtaba-komeili (approved):
LGTM.

parlai/agents/rag/dpr.py
klshuster: im not entirely sure what you're asking here --> we're only getting the document path from the `.opt` file. We should *never* actually build a FiD model in this scenario

PR #3881, opened 2021-07-30
Update retrievers.py in RAG

jianguoz:
Fixed a small error for handling `NaN` value in scores. 

**Patch description**

`scores.fill_(1)` will change all values in scores to 1, while there are values that are not `NaN`.

**Testing steps**

**Other information**


klshuster:
looks good, can you please run `bash autoformat.sh` from the top level ParlAI directory and resubmit?

cc @stephenroller we need to get precommit setup i believe

jianguoz:
> looks good, can you please run `bash autoformat.sh` from the top level ParlAI directory and resubmit?
> 
> cc @stephenroller we need to get precommit setup i believe

Thanks very much for the instructions! I have reformatted the code using pre-commit

klshuster (approved):
thank you!

PR #3883, opened 2021-07-30
Add "final_extra_valid_opt_filepath" to train_model.py

moyapchen:
I've been annoyed at having to kick off separate eval loops at the ends of trains, so here we are. 

Also added a way to configure saving outputs to json (cause it's something that I've wanted for a while and I didn't want to add another return value in case scripts elsewhere were expecting only v_report and t_report). 

Not super attached to names or to the default values I've got set. I tried using 'Optional[str]' for the cmdline arg, but the test was not happy with that...

**Testing steps**
Wrote a new test that exercises all of the new functionality.


parlai/scripts/train_model.py
```diff
@@ -87,6 +87,18 @@ def setup_args(parser=None) -> ParlaiParser:
         '--evaltask',
         help='task to use for valid/test (defaults to the one used for training)',
     )
+    train.add_argument(
+        '--final-extra-valid-opt-filepath',
```
stephenroller, line 91: Bike shedding but `--final-eval-opt`?

parlai/scripts/train_model.py
```diff
@@ -87,6 +87,18 @@ def setup_args(parser=None) -> ParlaiParser:
...
+        type=str,
+        default='',
+        help="A '.opt' file that is used for final eval. Useful for setting skip-generation to false. 'datatype' must be included as part of the opt.",
+    )
+    train.add_argument(
+        '--write-log-as-json',
```
stephenroller, line 97: Aren't the final evals stored in the trainstats file already? If not, my suggestion would be we include them in the trainstats.

parlai/scripts/train_model.py
```diff
@@ -919,6 +941,34 @@ def train(self):
 
         print_announcements(opt)
 
+        if opt['final_extra_valid_opt_filepath'] is not '':
```
stephenroller, line 944: Let's refactor this out plz

tests/test_train_model.py
```diff
@@ -13,12 +13,69 @@
...
 from parlai.core.metrics import AverageMetric
 from parlai.core.worlds import create_task
+from parlai.core.opt import Opt
 from parlai.core.params import ParlaiParser
 from parlai.core.agents import register_agent, Agent
 from parlai.utils.data import DatatypeHelper
 
 
 class TestTrainModel(unittest.TestCase):
+    def test_final_extra_eval_and_save_json(self):
```
stephenroller, line 23: Ty

parlai/scripts/train_model.py
stephenroller: And not even make it an option. Just do it.

stephenroller (commented):
Just want to finish discussion

parlai/scripts/train_model.py
klshuster: +1

klshuster (commented):
agree with all of @stephenroller's suggestions

parlai/scripts/train_model.py
```diff
@@ -656,19 +650,43 @@ def _run_eval(
...
-                    json.dump(dict_report(report), f)
-            else:
-                with PathManager.open(
-                    opt['model_file'] + log_suffix + '.' + datatype, 'a'
-                ) as f:
-                    f.write(f'{metrics}\n')
+            with PathManager.open(
+                opt['model_file'] + log_suffix + '.' + datatype + ".json", 'a'
+            ) as f:
+                json.dump(dict_report(report), f)
```
stephenroller, line 656: I think we'd prefer keeping it with the rest of the `.trainstats`?

moyapchen: The other valid/reports aren't saved to .trainstats right now.

stephenroller: Every validation run, except the final one, is stored in trainstats.

https://github.com/facebookresearch/ParlAI/blob/9b5f8c670d21964ed140683a4e3fd1ae7e1f5db3/parlai/scripts/train_model.py#L474-L481

I'm humbly requesting you expand it to also include the final valid/test reports. (They can just be None during the intermediate stages)

moyapchen: Alright, wording wasn't quite clear there. Anyway, done.

stephenroller (approved):
Seems like you have a bug for when model file isn't set (and we should be throwing trainstats into the void). Otherwise lgtm, thanks for making the changes.

PR #3888, opened 2021-08-02
Update Hallucination README.md

klshuster:
**Patch description**
Per #3872, we should include the `--model-file` parameter in the example training commands for the Hallucination project

### DIFF
--- a/parlai/agents/rag/README.md
+++ b/parlai/agents/rag/README.md
@@ -133,8 +133,6 @@ python generate_dense_embeddings.py -mf zoo:hallucination/multiset_dpr/hf_bert_b
 --outfile /tmp/wiki_passage_embeddings/wiki_passages --num-shards 50 --shard-id 0 -bs 32
 ```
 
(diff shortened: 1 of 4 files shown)


jianguoz:
> **Patch description**
> Per #3872, we should include the `--model-file` parameter in the example training commands for the Hallucination project

Per #3881 and #3873. Maybe we can merge them together later. In #3881, the model should immediately raise an `AssertionError` when there are `NaN` instead of logging errors and replacing `NaN` with other values. Besides, after careful investigation, I found the `NaN` error is not caused by either `--indexer-type exact` or `--indexer-type compressed.` The error comes from [num_shards](https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/rag/scripts/generate_dense_embeddings.py#L190). Where if end_idx<n_docs, it will raise `NaN` error. So it is better to set `--num-shards` to small values. For example, if `--shard-id` is 0, then `--num-shards is a value < int(len(rows of the passages file)/(n_maximum_passages_you_want_to_retrieve-1))`. In #3873, I will slightly update the readme soon to give more instructions.

klshuster:
Hi @jianguoz, though I appreciate the changes you have proposed, these seem more suitable for a separate PR (as now I cannot actually review the changes you've made)

jianguoz:
> Hi @jianguoz, though I appreciate the changes you have proposed, these seem more suitable for a separate PR (as now I cannot actually review the changes you've made)

@klshuster  Sure! I will do it:).

spencerp (approved):
lgtm!

PR #3889, opened 2021-08-02
[crowdsourcing] Forcing mock to shutdown in test cleanup

JackUrb:
**Patch description**
Adding a blocking forced shutdown for the Mephisto mock server to the cleanup of Mephisto-based testing. Attempts to ensure that the system doesn't hang indefinitely if a test fails and shutdown is unsuccessful.

**Testing steps**
Ran crowdsourcing tests locally 50 times. Ended up with a few failures, but none left pytest hanging.


EricMichaelSmith (approved):
Sounds good, great improvement!

PR #3892, opened 2021-08-04
[minor] Add DatatypeHelper func to split data by fold

moyapchen:
Use `tests/test_utils.py::TestDatatypeHelper::test_split_domains_by_fold` to check which tests a superset of this.

Doing so that I can make code for an internal class (one that uses data from a jsonl file) a little easier.


parlai/utils/data.py
```diff
@@ -140,9 +135,30 @@ def split_domains_by_fold(
...
+        cls,
+        fold: str,
+        domains: List[List],
+        train_frac: float,
+        valid_frac: float,
+        test_frac: float,
+        seed: int = 42,
+    ):
+        """
+        Need to be careful about how we setup random to not leak examples between trains
```
stephenroller, line 152: Please make a higher quality docstring

parlai/utils/data.py
```diff
@@ -115,20 +115,15 @@ def is_streaming(cls, datatype: str) -> bool:
...
         fold: str,
-        domains: List[List],
+        data: List,
         train_frac: float,
         valid_frac: float,
         test_frac: float,
         seed: int = 42,
     ):
-        """
-        Need to be careful about how we setup random to not leak examples between trains
```
stephenroller, line 128: Please make a  docstring for this method.

parlai/utils/data.py
```diff
@@ -140,9 +135,30 @@ def split_domains_by_fold(
             start = train_frac + valid_frac
             end = 1.0
 
+        random.Random(seed).shuffle(data)
+        return data[int(start * len(data)) : int(end * len(data))]
+
+    @classmethod
+    def split_domains_by_fold(
```
stephenroller, line 142: maybe shards? Or subsets? This is fine but `domains` is somewhat a specific name relative to how this function could be used?

stephenroller (approved):

parlai/utils/data.py
moyapchen: Changed to "subsets"

parlai/utils/data.py
moyapchen: Done.

parlai/utils/data.py
moyapchen: Done.

PR #3900, opened 2021-08-04
[params] Support options depending on options in parse_kwargs.

stephenroller:
**Patch description**
@ankitade reported an issue where systems using `parse_kwargs` (such as internal customers and the Colab notebooks) could not correctly use Mutator options. Debugging, it appears this derives from not having enough repeated iterations of `add_cmdline_args` being called with options that depend on other options.

Previously: Parse 'task=' and call the Task.add_cmdline_args. However, since `--mutators` hasn't been added yet, `partial_opt` doesn't contain any mutators and we don't load these in.

After: We repeatedly parse kwargs until there are remaining unparsed arguments, or we do not see any improvements (fixed point)

**Testing steps**
New CI examples that previously would not pass.

### DIFF
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -1246,9 +1246,6 @@ class ParlaiParser(argparse.ArgumentParser):
             new_unparsed_args = set()
             string_args = []
             for kwname, value in kwargs.items():
-                # note we don't have the if kwname not in kwname_to_action here.
-                # it MUST appear, or else we legitimately should be throwing a KeyError
-                # because user has provided an unspecified option
                 if kwname not in kwname_to_action:
                     new_unparsed_args.add(kwname)
                     continue
--- a/parlai/core/params.py
+++ b/parlai/core/params.py
@@ -1204,6 +1204,7 @@ class ParlaiParser(argparse.ArgumentParser):
                 kwname_to_action[kwname] = action
 
         # since we can have options that depend on options, repeat until convergence
+        string_args = []
         unparsed_args = set(kwargs.keys())
         while unparsed_args:
             string_args = []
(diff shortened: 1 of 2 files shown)


parlai/core/params.py
```diff
@@ -1203,62 +1203,81 @@ def _kwargs_to_str_args(self, **kwargs):
...
-                # during add_kw_Args
-                continue
-            action = kwname_to_action[kwname]
-            last_option_string = action.option_strings[-1]
-            if isinstance(action, argparse._StoreTrueAction):
-                if bool(value):
+        # since we can have options that depend on options, repeat until convergence
+        unparsed_args = set(kwargs.keys())
+        while unparsed_args:
+            string_args = []
```
stephenroller, line 1209: These 3 lines are the real change (the rest is just indentation)

parlai/core/params.py
```diff
@@ -1203,62 +1203,81 @@ def _kwargs_to_str_args(self, **kwargs):
...
+
+            new_unparsed_args = set()
+            string_args = []
+            for kwname, value in kwargs.items():
+                # note we don't have the if kwname not in kwname_to_action here.
+                # it MUST appear, or else we legitimately should be throwing a KeyError
+                # because user has provided an unspecified option
+                if kwname not in kwname_to_action:
+                    new_unparsed_args.add(kwname)
+                    continue
```
stephenroller, line 1251: These lines are new

parlai/core/params.py
```diff
@@ -1203,62 +1203,81 @@ def _kwargs_to_str_args(self, **kwargs):
...
+                # if we have converged to a fixed point with no improvements, we
+                # truly found some unreachable args
+                raise KeyError(
+                    f'Failed to parse one or more kwargs: {", ".join(new_unparsed_args)}'
+                )
             else:
-                raise TypeError(f"Don't know what to do with {action}")
+                # We've seen some improvements on the number of unparsed args,
+                # iterate again
+                unparsed_args = new_unparsed_args
```
stephenroller, line 1277: These lines are new

ankitade (approved):

PR #3901, opened 2021-08-04
Update retrievers and readme in RAG

jianguoz:
Update retriever to handle the NaN error and slightly add instructions to the readme 

**Patch description**

Per #3881 and #3873. In #3881, the model should immediately raise an AssertionError when there are NaN instead of logging errors and/or replacing NaN with other values. Besides, after careful investigation, I found the NaN error is not caused by either --indexer-type exact or --indexer-type compressed. The error comes from num_shards if the passages file format meets requirements and there are no repeated entries. Where if end_idx<n_docs, it will raise NaN error. So it is better to set --num-shards to small values. For example, if --shard-id is 0, then --num-shards is a value < int(len(rows of the passages file)/(n_maximum_passages_you_want_to_retrieve-1)). I had also slightly updated the readme to give more instructions.

**Testing steps**

**Other information**

### DIFF
--- a/parlai/agents/rag/README.md
+++ b/parlai/agents/rag/README.md
@@ -133,6 +133,8 @@ python generate_dense_embeddings.py -mf zoo:hallucination/multiset_dpr/hf_bert_b
 --outfile /tmp/wiki_passage_embeddings/wiki_passages --num-shards 50 --shard-id 0 -bs 32
 ```
 
+Note `--num-shards` should be a [small value](https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/rag/retrievers.py#L644) if the passages file only has a few entries; otherwise it may raise the `NaN` error when the number of retrieved passages, i.e., `--n-docs,` is a large value.
+
 ### 3. Index the Dense Embeddings
 
 The final step is to build the full FAISS index from these dense embeddings. You can use the [`index_dense_embeddings.py`](https://github.com/facebookresearch/ParlAI/blob/master/parlai/agents/rag/scripts/index_dense_embeddings.py) script to achieve this. You can choose one of the following options when indexing your embeddings for varying results, depending on the size of your dataset:
--- a/parlai/agents/rag/retrievers.py
+++ b/parlai/agents/rag/retrievers.py
@@ -1135,7 +1135,7 @@ class SearchQuerySearchEngineRetriever(SearchQueryRetriever):
 
     def _empty_docs(self, num: int):
         """
-        Generates the requsted number of empty documents.
+        Generates the requested number of empty documents.
         """
         return [BLANK_SEARCH_DOC for _ in range(num)]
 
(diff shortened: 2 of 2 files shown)


klshuster:
Could you provide some evidence that `--num-shards` has anything to do with this? I think the issue that's going on here is the **index construction**, not the dense embedding generation; num_shards only shards the passages such that you can parallelize the computation of dense embeddings

in reality, we should specify `--indexer-type exact` if the number of passages is small, as otherwise the compressed index default `IVF4096_HNSW128,PQ128` is clustering into 4096 clusters, which is not necessary for small passage files. We should probably warn users to instead decrease the value of `IVF<N>`. 

I had some internal notes/instructions that I found via trial and error that may help with this:

```
The "CLUSTERING" string will be of the form `IVFx_HNSW32`. We specify an index factory using the arg `--compressed-indexer-factory factory_string`

We'll be using the following terminology as we proceed:
- `T` -> as above, the total number of embeddings
- `d` -> the dimension of each embedding
- `M` -> `d`/4, or `d`/8, usually.
- `D` -> 2M, or 4M, usually.
- `K` -> 2^N, such that `2^(N-1) >= sqrt(T) * 16`. For example, suppose `T` = 1 billion. `sqrt(T) * 16 = 505964; you would set `K = 2^20 > 2^19 = 524288 > 505964`.

1. If T < 1m, you can simply set `--indexer-type exact`
2. Otherwise, use the string "OPQ`M`_`D`,IVF`K`_HNSW_32,PQ`D`x4fsr"

For example, if we had 1B vectors, we'd use the string, "OPQ128_256,IVF1048576_HNSW32,PQ128x4fsr"
```

We can ignore the `OPQ` and `x4fsr` for this case, but something to keep in mind

jianguoz:
> Could you provide some evidence that `--num-shards` has anything to do with this? I think the issue that's going on here is the **index construction**, not the dense embedding generation; num_shards only shards the passages such that you can parallelize the computation of dense embeddings
> 
> in reality, we should specify `--indexer-type exact` if the number of passages is small, as otherwise the compressed index default `IVF4096_HNSW128,PQ128` is clustering into 4096 clusters, which is not necessary for small passage files. We should probably warn users to instead decrease the value of `IVF<N>`.
> 
> I had some internal notes/instructions that I found via trial and error that may help with this:
> 
> ```
> We're going to be building a string of the form "COMPRESSION,CLUSTERING,QUANTIZATION". The "CLUSTERING" string will be of the form `IVFx_HNSW32`. We specify an index factory using the arg `--compressed-indexer-factory factory_string`
> 
> We'll be using the following terminology as we proceed:
> - `T` -> as above, the total number of embeddings
> - `d` -> the dimension of each embedding
> - `M` -> `d`/4, or `d`/8, usually.
> - `D` -> 2M, or 4M, usually.
> - `K` -> 2^N, such that `2^(N-1) >= sqrt(T) * 16`. For example, suppose `T` = 1 billion. `sqrt(T) * 16 = 505964; you would set `K = 2^20 > 2^19 = 524288 > 505964`.
> 
> 1. If T < 1m, you can simply set `--indexer-type exact`
> 2. Otherwise, use the string "OPQ`M`_`D`,IVF`K`_HNSW_32,PQ`D`x4fsr"
> 
> For example, if we had 1B vectors, we'd use the string, "OPQ128_256,IVF1048576_HNSW32,PQ128x4fsr"
> ```
> 
> We can ignore the `OPQ` and `x4fsr` for this case, but something to keep in mind

I have tried to use both --indexer-type exact/ compressed and --compressed-indexer-factory HNSW32. Both will still raise *NaN* errors if I set a large num_shards when my document is very small, e.g., num_shards=50 and the document has 400 entries, then you will have NaN errors when setting n-docs>10.

PR #3912, opened 2021-08-05
[TGA] Implement stochastic beam search (gumble sampling)

stephenroller:
**Patch description**
Just implement it and try it. Anecdotally, it's way too spicy.

**Testing steps**
Manual testing with BB2.7B


stephenroller:
Pushed the PR for posterity, but this doesn't seem worth merging.

PR #3917, opened 2021-08-06
[RAG] Update RAG readme

klshuster:
**Patch description**
Include more instructions for building new indices; mainly, that small datasets should use `--indexer-type exact`


jianguoz (approved):
Thanks very much for the added detailed instructions!

PR #3923, opened 2021-08-09
Preserve the order of loaded openers in selfchat

skywalker023:
Hello! 👋🏼
I found that we get non-deterministic order of openers even though we are loading them from fixed files or tasks.
So I fixed them to be deterministic.

**Patch description**

`set()` results in non-deterministic order of openers for every run.
Use `dict.fromkeys()` instead to deduplicate openers and preserve order.

**Testing steps**

**Other information**


stephenroller (approved):
Seems reasonable to me.

stephenroller:
Since CI isn't running in your fork, can you manually run `pytest tests/test_self_chat.py` for me?

skywalker023:
Done!

stephenroller:
Thanks!

Issue #3927, opened 2021-08-11
AttributeError: 'function' object has no attribute 'retriever' - Training ReGReT model


fabrahman:
Hi,

when trying to train a regret model using the below command (using a different model for initialization), I get the following error. Should I change anything in the train_model command? Thanks

command:
```
python parlai/scripts/train_model.py \
--model fid --task ranked_sim \
--indexer-type exact \
--generation-model bart --init-opt arch/bart_large \
--batchsize 16 --fp16 True --gradient-clip 0.1 --label-truncate 128 \
--log-every-n-secs 30 --lr-scheduler reduceonplateau --lr-scheduler-patience 1 \
--model-parallel True --optimizer adam --text-truncate 512 --truncate 512 \
--learningrate 1e-05 --validation-metric-mode min --validation-every-n-epochs 0.25 \
--validation-max-exs 1000 --validation-metric ppl --validation-patience 4 --n-docs 5 \
--model-file ${OUT_DIR}/model \
--regret True --regret-intermediate-maxlen 64 --regret-model-file dpr_outputs/fid_dpr_ranked/model \
--path-to-index /data/backup_tmp/passage_embeddings/my_passages \
--path-to-dpr-passages /data/retrieval_corpus_json/passages.tsv \
```

error:
```
Some weights of the model checkpoint at bert-base-uncased were not used when initializing BertModel: ['cls.seq_relationship.weight', 'cls.predictions.transform.dense.weight', 'cls.predictions.transform.LayerNorm.bias', 'cls.predictions.decoder.weight', 'cls.predictions.bias', 'cls.seq_relationship.bias', 'cls.predictions.transform.dense.bias', 'cls.predictions.transform.LayerNorm.weight']
- This IS expected if you are initializing BertModel from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing BertModel from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
2021-08-11 06:34:56,586 INFO     | Total parameters: 515,177,984 (514,784,768 trainable)
2021-08-11 06:34:56,586 INFO     | Loading existing model params from /amltdc91a6f0837d50a3bdf062e980f3fe7c/ParlAI/data/models/bart/bart_large/model
[...]

PR #3934, opened 2021-08-12
[RAG] Fix ReGReT

klshuster:
**Patch description**
Turns out the test I wrote for ReGReT was broken, so I didn't catch the bugs here. ReGReT should work now.

CC #3927 
**Testing steps**
Fixed the test and confirmed everything works

```bash
$ pytest -k TestRegret
====== test session starts ======
platform linux -- Python 3.7.9, pytest-6.2.1, py-1.10.0, pluggy-1.0.0.dev0
rootdir: /private/home/kshuster/ParlAI, configfile: pytest.ini
plugins: hydra-core-1.0.7, requests-mock-1.8.0, regressions-2.1.1, datadir-1.3.1
collected 666 items / 664 deselected / 2 selected

nightly/gpu/test_rag.py ..                                                                                                                                                     [100%]

====== slowest 10 durations ======
56.77s call     tests/nightly/gpu/test_rag.py::TestRegret::test_rag_regret_sep
11.50s call     tests/nightly/gpu/test_rag.py::TestRegret::test_rag_regret_same

(4 durations < 0.005s hidden.  Use -vv to show these durations.)
===== 2 passed, 664 deselected, 2 warnings in 76.86s (0:01:16) =====
```


mojtaba-komeili (approved):
LGTM.

spencerp (approved):
Nice! How broken was this? Are there any examples in docs that we need to update with this new flag for them to work?

also, plz teach me how to fix regret more generally thx

klshuster:
> How broken was this? 

So broken it literally didn't work.

> Are there any examples in docs that we need to update with this new flag for them to work?

I haven't included any examples in the docs for this, as it was not a highlighted model within the project; however, if more people are interested in the future this is something I can revisit.

> plz teach me how to fix regret more generally thx

Some code pointers, if you are indeed interested:

1. ReGReT only has [three (well, now four) flags](https://github.com/facebookresearch/ParlAI/blob/6a56b5c59f844162cd6076d3fe25e33fcca09b5f/parlai/agents/rag/args.py#L269): `--regret` controls using the model; `--regret-model-file` can be a separate retrieve+generate model; and `--regret-intermediate-maxlen` controls the maximum generation length of the intermediate generation (used in the second round of retrieval)
2. We build the ReGReT model in [`build_regret_model`](https://github.com/facebookresearch/ParlAI/blob/6a56b5c59f844162cd6076d3fe25e33fcca09b5f/parlai/agents/rag/rag.py#L184). This is essentially a no-op if we don't specify the model file, as the regret model is just the same as the normal, full model. Otherwise, we load up and create a new `RagModel` with the specified model file. There are a couple optimizations here that make this faster/more RAM-efficient; first, if the separate regret model has the same options for its index and passages, we share them between the two models so we don't load it twice; and, with this new flag I'm adding, you can now manually specify to share the same index and passages.
3. During training and eval, there are two functions required for using ReGReT: [`_regret_generate`](https://github.com/facebookresearch/ParlAI/blob/6a56b5c59f844162cd6076d3fe25e33fcca09b5f/parlai/agents/rag/rag.py#L709) and [`_regret_rebatchify`](https://github.com/facebookresearch/ParlAI/blob/6a56b5c59f844162cd6076d3fe25e33fcca09b5f/parlai/agents/rag/rag.py#L729). The former simply swaps around agent attributes and runs a full generation phase to get the outputs from the first round of ReGReT; the latter then creates a new batch where the query vector is no longer the human response but the generation from the first round of ReGReT; note that the final generation is still dependent on the input text, and it is just the _retrieval_ component that receives augmented input

PR #3945, opened 2021-08-16
[Crowdsourcing] edits for OSS personal knowledge project

jxmsML:
**Patch description**
- Add `ContextGenerator` based on the `task_name = 'msc'`
- Flexibility to build `ContextGenerator` for any given task:
    -  Modify `conversation_start_mode` to be either 'hi', or a 'task name' (e.g. 'blended_skill_talk', or 'msc'
    -  `get_context_generator ` in `parlai/crowdsourcing/tasks/model_chat/utils.py` will load `ContextGenerator` from the `<task_name>/agents.py` specified by `conversation_start_mode`
- Add `world_module` as an additional parameter for `run_task` in `parlai/crowdsourcing/tasks/model_chat/run.py`
- In `parlai/crowdsourcing/tasks/model_chat/worlds.py`, display fancy bot name with `self.bot.agent_id` rather than the default agent name from the 'id' field (mostly `TransformerGenerator`)
- Move "validate problem buckets" from `AbstractTurnAnnotationResultsCompiler` to `ModelChatResultsCompiler` init method.
```
if 'none_all_good' not in self.problem_buckets:
        # The code relies on a catchall "none" category if the user selects no other
         # annotation bucket
        raise ValueError(
       'There must be a "none_all_good" category in self.problem_buckets!'
         )
```

**Testing steps**
CI

**Other information**


parlai/crowdsourcing/tasks/model_chat/analysis/compile_results.py
```diff
@@ -63,6 +63,13 @@ def setup_args(cls):
     def __init__(self, opt: Dict[str, Any]):
 
         super().__init__(opt)
+        # Validate problem buckets
```
EricMichaelSmith, line 66: This check already exists in the superclass, `AbstractTurnAnnotationResultsCompiler`, but below it looks like you're moving the check to the subclasses - is there an advantage to this? (Do you have a case in which `none_all_good` is not a bucket? If so, maybe we can just move this check down to the top of the abstract class's `.compile_results()` method, because I'd like future subclasses of this class to still have this check)

parlai/crowdsourcing/tasks/model_chat/impl.py
```diff
@@ -15,10 +15,9 @@
 from parlai.crowdsourcing.tasks.model_chat.model_chat_blueprint import (
     SharedModelChatTaskState,
 )
-import parlai.crowdsourcing.tasks.model_chat.worlds as world_module
 
 
-def run_task(cfg: DictConfig, task_directory: str):
+def run_task(cfg: DictConfig, task_directory: str, world_module=None):
```
EricMichaelSmith, line 20: Yeah this makes sense, good for generality

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
```diff
@@ -307,8 +307,8 @@ class ModelChatBlueprintArgs(BaseModelChatBlueprintArgs):
     conversation_start_mode: str = field(
         default='hi',
         metadata={
-            "help": 'Whether to show "Hi!" or two previous utterances (as in BlendedSkillTalk) at the beginning of the conversation',
-            "choices": ['hi', 'bst'],
+            "help": 'Whether to show "Hi!" or two previous utterances (as in BlendedSkillTalk) at the beginning of the conversation '
+            'or specified by specific task name'
```
EricMichaelSmith, line 311: nit: maybe this would be clearer as `'Set to "hi" to show "Hi!" at the beginning of the conversation, or set to a task name to specify a custom context'`

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
```diff
@@ -384,7 +384,9 @@ def assert_task_args(
         args.blueprint.num_conversations = sum(conversations_needed.values())
         super().assert_task_args(args=args, shared_state=shared_state)
 
-        if args.blueprint.get("annotations_config_path", "") != "":
+        if args.blueprint.get(
```
EricMichaelSmith, line 387: 👍

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
```diff
@@ -424,9 +428,11 @@ def __init__(
         context_generator: Optional[ContextGenerator] = None
         if (
             args.blueprint.include_persona
-            or args.blueprint.conversation_start_mode == 'bst'
+            or args.blueprint.conversation_start_mode != 'hi'
```
EricMichaelSmith, line 431: Add: `# 'hi' mode does not use a context generator and instead just displays "Hi!" at the start of the conversation`

parlai/crowdsourcing/tasks/model_chat/model_chat_blueprint.py
```diff
@@ -440,7 +446,9 @@ def __init__(
                 'statistics_condition': statistics_condition,
                 'max_onboard_time': args.blueprint.max_onboard_time,
                 'onboard_task_data': self.onboard_task_data,
-                'onboarding_qualification': args.blueprint.onboarding_qualification,
+                "onboarding_qualification": args.blueprint.get(
+                    "onboarding_qualification", None
```
EricMichaelSmith, line 450: nit: you don't need the `None` arg in this case, it defaults to `None` if `"onboarding_qualification"` is not an attribute

parlai/crowdsourcing/tasks/model_chat/utils.py
```diff
@@ -421,7 +424,9 @@ def get_context_generator(
     if override_opt is not None:
         argparser.set_params(**override_opt)
     opt = argparser.parse_args([])
-    context_generator = ContextGenerator(opt, datatype='test', seed=0)
+    task_module = load_task_module(conversation_start_mode)
```
EricMichaelSmith, line 427: 👍

parlai/crowdsourcing/tasks/model_chat/analysis/compile_results.py
jxmsML, line 66: move out of `parlai/crowdsourcing/utils/analysis.py` here as it could be that "none_all_good" isn't necessary in `problem_buckets`.

parlai/crowdsourcing/tasks/model_chat/worlds.py
```diff
@@ -225,6 +225,15 @@ def parley(self):
         for idx, agent in enumerate([self.agent, self.bot]):
             if not self.chat_done:
                 acts[idx] = agent.act(timeout=self.max_resp_time)
+                if (
```
EricMichaelSmith, line 228: Yeah I think this is preferable to what was here previously, but now the Turker sees `self.bot_agent_id` as the name of the bot, right? Orion implemented a fix in his eval task to just set the bot display name for the Turkers as `YOUR PARTNER` so that they don't see the name of the bot - we might end up using that across all model chat tasks

parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
```diff
@@ -51,6 +51,14 @@ def setup_args(cls):
 
     def __init__(self, opt: Dict[str, Any]):
         super().__init__(opt)
+        # Validate problem buckets
```
EricMichaelSmith, line 54: nit: remove duplicated comment

parlai/crowdsourcing/tasks/model_chat/impl.py
jxmsML, line 20: adding `world_module` here so it doesn't have to call the `model_chat.worlds` by default but any `world_module` as long as it's consistent with `cfg.mephisto.blueprint.world_file`.

parlai/tasks/msc/agents.py
```diff
@@ -616,3 +639,125 @@ def __init__(self, opt, shared=None):
...
+    This generator was used to generate the context information shown to bots at the
+    beginning of a conversation, when crowdsourcing the conversations that for per-turn
+    human evaluation.
+    """
+
+    def __init__(self, override_opt, datatype='valid', seed: Optional[int] = None):
+        """
+        Initalize the context generator.
+
+        opt: only a 'datapath' key is required, to specify the ParlAI data folder
```
EricMichaelSmith, line 658: nit: you mean `override_opt`, not `opt`, right?

parlai/tasks/msc/agents.py
```diff
@@ -616,3 +639,125 @@ def __init__(self, opt, shared=None):
...
+        self._set_teacher_data_map()
+        self.context_done_statistics = copy.deepcopy(
+            override_opt.get('context_done_statistics', {})
+        )
+
+    def get_context(self, model_name: str = None) -> dict:
+        """
+        Get context information to be shown at the beginning of one conversation.
+
+        Values in return dict:
```
EricMichaelSmith, line 696: nit: it looks like there are more values in the return dict than just these, right? For instance, I see that `._extract_personas() ` returns `"context_for_bot_prompt"`, etc. Maybe it's worth just referring to the subclass's docstring instead of copying it here?

EricMichaelSmith (commented):
Yeah, good to have - thanks for generalizing `run_task`, etc.! I have a few nitpicky questions, and in particular, before approving, it might be good to decide where the `'none_all_good'` check should go

parlai/crowdsourcing/tasks/model_chat/analysis/compile_results.py
jxmsML: yea it's in my other pull request https://github.com/facebookresearch/ParlAI/pull/3946/files#diff-8edc807d80efc9567c466a08f1732192ad2cf8f6300f478e607ccf69e8936a1bR21

parlai/crowdsourcing/tasks/model_chat/worlds.py
jxmsML: sounds good. is the display name for bot configurable in Orion's implementation?

EricMichaelSmith: I don't think it currently is, but it'd be good for it to be

parlai/tasks/msc/agents.py
jxmsML: correct! sry will update the doc string

EricMichaelSmith (approved):

PR #3950, opened 2021-08-17
[BB2] Memory Only Fix

klshuster:
**Patch description**
We don't need to classify retrieval when we're dealing with `--knowledge-access-method <memory_only>/<none>`. We do not set the `query_generator_vec` unless we're in the other three regimes (`search_only`, `classify`, or `all`), and thus we should only generate the queries in these regimes as well.

**Testing steps**
Verified that `parlai i -mf zoo:blenderbot2/blenderbot2_400M/model` works with `--knowledge-access-method memory_only` (this would lead to a bug before)


jxmsML (approved):
LGTM! thanks for the fix

PR #3955, opened 2021-08-18
Allow more subclassing of self-chat world

EricMichaelSmith:
Allow for more subclassing of `SelfChatWorld` to modify/extend its logic:
- Abstract out the logic applied at the end of every episode, in order to allow subclasses to supply custom logic (in my case, resetting a random turn-index variable per episode)
- Allow for specifying custom logic for when to use seed utterances (in my case, only when the current turn index equals the value of the random turn-index variable specified above)

### DIFF
--- a/parlai/tasks/self_chat/worlds.py
+++ b/parlai/tasks/self_chat/worlds.py
@@ -140,7 +140,7 @@ class SelfChatWorld(DialogPartnerWorld):
             for a in agents:
                 a.reset()
 
-        if self._use_seed_utterances():
+        if self.turn_cnt == 0:
             self.acts = [None, None]
             # get the beginning of the conversation, which can include contexts
             # and/or any number of starting messages
@@ -160,7 +160,7 @@ class SelfChatWorld(DialogPartnerWorld):
                 self.agents[i].observe(validate(context))
             # clear contexts so they are only added once per episode
             self.contexts = None
-        elif self.seed_utterances:
+        elif self.seed_utterances and self._use_seed_utterances():
             # pop the next two seed messages (there may be less or more than 2 total)
             utts = self.seed_utterances[:2]
             self.seed_utterances = self.seed_utterances[2:]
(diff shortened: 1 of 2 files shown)


parlai/tasks/self_chat/worlds.py
```diff
@@ -186,3 +180,24 @@ def parley(self):
...
+        """
+        Apply logic to end the episode.
+        """
+        self.turn_cnt = 0
+        self.episode_cnt += 1
+        self.contexts = None
+        self.seed_utterances = None
+        agents = self.get_agents()
+        for a in agents:
+            a.reset()
```
stephenroller, line 194: maybe we should wrap this into self.reset?

stephenroller (approved):

parlai/tasks/self_chat/worlds.py
```diff
@@ -160,7 +154,7 @@ def parley(self):
                 self.agents[i].observe(validate(context))
             # clear contexts so they are only added once per episode
             self.contexts = None
-        elif self.seed_utterances:
+        elif self.seed_utterances and self._use_seed_utterances():
```
spencerp, line 157: I have a slight preference toward integrating this logic closer with the above:
```
self.seed_utterances = self._get_seed_utt_acts(
    self.episode_cnt, self.agents
)
```

spencerp (approved):

parlai/tasks/self_chat/worlds.py
EricMichaelSmith: I've created a `World.reset_agents()` method - is this what you were thinking? We can't call `World.reset()` directly here, because there's additional logic there - i.e. `World.reset()` also resets `self.total_parleys`, `self.total_exs`, `self.time`, etc.

EricMichaelSmith:
Thanks @stephenroller @spencerp for useful comments! I've tried to address your suggestions as best as possible, and since you've both approved the PR, I'm merging it in now - but if you feel like I didn't address your suggestions appropriately, lemme know and I'd be happy to make changes in another PR

PR #3957, opened 2021-08-18
[LIGHT] Jericho World dataset

mojtaba-komeili:
**Patch description**
Added ParlAI teachers for the [JerichoWorld Dataset](https://github.com/JerichoWorld/JerichoWorld) dataset. There are two sets of teachers for predicting (1) knowledge graph (2) actions. The focus of this patch is on knowledge graph teachers; although a simple action teacher is added as well. Here is a list of main teachers that are expected to use (not abstract etc.):

**StaticKGTeacher**: predict knowledge graph for a single state
Screen Shot 2021-08-18 at 5 49 54 PM

**ActionKGTeacher** predicts mutations to the knowledge graph after an action
Screen Shot 2021-08-18 at 5 51 14 PM

**StateToActionTeacher**: predicts player actions given the observation
Screen Shot 2021-08-18 at 5 52 18 PM


parlai/tasks/jericho_world/constants.py
```diff
@@ -13,10 +15,22 @@ class GraphMutations(Enum):
...
 
+try:
+    stopwords.words('english')
+except LookupError:
+    nltk.download('stopwords')
+
+
+# The set of words in the description of graph that we skip while checking
+# for overlap between knowledge grapj vertices and the description.
+GRAPH_VERT_SKIP_TOKEN = set(stopwords.words('english')) - {'you'}
```
JackUrb, line 26: Nit - it's useful to include reasonings for this kind of decision when external context drove the decision. Here the choice to leave out "you" is deliberate, but the note just describes what these tokens are used for.

parlai/tasks/jericho_world/agents.py
```diff
@@ -202,6 +210,62 @@ def _clean_example(self, game_step: Dict[str, Any]) -> Dict[str, Any]:
...
+                return False
+
+            for word in main_text.lower().split():
+                if not word or len(word) < 3 or word in consts.GRAPH_VERT_SKIP_TOKEN:
+                    continue
+                if word in content_text_tokens:
+                    return True
+            return False
+
+        def keep_edge(edge, text_tokens):
```
JackUrb, line 233: I'm not sure I follow what your `keep_edge` heuristic is here - so we keep all 'you in X' and 'you have X', but otherwise we keep edges where the subject and object are in the text? Probably could use a quick one line desc to this effect.

```
'''
Returns true if all the components of the edge are in the given text, or if it is the player location or unpruned inventory
'''
```

Sure this documentation isn't required, but it makes it easier to grok what's happening.

JackUrb (approved):
Overall this implementation is looking good - thanks for taking the time to break down the Jericho dataset and figure out what was not quite aligning. Did you come up with an answer to why some of the target actions weren't aligning with the given content?

I've left some nits, mostly about documentation for things that aren't super clear. You don't need to necessarily always be documenting at this level, but it certainly helps for getting code OSS ready early on, and makes it easier to return to this later when we need it again.

mojtaba-komeili:
> Overall this implementation is looking good - thanks for taking the time to break down the Jericho dataset and figure out what was not quite aligning. Did you come up with an answer to why some of the target actions weren't aligning with the given content?
> 
> I've left some nits, mostly about documentation for things that aren't super clear. You don't need to necessarily always be documenting at this level, but it certainly helps for getting code OSS ready early on, and makes it easier to return to this later when we need it again.

I couldn't figure out many of them, so ended up dropping them if them if they were detected as incomplete. But was able to "patch" some of them to some extent by checking the step before/after. I reached out to them about clarifying, but haven't received any response yet.

parlai/tasks/jericho_world/constants.py
mojtaba-komeili: Added extra explanation.

parlai/tasks/jericho_world/agents.py
mojtaba-komeili: Explained it in details in the method docstring.

PR #3964, opened 2021-08-20
Make the code compatible for fbcode

apravesh:
Make the code compatible for fbcode. We need to manually add some dependency model files and fetch them from datapath instead.

### DIFF
--- a/parlai/agents/rag/retrievers.py
+++ b/parlai/agents/rag/retrievers.py
@@ -17,10 +17,7 @@ import torch.cuda
 import torch.nn
 from tqdm import tqdm
 
-try:
-    from transformers import BertTokenizerFast as BertTokenizer
-except ImportError:
-    from transformers import BertTokenizer
+from transformers import BertTokenizer
 from typing import Tuple, List, Dict, Union, Optional, Any
 from typing_extensions import final
 from sklearn.feature_extraction.text import TfidfVectorizer
@@ -36,6 +33,7 @@ from parlai.core.torch_ranker_agent import TorchRankerAgent
 import parlai.utils.logging as logging
 from parlai.utils.torch import padded_tensor
 from parlai.utils.typing import TShared
+from parlai.utils.io import PathManager
 
 from parlai.agents.rag.dpr import DprQueryEncoder
 from parlai.agents.rag.polyfaiss import RagDropoutPolyWrapper
@@ -225,8 +223,11 @@ class RagRetrieverTokenizer:
     Wrapper for various tokenizers used by RAG Query Model.
     """
 
+    VOCAB_PATH = 'vocab.txt'
+
     def __init__(
         self,
+        datapath: str,
         query_model: str,
         dictionary: DictionaryAgent,
         max_length: int = 256,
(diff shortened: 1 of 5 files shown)


parlai/agents/rag/retrievers.py
```diff
@@ -17,10 +17,7 @@
 import torch.nn
 from tqdm import tqdm
 
-try:
-    from transformers import BertTokenizerFast as BertTokenizer
-except ImportError:
-    from transformers import BertTokenizer
+from transformers import BertTokenizer
```
klshuster, line 20: Is there a better way to make this change? Ideally we'd continue using `BertTokenizerFast` when it's available, since it is **significantly** faster

parlai/core/build_data.py
```diff
@@ -495,7 +495,6 @@ def modelzoo_path(datapath, path):
                 animal_ = '.'.join(animal.split(".")[:-1]) + '.build'
                 module_name_ = 'parlai.zoo.{}'.format(animal_)
                 my_module = importlib.import_module(module_name_)
-                my_module.download(datapath)
```
klshuster, line 498: what's the motivation for this change?

klshuster (changes requested):
Do you have any testing steps here to verify that nothing is broken on this end? I think something is broken

parlai/agents/rag/retrievers.py
stephenroller: Indeed. Given the try-catch, what is wrong with this?

parlai/core/build_data.py
stephenroller: +1, this feels wrong

apravesh: Was a local change that I forgot to remove. Thank you for catching it! Reverted this part

parlai/agents/rag/retrievers.py
apravesh: For some reason in fbcode ImportError was not thrown in this step but rather in the step below. Brought this back and added try catch in the section below to handle that.

apravesh:
> Do you have any testing steps here to verify that nothing is broken on this end? I think something is broken

I tested now with `parlai i -mf zoo:blenderbot2/blenderbot2_3B/model --search-server http://api.search.url` and it worked well

parlai/agents/rag/retrievers.py
```diff
@@ -259,7 +265,13 @@ def _init_tokenizer(
             ParlAI dictionary agent
         """
         if self.query_model in ['bert', 'bert_from_parlai_rag']:
-            return BertTokenizer.from_pretrained('bert-base-uncased')
+            try:
+                return BertTokenizer.from_pretrained('bert-base-uncased')
+            except ImportError or OSError:
```
stephenroller, line 270:
```suggestion
            except (ImportError, OSError):
```
I don't think or works here? I don't see anything in the python documentation about this syntax

stephenroller:
```python

In [1]: try:
   ...:     assert False
   ...: except OSError or AssertionError:
   ...:     print("Caught")
   ...:
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-1-c10757ab8e70> in <module>
      1 try:
----> 2     assert False
      3 except OSError or AssertionError:
      4     print("Caught")
      5

AssertionError:

In [2]: try:
   ...:     assert False
   ...: except (OSError, AssertionError):
   ...:     print("Caught")
   ...:
Caught

```

PR #4000, opened 2021-09-09
Disentangled checked_sentence_as_label.

leox1v:
Before it also added the label to the input. Now we have two separate mutators for this behavior: checked_sentence_as_label, add_label_to_input.

### DIFF
--- a/parlai/tasks/wizard_of_internet/agents.py
+++ b/parlai/tasks/wizard_of_internet/agents.py
@@ -624,7 +624,7 @@ class CheckedSentenceAsLabelLm(MessageMutator):
             label1 = ' '.join(ls[0:ind])
             label2 = ' '.join(ls[ind : len(ls)])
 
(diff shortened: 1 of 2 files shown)


jaseweston (commented):
ok that makes sense but then i guess it would be better to be consistent with the wizint ones i made as well, and the new lm one ... bit of a moving target rn (as i'm trying to run an experiment)! can you update and change the wizint ones and the lm too if you want to do this?

leox1v:
I fixed the mutators to not overwrite the original label and hence not copy the modified label to the next inputs. I also unified the mutators across WizWiki and WizInt.

The language modeling task can now be achieved by using the mutators add_label_to_input_lm and checked_sentence_as_label together, e.g.:

`parlai display_data -t wizard_of_internet -n 100 -dt valid --mutators flatten,add_label_to_input_lm,checked_sentence_as_label`

jaseweston (approved):

PR #4003, opened 2021-09-09
wow and wizint: can filter no passages used data

jaseweston:


stephenroller:
#4007 and #4008 pull out the side changes. Please remove them from this diff.

PR #4004, opened 2021-09-10
Force model chats starting with "Hi!" to use BST-style context

EricMichaelSmith:
**Patch description**
A recent refactor caused model chat to throw an error when starting a conversation with "Hi!" (`args.blueprint.conversation_start_mode == 'hi'`) and using personas: in that case, the code to pull the context would try to look for a `parlai.tasks.hi` module, which doesn't exist. This PR adds logic to state that, if we are starting our conversation with "Hi!", we want to use BST-style context.

**Testing steps**
CI checks on the model chat task


jxmsML (approved):
LGTM! Thanks for the fix.

PR #4005, opened 2021-09-10
Save model chat data in the Mephisto database

EricMichaelSmith:
**Patch description**
Save model-chat data in the Mephisto database via the `.get_custom_task_data()` method. Currently, this data will still *also* be stored on disk for compatibility; a future update will remove this.

**Testing steps**
Model chat CI checks


JackUrb (approved):
Happy to see motion towards using Mephisto standards here! Will definitely make the use cases for improving on the `DataBrowser` clearer for me as well.

PR #4006, opened 2021-09-12
[mutators] Prevent name collisions in mutators.

stephenroller:
**Patch description**
Recently @jaseweston had a spicy bug due to the global namespacing of mutators. This makes this failure case loud.

**Testing steps**
New CI.


EricMichaelSmith (approved):
Seems reasonable, good idea

jaseweston (approved):

klshuster (approved):

PR #4018, opened 2021-09-15
Minor: Print filename when missing in distributed.py

moyapchen:
...when you have a script that potentially includes file operations (ex when opening a manually-passed-in json file to load data for a task), this piece of code will eat the filename if it's missing.

** Testing Steps **
Have hit this empirically a few times over the past few days, lol.

### DIFF
--- a/parlai/utils/distributed.py
+++ b/parlai/utils/distributed.py
@@ -358,7 +358,9 @@ def slurm_distributed_context(opt):
         raise e
     except FileNotFoundError as e:
         # Slurm is not installed
-        raise RuntimeError('SLURM does not appear to be installed. Missing file: ' + e.filename)
+        raise RuntimeError(
+            'SLURM does not appear to be installed. Missing file: ' + e.filename
+        )
 
 
 def find_free_port() -> int:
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
This usually indicates a bad node

parlai/utils/distributed.py
```diff
@@ -356,9 +356,11 @@ def slurm_distributed_context(opt):
     except subprocess.CalledProcessError as e:
         # scontrol failed
         raise e
-    except FileNotFoundError:
+    except FileNotFoundError as e:
         # Slurm is not installed
-        raise RuntimeError('SLURM does not appear to be installed.')
+        raise RuntimeError(
```
stephenroller, line 361: hm maybe we should only try-catch the `subprocess.check_output` section. That way we won't eat the FileNotFoundError from way down in the stack

moyapchen: Don't have enough context to know what that means exactly... seeing as how it's not clear if anyone else has run into this, maybe I'll land this as is for now?

PR #4022, opened 2021-09-16
[RAG] Fix ReGReT Cuda Issue

klshuster:
**Patch description**
There was a small issue with tensors not being on appropriate devices; this did not show up in the RAG tests since the ReGReT tests are cpu-only (due to memory constraints...)

**Testing steps**
Tested a command from #3927 manually and confirmed this error 1) showed up before fix, and 2) does not show up after fix


mojtaba-komeili (approved):

stephenroller:
clean install is probably @Rebecca-Qian's issue being discussed internally

PR #3787, opened 2021-07-14
Fix an error for a 'DistributedDataParallel' object

jianguoz:
Fix error to address the `AttributeError: 'DistributedDataParallel' object has no attribute 'tokenize_query'` in distributed evaluation

**Patch description**

Fix error in the `RAG` model to address the `AttributeError: 'DistributedDataParallel' object has no attribute 'tokenize_query` in `distributed` evaluation. 

**Testing steps**

(referring to an internal task)
```console
parlai mp_eval  \
-t wizard_of_wikipedia \
--datatype test  \
--skip-generation False \
--batchsize 8 \
--model-file zoo:hallucination/bart_fid_rag/model \
--metrics all \
--n-docs 1 \
--model-parallel false \
--log-every-n-secs 30 
```

**Other information**

PR #4023, opened 2021-09-16
[BB2] Handle Distributed

klshuster:
**Patch description**
Make sure we don't directly access functions from `self.model` if we're in distributed mode. Fixes issue pointed out in #3872 

Follows very closely the solution in #3787 

**Testing steps**
Verified that bb2 tests still pass.

### DIFF
--- a/projects/blenderbot2/agents/blenderbot2.py
+++ b/projects/blenderbot2/agents/blenderbot2.py
@@ -47,6 +47,7 @@ from .modules import (
 )
 from .sub_modules import RetrievalType, KnowledgeAccessMethod
 from parlai.agents.fid.fid import SearchQuerySearchEngineFiDAgent
+from parlai.utils.fsdp import is_fsdp
 
 
 ZOO_QUERY_GENERATOR = 'zoo:blenderbot2/query_generator/model'
@@ -389,6 +390,13 @@ class BlenderBot2RagAgent(RagAgent):
         self._rag_model_type = model
         self._rag_model_interface = RAG_MODELS[model](self.opt, self.NULL_IDX)
 
+    @property
+    def model_api(self) -> BlenderBot2RagModel:
+        if hasattr(self.model, 'module') and not is_fsdp(self.model):
+            return self.model.module
+        else:
+            return self.model
+
     def build_model(self) -> BlenderBot2RagModel:
         """
         Build and return BlenderBot2RagModel.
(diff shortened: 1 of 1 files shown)


projects/blenderbot2/agents/blenderbot2.py
```diff
@@ -507,7 +515,7 @@ def _set_query_vec(self, observation: Message) -> Message:
             )
         if self.add_person_tokens:
             query_str = self._remove_person_tokens(query_str)
-        observation['query_vec'] = self.model.tokenize_query(query_str)
+        observation['query_vec'] = self.model_api.tokenize_query(query_str)
```
mojtaba-komeili, line 518: Shouldn't this be `self.model_api().tokenize_query(query_str)`?

projects/blenderbot2/agents/blenderbot2.py
```diff
@@ -541,7 +549,7 @@ def _set_memory_vec(self, observation: Message) -> Message:
                     m for m in memories if self.opt['memory_extractor_phrase'] in m
                 ]
             if memories:
-                mem_vecs = [self.model.tokenize_memory(mem) for mem in memories]
+                mem_vecs = [self.model_api.tokenize_memory(mem) for mem in memories]
```
mojtaba-komeili, line 552: same as above.

mojtaba-komeili: And for the rest of the instances of it as well.

projects/blenderbot2/agents/blenderbot2.py
stephenroller: idts, model_api is a `@property`

Kurt, maybe we should bump this all the way up to TA. We have this `module` stuff everywhere

mojtaba-komeili: Oh, yeah right. I missed that.

mojtaba-komeili (approved):

projects/blenderbot2/agents/blenderbot2.py
klshuster: > bump up to TA

We would need to think of an elegant solution so that we can still call `self.model()` directly (which _does_ work in distributed). We could solve it with `model_api` if you think that's appropriate

stephenroller: you're right, nvm

PR #4026, opened 2021-09-17
Allow the style classifier to save output probabilities

EricMichaelSmith:
**Patch description**
If `--print-scores True`, allow the classifier from the style-controlled generation project to save per-class probabilities in a `{class: prob}` dict. Useful for looking at the per-example per-class probabilities when doing `parlai eval_model`

**Testing steps**
Existing CI checks


klshuster (approved):
cool, this seems quite useful

PR #4032, opened 2021-09-21
Updating Cairaoke Dict Params.

kauterry:
**Patch description**

Making Torchscript compatible with a Bart model with lowercase vocabulary.

**Testing steps**

**Other information**


stephenroller:
cleaninstall could be fixed by rebasing.

PR #4033, opened 2021-09-21
Wizint the datset compiler

mojtaba-komeili:
**Patch description**
Added the `compile_results.py`. It retrieves the Wizard of Internet task's crowdsourced data from Mephisto and stores it in a josn file.

**Testing steps**
Running the data compiler with a subset of the data in Mephisto.


mojtaba-komeili:
There was a mistake in committing to the wrong branch.

PR #4034, opened 2021-09-21
[WizInt OSS] data compiler using Mephisto data browser

mojtaba-komeili:
**Patch description**
Added `compile_results.py`: It retrieves the Wizard of Internet task's crowdsourced data from Mephisto and stores it in a josn file.

**Testing steps**
Running the data compiler with a subset of the data in Mephisto.


EricMichaelSmith (approved):
Yeah this seems like a reasonable subclassing of `AbstractResultsCompiler`! No real comments, because this is code going into a `projects/` folder, but happy to review anything in more depth if desired :)

mojtaba-komeili:
> Yeah this seems like a reasonable subclassing of `AbstractResultsCompiler`! No real comments, because this is code going into a `projects/` folder, but happy to review anything in more depth if desired :)

Thanks Eric. Yeah, this code is not going to be reused very often and I tested it with my data. Its best use might be for having the complete code reproducing the Wizard of Internet dataset and more importantly providing a real world example of using the new `AbstractResultsCompiler`. At its current state it is perfectly fine for that purpose.

Issue #4049, opened 2021-10-03
No detect_offensive_language argument command 


yana-xuyan:
**Bug description**
Please enter a clear and concise description of what the bug is.
Hello, I'm following the paper "Recipes for Safety in Open-domain Chatbots" and trying to find the word-list based and transformer-based classifier to evaluate the safety factor. What I can find in Parlai github repo is the script called detect_offensive_language.py.

**Reproduction steps**
Enter steps to reproduce the behavior.

However, when I tried to run ```parlai detect_offensive_language --task "convai_chitchat" --display-examples True``` as it's shown in the comments, it shows that "Parse Error: argument COMMAND".

**Expected behavior**
Give a clear and concise description of what you expected to happen.
Could you please tell me how to find and use the two mentioned classifiers in the paper? Thank you very much!

**Logs**
Please paste the command line output:
usage: parlai [-h] [--helpall] [--version] COMMAND ...

       _
      /")
     //)
  ==//'=== ParlAI
   /

optional arguments:
  -h, --help               show this help message and exit
  --helpall                List all commands, including advanced ones.
  --version                Prints version info and exit.

Commands:

  eval_model (em, eval)    Evaluate a model
  display_data (dd)        Display data from a task
  display_model (dm)       Display model predictions.
  train_model (tm, train)  Train a model
  interactive (i)          Interactive chat with a model on the command line
  safe_interactive         Like interactive, but adds a safety filter
  self_chat                Generate self-chats of a model

Parse Error: argument COMMAND: invalid choice: 'detect_offensive_language' (choose from 'help', 'h', 'helpall', 'build_candidates', 'build_dict', 'eval_model', 'em', 'eval', 'convert_to_json', 'convert_to_parlai', 'convo_render', 'data_stats', 'detect_offensive', 'display_data', 'dd', 'display_model', 'dm', 'train_model', 'tm', 'train', 'eval_wordstat', 'extract_image_feature', 'interactive', 'i', 'interactive_web', 'iweb', 'multiprocessing_eval', 'mp_eval', 'multiprocessing_train', 'mp_train', 'party', 'parrot', 'profile_interactive', 'profile_train', 'safe_interactive', 'self_chat', 'token_stats', 'torchscript', 'vacuum', 'verify_data')

```
Output goes here
```

**Additional context**
Add any other context about the problem here. (like proxy settings, network setup, overall goals, etc.)

klshuster:
could you try running `parlai detect_offensive` instead?

[...]

PR #4054, opened 2021-10-05
Fix `detect_offensive_language` script

klshuster:
**Patch description**
The script would break if the `labels` or `text` fields within a task were empty strings; this fixes that.

I'm not really sure who to tag on this as a reviewer so lmk if someone else should review.

**Testing steps**
See #4049 for task that breaks the script. I've also added a new test to confirm this works.

### DIFF
--- a/parlai/scripts/detect_offensive_language.py
+++ b/parlai/scripts/detect_offensive_language.py
@@ -81,6 +81,8 @@ def detect(opt):
         return log
 
     def classify(text, stats):
+        if not text:
+            return
         offensive = False
         stats['total'] += 1
         if opt['safety'] == 'string_matcher' or opt['safety'] == 'all':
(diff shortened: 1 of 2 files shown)


parlai/scripts/detect_offensive_language.py
```diff
@@ -81,6 +81,8 @@ def report(world, stats):
         return log
 
     def classify(text, stats):
+        if not text:
```
emilydinan, line 84: i wonder if the right answer here is to just make our utils in `utils/safety.py` robust to empty text

klshuster: that's fair, would catch it at the source...

klshuster:
Updated the PR to make `utils/safety.py` robust to empty text

klshuster:
bump

klshuster:
bump

stephenroller (approved):
oct 2021?!

PR #4059, opened 2021-10-07
Update NLTK for security

stephenroller:
**Patch description**
Dependabot didn't create a patch, and NLTK team finally put out a release

**Testing steps**
CI


klshuster:
I think we need to require a higher version of `regex`... tried locally to access `regex.Pattern`, had to upgrade to `regex-2021.9.30`

stephenroller:
Delegated to Megan.

PR #4060, opened 2021-10-07
Never load datatype from opt files

stephenroller:
**Patch description**
After internal discussion, datatypes are a nasty thing to be pulling from opts and can create some weirdness. Always use the root datatype.

**Testing steps**
CI


stephenroller:
Abandoned bc #4134 comandeers.

PR #4061, opened 2021-10-07
[Modular] Update Transformer Layer `__init__`s

klshuster:
**Patch description**
The `TransformerDecoder(Layer)` and `TransformerEncoder(Layer)` support passing `**kwargs` for maximum customizability; however, these `**kwargs` are unnecessarily (and, actually, incorrectly) being passed to the `nn.Module`'s `super().__init__()`, which doesn't accept arbitrary `**kwargs`. This patch removes these from the super call

**Testing steps**
Subclassed a `TransformerEncoder`, attempted to initialize with arbitrary parameters, confirmed it works after fix.


jxmsML (approved):
LGTM! Thanks for the fix!

PR #4082, opened 2021-10-12
Task-oriented dialogue core, agents, world, world script, and tests

moyapchen:
[Will write better documentation for this in a bit]


moyapchen:
Splitting and simplifying in stack of: #4174 #4178 #4179

PR #4093, opened 2021-10-14
Prevent tensorboardX SummaryWriter from cluttering logs with warnings.

kauterry:
**Patch description**

We remove the invalid characters from the tag name beforehand which the TensorboardX SummaryWriter disallows.

**Testing steps**

**Other information**


moyapchen (approved):
Feel like would be better to sub with some other character if the goal is just to replace ":", but accepting to unblock.

PR #4095, opened 2021-10-15
Move defaults to config yaml files for crowdsourcing tasks

meganung:
Moved configs from run scripts to yaml config files for the crowdsourcing tasks. 

Checked that the configs set are respected by running the run scripts with the flag [ --cfg job ](https://hydra.cc/docs/advanced/defaults_list/#debugging-the-defaults-list) and inspecting the defaults list by running the run scripts with the flag [--info defaults](https://hydra.cc/docs/advanced/defaults_list/#debugging-the-defaults-list)

~~crowdsourcing tests pass except for:~~

* ~~test_chat_demo which will pass when the corresponding [PR](https://github.com/facebookresearch/Mephisto/pull/587) in Mephisto is merged~~ done- pointed to new mephisto release tag
* ~~test_turn_annotations_static tests - failing because need to add onboarding testing (in a separate PR and then rebase this branch ontop of it). Config changes make the onboarding_qualification parameter to be non optional so cannot be null.~~ this is done.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,7 @@ commands:
             command: |
               cd ..
               git clone git@github.com:facebookresearch/Mephisto.git Mephisto
-              cd Mephisto; git checkout v0.4.0 -b stable
+              cd Mephisto; git checkout v0.4.1 -b stable
               pip install -r requirements.txt
               python setup.py develop
               # `echo` so that ENTER will be pressed at the prompt


meganung:
Everything should be in order now, I think it's failing the one crowdsourcing TestChatDemo on CI because the [CI config for ParlAI uses Mephisto v0.4.0](https://github.com/facebookresearch/ParlAI/blob/main/.circleci/config.yml#L170). I merged the corresponding mephisto [PR](https://github.com/facebookresearch/Mephisto/pull/587) so locally when I'm on the latest Mephisto main commit and on this branch the crowdsourcing tests pass (including TestChatDemo). 

What's the proper or preferred procedure so we can merge this PR and also have the test pass?

parlai/crowdsourcing/tasks/acute_eval/fast_eval.py
```diff
@@ -547,14 +547,7 @@ def analyze_results(self, args: Optional[str] = None):
...
 
-defaults = [
-    '_self_',
-    {"mephisto/blueprint": FAST_ACUTE_BLUEPRINT_TYPE},
-    {"mephisto/architect": "local"},
-    {"mephisto/provider": "mock"},
-    'conf/base_fast_acute',
-    {"conf": "example_fast_acute"},
-]
+defaults = ['_self_', {"conf": "example_fast_acute"}]
```
EricMichaelSmith, line 550: hmm, there's a lint error on line 35 here because FAST_ACUTE_BLUEPRINT_TYPE is now unused. You can try just removing the import, but if you find that the import can't be removed because it triggers registering the Blueprint, you can copy how I've worked around this in another file by adding https://github.com/facebookresearch/ParlAI/blob/e04a35cff3f8f11dff0cbcde6d262a3fc0adbffc/parlai/crowdsourcing/tasks/acute_eval/analysis.py#L38

parlai/crowdsourcing/tasks/acute_eval/run.py
```diff
@@ -30,12 +30,7 @@
 
 """
 
-defaults = [
-    {"mephisto/blueprint": BLUEPRINT_TYPE},
-    {"mephisto/architect": "local"},
-    {"mephisto/provider": "mock"},
-    {"conf": "example"},
-]
+defaults = ["_self_", {"conf": "example"}]
```
EricMichaelSmith, line 33: (the same lint error as above appears here)

parlai/crowdsourcing/utils/tests.py
```diff
@@ -89,10 +89,10 @@ def _set_up_config(
             self.config = compose(
                 config_name="example",
                 overrides=[
-                    f'+mephisto.blueprint._blueprint_type={blueprint_type}',
-                    f'+mephisto.blueprint.link_task_source=False',
-                    f'+mephisto/architect=mock',
-                    f'+mephisto/provider=mock',
+                    f'mephisto.blueprint._blueprint_type={blueprint_type}',
```
EricMichaelSmith, line 92: Ooof, fiddling with the pluses is painful -__- Hmm what does the double-plus mean? Does that allow for either adding *or* overwriting the param?

PR #4096, opened 2021-10-15
Add onboarding to crowdsourcing tests

meganung:
Added support for onboarding in crowdsourcing tests. Used [this example](https://github.com/facebookresearch/Mephisto/blob/dfdfc7592c8e401be5cf14e4a60a4ff88cf6217f/test/core/test_supervisor.py#L525) in the Mephisto repo as a guide. 

CI crowdsourcing tests.

(will rebase this [PR](https://github.com/facebookresearch/ParlAI/pull/4095) after)

### DIFF
--- a/parlai/crowdsourcing/utils/tests.py
+++ b/parlai/crowdsourcing/utils/tests.py
@@ -132,67 +132,12 @@ class AbstractCrowdsourcingTest:
         else:
             raise ValueError('No channel could be detected!')
 
-    def _register_and_onboard_mock_agents(self, num_agents: int = 1) -> List[str]:
+    def _register_mock_agents(
+        self, num_agents: int = 1, assume_onboarding: bool = False
+    ) -> List[str]:
         """
-        Register mock agents and onboard them for testing, taking the place of crowdsourcing workers.
-
-        Specify the number of agents to register. Return the agents' IDs after creation.
-        """
-        for idx in range(num_agents):
-
-            mock_worker_name = f"MOCK_WORKER_{idx:d}"
-            max_num_tries = 6
-            initial_wait_time = 0.5  # In seconds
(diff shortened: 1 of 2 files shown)


parlai/crowdsourcing/utils/tests.py
```diff
@@ -215,7 +273,10 @@ def _get_agent_state(self, task_data: Dict[str, Any]):
         """
 
         # Set up the mock human agent
-        agent_id = self._register_mock_agents(num_agents=1)[0]
+        if self.config.mephisto.blueprint.get("onboarding_qualification", None):
```
meganung, line 276: is there a better way to check if we should test onboarding? Would it be better to pass an onboarding bool variable?

JackUrb: At the moment this is how we specify onboarding in Mephisto:
https://github.com/facebookresearch/Mephisto/blob/main/mephisto/abstractions/blueprints/mixins/onboarding_required.py#L103-L107
So I think this is really the only way to check that a config is trying to use onboarding.

parlai/crowdsourcing/utils/tests.py
```diff
@@ -132,6 +132,64 @@ def _get_channel_info(self):
         else:
             raise ValueError('No channel could be detected!')
 
+    def _register_and_onboard_mock_agents(self, num_agents: int = 1) -> List[str]:
```
EricMichaelSmith, line 135: Yeah, I think this functionality is useful! But because most of this code is duplicated from `._register_mock_agents()`, could you instead enable this behavior with an `assume_onboarding=True` flag perhaps?

EricMichaelSmith (commented):
Useful to have this feature - thanks for building it in! I have a comment about structure

PR #4098, opened 2021-10-15
Sanitize the file names so that they don't cause error in different operating systems

### LINKED ISSUE #4097: Windows - OSError: [Errno 22] Invalid argument: '/tmp/babi_memnn.cands-babi:task10k:1.cands'
When you run the following command

`parlai interactive --model-file /tmp/babi_memnn --eval-candidates vocab`

The path generated by _get_task_candidates_path_ function contains a colon ":" which invalid character for a windows file/directory name. So we get the error below

`OSError: [Errno 22] Invalid argument: '/tmp/babi_memnn.cands-babi:task10k:1.cands'`

Invalid characters should be removed from the file name that is generated so that it can work without errors.

sherazlodhi:
Contains a fix to sanitize the file names. Closes #4097


stephenroller:
This just needs someone to regenerate the teacher most likely

PR #4132, opened 2021-11-01
Concatenating the conversations from a task into a single string

mojtaba-komeili:
**Patch description**
Generates a json file that has a single string made from all the conversations in the given task.
example usage:
```
python conv_concat.py --task convai2 
```


stephenroller:
Just need it as a one-off, we'll avoid landing this.

PR #4134, opened 2021-11-01
Never load datatype from opt files

moyapchen:
Stealing from https://github.com/facebookresearch/ParlAI/pull/4060:

> Patch description
> After internal discussion, datatypes are a nasty thing to be pulling from opts and can create some weirdness. Always use the root datatype.
> 
> Testing steps
> CI 

Also added some changes to support "final extra eval opt"

Made separate change cause there's not a nice way to commandeer, as per https://stackoverflow.com/questions/66539231/how-to-change-the-owner-of-a-pr-on-github-how-to-commandeer-an-open-github-pr


stephenroller (commented):
We can do this but I was hoping we would do what we've talked about before, and simply eliminating this "final eval" and making the extra opt happen during valid/test.

PR #4144, opened 2021-11-05
dicts to messasges

mojtaba-komeili:
**Patch description**
Changed the Wizard of Wikipedia examples from `dict` to `Message`. Running WoW with `parlai dd` used to show the following warning about the teacher examples being `dict` instead of `Message`. 

Screen Shot 2021-11-05 at 1 47 13 PM

**Testing steps**
1. The existing teacher tests.
2. Checking that warning doesn't appear anymore.


klshuster (approved):
thanks for this!

PR #4146, opened 2021-11-05
FiD gold retrieved docs agent n-docs debug

mojtaba-komeili:
**Patch description**
Working with `WizIntGoldDocRetrieverFiDAgent`, I noticed it OOMing more often that it should. Looking into it, I realized that the agent was including all the retrieved docs. WoI doc chunk mutators split a document into 100 chunks or more and `ObservationEchoRetriever` returns these all docs---ignoring the `n_docs` parameter.

This patch fixed this issue by picking `n_docs` number of docs from the list of docs and then adding some filler docs.

NOTE: this problem could be fixed in the upstream by setting `--woi-doc-max-chunks` from `woi_dropout_retrieved_docs` as well. But this makes sure that it is safe to run it with other mutators, teachers as well.

**Testing it**
Ran 100 steps of training with a debug level message that shows the number of docs before and after trimming.

Screen Shot 2021-11-05 at 6 09 34 PM

### DIFF
--- a/parlai/agents/fid/fid.py
+++ b/parlai/agents/fid/fid.py
@@ -353,6 +353,14 @@ class GoldDocRetrieverFiDAgent(SearchQueryFiDAgent):
 
     def _set_query_vec(self, observation: Message) -> Message:
         retrieved_docs = self.get_retrieved_knowledge(observation)
+        if len(retrieved_docs) > self._n_docs:
+            logging.warning(
+                f'Your `get_retrieved_knowledge` method returned {len(retrieved_docs)} Documents, '
+                f'instead of the expected {self._n_docs}. '
+                f'This agent will only use the first {self._n_docs} Documents. '
+                'Consider modifying your implementation of `get_retrieved_knowledge` to avoid unexpected results.'
+            )
+            retrieved_docs = retrieved_docs[: self._n_docs]
         self.model_api.retriever.add_retrieve_doc(
             observation[self._query_key], retrieved_docs
         )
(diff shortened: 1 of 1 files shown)


klshuster (commented):
This looks good for the `WizIntGoldDocRetriever`, but this should probably be respected at the top level as well --> i.e., in `_set_query_vec`, we should cut the retrieved documents to be no more than `opt['n_docs']`

parlai/agents/fid/fid.py
```diff
@@ -352,6 +353,14 @@ def get_retrieved_knowledge(self, message):
 
     def _set_query_vec(self, observation: Message) -> Message:
         retrieved_docs = self.get_retrieved_knowledge(observation)
+        if len(retrieved_docs) > self._n_docs:
```
mojtaba-komeili, line 356: @klshuster picking the right documents, if there are more than `n_docs`, depends a lot on the data set. I add this warning here and trim it down to the first `n_docs`. I hope this properly addresses your comment. LMK.

klshuster: sounds good to me. given how interactive this retriever is (i.e., interactive with the user --> the user has total control over what documents are being passed), it makes sense that the onus is on them to either 1) provide the right number of documents, or 2) set `--n-docs` correctly

parlai/agents/fid/fid.py
```diff
@@ -352,6 +353,14 @@ def get_retrieved_knowledge(self, message):
...
+            logging.warning(
+                f'Your `get_retrieved_knowledge` method returned {len(retrieved_docs)} Documents, '
+                f'instead of the expected {self._n_docs}. '
+                f'This agent will only use the first {self._n_docs} Documents. '
+                'Consider modifying your implementation of `get_retrieved_knowledge` to avoid unexpected results.'
```
klshuster, line 361: can we also say `'or set the --n-docs parameter accordingly'` to inform the user that they can do that too?

klshuster (approved):

PR #4149, opened 2021-11-08
Minor variable name fix in error code

moyapchen:
Super minor one char change; ran into while dealing with MultiWoz.


stephenroller (approved):
cheers

stephenroller:
Maybe we should add a test for this code path by downloading some file and lying about its hash.

PR #4151, opened 2021-11-08
Improve error message in Teachers

stephenroller:
**Patch description**
This is a fairly common mistake, but the old error message is quite obtuse and far away from the actual solution. Provide a better error message that gives more hints about where the issue actually is.

**Testing steps**
CI, reading.


stephenroller:
No don't prune!

moyapchen (approved):
lgtm, ty.

PR #4162, opened 2021-11-10
unify turn annotation tasks (model chat and turn annotations static tasks)

meganung:
**Patch description**
* makes the schema of JSON files for defining the annotation buckets the same between model chat task and turn annotations static task
    * flag name: annotations_config_path
    * schema [example](https://github.com/facebookresearch/ParlAI/blob/3ff5cfa4ecee2cf84a780f80ec5f0e2edcf84d48/parlai/crowdsourcing/tasks/model_chat/task_config/annotations_config.json)
* removes the unused task_description flag in model chat task
* chat_title → title_task in model chat task
* make task description a html file in turn annotations static task to match model chat task

**Testing steps**
Circle CI


parlai/crowdsourcing/tasks/turn_annotations_static/hydra_configs/conf/example.yaml
```diff
@@ -11,20 +11,14 @@ mephisto:
     extra_source_dir: ${task_dir}/webapp/src/static
     units_per_assignment: 5
     onboarding_qualification: turn_annotations_static
-    annotation_buckets: ${task_dir}/task_config/annotation_buckets.json
+    annotations_config_path: ${task_dir}/task_config/annotations_config.json
```
EricMichaelSmith, line 14: Hmm, here `annotations_config.json` has an `s` at the end of `annotations`, but in `parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py` it doesn't

parlai/crowdsourcing/tasks/turn_annotations_static/hydra_configs/conf/example.yaml
```diff
@@ -11,20 +11,14 @@ mephisto:
...
     response_field: False
+    task_description_file: ${task_dir}/task_config/task_description.html
```
EricMichaelSmith, line 16: It'd be good to add in a sample `task_description.html` file as well here. Basically, the principle behind all tasks in `parlai/crowdsourcing/tasks/` is that they should be immediately runnable on the command-line as-is by just calling the `run.py` file without any other args, in order to let the user quickly understand what the task looks like and how it works. However, this requires all config files specified in the default YAML file to exist in the `task_config/` folder

parlai/crowdsourcing/tasks/turn_annotations_static/turn_annotations_blueprint.py
```diff
@@ -184,18 +189,31 @@ def get_frontend_args(self) -> Dict[str, Any]:
         for use by the task's frontend.
         """
 
+        # Load task description from file
+        MISSING_TASK_DESCRIPTION_TEXT = (
```
EricMichaelSmith, line 193: hmm what is the advantage of defining `MISSING_TASK_DESCRIPTION_TEXT` and then defining `task_description` as equal to it?

tests/crowdsourcing/tasks/model_chat/expected_states/final_chat_data.json
```diff
@@ -132,7 +132,7 @@
...
+        "annotations_config": "{\n    \"config\": {\n        \"bucket_0\": {\n            \"name\": \"Bucket 0\",\n            \"description\": \"this response implies something...0\"\n        },\n        \"bucket_1\": {\n            \"name\": \"Bucket 1\",\n            \"description\": \"this response implies something...1\"\n        },\n        \"bucket_2\": {\n            \"name\": \"Bucket 2\",\n            \"description\": \"this response implies something...2\"\n        },\n        \"bucket_3\": {\n            \"name\": \"Bucket 3\",\n            \"description\": \"this response implies something...3\"\n        },\n        \"bucket_4\": {\n            \"name\": \"Bucket 4\",\n            \"description\": \"this response implies something...4\"\n        },\n        \"none_all_good\": {\n            \"name\": \"None, all good\",\n            \"description\": \"This response implies that there are no problems with the data\"\n        }\n    }\n}\n",
```
EricMichaelSmith, line 135: oy yeah these strings are the worst -__-

EricMichaelSmith (commented):
Yeah these improvements are super super useful to have, thanks a lot! Some comments about loose ends

parlai/crowdsourcing/tasks/turn_annotations_static/hydra_configs/conf/example.yaml
meganung: added!

parlai/crowdsourcing/tasks/model_chat/hydra_configs/conf/example.yaml
```diff
@@ -22,13 +22,6 @@ mephisto:
     assignment_duration_in_seconds: 600
     max_num_concurrent_units: 0  # 0 means infinite; set this to a positive integer to limit concurrent HITs and prevent crashes
     maximum_units_per_worker: 3
-    task_description: |
```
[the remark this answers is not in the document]
jxmsML: this part always confused me in the past. "task_description.html", "left_pane.html", `task_description` in the yaml file

parlai/crowdsourcing/tasks/model_chat/frontend/components/checkboxes.jsx
```diff
@@ -41,13 +41,13 @@ function Checkboxes({
               onChange={(evt) => {
```
jxmsML, line 41: @EricMichaelSmith  @meganung  I wonder if this logic work for non-checkbox inputtype, e.g. "radio" button,

suppose you click radio_bucket1, and then changed it to radio_bucket2,
I guess in this case, both radio_bucket1, and radio_bucket2 are `True`, which is not exatly what we want

jxmsML: shall we remove L31? and always set  `inputType = checkbox `

parlai/crowdsourcing/tasks/model_chat/task_config/annotations_config.json
```diff
@@ -1,32 +1,28 @@
...
-        "value": "bucket_4",
-        "name": "Bucket 4",
-        "description": "this response implies something...4"
-    },
-    {
-        "value": "none_all_good",
-        "name": "None, all good",
-        "description": "This response implies that there are no problems with the data"
+{
+    "config": {
```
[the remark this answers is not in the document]
jxmsML: like that this is a dictionary !!! so we can pass our own config as well without have to change too much in the blueprint!

parlai/crowdsourcing/tasks/model_chat/frontend/components/checkboxes.jsx
meganung: it does work if you specify type: radio in the config - i've done it in turn annotations static tasks at least. Radio buttons you can only select one so only one will be `True`.

EricMichaelSmith: Hmm interesting - @meganung in that case, do you expect that radio boxes will work correctly in the current version of model_chat as well, given this PR? I suppose I see two possibilities:
(1) we test out that radio buttons render correctly in the current version of model chat (i.e. with only one being checked at once as @jxmsML notes) and then add documentation about how to enable that to the README. In this case we should also add a unit test for radio buttons (or a TODO for a unit test at the very least, although a unit test is of course better) 
(2) we just fix `input_type = checkbox` for now, as @jxmsML suggests, and maybe add a TODO to add and test out radio box support for this task in the future

Either possibility is fine with me, but @meganung lemme know what you think :)

meganung: good catch @jxmsML and thanks for the input @EricMichaelSmith -- I manually tested it and the frontend is fine (visually you can only check one) but our code for updating the data doesn't do that so ends up saving multiple different values as true if the user switches between selecting different radio buttons. I added a TODO for now since adding support for radio buttons in model chat task can be a separate PR.

jxmsML (approved):
LGTM!! awesome job @meganung

parlai/crowdsourcing/tasks/model_chat/frontend/components/checkboxes.jsx
EricMichaelSmith: @jxmsML Hmm, anything that affects the final data will break one of the tests, yes - I see that Megan fixed the tests in this PR, but it doesn't look like she needed to fix anything related to the checkbox type, so presumably the checkbox type doesn't get saved in the final data :)

EricMichaelSmith (approved):
Yeah thanks a lot @meganung , this is super useful!

Issue #4148, opened 2021-11-07
Can't run T5 agent on latest transformers


zhangmozhi:
**Bug description**
When training T5 model with Transformer version 4.10+, ParlAI will throw a runtime error `Must use transformers package >= 4.3 to use t5`.

**Reproduction steps**
Install the latest Transformer and run any T5 training command. For example:
`parlai train_model -m hugging_face/t5 -mf $HOME/project/model/t5_test -t convai2 -bs 24 --fp16 true -eps 1 -lr 1e-5 --optimizer adam --t5-model-arch t5-3b --t5-model-parallel true`

**Expected behavior**
There should be no runtime error.

**Logs**

```
Traceback (most recent call last):
  File "/home/mozhi/miniconda3/bin/parlai", line 33, in <module>
    sys.exit(load_entry_point('parlai', 'console_scripts', 'parlai')())
  File "/mnt/mozhi/ParlAI/parlai/__main__.py", line 14, in main
    superscript_main()
  File "/mnt/mozhi/ParlAI/parlai/core/script.py", line 325, in superscript_main
    return SCRIPT_REGISTRY[cmd].klass._run_from_parser_and_opt(opt, parser)
  File "/mnt/mozhi/ParlAI/parlai/core/script.py", line 108, in _run_from_parser_and_opt
    return script.run()
  File "/mnt/mozhi/ParlAI/parlai/scripts/train_model.py", line 997, in run
    self.train_loop = TrainLoop(self.opt)
  File "/mnt/mozhi/ParlAI/parlai/scripts/train_model.py", line 353, in __init__
    self.agent = create_agent(opt)
  File "/mnt/mozhi/ParlAI/parlai/core/agents.py", line 479, in create_agent
    model = model_class(opt)
  File "/mnt/mozhi/ParlAI/parlai/core/torch_generator_agent.py", line 484, in __init__
    self.model = fsdp_utils.fsdp_wrap(self.build_model())
  File "/mnt/mozhi/ParlAI/parlai/agents/hugging_face/t5.py", line 105, in build_model
    model = ParlaiT5Model(self.opt, self.dict)
  File "/mnt/mozhi/ParlAI/parlai/agents/hugging_face/t5.py", line 285, in __init__
    self.t5 = build_t5(opt)
  File "/mnt/mozhi/ParlAI/parlai/agents/hugging_face/t5.py", line 34, in build_t5
    raise RuntimeError('Must use transformers package >= 4.3 to use t5')
[...]

PR #4164, opened 2021-11-11
[T5] Fix HF versioning issue

klshuster:
**Patch description**
Addresses issue #4148; now correctly check HF versioning

**Testing steps**
Locally tested that various versions worked; also CI for T5

### DIFF
--- a/parlai/agents/hugging_face/hugging_face.py
+++ b/parlai/agents/hugging_face/hugging_face.py
@@ -17,7 +17,10 @@ except ImportError:
     raise ImportError('Please run `pip install transformers`.')
 
 
-HF_VERSION = float('.'.join(transformers.__version__.split('.')[:2]))
+HF_VERSION = (
+    int(transformers.__version__.split('.')[0]),
+    int(transformers.__version__.split('.')[1]),
+)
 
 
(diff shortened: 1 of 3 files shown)


mojtaba-komeili (approved):
LGTM.

parlai/agents/hugging_face/hugging_face.py
```diff
@@ -17,7 +17,10 @@
     raise ImportError('Please run `pip install transformers`.')
 
 
-HF_VERSION = float('.'.join(transformers.__version__.split('.')[:2]))
+HF_VERSION = (
+    int(transformers.__version__.split('.')[0]),
+    int(transformers.__version__.split('.')[1]),
```
mojtaba-komeili, line 22: nit (question): looking at their version history it looks like they only add letter like `rc` to the third digit. Is that always the case? Should we worry about `int` transformation here ever?

mojtaba-komeili (approved):

parlai/agents/hugging_face/hugging_face.py
klshuster: i think we can worry about that later, if it becomes an issue. but good to know ahead of time

PR #4169, opened 2021-11-12
Logging token level losses at inference time

c-flaherty:
## Intro 
In this PR, I update `TorchGeneratorAgent` to support logging token level conditional probabilities and ranks at inference time when in `verbose` mode. This logging works whether the agent is using greedy, beam, or any other kind of token generation supported by the agent currently. 

Here are two examples:

`parlai dm --model-file zoo:unittest/transformer_generator2/model --truncate 1024 -v --task integration_tests:multiturn_nocandidate -ne 1 --inference beam --beam-size 3`

Screen Shot 2021-11-12 at 1 45 35 PM

`parlai dm --model-file zoo:unittest/transformer_generator2/model --truncate 1024 -v --task integration_tests:multiturn_nocandidate -ne 1 --inference greedy`

Screen Shot 2021-11-12 at 1 46 40 PM

## A brief explanation with code pointers:
The scores are here: https://github.com/facebookresearch/ParlAI/blob/8094996b09c49f436255d0232a13d3f7201c36e2/parlai/core/torch_generator_agent.py#L1135

To my understanding score is a tensor of shape (batch size, num of beams, vocab size) and score[b, i, :] for example contains the conditional probabilities of each token in vocab being the next token in the i-th beam of batch b. However, generation candidates are added to beam objects whenever an EOS token is found. Therefore, accumulating score does not really get us all the way there. We need to have each beam accumulate token probabilities of beams and store them whenever finished hypotheses are found. This requires 

(1) adding an additional parameter to the TreeSearch object to store them:
https://github.com/facebookresearch/ParlAI/blob/8094996b09c49f436255d0232a13d3f7201c36e2/parlai/core/torch_generator_agent.py#L1276-L1277

(2) updating TreeSearch:select_paths method to output token probabilities of next paths in beam: 
https://github.com/facebookresearch/ParlAI/blob/8094996b09c49f436255d0232a13d3f7201c36e2/parlai/core/torch_generator_agent.py#L1327

(3) updating store of beam token probabilities each time TreeSearch:advance method is called:
https://github.com/facebookresearch/ParlAI/blob/8094996b09c49f436255d0232a13d3f7201c36e2/parlai/core/torch_generator_agent.py#L1423-L1436

and finally (4) storing token probabilities along with candidate utterances in TreeSearch:finished parameter: 
https://github.com/facebookresearch/ParlAI/blob/8094996b09c49f436255d0232a13d3f7201c36e2/parlai/core/torch_generator_agent.py#L1438-L1450

Once, we do this, we will be able to easily able to pass through token probabilities through get_rescored_finished (the method introducing length penalty to utterance level probability) while keeping them stored alongside candidates, output them from TorchGeneratorAgent:_generate in both beam_preds_scores and in beams for free, and finally assign them to token_losses variable in TorchGeneratorAgent:eval_step, so we can output them in the same way they are outputted for examples with labels

## Tests:
`pytest tests/test_tga.py`


parlai/core/torch_generator_agent.py
```diff
@@ -875,15 +877,26 @@ def eval_step(self, batch):
...
             self._add_generation_metrics(batch, preds)
 
             # bsz x beamsize
             beam_texts: List[List[Tuple[str, float]]] = []
+            beam_generated_text_token_info: List[List[Tuple[str, float, int]]] = []
+
             for beam in beams:
                 beam_texts.append([])
-                for tokens, score in beam.get_rescored_finished():
+                if self.output_token_losses:
```
stephenroller, line 889: how is this getting set

stephenroller: should it just be always true?

parlai/core/torch_generator_agent.py
```diff
@@ -875,15 +877,26 @@ def eval_step(self, batch):
...
+                    beam_generated_text_token_info.append([])
+
+                for tokens, score, token_scores, token_ranks in beam.get_rescored_finished():
                     try:
+                        token_scores_pretty = None
+                        if self.output_token_losses:
+                            tokens_as_txt = [self.dict[int(token)] for token in tokens]
+                            token_scores_pretty = list(zip(tokens_as_txt, token_scores.tolist(), token_ranks.tolist()))
+                            beam_generated_text_token_info[-1].append(token_scores_pretty)
```
stephenroller, line 898: I'm nervous that we're doing some slow per-item gpu copies here. Can you run a benchmark of `eval_model` on some dataset before and after this change?

parlai/core/torch_generator_agent.py
```diff
@@ -894,15 +907,22 @@ def eval_step(self, batch):
         if self.rank_candidates:
             cand_choices, cand_scores = self.rank_eval_label_candidates(batch, bsz)
 
-        text = [self._v2t(p) for p in preds] if preds is not None else None
+        text = [self._v2t(pred_data[0]) for pred_data in beam_preds_scores] if beam_preds_scores is not None else None
+
+        if self.output_token_losses:
+            generated_text_token_info = [list(zip([self.dict[int(token)] for token in pred_data[0]],pred_data[2].tolist(), pred_data[3].tolist())) for pred_data in beam_preds_scores] if beam_preds_scores is not None else None
```
stephenroller, line 913: this is far too hard to read. Let's just be explicit in the loop and avoid zip madness

parlai/core/torch_generator_agent.py
```diff
@@ -1164,6 +1184,7 @@ def _generate(
         # get all finalized candidates for each sample (and validate them)
         n_best_beam_preds_scores = [b.get_rescored_finished() for b in beams]
 
+
```
stephenroller, line 1187: nit, clear

parlai/core/torch_generator_agent.py
```diff
@@ -1482,13 +1515,16 @@ def _get_hyp_from_finished(self, hypothesis_tail):
...
         for i in range(hypothesis_tail.timestep, -1, -1):
             hyp_idx.append(
                 _HypothesisTail(
                     timestep=i,
                     hypid=endback,
                     score=self.all_scores[i][endback],
                     tokenid=self.outputs[i][endback],
+                    token_scores=self.token_scores[endback,:i] if self.token_scores is not None and endback in range(self.token_scores.size()[0]) else None,
+                    token_ranks=self.token_ranks[endback,:i] if self.token_ranks is not None and endback in range(self.token_ranks.size()[0]) else None
```
stephenroller, line 1527: ditto. too hard to read

parlai/core/torch_generator_agent.py
```diff
@@ -1512,9 +1548,11 @@ def get_rescored_finished(self, n_best=None):
             number of finalized hypotheses to return
 
         :return:
-            list of (tokens, score) pairs, in sorted order, where:
+            list of (tokens, score, token_scores, token_ranks) pairs, in sorted order, where:
```
stephenroller, line 1551: Is it worth us refactoring this away from a tuple and to some sort of Named class at this point?

parlai/core/torch_generator_agent.py
```diff
@@ -1584,7 +1626,12 @@ def select_paths(self, logprobs, prior_scores, current_length):
         tok_scores, tok_ids = logprobs.max(1)
         best_scores = tok_scores + prior_scores
         hyp_ids = torch.arange(logprobs.size(0)).to(logprobs.device)
-        return (hyp_ids, tok_ids, best_scores)
+
+        vocab_size = logprobs.size()[1]
+        ranks = logprobs.scatter(1, logprobs.argsort(descending=True), torch.arange(vocab_size).type(torch.FloatTensor).to(logprobs.device)[None,:].repeat((self.beam_size,1)) + 1)
+        tok_ranks = torch.gather(ranks, 1, tok_ids[None, :]).view(-1)
+
+        return (hyp_ids, tok_ids, best_scores, tok_scores.view(-1), tok_ranks)
```
stephenroller, line 1634: ah I would prefer if we didn't have to change the signature of select_paths. Is there a way that is possible?

parlai/core/torch_generator_agent.py
c-flaherty: This is a flag that's already being used in `main` to determine whether to show token level probabilities for labels

parlai/core/torch_generator_agent.py
c-flaherty: Will do. Will something like `time parlai eval_model --task "babi:Task1k:2" -m "repeat_label"` to check how long `eval_model` took be okay, or did you have something else in mind?

parlai/core/torch_generator_agent.py
c-flaherty: Probably, yes. Before we do that though, we should probably add type hints to all of the methods?

c-flaherty: That could make things less error prone definitely

parlai/core/torch_generator_agent.py
c-flaherty: I don't think so. Since select paths is where token level probabilities are actually computed, we need some way to get them out of this method and persist them through the search process. The two approaches that come to mind are updating the type signature or have a side effect inside `select_paths` where we update a property attached to the `TreeSearch` object with the token level probabilities. If we do the latter for token probabilities, we should do it for all of these fields though, to keep these consistent. The latter option seems easier for now. There's only one callsite for `select_paths` (in `TreeSearch::advance`), so I am not clear as to why we don't want to change this signature. Why don't you want to change it?

stephenroller: Only reason I'm resistant is that it adds more overhead to implementing a novel `select_paths` implementation.

Okay, this is fine. Please refactor it to return a nice object instead of an ugly 5-tuple.

parlai/core/torch_generator_agent.py
c-flaherty: There are only two callsites for this method, and they are in the same method. I think it is okay to leave as is for now.

parlai/core/torch_generator_agent.py
c-flaherty: Not sure 

On my code:
```
> time parlai eval_model --task "babi:Task1k:2" -m "repeat_label"
...
real    0m3.068s
user    0m3.349s
sys     0m5.302s
```
On `main`:
```
> time parlai eval_model --task "babi:Task1k:2" -m "repeat_label"
...
real    0m3.394s
user    0m3.341s
sys     0m5.162s
```

stephenroller: This must be measuring noise, because `-m repeat_label` isn't hitting any of your code paths.

You need to test with an actual generative model.

c-flaherty: Updated! I test with a transformer model now and turn on verbose mode with `-v` to ensure the generative code is run

c-flaherty: It looks like you were right. There seems to be a nontrivial degradation in performance. Any tips for how to improve this? What kinds of difference in performance would you be okay with?

c-flaherty: Also, my code updates only run if `verbose` mode is on. If you really care about performance, you would probably turn this mode off anyways.

parlai/core/torch_generator_agent.py
```diff
@@ -1611,7 +1735,27 @@ def select_paths(self, logprobs, prior_scores, current_length):
...
-        return (hyp_ids, tok_ids, best_scores)
+        tok_scores = torch.gather(logprobs, 1, tok_ids[None, :]).view(-1)
+
+        vocab_size = logprobs.size()[1]
+        ranks = logprobs.scatter(
+            1,
+            logprobs.argsort(descending=True),
+            torch.arange(vocab_size)
+            .type(torch.FloatTensor)
+            .to(logprobs.device)[None, :]
```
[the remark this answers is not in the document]
emilydinan: (i'll just leave this one comment but i see you do it elsewhere too haha, I assume you'll just change it throughout)

parlai/core/torch_generator_agent.py
```diff
@@ -1655,15 +1799,34 @@ def __init__(self, k, *args, **kwargs):
...
+            .repeat((self.beam_size, 1))
+            + 1,
+        )
+        tok_ranks = torch.gather(ranks, 1, tok_ids[None, :]).view(-1)
```
emilydinan, line 1821: also here you don't actually need to go through the trouble of resorting the logprobs. `torch.topk` actually returns the values and indices of teh top-k log probs in sorted order for you already, so you can make use of that

parlai/core/torch_generator_agent.py
```diff
@@ -1698,4 +1861,23 @@ def select_paths(self, logprobs, prior_scores, current_length):
...
```
emilydinan, line 1875: same as in top-k, you are actually already sorting the probs above, so use that instead of re-sorting

parlai/core/torch_generator_agent.py
```diff
@@ -1655,15 +1799,34 @@ def __init__(self, k, *args, **kwargs):
...
+        vocab_size = logprobs.size()[1]
+        ranks = logprobs.scatter(
+            1,
+            logprobs.argsort(descending=True),
+            torch.arange(vocab_size)
+            .type(torch.FloatTensor)
+            .to(logprobs.device)[None, :]
+            .repeat((self.beam_size, 1))
+            + 1,
+        )
```
emilydinan, line 1820: i think there is actually a bug here -- say beam size is 3 and the tok ids are 1, 2, and 3 -- you are returning the rank of tokens 1,2, and 3 in the first beam. so, instead of `tok_ids.unsqueeze(0)` here you need `tok_ids.unsqueeze(1)`.

see here for example:
```
(Pdb) test  = torch.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]])
(Pdb) choice = torch.tensor([1,1,1,1])
(Pdb) test.gather(1, choice.unsqueeze(0))
tensor([[2, 2, 2, 2]])
(Pdb) test.gather(1, choice.unsqueeze(1))
tensor([[ 2],
        [ 5],
        [ 8],
        [11]])
   ```

you want the second thing here

File: parlai/core/torch_generator_agent.py [hunk 12]
```diff
@@ -1611,7 +1735,27 @@ def select_paths(self, logprobs, prior_scores, current_length):
...
+        tok_ranks = torch.gather(ranks, 1, tok_ids[None, :]).view(-1)
```
emilydinan, line 1750: i think there is a bug here, this needs to be gathered along the `hyp_id` -- see code above, the scores are flattened and then the best indexes are chosen from that flattened score. then we back out which hypothesis id the indices were taken from.

emilydinan: (also see comment below about bug in gathering! may be helpful here)

emilydinan (changes requested):
thanks for the well-documented PR description @c-flaherty and nice job so far!! this is going to be a super useful feature. 

i didn't finish reviewing but found some things that need fixing so i wanted to unblock you before EOD!

parlai/core/torch_generator_agent.py
```diff
@@ -875,34 +878,75 @@ def eval_step(self, batch):
...
+                                )
+                            )
+                            beam_generated_text_token_info[-1].append(
+                                token_scores_pretty
+                            )
+
                         beam_texts[-1].append((self._v2t(tokens), score.item()))
-                    except KeyError:
-                        logging.error("Decoding error: %s", tokens)
+                    except KeyError as e:
```
c-flaherty, line 909: Remove

parlai/core/torch_generator_agent.py
```diff
@@ -875,34 +878,75 @@ def eval_step(self, batch):
...
         if text and self.compute_tokenized_bleu:
             # compute additional bleu scores
             self._compute_fairseq_bleu(batch, preds)
         retval = Output(
             text, cand_choices, token_losses=token_losses, cand_scores=cand_scores
         )
+
         if not self.skip_generation:
             retval.beam_texts = beam_texts
+            retval.beam_generated_text_token_info = beam_generated_text_token_info
```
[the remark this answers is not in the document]
c-flaherty: 141519071-681b210c-d002-40b6-abe0-398240286201

This is an example output (also in PR description) with these changes. The existing code had two fields `text` and `beam_texts`. `text` contained the highest scoring generated utterance. `beam_texts` contained the highest scoring generated utterance from each beam. I mimicked this structure.  `generated_text_token_info` contains the token probs and ranks for the utterance in `text` and `beam_generated_text_token_info` contains teh token probs and ranks for the utterances in `beam_texts`. 

You are right though -- we don't need to reprettify it. That logic could be simplified.

parlai/core/torch_generator_agent.py
c-flaherty: Curious -- we're not copying any big param matrices. We're just copying utterances that will probably be pretty short (the number of beams will be pretty small too). What made you think this could be an issue?

parlai/core/torch_generator_agent.py
c-flaherty: Hm I'm not sure I'm following here. Will ping you and then update this once we chat about it

parlai/core/torch_generator_agent.py
```diff
@@ -1698,4 +1822,19 @@ def select_paths(self, logprobs, prior_scores, current_length):
         # Convert back to logspace.
         scores = sprobs[hyp_ids, choices].log()
         best_scores = prior_scores.expand_as(scores) + scores
-        return (hyp_ids, tok_ids, best_scores)
+
+        # need to resort because masking out of part of distribution can affect ranks
+        tok_ranks = (
```
c-flaherty, line 1827: Here we use the double argsorting trick to compute ranks. We cannot use `sinds` as the first argsort computation because after it is computed, some values may be masked to `0` in `sprobs` on line 1817. Thus, we must run both `argsort` computations. 

We want the highest probability to have the lowest rank, so we set `descending=True`.

`choices` only contains one value for NucleusSampling since there are no beams (obviously) and this code is not batched. Thus, we flatten it to get the dimensions we want.

The lowest rank should be `1` not `0`, so we `+ 1` (this is also true for the other decoding schemes).

parlai/core/torch_generator_agent.py
```diff
@@ -1611,7 +1718,17 @@ def select_paths(self, logprobs, prior_scores, current_length):
         # get the actual word id from residual of the same division
         tok_ids = best_idxs % voc_size
 
-        return (hyp_ids, tok_ids, best_scores)
+        tok_scores = torch.gather(logprobs, 1, tok_ids.unsqueeze(0)).view(-1)
+
+        tok_ranks = torch.arange(self.beam_size).to(logprobs.device) + 1
```
c-flaherty, line 1723: Observe that `topk` returns elements in sorted order so the rank of an element in this returned list is simply its index plus 1.

parlai/core/torch_generator_agent.py
```diff
@@ -1655,15 +1772,22 @@ def __init__(self, k, *args, **kwargs):
...
         scores = values[hyp_ids, choices]
         best_scores = prior_scores.expand_as(scores) + scores
-        return (hyp_ids, tok_ids, best_scores)
+
+        return _PathSelection(
+            hypothesis_ids=hyp_ids,
+            token_ids=tok_ids,
+            scores=best_scores,
+            token_scores=scores.view(-1),
+            token_ranks=choices.view(-1) + 1,
```
c-flaherty, line 1789: (pointers for review):

Observe that `topk` returns elements in sorted order so the (single) choice has a rank equal to itself.

emilydinan: thanks for the pointer!

tests/test_tga.py
```diff
@@ -171,6 +172,69 @@ def test_prefix_tokens(self):
...
+            'greedy': {
+                'text_token_info': [
+                    ('__start__', 0.0, 1.0),
+                    ('5', -2.5510462364763953e-05, 1.0),
+                    ('__end__', -1.1920922133867862e-06, 1.0),
+                ],
+                'extra_args': [],
+            },
+            # sampling based token selection will produce non-deterministic output
+            'topk': {'extra_args': ['--topk', '2']},
```
emilydinan, line 201: can we try these methods with beam_size > 1 as well

emilydinan (commented):
great! this is looking so much simpler. added a few more questions/comments!

parlai/core/torch_generator_agent.py
```diff
@@ -1420,20 +1488,26 @@ def advance(self, logprobs):
...
-        tok_id_list = tok_ids.tolist()
+        self.outputs.append(path_selection.token_ids)
+        self.bookkeep.append(path_selection.hypothesis_ids)
+        tok_id_list = path_selection.token_ids.tolist()
         self.partial_hyps = [
-            self.partial_hyps[hyp_ids[i]] + [tok_id_list[i]]
+            self.partial_hyps[path_selection.hypothesis_ids[i]] + [tok_id_list[i]]
             for i in range(self.beam_size)
         ]
+        self.token_scores = torch.cat(
```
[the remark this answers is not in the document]
c-flaherty: Won't it be a little faster this way because it keeps the score on the GPU?

parlai/core/torch_generator_agent.py
c-flaherty: Ah right, can we just do `tok_ranks = logprobs.view(-1).argsort().argsort()`?

parlai/core/torch_generator_agent.py
c-flaherty: Ah right, they will always be lower prob than chosen tokens, so they won't affect their ranks. So we can do `tok_ranks = torch.arange(sprobs.size[1]).to(logprobs.device)[choices.view(-1)] + 1`

tests/test_tga.py
c-flaherty: Sorry if I'm misunderstanding but how can we have multiple beams in vanilla topk sampling or vanilla nucleus sampling?

parlai/core/torch_generator_agent.py
```diff
@@ -1611,7 +1718,17 @@ def select_paths(self, logprobs, prior_scores, current_length):
         # get the actual word id from residual of the same division
         tok_ids = best_idxs % voc_size
 
-        return (hyp_ids, tok_ids, best_scores)
+        tok_scores = torch.gather(logprobs, 1, tok_ids.unsqueeze(0)).view(-1)
+
```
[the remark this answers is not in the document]
c-flaherty: Agreed

parlai/core/torch_generator_agent.py
```diff
@@ -1501,6 +1580,18 @@ def _get_pretty_hypothesis(self, list_of_hypotails):
         """
         return torch.stack([ht.tokenid for ht in reversed(list_of_hypotails)])
 
+    def _get_pretty_token_metadata(self, list_of_hypotails):
+        """
+        Return token probabilities and ranks as two tensors.
```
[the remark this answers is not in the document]
c-flaherty: Hm, I don't think it would be any quicker here due to the list comprehension happening in system RAM not GPU ram

tests/test_tga.py
c-flaherty: Emily and I discussed this in a meeting. The answer is no, you cannot, but ParlAI is written to allow for multiple beams in topk sampling and nucleus sampling. When there are multiple beams for either of these methods, the methods will decode sequences independently along each beam. You can think of beam size as a kind of batch size. It cannot be set for greedy decoding. This makes sense because greedy decoding always produces the same thing.

parlai/core/torch_generator_agent.py
```diff
@@ -1611,7 +1714,22 @@ def select_paths(self, logprobs, prior_scores, current_length):
         # get the actual word id from residual of the same division
         tok_ids = best_idxs % voc_size
 
-        return (hyp_ids, tok_ids, best_scores)
+        tok_scores = torch.gather(logprobs, 1, tok_ids.unsqueeze(0)).view(-1)
+
+        tok_ranks = (
```
c-flaherty, line 1719: I think this should work. We compute ranks along each beam by using double argsort along beams (dimension 1). Then, we flatten to align the output of this with `flat_beam_scores`. Finally, we use `best_idxs` to extract the relevant ranks.

parlai/core/torch_generator_agent.py
```diff
@@ -1698,4 +1823,11 @@ def select_paths(self, logprobs, prior_scores, current_length):
...
         scores = sprobs[hyp_ids, choices].log()
         best_scores = prior_scores.expand_as(scores) + scores
-        return (hyp_ids, tok_ids, best_scores)
+
+        return _PathSelection(
+            hypothesis_ids=hyp_ids,
+            token_ids=tok_ids,
+            scores=best_scores,
+            token_scores=scores.view(-1),
+            token_ranks=choices.view(-1),
```
c-flaherty, line 1832: Ranks are just choices because `sprobs` is sorted in descending order.

parlai/core/torch_generator_agent.py
```diff
@@ -1655,15 +1773,22 @@ def __init__(self, k, *args, **kwargs):
...
         scores = values[hyp_ids, choices]
         best_scores = prior_scores.expand_as(scores) + scores
-        return (hyp_ids, tok_ids, best_scores)
+
+        return _PathSelection(
+            hypothesis_ids=hyp_ids,
+            token_ids=tok_ids,
+            scores=best_scores,
+            token_scores=scores.view(-1),
+            token_ranks=choices.view(-1),
```
c-flaherty, line 1790: Ranks are just choices because `probs` is sorted in descending order.

parlai/core/torch_generator_agent.py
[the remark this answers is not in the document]
emilydinan: I don't necessarily see why it would be faster because we'll have to move it off GPU eventually and AFAICT we are not doing any operations which would be faster on GPU, nevertheless I don't have a strong opinion on this so I will defer to what you think is best here for future work

parlai/core/torch_generator_agent.py
```diff
@@ -1611,7 +1714,22 @@ def select_paths(self, logprobs, prior_scores, current_length):
         # get the actual word id from residual of the same division
         tok_ids = best_idxs % voc_size
 
-        return (hyp_ids, tok_ids, best_scores)
+        tok_scores = torch.gather(logprobs, 1, tok_ids.unsqueeze(0)).view(-1)
```
emilydinan, line 1717: there is still a bug here, because we need to select along the right hypothesis IDs

i think you need something like `tok_scores = torch.index_select(logprobs, 0, hyp_ids).gather(1, tok_ids.unsqueeze(1))`, which first selects the logprobs along the hypothesis ids, and then gathers along the specific token ids within each hypothesis id

parlai/core/torch_generator_agent.py
```diff
@@ -1611,7 +1714,22 @@ def select_paths(self, logprobs, prior_scores, current_length):
...
+
+        tok_ranks = (
+            logprobs.argsort(1, descending=True)
+            .argsort(1, descending=True)
```
emilydinan, line 1721: similarly here, the best indices need to be selected along the hypothesis ids

emilydinan (changes requested):
looks *nearly* good to go, but need one last fix in beam search.

i'll also add one last request that you add a test that would capture some of the bugs you ran into here. For example, if you initialize the `BeamSearch` object and pass a set of (fake) logprobs to `select_paths`, can you ensure that it returns you the correct token scores? (& similarly for the other tree search objects)

parlai/core/torch_generator_agent.py
c-flaherty: Ah, right. Somehow this one got lost in the cracks. What you suggested makes sense to me (adding `.view(-1)` at the end though)

parlai/core/torch_generator_agent.py
c-flaherty: Hm, I think this actually works as I did because `best_idx` gives the indices along the flattened concatenation of all of the beams in `logprobs`. It's not the same situation as what you pointed out with me using `tok_ids` when computing `tok_scores`, where the `tok_ids` represent indices along a single beam of probabilities. Since `best_idxs` represents indices along the concatenation of the beams in logprobs, we can first double arg sort logprobs along the dim 1 (along each beam), flatten the result, and then index by `best_idxs`. Am I missing something here?

parlai/core/torch_generator_agent.py
c-flaherty: @emilydinan thoughts?

parlai/core/torch_generator_agent.py
c-flaherty: The fix is I need to not have `descending=True` flag in second `argsort`. I wrote some tests to check for errors like this.

tests/test_tga.py
```diff
@@ -171,6 +179,203 @@ def test_prefix_tokens(self):
...
+                        == tok_data[0]
+                    ), f"failed token prediction for inference type {inference_type} at token {gold_data[inference_type]['text_token_info'][i][0]}"
+                    assert math.isclose(
+                        gold_data[inference_type]['text_token_info'][i][1], tok_data[1]
+                    ), f"failed token probability prediction for inference type {inference_type} at token {gold_data[inference_type]['text_token_info'][i][0]}"
+                    assert math.isclose(
+                        gold_data[inference_type]['text_token_info'][i][2], tok_data[2]
+                    ), f"failed token rank prediction for inference type {inference_type} at token {gold_data[inference_type]['text_token_info'][i][0]}"
+
+    def test_tree_search(self):
```
c-flaherty, line 249: > i'll also add one last request that you add a test that would capture some of the bugs you ran into here. For example, if you initialize the BeamSearch object and pass a set of (fake) logprobs to select_paths, can you ensure that it returns you the correct token scores? (& similarly for the other tree search objects)

Here is a test that should capture this for all the different decoding schemes w/ and w/o beams.

c-flaherty:
I couldn't get the code to go any faster after playing around with some suggestions from Stephen, so instead I moved all token level logging related code behind guard statements. This means that none of the tensor operations related to token level logging run unless verbose mode is on. As expected, this change means that running this code without verbose mode on is not any slower than running this code on `main`.

On `main`, running `CUDA_LAUNCH_BLOCKING=1  perf stat -r 10 -d parlai eval_model --task "babi:Task1k:2" --model-file "zoo:tutorial_transformer_generator/model" --optimizer adam`:
```
Performance counter stats for 'parlai eval_model --task babi:Task1k:2 --model-file zoo:tutorial_transformer_generator/model --optimizer adam' (10 runs):

          45868.78 msec task-clock                #    1.159 CPUs utilized            ( +-  1.66% )
           1986695      context-switches          #    0.043 M/sec                    ( +-  3.08% )
               146      cpu-migrations            #    0.003 K/sec                    ( +-  2.83% )
           1326937      page-faults               #    0.029 M/sec                    ( +-  0.29% )
      164034479904      cycles                    #    3.576 GHz                      ( +-  1.65% )  (49.98%)
      197481840141      instructions              #    1.20  insn per cycle           ( +-  0.41% )  (62.58%)
       41177732097      branches                  #  897.729 M/sec                    ( +-  0.48% )  (62.69%)
         414964167      branch-misses             #    1.01% of all branches          ( +-  0.57% )  (62.78%)
       61351690838      L1-dcache-loads           # 1337.548 M/sec                    ( +-  0.60% )  (62.66%)
        3185721966      L1-dcache-load-misses     #    5.19% of all L1-dcache hits    ( +-  0.70% )  (62.42%)
         204441820      LLC-loads                 #    4.457 M/sec                    ( +-  0.52% )  (49.72%)
          32770298      LLC-load-misses           #   16.03% of all LL-cache hits     ( +-  1.03% )  (49.74%)

            39.568 +- 0.758 seconds time elapsed  ( +-  1.92% )
```

On `c-flaherty:add_token_level_probability_logging_TGA`, running `CUDA_LAUNCH_BLOCKING=1  perf stat -r 10 -d parlai eval_model --task "babi:Task1k:2" --model-file "zoo:tutorial_transformer_generator/model" --optimizer adam`:
```
Performance counter stats for 'parlai eval_model --task babi:Task1k:2 --model-file zoo:tutorial_transformer_generator/model --optimizer adam' (10 runs):

          45668.64 msec task-clock                #    1.165 CPUs utilized            ( +-  1.09% )
           1804133      context-switches          #    0.040 M/sec                    ( +-  4.24% )
               139      cpu-migrations            #    0.003 K/sec                    ( +-  3.24% )
           1326329      page-faults               #    0.029 M/sec                    ( +-  0.13% )
      163236060522      cycles                    #    3.574 GHz                      ( +-  1.18% )  (49.82%)
      197468618586      instructions              #    1.21  insn per cycle           ( +-  0.55% )  (62.27%)
       41286514451      branches                  #  904.045 M/sec                    ( +-  0.59% )  (62.28%)
         423899974      branch-misses             #    1.03% of all branches          ( +-  1.77% )  (62.37%)
       61301661616      L1-dcache-loads           # 1342.314 M/sec                    ( +-  0.61% )  (62.54%)
        3163267796      L1-dcache-load-misses     #    5.16% of all L1-dcache hits    ( +-  1.22% )  (62.76%)
         204609470      LLC-loads                 #    4.480 M/sec                    ( +-  0.53% )  (50.20%)
          32725509      LLC-load-misses           #   15.99% of all LL-cache hits     ( +-  0.53% )  (50.04%)

            39.206 +- 0.522 seconds time elapsed  ( +-  1.33% )
```

On `c-flaherty:add_token_level_probability_logging_TGA`, running `CUDA_LAUNCH_BLOCKING=1  perf stat -r 10 -d parlai eval_model --task "babi:Task1k:2" --model-file "zoo:tutorial_transformer_generator/model" --optimizer adam -v`:
```
Performance counter stats for 'parlai eval_model --task babi:Task1k:2 --model-file zoo:tutorial_transformer_generator/model --optimizer adam -v' (10 runs):

          48766.12 msec task-clock                #    1.156 CPUs utilized            ( +-  1.02% )
           1852279      context-switches          #    0.038 M/sec                    ( +-  3.12% )
               145      cpu-migrations            #    0.003 K/sec                    ( +-  2.74% )
           1325857      page-faults               #    0.027 M/sec                    ( +-  0.22% )
      174661626177      cycles                    #    3.582 GHz                      ( +-  1.11% )  (50.21%)
      214397405556      instructions              #    1.23  insn per cycle           ( +-  0.42% )  (62.70%)
       45077919773      branches                  #  924.370 M/sec                    ( +-  0.43% )  (62.63%)
         432371445      branch-misses             #    0.96% of all branches          ( +-  0.99% )  (62.49%)
       67462839403      L1-dcache-loads           # 1383.396 M/sec                    ( +-  0.44% )  (62.33%)
        3388401715      L1-dcache-load-misses     #    5.02% of all L1-dcache hits    ( +-  0.62% )  (62.30%)
         210949249      LLC-loads                 #    4.326 M/sec                    ( +-  0.44% )  (49.97%)
          33144635      LLC-load-misses           #   15.71% of all LL-cache hits     ( +-  0.72% )  (50.08%)

            42.195 +- 0.429 seconds time elapsed  ( +-  1.02% )
```

emilydinan (approved):
Thanks for persisting through several rounds of edits @c-flaherty!!! This will be a super helpful change 😄 

This looks good to go from my perspective, pending tests passing. 

@stephenroller  -- did you want to take a look before merge?

tests/nightly/gpu/test_bb2.py
```diff
@@ -88,8 +88,9 @@ def test_retrieval_none(self):
         _test_bb2_rag(KnowledgeAccessMethod.NONE, n_docs=1)
 
 
-@testing_utils.skipUnlessGPU
-@unittest.skipIf(LOCAL, "Skipping Test because its slow and mem intensive")
+# @testing_utils.skipUnlessGPU
```
emilydinan, line 91: you may have left these comments in by mistake

tests/nightly/gpu/test_rag.py
```diff
@@ -110,7 +110,7 @@
 }
 
 
-@testing_utils.skipUnlessGPU
+@unittest.skip("Cache too large")
```
emilydinan, line 113: Do we actually need to skip these all of the time? Or just when they run on Circle CI? If so, you can use the decorator `skipIfCircleCI` (in `utils/testing.py`).

emilydinan (commented):
just to confirm -- do the rag tests you skip pass locally?

tests/nightly/gpu/test_bb2.py
c-flaherty: I'm just testing to right now to see if skipping these tests fixes the cache issue. If it does, then yea will remove this comment, but if it doesn't, then I'll revert these changes related to skipping tests

c-flaherty: (And address the cache issue in a separate pr)

tests/nightly/gpu/test_rag.py
c-flaherty: Ahh, yeah I'll make that change if this indeed does fix the cache issue. I'll probably have to remove all these changes to the test decorators anyways though, as they probably won't fix the cache issues. Will keep this in mind!

PR #4172, opened 2021-11-15
[BB2] FAQ

klshuster:
**Patch description**
I've included a README for BB2 with FAQ that I've seen in our issues. Hopefully we can direct people towards this if it has the answer...

### DIFF
--- a/projects/blenderbot2/agents/README.md
+++ b/projects/blenderbot2/agents/README.md
@@ -45,7 +45,13 @@ You'll need to do two things:
 1. Set `--knowledge-access-method search_only`
 2. Set `--query-generator-model-file zoo:sea/bart_sq_gen/model`
 
-### How can I train with gold documents provided?
+### How can I manually provide documents for my model to use?
+
+You may want to provide the model with "Gold" documents, or documents on which the generator *should condition*. We offer two ways of doing this:
+1. **Only Gold**: Providing the model with **only** the documents you want, bypassing search.
+2. **Insert Gold**: Mixing retrieved documents with gold documents.
+
+This may be useful both for ensuring reproducibility between experiments for retrievers with stochastic responses (e.g., the internet). For example, see the [`WizIntGoldDocRetrieverFiDAgent`](https://github.com/facebookresearch/ParlAI/blob/f1a46aad3dbae55f8a7f8aaa70b2330135c23e35/parlai/agents/fid/fid.py#L374) model with wizard of internet dataset).
 
 **ONLY GOLD DOCS**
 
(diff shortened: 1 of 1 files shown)


projects/blenderbot2/agents/README.md
```diff
@@ -0,0 +1,118 @@
...
+IndexError: index 1 is out of bounds for dimension 0 with size 1
+```
+If you notice this during training, please set `--memory-decoder-model-file ''`.
+
+### How can I use BlenderBot2 with **only** search?
+
+You'll need to do two things:
+
+1. Set `--knowledge-access-method search_only`
+2. Set `--query-generator-model-file zoo:sea/bart_sq_gen/model`
```
mojtaba-komeili, line 46: Is this one necessary? Isn't that the default atm?

projects/blenderbot2/agents/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+### How can I train with gold documents provided?
```
mojtaba-komeili, line 48: Should we add a note here that what we mean by gold docs. Maybe to keep it general enough we could say something like this:

Gold documents are any set of documents that you need you retriever to surface. We add them to make sure the generator model sees (is conditioned on) a certain set of document. Example use cases are as follows:
1- Adding documents to the mix of retrieved that you are certain have useful knowledge that the model should use for generating response.
2- Ensure reproducibility between experiments for retrievers that have randomized responses. For example,  if you wanted your model to see the same documents that was shown to crowdsourcing agent 
 (eg `WizIntGoldDocRetrieverFiDAgent` model with `wizard of internet` dataset).

mojtaba-komeili (approved):
Thanks for adding this. Reading it was useful for me too.

projects/blenderbot2/agents/README.md
klshuster: No, this is not the default. The default uses `--knowledge-access-method classify`, with the query generator from BB2 (trained to either search or retrieve from memory)

projects/blenderbot2/agents/README.md
klshuster: Great idea, i'll add that note

projects/blenderbot2/agents/README.md
mojtaba-komeili: Sorry, I was referring to `--query-generator-model-file zoo:sea/bart_sq_gen/model`. That can be default, right?

klshuster: At the moment it's not, the default is `zoo:blenderbot2/query_generator/model`, which determines whether to search or retrieve from memory

mojtaba-komeili: Oh, yeah, I remember now.

PR #4175, opened 2021-11-16
Fix WandB requeueing

c-flaherty:
**Patch description**

If a job is pre-empted and then re-queued, the current logic produces multiple runs in WandB, each corresponding a respective execution of the job. We should tell WandB to automatically group runs of the same job into a single run if a process exited unsuccessfully (i.e. was pre-empted). 

For more detail, see: https://docs.wandb.ai/guides/track/advanced/resuming.

**Testing steps**

I ran the following command twice, Ctrl+C'ing after the first one, using my WandB account to verify that it did not produce two runs, but instead joined the logs from both runs. If you want to run this yourself, you need to login to your WandB account and update ` --wandb-entity` arg to your username below.

```
parlai train_model -t personachat -m transformer/ranker --n-layers 1 --embedding-size 300 --ffn-size 600 --n-heads 4 --num-epochs 10 -veps 0.25 -bs 64 -lr 0.001 --dropout 0.1 --embedding-type fasttext_cc --candidates batch -wblog True --wandb-name a_test --wandb-project test2 --wandb-entity colinflaherty --model-file './tmp/model'
```

This is what things look like if for the code currently on `main`:
Screen Shot 2021-11-16 at 3 49 10 PM

Screen Shot 2021-11-16 at 3 49 35 PM

Screen Shot 2021-11-16 at 3 49 26 PM

This is what things look like with this PR's changes added:

Screen Shot 2021-11-16 at 3 52 55 PM

Screen Shot 2021-11-16 at 3 53 16 PM

### DIFF
--- a/parlai/core/logs.py
+++ b/parlai/core/logs.py
@@ -180,7 +180,7 @@ class WandbLogger(object):
             notes=f"{opt['model_file']}",
             entity=opt.get('wandb_entity'),
             reinit=True,  # in case of preemption
-            resume=True, # requeued runs should be treated as single run
+            resume=True,  # requeued runs should be treated as single run
         )
         # suppress wandb's output
         logging.getLogger("wandb").setLevel(logging.ERROR)
--- a/parlai/core/logs.py
+++ b/parlai/core/logs.py
@@ -184,9 +184,12 @@ class WandbLogger(object):
         )
         # suppress wandb's output
         logging.getLogger("wandb").setLevel(logging.ERROR)
-        for key, value in opt.items():
-            if value is None or isinstance(value, (str, numbers.Number, tuple)):
-                setattr(self.run.config, key, value)
+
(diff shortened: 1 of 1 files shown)


c-flaherty:
@klshuster

klshuster (approved):
thanks for the fix!

c-flaherty:
@klshuster Are you familiar with what might be causing the failing tests? I looked through the logs and didn't see anything suggesting they were related to my code changes. At the same time, they're not failing on main. Could they be flaky?

stephenroller:
Test looks unrelated. Go ahead and merge! Great fix!

c-flaherty:
> Test looks unrelated. Go ahead and merge! Great fix!

@stephenroller Thanks and sounds good! Since I don't have write access for the ParlAI repo, I can't merge. Can you merge it?

PR #4176, opened 2021-11-16
[Tod] Agents, teacher metrics, and tests for these

moyapchen:
See documentation block in `tod_agents.py`

Fairly open to documenting/clearing up more information in this code as well.

(As an aside - I'm not 100% sure if `conftest.py` is a right file to change, though I did notice that `pytest.ini` was necessary to get pytest to run.)


parlai/core/tod/tod_agents.py
```diff
@@ -0,0 +1,794 @@
...
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Agents (used for dumping data) and Teachers (for training models) related to the TOD
+conversation setup.
+
+# Usage
+
+For a given dataset, extend `TodStructuredDataParser` and implement `generate_episodes()` and `get_id_task_prefix()`. The former of these is expected to do the data processing to convert a dataset to `List[TodStructuredEpisode]`. From here, multiple inheritance can be used to define Agents and Teachers that utilize the data.
```
moyapchen, line 12: I don't get why `docformatter` isn't catching these...

parlai/core/tod/teacher_metrics.py
```diff
@@ -0,0 +1,159 @@
...
+
+class SlotMetrics(Metrics):
+    """
+    Helper container which encapsulates standard slot metrics in task oriented learning
+    (jga, slot_p, slot_r, etc).
+
+    Due to differences in dialogue representations between tasks, the input is pre-
+    parsed ground truth and predicted slot dictionaries.
+
+    The 'jga+nlg' metric assumes a balanced set of JGA and NLG scores such that
```
EricMichaelSmith, line 22: Oh hmm - can you explain this a bit more? i.e. when you say `2 * Avg(JGA, NLG_BLEU)`, do you mean the average between JGA and NLG_BLEU, further averaged over the population, i.e. `2 * sum(mean(JGA,NLG_BLEU))/num_samples`? Because if so, wouldn't that just be the same as `2 * sum((JGA+NLG_BLEU)/2)/num_samples`, i.e. `sum(JGA+NLG_BLEU)/num_samples` or `Avg(JGA + NLG_BLEU)`?

parlai/core/tod/teacher_metrics.py
```diff
@@ -0,0 +1,159 @@
...
+        2 * Avg(JGA, NLG_BLEU) = Avg(JGA + NLG_BLEU)
+    The `jga+nlg` metric assumes that `NlgMetrics` is used to calculated the other side.
```
EricMichaelSmith, line 24: Can you explain what you mean by the other side in this case?

parlai/core/tod/tod_agents.py
```diff
@@ -0,0 +1,787 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""
```
EricMichaelSmith, line 6: wow, a lot of detail here! Honestly, this might be more readable in a separate README.md so that it can get formatted in a cleaner manner

parlai/core/tod/tod_agents.py
```diff
@@ -0,0 +1,787 @@
...
+class TodStructuredDataParser(Agent):
+    """
+    Base class that specifies intermediate representations for Tod conversations.
+    """
+
+    @classmethod
+    def add_cmdline_args(
+        cls, parser: ParlaiParser, partial_opt: Optional[Opt] = None
+    ) -> ParlaiParser:
+        if hasattr(super(), "add_cmdline_args"):
```
EricMichaelSmith, line 85: Hmm Agent doesn't have `self.add_cmdline_args()`, it looks like, but presumably this check is for multiple inheritance?

tests/tod/test_tod_agents_and_teachers.py
```diff
@@ -0,0 +1,315 @@
...
+"""
+
+import unittest
+
+import copy
+import parlai.core.tod.tod_core as tod_core
+import parlai.core.tod.tod_test_utils.test_agents as test_agents
+
+
+class TestTodAgentsAndTeachersBase(unittest.TestCase):
```
EricMichaelSmith, line 18: Hmm these classes might be especially good to add docstrings for, because if these tests break it'd be useful to know what they do in order to help debug them

EricMichaelSmith (commented):
A few general comments - you can tell I started strong in the beginning and then petered out -__- Feel free to call out if there's a particular section you'd especially like feedback on! Leaving to others with more context to approve (ideally)

tests/tod/test_tod_agents_and_teachers.py
moyapchen: ...good call, thanks!

parlai/core/tod/tod_agents.py
moyapchen: Yeah, pretty much. Can add a comment to document this.

parlai/core/tod/tod_agents.py
moyapchen: That's a good call - yeah will do.

parlai/core/tod/teacher_metrics.py
moyapchen: Ahh this was a copy/pasta of a comment from something that @skiingpacman added. 

But yeah, the intent of the jga+nlg metric is to get an equal weighing between the two. In hindsight, for the purposes of open sourcing, especially since `jga+nlg` isn't being used anywhere, I'll just remove this.

PR #4178, opened 2021-11-16
[TOD] World, world metrics, script, tests

moyapchen:
See documentation in `tod_world_script.py` for usage of script overall.

This is the diff in this stack that I'm the most unsure of in terms of documentation. Reading `tod_world_script.py` is good enough to get a good "I don't know what's going on but can run stuff" understanding for this code. Some of the details about the internals (ex. the metrics setup for the world) I've tried to document, but might be a tad confusing for an outsider. 

# Test Plan
Code tested in the tests included which tests:
1. TodWorld class on its own
2. TodWorld inside of `tod_world_script`
3. Metrics for TodWorld on their own 
4. TodWorld's calculations of Metrics in simulated conversations using `tod_world_script`

I also run a quick end-to-end test with the last diff in this stack, which includes project code.


stephenroller (approved):

skiingpacman (approved):
LGTM - as requested focused mainly on comments.

PR #4180, opened 2021-11-16
[TOD][straightforward] Tod world output jsonl structure to teacher task

moyapchen:
As noted in the README, this agent takes data generated from `tod_world_script.py` and dumps it out to a teacher.

(Note that I tried setting up a regression test for this teacher, but I ran into issues getting it to save the output directory to not be something that included my local homedir name in it..)


stephenroller (approved):
Stamped

PR #4183, opened 2021-11-16
[TOD][Datasets][Easy] MetalWoz into ParlAI (User + System utterances)

moyapchen:
Code for process MetalWoz into System + User Simulator teachers

Getting it to be in the Conversations format is a pain, so I don't even try here. (It's documented this way in the User Simulator paper as well)

----------------------------
* Google SGD
* **MetalWoz**
* MSR_E2E
* Multidogo
* MultiWoz V2.2
* Taskmaster
* Taskmaster2
* Taskmaster3 (TicketTalk)

Test plan:


moyapchen:

stephenroller (approved):
Stamped

PR #4187, opened 2021-11-16
[TOD][Datasets][Easy] Taskmaster3 (TicketTalk) to TOD Structured format

moyapchen:

Sooo many datasets.

---------------------------------

* Google SGD
* MetalWoz
* MSR_E2E
* Multidogo
* MultiWoz V2.2
* Taskmaster
* Taskmaster2
* **Taskmaster3 (TicketTalk)**

Test plan:

**Patch description**

**Testing steps**

**Other information**


moyapchen:

(Same with lint)

moyapchen:
Not able to reproduce the test failure locally hmm....

stephenroller (approved):

PR #4188, opened 2021-11-16
[TOD][Datasets][Easy] Taskmaster2 to TOD Conversations format

moyapchen:

There's a legacy implementation of Taskmaster2 here, but doesn't seem to be anyone using, so clobbering.

---------------------------------

* Google SGD
* MetalWoz
* MSR_E2E
* Multidogo
* MultiWoz V2.2
* Taskmaster
* **Taskmaster2**
* Taskmaster3 (TicketTalk)

Test plan:


parlai/tasks/taskmaster2/README.md
```diff
@@ -1,5 +1,5 @@
 # Taskmaster 2
 
 Originally from the
-[Google Research Datasets](https://github.com/google-research-datasets/Taskmaster/blob/main/TM-2-2020/README.md).
+[Google Research Datasets](https://github.com/google-research-datasets/Taskmaster/blob/master/TM-2-2020/README.md).
```
moyapchen, line 4: This is indeed the correct path, interestingly enough. Not sure exactly where this changed though hmm...

moyapchen:

(Same with lint)

parlai/tasks/taskmaster2/README.md
stephenroller: Wild to see Google make the opposite push as us.

stephenroller (approved):
Stamp

parlai/tasks/taskmaster2/README.md
moyapchen: Looking at the history of that file, I think the `master->main` change came in from a regex from when we did that. (No intentional moving from `main->master` on their part.

PR #4189, opened 2021-11-16
[TOD][Datasets][Easy] Taskmaster(1) in Conversations format

moyapchen:

There's non-fb people that made edits in the original version of Taskmaster, so keep those teachers around too.
---------------
* Google SGD
   * MetalWoz
   * MSR_E2E
   * Multidogo
   * MultiWoz V2.2
   * **Taskmaster**
   * Taskmaster2
   * Taskmaster3 (TicketTalk)

   Test plan:


moyapchen:

(Same with lint)

stephenroller (approved):
Stamp

PR #4194, opened 2021-11-17
[TOD] Projects folder for tod_simulator + scripts + documentation

moyapchen:
See the long `README.md` file.

I'd tested the two mega-scripts (`tod_distributed_uber{,_multiwoz}_script.py`) on the FAIR cluster a few weeks ago and haven't since, but I have no idea how well/not well those would run outside of the FAIR cluster anyhow, so... 

(Incidentally, there's lot of repeated-looking things in the `tod_distributed..` scripts; I'm actually fairly convinced abstracting most of it out would just be another level of indirection, so didn't bother. Can if it seems egregiously confusing though.)


projects/tod_simulator/README.md
```diff
@@ -0,0 +1,79 @@
+# Task Oriented Dialogue (TOD) Script + World
+
```
stephenroller, line 2: start off with a link to the paper, the abstract, and a bibtex entry.

projects/tod_simulator/scripts/get_quick_eval_stats.py
```diff
@@ -0,0 +1,208 @@
...
+            )
+            metrics = [
+                orig_path,
+                base,
+                str(ppl),
+                str(token_em),
+                str(jga),
+                str(jga_n),
+                str(jga_e),
+            ]
```
stephenroller, line 113: eee fstring?

projects/tod_simulator/scripts/get_quick_eval_stats.py
```diff
@@ -0,0 +1,208 @@
...
+                    if "1eb" in line:
+                        base = "AllNoApiRL1"
+                    if "076" in line:
+                        base = "AllYesApiRL1"
+                    if "79a" in line:
+                        base = "SysGsgdOnly"
+                    if "c31" in line:
+                        base = "SysGood"
+                    if "ba9" in line:
+                        base = "SysAll"
```
stephenroller, line 63: I don't think this should stand

projects/tod_simulator/README.md
moyapchen: Did in https://github.com/facebookresearch/ParlAI/pull/4201

projects/tod_simulator/scripts/get_quick_eval_stats.py
moyapchen: Yeah I can change it to that lol.

Did it this way initially cause having everything on separate lines was slightly more readable when I was initially going through this code...

projects/tod_simulator/scripts/get_quick_eval_stats.py
moyapchen: Good catch

stephenroller (approved):

Issue #4195, opened 2021-11-18
Warmup scheduler is loaded from the state dict, even when initializing fresh


DrMatters:
**Bug description**
Warmup scheduler is initialized from the checkpoint states dict even when parameter "hard_reset" is set to "True". This messes up the learning rate.

**Reproduction steps**
```shell
parlai train_model -t blended_skill_talk,wizard_of_wikipedia,empathetic_dialogues --multitask-weights 1,1,1 -validation-every-n-epochs 0.05 --attention-dropout 0.0 --batchsize 16 --eval-batchsize 64 --update-freq 2 --model transformer/generator --embedding-size 1280 --ffn-size 5120 --variant prelayernorm --n-heads 32 --n-positions 128 --n-encoder-layers 2 --n-decoder-layers 12 --history-add-global-end-token end --delimiter "  " --dict-tokenizer bytelevelbpe --dropout 0.1 --fp16 True --init-model zoo:blender/blender_400Mdistill/model --dict-file zoo:blender/blender_400Mdistill/model.dict --num-epochs 3.0 --label-truncate 128 --log_every_n_secs 30 --learningrate 7e-06 --lr-scheduler reduceonplateau --lr-scheduler-patience 3 --optimizer mem_eff_adam --relu-dropout 0.0 --activation gelu --text-truncate 128 --truncate 128 --warmup_updates 100 --fp16-impl mem_efficient --gradient-clip 0.1 --skip-generation False --validation-patience 20 --validation-metric ppl --validation-metric-mode min --save-after-valid True --model-file /home/user/models/test_LR_bug_model/model -wblog True --wandb-project test_LR_bug --verbose
```

**Expected behavior**
As it can be seen from the cli args, the learning rate should be 7e-06, but in reality (as reported to WandB) it's 4e-4

**Additional context**
I've looked into the bug and proposed a solution

DrMatters:
PR: https://github.com/facebookresearch/ParlAI/pull/4196

PR #4196, opened 2021-11-18
Fix "Warmup scheduler is loaded from the state dict, even when initializing fresh"

DrMatters:
**Patch description**
This patch fixes the bug described in the issue: https://github.com/facebookresearch/ParlAI/issues/4195

**Testing steps**
I don't know much about testing, but probably we should have a test that checks the bug described in the issue above.

**Other information**
lr_scheduler.py needs a huge refactoring

### DIFF
--- a/parlai/nn/lr_scheduler.py
+++ b/parlai/nn/lr_scheduler.py
@@ -51,14 +51,11 @@ class ParlAILRScheduler(object):
         self.hard_reset = hard_reset
         self.warmup_scheduler = None
 
-    def _init_warmup_scheduler(self, optimizer, states):
+    def _init_warmup_scheduler(self, optimizer):
         """
         :param optimizer optimizer:
             Optimizer being used for training. May be wrapped in
             fp16_optimizer_wrapper depending on whether fp16 is used.
-        :param state_dict states:
-            Possible state_dict provided by model checkpoint, for restoring
-            LR state.
         """
         if self.warmup_updates > 0:
             self.warmup_scheduler = optim.lr_scheduler.LambdaLR(
--- a/parlai/nn/lr_scheduler.py
+++ b/parlai/nn/lr_scheduler.py
@@ -60,8 +60,9 @@ class ParlAILRScheduler(object):
         if self.warmup_updates > 0 and (
             updates_so_far < self.warmup_updates or hard_reset
         ):
+            last_epoch = -1 if updates_so_far == 0 else updates_so_far
             self.warmup_scheduler = optim.lr_scheduler.LambdaLR(
-                optimizer, self._warmup_lr, last_epoch=updates_so_far
+                optimizer, self._warmup_lr, last_epoch=last_epoch
             )
 
     def get_last_lr(self):
--- a/parlai/nn/lr_scheduler.py
+++ b/parlai/nn/lr_scheduler.py
@@ -284,12 +284,12 @@ class ParlAILRScheduler(object):
             )
             hard_reset = True
 
+        # setup warmup scheduler after loading saved scheduler
+        scheduler._init_warmup_scheduler(optimizer)
+
         if not hard_reset:
             # do the actual loading (if possible)
             scheduler.load_state(states)
-
-        # setup warmup scheduler after loading saved scheduler
-        scheduler._init_warmup_scheduler(optimizer, states)
 
         return scheduler
 
--- a/parlai/nn/lr_scheduler.py
+++ b/parlai/nn/lr_scheduler.py
@@ -38,12 +38,6 @@ class ParlAILRScheduler(object):
         Initialize warmup scheduler. Specific main schedulers should be initialized in
         the subclasses. Do not invoke this method diretly.
 
-        :param optimizer optimizer:
-            Optimizer being used for training. May be wrapped in
-            fp16_optimizer_wrapper depending on whether fp16 is used.
-        :param state_dict states:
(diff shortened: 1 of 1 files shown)


DrMatters:
Hello! Is CI "cleaninstall_37" working properly on the current main?
It ends with "error: importlib-metadata 4.2.0 is installed but importlib-metadata>=4.4 is required by {'markdown'}", doesn't seem to be related to my changes

stephenroller:
Yeah you can ignore clean install.

Can we talk more about testing steps? What have you tried? (Can you successfully resume both during and after warmup, as well as starting fresh?)

DrMatters:
@stephenroller I've tested the code using some cases as you asked me to. This helped me to catch some bugs, but now everything looks fine.

DrMatters:
Except for the tests themselves:
https://github.com/facebookresearch/ParlAI/blob/main/tests/test_lr_schedulers.py#L205
`if 'warump_updates' in kwargs:`
There's a typo in the word 'warmup', which renders all "warmup" tests inactive.
If I correct the typo this makes a whole bunch of tests in the "lr_schedulers" fail. It's up to you.

DrMatters:
I also suggest refactoring the "lr_scheduler.py" base class to use the PyTorch's either ChainedScheduler or SequentialLR instead of reimplementing the existing functionality.
It may help make warmup schedulers more comprehensible and easier to handle (storing/loading). What do you think of this?

stephenroller:
Can you fix the typo and also try running things without your changes and fixing the typo?

DrMatters:
@stephenroller 

> Can you fix the typo and also try running things without your changes and fixing the typo?

I've created a PR and an Issue for this: #4242 #4243 .
Can you review this PR independently from the issue with tests that I've found?

stephenroller:
I think fixed tests need to be part of the PR. This has always been a tricky part of code.

DrMatters:
> I think fixed tests need to be part of the PR. This has always been a tricky part of code.

Fixed tests make 80% of tests fail, even on the current 'main' branch, you can check it here #4242. This is because now the code for warmup is slightly incorrect. So if I fix tests, I'll have to fix all LR schedulers, which is far beyond the scope of this PR, which is related to #4195.

If you want to, I can try to rewrite schedulers so that they work properly and are less hard to understand, but as I've pointed this is a huge refactoring, not related to the issue.

DrMatters:
Let me clarify things as I see them.
I've made a fix for the issue #4159 `Warmup scheduler is loaded from the state dict, even when initializing fresh`.
All tests (except for the cleaninstall_37, that you told to ignore) are passing.
My current PR is dedicated to fix this issue.

There is another issue (#4242) I discovered while working on this PR. This issue is somewhat related to this PR, but fixing it is a whole different topic. It consist of 2 parts:
1) Fixing the typo in LR scheduler tests. I did it here: #4242.
2) Fixing a bunch of bugs which was exposed by the fixed tests. I think it may include huge refactoring including rewriting all schedulers to work with [SequentialLR](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.SequentialLR.html#torch.optim.lr_scheduler.SequentialLR) or [ChainedScheduler](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ChainedScheduler.html#torch.optim.lr_scheduler.ChainedScheduler) instead of custom implementations. This is far beyond the scope of this PR.

Do you think that it's possible to merge this PR as it is and create separate issues (like #4242) for the other problems?

meganung:
This PR doesn't seem to affect the tests at all- fixing the typo in the test results in 8 failed tests in `test_lr_schedulers.py` without the changes in the PR and with these changes as well.

stephenroller:
I'm not sure to be honest. I haven't had time to look at this deeply.

I respect that our broken tests aren't your problem, but I'm  nervous about changing the behavior of any of this without strong verification of correctness. At best we are keeping a second bug and fixing one; at worst we are replacing one buggy behavior with a DIFFERENT buggy behavior, which I disfavor for stability reasons.

The refactor may be an option because of newer pytorches. Historically it wasn't because LR schedulers load their LR from state_dict (since torch 1.7 I believe) which prevented us from using the clean abstractions, as changing LR mid flight is a pretty common thing to do.

So without tests or graphs, I'm left without confidence in either implementation, and so feel inclined to just hold until we can figure out exactly how the tests are broken.

I've asked @meganung to see if she can identify the root cause this week. Of course, if you want to help, any analysis (or a fix!) you provided would speed things up. Thanks for your patience!

DrMatters:
@meganung,
Integration tests are passing with changes described in #4242 (I've maybe warmup scheduler to warmup until 1.0 multiplier), but unit tests start to fail

DrMatters:
@meganung,
@stephenroller,
I've fixed lr_schedulers by updating [def _is_lr_warming_up](https://github.com/facebookresearch/ParlAI/blob/main/parlai/nn/lr_scheduler.py#L82) (so they pass tests) in my PR #4242 , but it is a kludge meant to help to identify the root cause of the incorrect behavior. Using the kludge, you can see and verify that the problem is coupled with the moment of switching the warmup scheduler to the main scheduler.

With the changes presented there (#4242), this PR (#4196) is also passing tests succesfully

DrMatters:
Haven't you noticed before that the warmup is not finished properly? The target LR for this plot is 7e-6, but it only reaches 6.93e-6 due to warmup discontinuing one step earlier

PR #4197, opened 2021-11-18
Curated response generators

mojtaba-komeili:
**Patch description**
Adds two functions for generating canned curated responses that may replace an agent utterances. This may happen in the beginning of a conversation to initiate a good topic (`generate_init_prompot` function), or when the safety classifier detects an inappropriate dialogue (`generate_safe_response` changes the topic to something safer).

**Testing steps**
**WILL ADD IN THE NEXT COMMIT**


melanierk (approved):

parlai/utils/curated_response.py
```diff
@@ -0,0 +1,79 @@
...
+    "I know many people are intereseted in",
+    "I think it is excting to talk about",
+)
+
+TOPICS = (
+    "new marvel movies.",
+    "tonight game.",
+    "next year Oscar.",
+    "traveling to exotic destinations.",
+    "read new books.",
```
stephenroller, line 66: "Have you heard anything about read new books."

parlai/utils/curated_response.py
```diff
@@ -0,0 +1,79 @@
...
+    "I don't find that topic interesting.",
+    "Let's talk about something else.",
+)
+
+TEMPLATE = (
+    "Have you heard anything about",
+    "I was recently read about",
+    "I am really excited about",
+    "I know many people are intereseted in",
+    "I think it is excting to talk about",
```
stephenroller, line 58: typos

parlai/utils/curated_response.py
mojtaba-komeili: Yeah, it may generate non-sense like that too. I will have a second look and change them for a better template.

mojtaba-komeili:
Thanks to @melanierk, we have an extended version of the safe responses sentences. Looking at some of them they look a lot more natural than before, although there is still change of having weird combinations, but it seems much less likely.

stephenroller (approved):

PR #4204, opened 2021-11-22
WoI + Wow Mutator cleanup

jaseweston:
- moved to mutators.py files
- moved "wow" and "woi" always to the front of mutator names, it was inconsistent sometimes at front or back
- new mutator to convert to "woi" format, so then you can e.g. use all the woi mutators for wow data


parlai/tasks/wizard_of_internet/mutators.py
```diff
@@ -0,0 +1,255 @@
...
+            else:
+                pass
+        return out_episodes
+
+
+def chunk_docs_in_message(message, chunk_sz):
+    if CONST.RETRIEVED_DOCS not in message:
+        return message
+    new_message = message.copy()
+    docs = message.get(CONST.RETRIEVED_DOCS)
```
mojtaba-komeili, line 135: nit: maybe it's not a bad idea to use `message[CONST.RETRIEVED_DOCS]` to make the code crash here, if this mutator was accidentally used with other teachers. Seeing why it goes wrong later might be a bit harder for someone who doesn't know these mutators to debug.

parlai/tasks/wizard_of_internet/mutators.py
```diff
@@ -0,0 +1,255 @@
...
+        # Guarantees that checked sentences are not split in half (as we split by space).
+        for i in range(len(checked_sentences)):
+            d = d.replace(checked_sentences[i], "||CHECKED_SENTENCE_" + str(i) + "||")
+        while True:
+            end_chunk = d.find(' ', chunk_sz)
+            if end_chunk == -1:
+                # last chunk
+                for i in range(len(checked_sentences)):
+                    d = d.replace(
+                        "||CHECKED_SENTENCE_" + str(i) + "||", checked_sentences[i]
```
mojtaba-komeili, line 162: same as above for the string.

parlai/tasks/wizard_of_internet/mutators.py
```diff
@@ -0,0 +1,255 @@
...
+    ) -> ParlaiParser:
+        parser.add_argument(
+            '--woi-doc-chunk-size',
+            default=500,
+            type=int,
+            help='Document chunk size (in characters).',
+        )
+
+    def message_mutation(self, message: Message) -> Message:
+        chunk_sz = self.opt.get('woi_doc_chunk_size', 500)
```
mojtaba-komeili, line 202: Given the mistakes that we had with setting flags before, may be we should have this as something like 

```.py
if not message.get('woi_doc_chunk_size'):
    logging.warning("The --woi-doc-chunk-size flag is received as None, using the default value of X")
```

It has happened to me that I set a flag without knowing that it was ignored. This at least warns about issue like that.

parlai/tasks/wizard_of_internet/mutators.py
jaseweston: ok

parlai/tasks/wizard_of_internet/mutators.py
```diff
@@ -0,0 +1,255 @@
...
+    )
+    for i in range(len(checked_sentences)):
+        checked_sentences[i] = checked_sentences[i].lstrip(' ').rstrip(' ')
+    if ' '.join(checked_sentences) == CONST.NO_SELECTED_SENTENCES_TOKEN:
+        checked_sentences = []
+    for ind in range(len(docs)):
+        d = docs[ind]
+        # Guarantees that checked sentences are not split in half (as we split by space).
+        for i in range(len(checked_sentences)):
+            d = d.replace(checked_sentences[i], "||CHECKED_SENTENCE_" + str(i) + "||")
```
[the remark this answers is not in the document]
jaseweston: fyi in my opinion, asking for these kind of nit changes just increases the chance of already working code suddenly becoming buggy by accident for not much gain, + taking time to check if it still works when we're trying to like do some research, but i can try..

parlai/tasks/wizard_of_internet/mutators.py
```diff
@@ -0,0 +1,255 @@
...
+
+        keep = torch.randperm(len(docs))[0:max_chunks]
+        remove = torch.ones(len(docs))
+        remove[keep] = 0
+
+        for i in range(len(docs)):
+            if remove[i] == 0:
+                new_docs.append(docs[i])
+            else:
+                # We may still keep the doc if it contains the gold checked sentence(s).
```
mojtaba-komeili, line 239: In that case should we mark one of the other docs to be removed instead?

mojtaba-komeili (approved):

parlai/tasks/wizard_of_internet/mutators.py
jaseweston: well, what we actually need to do is fix the bug in parlai .. i was hoping kurt would do this...

parlai/tasks/wizard_of_internet/mutators.py
jaseweston: we'll just have one less, i dont think it's a big deal

PR #4205, opened 2021-11-22
[NQ] Hard Code # Samples

klshuster:
**Patch description**
It takes forever to load and count the number of samples of NQ for each split; we're just hardcoding now.

**Testing steps**
Confirmed `parlai dd` works


jaseweston (approved):
thank you!

Issue #4202, opened 2021-11-20
BlenderBot2 - Self chat - no search -  KeyError: 'full_text'


louis030195:
**Bug description**

I'm trying to run self-chat of BlenderBot2 on Google Colab, it fails.

**Reproduction steps**

```bash
echo "Hi, my name is Bob" >> foo.txt
echo "Hi, my name is Alice" >> foo.txt
```

```bash
!pip install parlai fairseq transformers
```

```bash
!parlai self_chat --model projects.blenderbot2.agents.blenderbot2:BlenderBot2FidAgent --model-file zoo:blenderbot2/blenderbot2_400M/model --task empathetic_dialogues --num-self-chats 1000 --display-examples True --knowledge_access_method "memory_only" --search_server None --seed_messages_from_file="./foo.txt" --outfile="./out"
```

**Expected behavior**

Run self chat using my messages as seed

**Logs**
Please paste the command line output:

```
/usr/local/lib/python3.7/dist-packages/nltk/twitter/__init__.py:20: UserWarning: The twython library has not been installed. Some functionality from the twitter package will not be available.
  warnings.warn("The twython library has not been installed. "
17:32:44 | Overriding opt["model_file"] to /root/ParlAI/data/models/blenderbot2/blenderbot2_400M/model (previously: /checkpoint/kshuster/projects/knowledge_bot/kbot_memfix_sweep25_Fri_Jul__9/338/model.oss)
17:32:44 | Overriding opt["task"] to empathetic_dialogues (previously: None)
17:32:44 | Overriding opt["knowledge_access_method"] to memory_only (previously: classify)
17:32:44 | Using CUDA
17:32:44 | loading dictionary from /root/ParlAI/data/models/blenderbot2/blenderbot2_400M/model.dict
17:32:44 | num words = 50264
17:32:44 | BlenderBot2Fid: full interactive mode on.
17:33:31 | Creating the search engine retriever.
17:33:31 | No protocol provided, using "http://"
Some weights of the model checkpoint at bert-base-uncased were not used when initializing BertModel: ['cls.predictions.decoder.weight', 'cls.seq_relationship.weight', 'cls.predictions.transform.dense.weight', 'cls.predictions.transform.LayerNorm.bias', 'cls.predictions.bias', 'cls.seq_relationship.bias', 'cls.predictions.transform.LayerNorm.weight', 'cls.predictions.transform.dense.bias']
[...]

PR #4207, opened 2021-11-22
[RAG] Handle Self Chat

klshuster:
**Patch description**
Update RAG to ensure it handles self-chat. This primarily involves ensuring that the `query_key` is in the observation before accessing it.

Addresses #4202

**Testing steps**
Added test, confirmed it failed pre-fix and works post-fix


jxmsML:
btw, I've notice the  KeyError when running interactive (in human-model chat, not selfchat) before, but unable to reproduce.

jxmsML (approved):
LGTM!

klshuster:
> btw, I've notice the KeyError when running interactive (in human-model chat, not selfchat) before, but unable to reproduce.

indeed it's a related error

PR #4218, opened 2021-11-30
Save conversations check for parent directory

mojtaba-komeili:
**Patch description**
Saving the conversation log was crashing if the parent directory of the provided path to save the file didn't exist. This patch checks for the dir that is going to contain the output file. It creates the parent directories if they do not exist.

**Testing steps**
Had a script that crashed while trying to save the data. The problem was solved after adding this code. It created the parent directory and saved the data.


parlai/utils/conversations.py
```diff
@@ -276,6 +276,14 @@ def _get_path(datapath):
...
         return fle + '.jsonl'
 
+    @staticmethod
+    def _check_parent_dir_exits(datapath):
+        parent_dir = os.path.dirname(datapath)
+        if os.path.exists(parent_dir) and os.path.isdir(parent_dir):
+            return
+        logging.info(f'Parent directory ({parent_dir}) did not exist and was created.')
+        os.makedirs(parent_dir)
+
```
stephenroller, line 286: needs to be migrated to pathmanager, thx

mojtaba-komeili:
@stephenroller switched to `PathManager`.

stephenroller (approved):
Lgtm. Prolly don't need the log but it's fine too


Moya has a separate lint pr

PR #4221, opened 2021-12-01
[BlenderBot2] A hybrid mode (skip search when needed)

jxmsML:
**Patch description**
Pass a new batch field `skip_search` to classify_retrieval: to skip search if that field is true.

This can be useful for example, when gold knowledge is prepended to the text.

**Testing steps**

**Other information**


projects/blenderbot2/agents/sub_modules.py
```diff
@@ -181,7 +181,7 @@ def __init__(self, opt: Opt):
             )
             assert isinstance(base_agent, TorchAgent)
             self.agents = [base_agent]
-            bsz = max(opt.get('batchsize', 1), opt.get('eval_batchsize', 1))
+            bsz = max(opt.get('batchsize') or 1, opt.get('eval_batchsize') or 1)
```
mojtaba-komeili, line 184: Oddly, I was getting this error when trying this agent. This change was a solution to avoid it. But maybe we should 

Screen Shot 2021-12-01 at 11 24 42 AM

projects/blenderbot2/agents/blenderbot2.py
```diff
@@ -894,6 +914,13 @@ def build_model(self) -> Union[BlenderBot2FidModel, T5BlenderBot2FidModel]:
         return model
 
 
+class BlenderBot2SearchQueryFiDAgent(BlenderBot2FidAgent):
```
jxmsML, line 917: 👍

projects/blenderbot2/agents/sub_modules.py
jxmsML: hmm interesting, I thought anything return from opt.get(XXX, 1) would never be None

klshuster: it can be `None` if the arg is in `opt` but no value is given

klshuster (approved):
cool, looks good to me!

parlai/tasks/wizard_of_internet/agents.py
```diff
@@ -443,10 +452,16 @@ def teacher_setup_data(self, datafile) -> Message:
...
                 )
+                if self.add_skip_search_if_gold_prepended:
+                    message[CONST.SKIP_SEARCH] = True
             yield message, episode_started
 
 
 class WizardDialogGoldKnowledgeTeacher(WizardDialogTeacher):
+    def __init__(self, opt, shared=None):
+        super().__init__(opt, shared=shared)
+        self.id = 'WizardDialogGoldKnowledgeTeacher'
```
jxmsML, line 463: ah thanks for adding this, otherwise the train_model also shows this red flag ;)

PR #4224, opened 2021-12-01
Fix cleaninstall test breakage

moyapchen:
...gonna guess this might have to be done again at some point, but `python setup.py develop` runs cleanly in a new conda environment when I use this.

(Going to guess it'll involve changing all the `==` to `<=` or something. Guessing something changed `setuptools` under the hood at some point maybe... or the dependencies in the other requirements got upgraded enough where some of the documented undefined behavior about how it resolves versioning that setuptools has is biting us. Figuring that out is gonna take much longer than this brute force so w/e.)

**Testing steps**
Make a bunch of clean conda envs and finagle with requirements until everything installs cleanly lol.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,6 @@ ipython==7.19.0
 torch>=1.4.0
 joblib==0.14.1
 nltk==3.6.4
-numpy==1.17.5
 omegaconf~=2.1.1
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
ty moya!!!!

PR #4236, opened 2021-12-09
Fix Crowdsourcing Tests after mephisto hydra upgrade

meganung:
**Patch description**
After hydra upgrade, blueprints are specified in yaml files so instead of specifying the blueprint, we point to the right yaml file (which specifies the blueprint) and remove any overrides that aren’t necessary (e.g. those that are defined in the yaml file or defaults in the blueprint).  Some overrides are still there since they do indeed override values in the yaml file or the default ones in the blueprint. 

**Testing steps**
`pytest tests/crowdsourcing/tasks`


EricMichaelSmith (approved):
Oh wow - this is fantastic! This is gonna make it so much easier for us to make new tests or change existing ones :) Thanks so much for making this change!

PR #4238, opened 2021-12-09
[BB2] Handling increase in batch size from dynamic batching

mojtaba-komeili:
**Patch description**
Dynamic batching in ParlAI increases the batch size. This causes some issues with empty `doc_score` place-holders that BB2 uses: they don't get filled and remain empty, which results in run-time error during tensor operations. This patch filters the empty `doc_score` place-holders.

### DIFF
--- a/projects/blenderbot2/agents/modules.py
+++ b/projects/blenderbot2/agents/modules.py
@@ -437,6 +437,9 @@ class BlenderBot2RagModel(RagModel):
             input_lengths = input_lengths.repeat_interleave(
                 input_turns_cnt, dim=0
             )  # type: ignore
+
+        # Filtering empty doc_scores added due to dynamic batching (if used)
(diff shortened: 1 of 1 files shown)


klshuster:
could you elaborate why there are "empty" doc scores?

mojtaba-komeili:
> could you elaborate why there are "empty" doc scores?
> 
> edit: nvm, think I understand. does this not mess up the scores computations down the road?

Do we even use the empty ones even. I thought they were ignored. I don't really know much about how dynamic batching works.

stephenroller (approved):

stephenroller:
https://parl.ai/docs/tutorial_worlds.html see here for an explanation of dynamic batching

mojtaba-komeili:
> https://parl.ai/docs/tutorial_worlds.html see here for an explanation of dynamic batching

Actually reading that helped me find the bug, but I am not sure if ignoring the empty batches is safe for BB2.

PR #4239, opened 2021-12-09
[LIGHT] AIMOY

klshuster:
**Patch description**
Code and models. Mostly looking for comments on the respective READMEs.

**Testing steps**
Added CI, with both local tests for pre-trained models and CI tests for code functionality; both work as expected.

CI Tests:
Screen Shot 2021-12-09 at 5 49 47 PM

W/ Pre-trained Models:

Screen Shot 2021-12-09 at 5 49 55 PM


mojtaba-komeili:
WizInt again!?

JackUrb (approved):
Wow this is a lot of agents. I found the documentation really thorough on how each of them can be extended and used in other contexts in the future, so this looks really good to me!

As a nitty aside, it seems `autoformat.sh` really wants to make a lot of changes, could be worthwhile to include, but is likely just dropping spaces out or something.

stephenroller (approved):
Aimoy, matey

PR #4242, opened 2021-12-10
Fix lr scheduler tests & lr schedulers

DrMatters:
**Patch description**
This patch fixes a wrong parameter name in LR schedulers test

**Testing steps**
This is a fix for tests

**Other information**
Seems, like warmup implementation for all schedulers has never been tested. This was because the testing code was unreachable due to the wrong parameter name ('warump_updates'). After fixing the parameter of the test, some tests started to fail.

### DIFF
--- a/parlai/nn/lr_scheduler.py
+++ b/parlai/nn/lr_scheduler.py
@@ -86,7 +86,7 @@ class ParlAILRScheduler(object):
         return (
             hasattr(self, 'warmup_scheduler')
             and self.warmup_scheduler is not None
-            and self._number_training_updates <= self.warmup_updates
+            and self.warmup_scheduler.get_last_lr()[0] < 1.0
         )
 
     def _warmup_lr(self, step):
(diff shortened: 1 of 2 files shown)


DrMatters:
Have you noticed before that the warmup is not finished properly? The target LR for this plot is 7e-6, but it only reaches 6.93e-6 due to warmup discontinuing one step earlier

DrMatters:
I've fixed lr_schedulers by updating def _is_lr_warming_up (so they pass tests), but it is a kludge meant to help to identify the root cause of the incorrect behavior. Using the kludge, you can see and verify that the problem is coupled with the moment of switching the warmup scheduler to the main scheduler.

stephenroller (approved):
Awesome, this is great. Thank you so much.

I'm going to smash land, and then if you just rebase your other PR, I'll be happy to accept it.

Thanks so much for your hard work!!!

PR #4256, opened 2021-12-15
ConversationTeacher parent class changed 

mojtaba-komeili:
**Patch description**
Changed`ConversationTeacher`'s parent class to `ParlAIDialogTeacher`. 

**Testing steps**
1- Standard teacher tests in ParlAI that rely on `ConversationTeacher`. 
2- `parlai dd` some internal classes that use this teacher


parlai/core/teachers.py
```diff
@@ -1611,12 +1611,12 @@ def setup_data(self, datafile):
                 yield act, next_episode_new
 
 
-class ConversationTeacher(FixedDialogTeacher):
+class ConversationTeacher(ParlAIDialogTeacher):
```
stephenroller, line 1614: I think it might need to be DialogTeacher, not ParlAIDialogTeacher

mojtaba-komeili: What makes `DialogTeacher` preferable to `ParlAIDialogTeacher` here?

stephenroller: ParlAI DT should be assuming a particular format. It might be that we successfully override everything but really we shouldn't.

mojtaba-komeili:
@stephenroller I changed the base class to `DialogTeacher`. There were more changes and I ended up directly changing `JsonTeacher` as well. Teacher tests pass and I tried to run tasks that were using these teachers.
There is going to be a separate PR on adapting one of the heavily customized teachers to the new format. But the rest should be fine.

parlai/core/teachers.py
```diff
@@ -1719,27 +1704,24 @@ def _setup_data(self, path):
             if self.label_turns in ['firstspeaker', 'both']:
                 eps = self._get_ep_from_turns(turns[::2], turns[1::2])
                 if eps:
-                    self.episodes.append(eps)
-                    self.num_exs += len(eps)
+                    for example, example_begins in self._return_episode_examples(eps):
```
stephenroller, line 1707: looks good

parlai/core/teachers.py
```diff
@@ -1719,27 +1704,24 @@ def _setup_data(self, path):
...
+                        yield example, example_begins
 
             # train on even turns as labels (turns w/ second speaker)
             if self.label_turns in ['secondspeaker', 'both']:
                 eps = self._get_ep_from_turns(turns[1::2], turns[2::2])
                 if eps:
-                    self.episodes.append(eps)
-                    self.num_exs += len(eps)
+                    for example, example_begins in self._return_episode_examples(eps):
```
stephenroller, line 1714: looks good!

stephenroller (approved):

PR #4264, opened 2021-12-20
changes for beam forcing, in-progress

Alex-Gurung:
Two new parameters for use with the common sense world builder:

`force-start-list`
* Forces all beams created from beam searches to begin with a given list of tokens
* At each appropriate index sets the likelihood of those tokens to infinity
* In the future could be converted to taking in text instead of tokens 

`beam-block-list-strings`
* In-progress, still requires modifications to work properly
* Prevents beam search from using any string (and modifications of those strings) in the provided list by adding it to a SearchBlocklist used at generation time
* Should be modifiable - as the opt changes in-between searches the block list and future generations changes as well


parlai/core/torch_generator_agent.py
```diff
@@ -567,11 +581,13 @@ def _v2t(self, vec):
         new_vec = []
         if hasattr(vec, 'cpu'):
             vec = vec.cpu()
+        start = True
```
mojtaba-komeili, line 584: I don't understand why we need `start`. Is that because the first index can be `END_IDX`? Could you explain in a comment?

mojtaba-komeili:
Thanks for adding this. It look good. I also add @stephenroller here for more comments. Here are mine:
1. Your force start list works great if there is only one token you want to add. How about when there are multiple ones in the list? The current code seems to add them all in a random order.
2. What happens if the start list strings are longer than a full token list of the tokenizer? (might be because it is not in the list of tokens).
3. Make sure you address lint errors.

klshuster:
I won't pretend to know what is fully going on in this PR, but have you explored possibly using the `prefix_tokens` functionality built into `TorchGeneratorAgent._generate`? You can [override an agent to return a list of prefix tokens](https://github.com/facebookresearch/ParlAI/blob/66c71e01b79b118a717c1115fecea5e99b12429a/parlai/core/torch_generator_agent.py#L1051), which are enforced by `_generate` [to start each generation](https://github.com/facebookresearch/ParlAI/blob/66c71e01b79b118a717c1115fecea5e99b12429a/parlai/core/torch_generator_agent.py#L1136-L1144)

stephenroller:
Yeah I haven't read the code but from the description, it seems like we should either use prefix tokens or generalize it appropriately.

mojtaba-komeili:
Yeah, looks like `prefix_tokens` can solve the first problem here. 

(a cleaner implementation of) The blocklist part may still be needed though. To provide some context, we have a script that calls the model to generate a sequence multiple times. Each time, depending on the previously generate sequence, there is a new list of words to avoid. We might be able to use `--beam-block-list-filename`, but passing the sequence directly to the model makes things more convenient. 
@Alex-Gurung maybe you can make sure that your blocklist and `--beam-block-list-filename` share most of their code and the block list is their union (?).

parlai/core/torch_generator_agent.py
Alex-Gurung: Think this was leftover by some testing, removed

Alex-Gurung:
New changes:
* Removed `beam-block-list-strings` for now since it's not necessary for current worldbuilder
* Renamed `force-start-list` to `opt-prefix-tokens` and reorganized to use prefix_tokens method
Note: Currently it uses only one of subclass-prefix-tokens or opt-prefix-tokens depending on which is set, defaulting to opt-prefix-tokens as that would need to be explicitly set by end-users

Having both is useful in the case that you want to change or add to prefix tokens on pre-trained models/agent types without needing to add an entire intermediary class, but there may be a better way of distinguishing between the two methods.

parlai/core/torch_generator_agent.py
```diff
@@ -1059,6 +1065,36 @@ def get_prefix_tokens(self, batch: Batch) -> Optional[torch.LongTensor]:
...
+        """
+        prefix_toks = self.opt.get("opt_prefix_tokens", [])
+        if len(prefix_toks) == 0:
+            return None
+        bsz = batch.batchsize
+        dev = batch.text_vec.device
+        prefix_toks_batch = [prefix_toks for _ in range(bsz)]
+        return torch.LongTensor(prefix_toks_batch).to(dev)
+
+    def get_all_prefix(self, batch: Batch) -> Optional[torch.LongTensor]:
```
mojtaba-komeili, line 1085: IDK, this looks to make the code a bit patchy for general use. Maybe we should keep this in a branch for our own project and not merge it into ParlAI, until we have a cleaner solution. In other words, just keep this branch updated (with rebase) with the main branch and use it for your project.

PR #4269, opened 2021-12-21
adjust data licenses for WOW, WOI, ED, and BST

melanierk:
**Patch description**
Updating our data licenses to CC-BY (i.e., commercial use allowed) for several popular datasets that were either CC-BY-NC or not explicitly licensed. Have the appropriate approvals internally to do this.

**Testing steps**
Linter/automatic tests.

**Other information**
Let me know if anyone has opinions on replicating the data licenses vs having them in a shared place. I chose replicate because I think we still don't have consistency across all datasets and putting licenses in a shared place might be confusing, but it does duplicate a lot of text.


EricMichaelSmith (approved):
Yeah this is great! Hmm, I imagine that replicating the license per-task is not the best long-term solution, yeah, but I agree that it's a good stopgap for now

stephenroller (approved):
Whoo

klshuster (approved):
🚀

PR #4275, opened 2021-12-23
Model chat: space out annotation buckets

EricMichaelSmith:
**Patch description**
- For the `model_chat` crowdsourcing task, if the text of the annotation categories is very long, you can set the `"show_line_breaks"` field (see `task_config/annotations_config.json`) to `true` in order to put a line break between the checkbox and text string of each category. This was useful recently when we had a test with very long questions for each checkbox.

**Testing steps**
Test command:
```
python parlai/crowdsourcing/tasks/model_chat/run.py \
mephisto.architect.port=3005 \
mephisto.task.maximum_units_per_worker=1000000 \
mephisto.task.task_name=${TASK_NAME}
```

`"show_line_breaks": false`:
Screen Shot 2021-12-23 at 4 54 37 PM

`"show_line_breaks": true`:
Screen Shot 2021-12-23 at 4 55 40 PM


meganung (approved):
ah this is a great option that improves the UX nicely!

PR #4283, opened 2022-01-04
Update receive_metrics for all agents including teachers

skiingpacman:
Summary:
"There's a method called receive_metrics that Agents get during
validation (mostly for lowering LR on plateau), but Teachers do not.
However, I've wanted to have teachers receive that for some time. That
would be the ideal time for you to adjust ratios [in the training data].
(I've wanted to do it for dynamic sampling during multitask learning in
order to start undersampling when one dataset overfits)" [S. Roller]

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

**Patch description**

**Testing steps**

**Other information**


stephenroller:
Oof didn't mean to drop this

PR #4289, opened 2022-01-06
[BB2] loading BB2 without skip-search-key flag.

mojtaba-komeili:
**Patch description**
Trying to load a BB2 model, because the new flag is not in the main opt, it was crashing (dict key error).
 
Screen Shot 2022-01-06 at 11 47 36 AM

Using `get` from the `opt` to avoid the crash.


jxmsML (approved):
LGTM! Thanks for the fix

PR #4291, opened 2022-01-06
classifier_reranker

jaseweston:


klshuster (approved):
Maybe you can also add a quick note to the README somewhere saying that this is an implementation of the `AbstractReranker` that works for just about any classifier?

jaseweston:
> Maybe you can also add a quick note to the README somewhere saying that this is an implementation of the `AbstractReranker` that works for just about any classifier?

added

PR #4297, opened 2022-01-10
[crowdsourcing] add pointer to documentation in run scripts

meganung:
just adding comments in the run.py scripts to read `parlai/crowdsourcing/README.md` for more info for running tasks after all the updates following hydra 1.1


EricMichaelSmith (approved):
Yeah this is great - so useful to be able to point to a single source of truth (i.e. the README) for specifying how to run these tasks! Thanks for making this change :)

PR #4298, opened 2022-01-10
Add help string to crowdsourcing task

EricMichaelSmith:
Adds a help string to a crowdsourcing task to let users know how to run it. Based on https://github.com/facebookresearch/ParlAI/pull/4297


meganung (approved):
ah looks like i missed one thanks!

PR #4306, opened 2022-01-13
Delete chat services

stephenroller:
**Patch description**
Chat Services seems to be frequently a source of confusion and trouble for external users, who don't realize it is highly deprecated and not actively maintained.

This PR deletes all chat services to discourage users from building on top of it.

**Testing steps**
CI


klshuster:
I'm certainly not opposed to this, but it might be better to include some sort of note about how to recover and use? I do think this is being used...

stephenroller:
Sorry, I feel better after pretending I could

PR #4329, opened 2022-01-27
Decoder-Only Transformer

spencerp:
Supporting decoder-only transformer training:
```
parlai train_model --model transformer/decoder
```

**Summary of changes**
- New `TransformerDecoderLayer` without encoder attention
- New `TransformerDecoder` that concatenates input and "encoder state"
        - Needed to pass the dictionary to `TransformerDecoder` to properly do concatenation
	- Right-pad the label, left-pad the context
- New `PassThroughEncoder`, similar to `IdentityLayer` but compatible with TransformerEncoder API
	- There are many assumptions about there being an "encoder" in `TorchGeneratorAgent`, so the path of most code reuse led to using a dummy encoder to satisfy those assumptions
- New `DecoderAgent` to override `build_model`
- Extract label from scores to calculate loss (done at end of `TransformerDecoderOnly.forward`)
- Only use last `query_len` rows from incremental attention to account for incremental step from context to first generated token
- Added `DecoderIncrState` and `DecoderLayerIncrState` type aliases
- Added unit tests

**Summary of structural changes in `decoder.py`**
decoder_only_refac

**Testing steps**
```
python -m pytest tests/test_transformers.py
```
Screen Shot 2022-02-08 at 12 37 03

```
python -m pytest tests/test_light_whoami.py
```
Screen Shot 2022-02-08 at 12 12 04

Also ran some small training runs locally to sanity check:
```
TF_ARGS="--embedding-size 128 --ffn-size 512 --batchsize 16 --eval-batchsize 16 --model-parallel true --variant prelayernorm --n-heads 8 --n-positions 512 --activation gelu --text-truncate 256 --label-truncate 128 -lr 7e-05 --lr-scheduler invsqrt --optimizer adam --warmup_updates 1000 -vp 10 -vmt ppl -vmm min --load-from-checkpoint false -vstep 500 --validation-max-exs 1000 -tstep 150000 --log-every-n-secs 30 --update-freq 1 --dynamic-batching full"
parlai train_model -t convai2 --model transformer/decoder --model-file /tmp/test_decoder_only_model --n-layers 2 $TF_ARGS
parlai train_model -t convai2 --model transformer/generator --model-file /tmp/test_enc_decoder_model --n-encoder-layers 1 --n-decoder-layers 1 $TF_ARGS
```

**Benchmark Comparison with Encoder-Decoder**
Both trained to 10k steps with 120M parameters (8 layers) on 2x Quadro GP100 GPUs. The encoder/decoder is closer to 140M parameters due to the cross-attention.
```
TF_ARGS="--embedding-size 1024 --ffn-size 4096 --batchsize 16 --eval-batchsize 16 --model-parallel true --variant prelayernorm --n-heads 16 --n-positions 512 --activation gelu --text-truncate 256 --label-truncate 128 -lr 7e-05 --lr-scheduler invsqrt --optimizer adam --warmup_updates 1000 -vp 10 -vmt ppl -vmm min --load-from-checkpoint false -vstep 1000 --validation-max-exs 1000 -tstep 10000 --log-every-n-secs 30 --update-freq 1 --dynamic-batching full --fp16 true"
```
```
parlai train_model -t internal:new_reddit --model transformer/decoder --model-file /tmp/test_decoder_only_model --n-layers 8 $TF_ARGS
...
13:17:37 | time:6960s total_exs:433912 total_steps:9950 epochs:0.00 time_left:35s
    clen  clip  ctpb  ctps  ctrunc  ctrunclen  exps  exs  fp16_loss_scalar  gnorm  gpu_mem  llen  loss        lr  ltpb  ltps  ltrunc  ltrunclen   ppl  token_acc  token_em  total_train_updates  tpb  tps   ups
   58.72     1  2603  4957  .01423      .8327 85.63 2248            131072  1.711    .2126 26.09 4.036 2.219e-05  1173  2234       0          0 56.59      .2843         0                 9950 3776 7191 1.905

13:18:04 | Stopping from Maximum LR steps
```
```
parlai train_model -t internal:new_reddit --model transformer/generator --model-file /tmp/test_enc_decoder_model --n-encoder-layers 4 --n-decoder-layers 4 $TF_ARGS
...
10:52:31 | time:5367s total_exs:431792 total_steps:9950 epochs:0.00 time_left:27s
    clen  clip  ctpb  ctps  ctrunc  ctrunclen  exps  exs  fp16_loss_scalar  gnorm  gpu_mem  llen  loss        lr  ltpb  ltps  ltrunc  ltrunclen   ppl  token_acc  token_em  total_train_updates  tpb   tps  ups
   64.57     1  2711  6939  .02041      1.706 110.4 2156            131072  1.633    .1838 28.03 4.066 2.219e-05  1208  3094       0          0 58.33      .2803         0                 9950 3919 10033 2.56

10:52:51 | Stopping from Maximum LR steps
```


parlai/agents/transformer/modules/attention.py
```diff
@@ -224,9 +219,13 @@ def prepare_head(tensor):
...
                 mask = incr_state['prev_mask']
             else:
-                mask = torch.cat([incr_state['prev_mask'], mask], dim=2)
-                # Prepend along the key_len dimension (analogous to
-                # incr_state['prev_key'])
+                # Mask will be of size (B x query_len x key_len)
+                # During incremental decoding the query will only represent the next token,
+                # whereas the key/value will represent the entire sequence thus far.
+                # In such a case, we only want to look at the last element of the mask in the query dimension.
+                prev_mask = incr_state['prev_mask'][:, -query_len:, :]
```
klshuster, line 226: interesting, was this broken before?

parlai/agents/transformer/modules/decoder.py
```diff
@@ -451,3 +234,463 @@ def _apply_model_parallel(self, tensor, encoder_output, encoder_mask, incr_state
...
+# ENCODER-DECODER MODULES #
+###########################
+
+
+@swappable(
+    self_attention=MultiHeadAttention,
+    encoder_attention=MultiHeadAttention,
+    feedforward=TransformerFFN,
+)
+class TransformerDecoderLayer(BaseTransformerDecoderLayer):
```
klshuster, line 328: i dont think this class implements building self attention; is that necessary?

also, why is that function required to be implemented in the first place?

parlai/agents/transformer/modules/attention.py
spencerp: We never ran into this specific case before. This only breaks when 1) we are incremental decoding, and 2) sequences of different lengths are used between time steps.

When incremental decoding with decoder-only, the first step is the entire context. Then all following steps are of length one. So when transitioning from context to incremental decoding, `incr_state['prev_mask'].size(1)` will be the length of the context, but `mask.size(1)` will be one.

parlai/agents/transformer/modules/decoder.py
spencerp: Good catch! Will update.

spencerp: > also, why is that function required to be implemented in the first place?

The way `@swappable` is currently implemented, you can't have subclass where the parent and child are both swappable.

So `BaseTransformerDecoderLayer` doesn't have access to `self.swappables.feedforward`, and therefore these swappable modules need to be instantiated in the concrete `DecoderLayer` classes.

klshuster: cool, could you possibly leave a comment (if you haven't already) explaining that in the code?

spencerp: Added some comments to hopefully make it clearer, and went ahead and added a default implementation to `BaseTransformerDecoderLayer` in case someone wants to subclass it and _not_ use `@swappable`.

parlai/agents/transformer/decoder_only.py
```diff
@@ -0,0 +1,92 @@
...
+from parlai.utils.torch import padded_tensor
+
+from .modules import (
+    PassThroughEncoder,
+    TransformerDecoderOnly,
+    TransformerGeneratorModel,
+)
+
+
+class DecoderOnlyAgent(TorchGeneratorAgent):
```
stephenroller, line 24: nit, maybe just `transformer/decoder`?

parlai/agents/transformer/decoder_only.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    DecoderOnlyAgent.
+
+    Implementation of TorchGeneratorAgent, where the model is a Decoder-Only Transformer.
```
stephenroller, line 28: lint

parlai/agents/transformer/decoder_only.py
```diff
@@ -0,0 +1,92 @@
...
+
+        super().add_cmdline_args(parser, partial_opt=partial_opt)
+        return agent
+
+    def build_model(self, states=None):
+        """
+        Override of ``TorchAgent.build_model``.
+        """
+        assert (
+            self.opt['n_encoder_layers'] == -1
```
stephenroller, line 50: wishlist: this would be a really great opportunity to finally implement the `parser.remove_arg` functionality we've desired for a long time

parlai/agents/transformer/decoder_only.py
```diff
@@ -0,0 +1,41 @@
...
+        )
+        return wrapped_class(self.opt, self.dict)
+
+    def _pad_tensor(self, items, is_label=False):
+        """
+        Override of ``TorchAgent._pad_tensor``.
+        """
+        if is_label:
+            return padded_tensor(
+                items, pad_idx=self.NULL_IDX, left_padded=False, fp16friendly=False
```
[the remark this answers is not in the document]
stephenroller: We should address this

parlai/agents/transformer/modules/decoder.py
```diff
@@ -8,7 +8,8 @@
 """
 
 from __future__ import annotations
-from typing import Dict, Optional, Tuple
+from abc import ABC
+from typing import Dict, Optional, Tuple, Union
```
stephenroller, line 12: lint

parlai/agents/transformer/modules/decoder.py
```diff
@@ -23,20 +24,403 @@
...
+################
+# BASE CLASSES #
+################
+
+DecoderIncrState = Dict[int, Dict[str, Dict[str, torch.Tensor]]]
+
+
+class BaseTransformerDecoder(nn.Module, ABC):
+    """
+    Subclasses are require to implement ``forward``.
```
stephenroller, line 44: add some more docstrings describing the context of this

parlai/core/params.py
```diff
@@ -1393,3 +1393,10 @@ def error(self, message):
         self.print_help()
         _sys.stderr.write('\nParse Error: %s\n' % message)
         _sys.exit(2)
+
+
+def default(val, default):
```
stephenroller, line 1398: (y)

parlai/agents/transformer/decoder_only.py
spencerp: It'll be a liiiittle confusing when navigating through files, since there's also a `transformer/modules/decoder.py`. But I agree with you that we should prioritize the user-facing model name! Will change.

stephenroller:
Alright it's a little slower but good enough for today. Let Jason know it's done!

spencerp:
Hm, there's a distillation test failing which I don't understand yet... `dec_hid_loss` and `enc_dec_attn_loss` don't match what the test expects which is suspicious, so I'm going to wait to merge until I understand that better.

klshuster:
cc @EricMichaelSmith: there are two distillation tests failing due to this PR, do you have any idea why that might be the case? it looks like very small (but consistent) differences in testing (that is, these tests *do not* seem to be failing on main but consistently fail on this branch)

stephenroller:
Let's just let this sit a while. The regression failure is _a little_ concerning. It's enough to make me worry the code has changed in a subtle way, but close enough that it's _tempting_ to just --force-regen and forget about it.

EricMichaelSmith:
> Let's just let this sit a while. The regression failure is _a little_ concerning. It's enough to make me worry the code has changed in a subtle way, but close enough that it's _tempting_ to just --force-regen and forget about it.

Hmm seconded - yeah, @klshuster I'm not sure offhand, but perhaps this does indicate that the forward pass has indeed changed in some subtle way?

spencerp:
I will resuscitate it next week :)

spencerp:
Finally got back to this today and figured out the issue!

Apparently the distillation code is sensitive to the order in which modules are initialized 😔. I made a branch to demonstrate the minimal change needed to repro the test break: https://github.com/facebookresearch/ParlAI/pull/4526

To my knowledge, module initialization order has no implications on training dynamics/model performance. So I'm inclined to call this a distillation bug and outside the scope of this PR. Anyone know differently? @EricMichaelSmith 

If I don't hear any objections by Thursday I'll merge as-is with the broken test and file an issue to fix the bug in the distillation code.

spencerp:
Chatted with @EricMichaelSmith offline. He pointed out that this could be a result of a different order of random operations done during module initialization. I was able to confirm that's the problem, details in https://github.com/facebookresearch/ParlAI/pull/4526

So I'm just going to update the test numbers and merge.

PR #4330, opened 2022-01-27
[RAG] Fix Doc Chunker Assertion

klshuster:
**Patch description**
Fixes an assertion within the `RetrievedChunkRanker`; requires the doc chunks to be a str.

**Testing steps**
local testing


jxmsML (approved):
LGTM!!

PR #4331, opened 2022-01-27
BlenderBot2 memory generation batching

mojtaba-komeili:
**Patch description**
The offset value in `MemoryDecoder` was generating empty memories for all the examples in a batch (except the first one). This patch fixes that by removing the offset and passing the whole input.


klshuster (approved):
thanks for flagging this!

PR #4360, opened 2022-02-15
Add support for any GPT-2 model hosted in Huggingface

cahya-wirawan:
**Patch description**
As mentioned in https://github.com/facebookresearch/ParlAI/discussions/4357, the current Huggingface agent only supports few english GPT2 models (small, medium, large, xl and distilgpt2), therefore I add the functionality to specify arbitrary GPT-2 model hosted in Huggingface such as **anonymous-german-nlp/german-gpt2** or **indonesian-nlp/gpt2**.

**Testing steps**
Following is an example to fine tune indonesian-nlp/gpt2 model with convai2 dataset:
```
parlai train_model -m hugging_face/gpt2 --model_name indonesian-nlp/gpt2 --add-special-tokens True --add-start-token True -t convai2 -bs 2 -mf <modelfile>
```

### DIFF
--- a/parlai/agents/hugging_face/gpt2.py
+++ b/parlai/agents/hugging_face/gpt2.py
@@ -58,23 +58,26 @@ class GPT2Decoder(torch.nn.Module):
 
     def _init_from_pretrained(self, opt):
         # load model
-        model_sz = opt["gpt2_size"]
-        if model_sz == "small":
-            model_key = "gpt2"
-        elif model_sz == "distilgpt2":
-            model_key = "distilgpt2"
+        if opt["model_name"]:
+            fle_key = opt["model_name"]
         else:
-            model_key = f"gpt2-{model_sz}"
+            model_sz = opt["gpt2_size"]
+            if model_sz == "small":
+                model_key = "gpt2"
+            elif model_sz == "distilgpt2":
+                model_key = "distilgpt2"
+            else:
+                model_key = f"gpt2-{model_sz}"
 
-        # check if datapath has the files that hugging face code looks for
-        hf_dir = os.path.join(opt["datapath"], "hf", model_key)
-        if all(
-            PathManager.exists(os.path.join(hf_dir, file_name))
-            for file_name in ["pytorch_model.bin", "config.json"]
-        ):
-            fle_key = PathManager.get_local_path(hf_dir, recursive=True)
-        else:
-            fle_key = model_key
+            # check if datapath has the files that hugging face code looks for
+            hf_dir = os.path.join(opt["datapath"], "hf", model_key)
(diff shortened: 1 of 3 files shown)


parlai/agents/hugging_face/gpt2.py
```diff
@@ -237,6 +240,12 @@ def add_cmdline_args(
         cls, parser: ParlaiParser, partial_opt: Optional[Opt] = None
     ) -> ParlaiParser:
         agent = parser.add_argument_group("Gpt2 Args")
+        agent.add_argument(
+            "--model_name",
+            type=str,
+            default=None,
```
stephenroller, line 246: can we provide a default here that would fall back to the existing behavior?

parlai/agents/hugging_face/gpt2.py
```diff
@@ -237,6 +240,12 @@ def add_cmdline_args(
         cls, parser: ParlaiParser, partial_opt: Optional[Opt] = None
     ) -> ParlaiParser:
         agent = parser.add_argument_group("Gpt2 Args")
+        agent.add_argument(
+            "--model_name",
```
stephenroller, line 244:
```suggestion
            "--model-name",
```

stephenroller (commented):
Cool! seems like a nice extension. Flagging for @jxmsML to also review please.

parlai/agents/hugging_face/gpt2.py
cahya-wirawan: Actually, it is already the case. If we don't use the argument "--model_name" at all, it will fall back to the old behavior.

parlai/agents/hugging_face/gpt2.py
cahya-wirawan: Agree, I like this argument name more. I used --model_name because this agent uses already the argument "--gpt2_size".
If we change it to "--model-name", we should do the same in other place too, isn't it?

cahya-wirawan:
Hi @stephenroller , sorry I am not familiar with PR procedure of this repo yet,  I just don't know about the error in ci/circleci: unittests_38 above, should I create a unit test for the additional argument? and if yes, where should I put it? Thanks.

stephenroller:
Our CircleCI is broken right now. I think this PR is fine.

stephenroller (approved):

stephenroller:
Last thing: can you run "black" on the code?

cahya-wirawan:
> 
I did it before, but here again:
% black parlai/agents/hugging_face
All done! ✨ 🍰 ✨
6 files left unchanged.

stephenroller:
Thanks!

PR #4366, opened 2022-02-17
[BB2] Add check for correct query generator

klshuster:
**Patch description**
If we set `--knowledge-access-method search_only` for BB2, we *cannot* use the default query generator. This patch adds a runtime check for this, and raises an error if it occurs.

Also checking in an override option.

**Testing steps**
local testing


jxmsML (approved):
LGTM!!! Thanks

PR #4369, opened 2022-02-18
Add WorldLogger to train_model script.

kauterry:
**Patch description**

We'd like to dump the world logs for the "test" fold in the final evaluation during training. This way we can skip running the eval_model script while running training and evaluation back to back, and we can retrieve all the evaluation information from the world logs.

**Testing steps**

Added unittests in tests/test_train_model.py

**Other information**

### DIFF
--- a/parlai/scripts/train_model.py
+++ b/parlai/scripts/train_model.py
@@ -25,6 +25,7 @@ parlai train_model --model seq2seq --task babi:Task10k:1 --model-file '/tmp/mode
 # * More logging (e.g. to files), make things prettier.
 import copy
 import json
+import os
 import numpy as np
 import signal
 from typing import Tuple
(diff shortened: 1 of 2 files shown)


tests/test_train_model.py
```diff
@@ -223,6 +225,52 @@ def test_opt_step(self):
...
+                }
+            )
+
+            for task in multitask.split(','):
+                task_log_report = get_task_world_logs(
+                    task, log_report, is_multitask=True
+                )
+                with PathManager.open(task_log_report) as f:
+                    json_lines = f.readlines()
+                assert len(json_lines) == 5
```
jxmsML, line 272: why is there only 5 lines instead of 10?

jxmsML (commented):
LGTM! a minor point.
Also it might be nice to take a look at the CI test failures.

parlai/scripts/train_model.py
```diff
@@ -647,8 +687,11 @@ def _run_eval(
         reports = []
 
         max_exs_per_worker = max_exs / (len(valid_worlds) * num_workers())
```
kauterry, line 689: @jxmsML This is why we have only 5 lines instead of 10, cuz during multitask `len(valid_worlds)` > 1

jxmsML (approved):

Issue #4372, opened 2022-02-21
Using --no-cuda option still uses GPU memory


tungnt55:
**Bug description**
Running BlenderBot2 in interactive mode with --no-cuda option. GPU utilization is at 0~1% but the program still uses GPU memory.

**Reproduction steps**
Running BlenderBot 2 with the following command: 
parlai interactive --model-file zoo:blenderbot2/blenderbot2_400M/model --search-server None --knowledge-access-method memory_only --memory-key full_text --inference topk --no-cuda

**Expected behavior**
The program should not use any GPU memory.

**Logs**
Please paste the command line output:
Below is the nvidia-smi output
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 470.103.01   Driver Version: 470.103.01   CUDA Version: 11.4     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA GeForce ...  Off  | 00000000:02:00.0  On |                  N/A |
|  0%   59C    P2    33W / 170W |   8083MiB / 12031MiB |      1%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|    0   N/A  N/A      1118      G   /usr/lib/xorg/Xorg                 53MiB |
|    0   N/A  N/A      1674      G   /usr/lib/xorg/Xorg                294MiB |
|    0   N/A  N/A      1819      G   /usr/bin/gnome-shell               65MiB |
|    0   N/A  N/A      3188      G   /usr/lib/firefox/firefox           10MiB |
|    0   N/A  N/A    170635      C   python                           7627MiB |
+-----------------------------------------------------------------------------+
```

Output of running parlai:
```
[...]

PR #4377, opened 2022-02-23
[BB2] Respect `no_cuda`

klshuster:
**Patch description**
Previously, when instantiating bb2 with `--no-cuda`, the sub modules (query generator, memory decoder) were not respecting this flag. This patch fixes that (addressing #4372)

**Testing steps**
Tested locally, confirmed via `nvidia-smi`


mojtaba-komeili (approved):
LGTM

jxmsML (approved):
LGTM!

PR #4379, opened 2022-02-23
[BlenderBot2] clean up reply during interactive mode

jxmsML:
**Patch description**
Right now, during interactive mode, the `'_POTENTIALLY_UNSAFE__'` from blenderbot2 generation would go into blenderbot 2 history and possibly affects its subsequent generation. However the model is never trained on observing such specially tokens, only generating them.

After the change, we still keep the `'_POTENTIALLY_UNSAFE__'`  in the reply text, and only remove that from the blenderbot2 history.

**Testing steps**

**Other information**


projects/blenderbot2/agents/blenderbot2.py
```diff
@@ -898,6 +905,39 @@ def compute_loss(
...
+        """
+        Observe one's own utterance.
+        Override TorchAgent.self_observe with the optional cleaned text
+
+        :param self_message:
+            The message corresponding to the output from batch_act.
+        """
+
+        if (
+            self.opt('clean_reply', False)
```
mojtaba-komeili, line 922: So it returns here and doesn't go to cleaning the text if `clean_reply` is `True`?

jxmsML: I'd love to ALWAYS clean the reply and I think it's the right thing to do?

projects/blenderbot2/agents/blenderbot2.py
```diff
@@ -898,6 +905,39 @@ def compute_loss(
...
+            or self.observation['episode_done']  # last example in the episode
+            or use_reply == 'none'  # not including our own responses anyway
```
mojtaba-komeili, line 924: Where does `use_reply` come from?

projects/blenderbot2/agents/blenderbot2.py
jxmsML: I'd love to ALWAYS clean the reply and I think it's the right thing to do?

projects/blenderbot2/agents/blenderbot2.py
jxmsML: good catch, I've added that back now.

klshuster (approved):

klshuster:
could you maybe add some test cases to make sure that `_clean_text` works?

projects/blenderbot2/agents/blenderbot2.py
```diff
@@ -898,6 +899,38 @@ def compute_loss(
...
+                and any([l in self.observation for l in ['labels', 'eval_labels']])
+            )  # has true label
+        ):
+            return super().self_observe(self_message)
+
+        # otherwise, we use the CLEANED last output the model generated
+        if self_message is not None:
+            last_reply = self_message['text']
+            clean_reply = self._clean_text(last_reply)
+            self.history.add_reply(clean_reply)
```
stephenroller, line 929: perhaps we should just override the History class and do this logic there?

stephenroller:
Go ahead and merge main into this branch

PR #4384, opened 2022-02-28
Warmup updates bug for LR < 1

emilydinan:
**Patch description**
#4242 introduced a bug in which `_is_warming_up` returns True if the last LR is < 1. This will introduce a bug for all initial LRs != 1. In particular, for models with initial learning rate < 1, the model will be "warming up" forever, and the LR will remain constant after the warmup updates have finished rather than starting to decay according to the provided schedule.

**Testing steps**
```
parlai tm -t convai2 -m transformer/generator --lr-scheduler linear --warmup-updates 10 -lstep 1 -vstep 10000000 --max-lr-steps 100 --skip-generation True --warmup-rate 0.01 -lr 0.00001 --dict-file /tmp/test123.dict -mf /tmp/test1234dsfsdf5
```

I had to relax restrictions to get tests to pass. If we change `self._number_training_updates < self.warmup_updates` -->, `self._number_training_updates <= self.warmup_updates`, we hit the exact max LR, but don't quite anneal to zero. Will leave it to follow up PR (Jude) to test this more robustly

### DIFF
--- a/parlai/nn/lr_scheduler.py
+++ b/parlai/nn/lr_scheduler.py
@@ -83,7 +83,7 @@ class ParlAILRScheduler(object):
         """
         return (
             self.warmup_scheduler is not None
-            and self.warmup_scheduler.get_last_lr()[0] < 1.0
+            and self._number_training_updates < self.warmup_updates
         )
 
     def _warmup_lr(self, step):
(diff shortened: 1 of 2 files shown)


stephenroller (approved):
Ty

emilydinan:
Tests pass locally

stephenroller:
@klshuster and @jxmsML confirm this does NOT affect reduceonplataeu

stephenroller:
@spencerp found this did NOT affect invsqrt

PR #4389, opened 2022-03-02
[RAG] Fix Extra Positions Bug

klshuster:
**Patch description**
`--n-extra-positions` can be used to expand the total number of positions used when encoding documents + input in pre-trained transformers. If we set this >0, then we are *supposed* to fill up all of end positions with the doc tokens, leaving the beginning positions for the input text. 

In the current RAG implementation, if we specify `--n-extra-positions M` < `--n-positions N`, we encounter an issue where the documents were not properly truncated. This is found in the `concat_docs_and_input` function, where `self.expanded_input_truncate` is used to determine the maximum length of the documents. This is an overloaded attribute, hence why the bug was a bit subtle.

This bug was not previously caught, as in all of my experiments, `--n-extra-positions > --n-positions`. 

**Testing steps**
Added a test to confirm that `concat_docs_and_input` respects the `n_positions` available in the model.

### DIFF
--- a/parlai/agents/rag/modules.py
+++ b/parlai/agents/rag/modules.py
@@ -74,9 +74,10 @@ class RagModel(TorchGeneratorModel):
             opt['text_truncate'] or opt['truncate'], get_n_positions_from_options(opt)
         )
         if self.n_extra_positions > 0:
-            self.expanded_input_truncate = max(
-                self.expanded_input_truncate, self.n_extra_positions
-            )
+            # This attribute is overloaded.
+            # when n_extra_positions == 0, it is the truncation of the full expanded input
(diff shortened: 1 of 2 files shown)


mojtaba-komeili (approved):
LGTM

mojtaba-komeili:
nit: a general comment on `self.n_extra_positions`: In `RagModel` constructor we have ` assert opt['n_extra_positions'] >= 0`. Yet there are line in code that we check for `self.n_extra_positions <= 0`.

jxmsML (approved):
👍

klshuster:
The failing test is fixed in #4395

PR #4392, opened 2022-03-03
Lrsched missing step

juderoque:
**Patch description**
Follow up from #4384: Unrelaxed relaxed conditions in tests and modified scheduler logic to fit unrelaxed conditions.

- changed `self._number_training_updates < self.warmup_updates` `-->` `self._number_training_updates <= self.warmup_updates` to hit the exact `max_lr`. 
   - In this case the lr doesn't anneal to 0 due to a missing step.
- Modified stopping conditions in `parlai/nn/lr_scheduler.py` and `parlai/scripts/train.py` to allow the missing step to be taken
- Modified `test_lr_schedulers.py` to step from 1 -> total_steps rather than 0 -> total_steps - 1 to match the behavior in `lr_scheduler.py`

**Testing steps**
```
parlai tm -t convai2 -m transformer/generator --lr-scheduler linear --warmup-updates 10 -lstep 1 -vstep 10000000 --max-lr-steps 100 --skip-generation True --warmup-rate 0.01 -lr 0.00001 --dict-file /tmp/test123.dict
```
should now show 100 steps instead of 99

**Other information**
Still not sure if this is the desired behavior, is there an implicit step (0) taken?

### DIFF
--- a/parlai/nn/lr_scheduler.py
+++ b/parlai/nn/lr_scheduler.py
@@ -83,7 +83,7 @@ class ParlAILRScheduler(object):
         """
         return (
             self.warmup_scheduler is not None
-            and self._number_training_updates < self.warmup_updates
+            and self._number_training_updates <= self.warmup_updates
         )
 
     def _warmup_lr(self, step):
@@ -414,7 +414,7 @@ class InvSqrtLRScheduler(ParlAILRScheduler):
         return self.decay_factor / np.sqrt(max(1, self.invsqrt_lr_decay_gamma + step))
 
     def train_step(self, scheduler_steps):
-        if self.max_lr_steps > 0 and scheduler_steps >= self.max_lr_steps:
+        if self.max_lr_steps > 0 and scheduler_steps > self.max_lr_steps:
             raise StopTrainException('Maximum LR steps')
         self.scheduler.step()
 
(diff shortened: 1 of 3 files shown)


emilydinan (commented):
Thanks for fix --great work!! 

Can you add some tests with `max_lr` != 1? Also can you add additional tests to check that LR < max_lr at output[warmup_updates]

juderoque:
Tests added!

juderoque:
I noticed that the step counter used by this function (https://github.com/facebookresearch/ParlAI/blob/main/parlai/nn/lr_scheduler.py#L294) starts out at 1. This means the first step of the warmup value is 1 step ahead of the specified value, and the first step of the regular scheduler is also one step ahead of the specified max_lr value. In this patch I have made it so the last step of the warmup scheduler hits the max_lr. An alternative behavior could be that the steps start from 0, the last step of the warmup scheduler is "one step before" the max_lr, and the regular scheduler starts out at max_lr. This distinction is important in the case where there are 0 warmup-updates because here we never actually the specified max_lr, rather start at one step after it.
Thoughts @stephenroller @emilydinan ?

stephenroller:
Generally I prefer backwards compatibility even if it’s esoteric.

juderoque:
> Generally I prefer backwards compatibility even if it’s esoteric.

Do you think adding the extra step would ruin backwards compatibility?

emilydinan (approved):
great job @juderoque !

stephenroller:
I defer to Emily. She's thought way more about this

stephenroller:
Bump

stephenroller:
(go ahead and merge main into this to fix tests, please)

stephenroller:
Should we reconsider this?

PR #4395, opened 2022-03-03
[🤗] Safely get model_name from opt

spencerp:
**Patch description**
I used `HuggingFaceDictionaryAgent` in my own agent, which crashed because my agent doesn't have the `--model-name` flag. By accessing `"model_name"` with `.get()`, these conditions work as expected in agents other than `Gpt2Agent`.

Context of breaking change: https://github.com/facebookresearch/ParlAI/pull/4360

**Testing steps**
Used my agent (code not checked in) and verified code did not crash.
CircleCI

### DIFF
--- a/parlai/agents/hugging_face/dict.py
+++ b/parlai/agents/hugging_face/dict.py
@@ -140,7 +140,7 @@ class Gpt2DictionaryAgent(HuggingFaceDictionaryAgent):
         """
         Instantiate tokenizer.
         """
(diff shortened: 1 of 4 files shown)


klshuster (approved):
thank you!

klshuster:
it looks like teacher tests are failing when testing the LCCC dataset?

stephenroller:
LCCC is another issue. This is great, thanks

PR #4419, opened 2022-03-11
[BB2] Support (in theory) BB2 Distributed

klshuster:
**Patch description**
A few small changes:

1. Explicitly setting `param.requires_grad=False` for some FiD modules; this is to, in theory, support something like zero2 ddp.
2. provide backwards compat for gold doc agent with agent that was trained with a non-gold-doc agent.

**Testing steps**
Tested gold document functionality locally.

### DIFF
--- a/projects/blenderbot2/agents/modules.py
+++ b/projects/blenderbot2/agents/modules.py
@@ -828,6 +828,8 @@ class BlenderBot2Fid(Fid):
 class BlenderBot2FidModelMixin:
     embedding_size: int
     pad_idx: int
+    long_term_memory: LongTermMemory
+    retriever: RagRetriever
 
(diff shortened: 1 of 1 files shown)


projects/blenderbot2/agents/modules.py
```diff
@@ -837,6 +839,12 @@ def __init__(self, opt: Opt, dictionary: DictionaryAgent, retriever_shared=None)
             opt, dictionary[dictionary.null_token]
         )
         self.embedding_size = opt['embedding_size']
+        for param in self.long_term_memory.query_encoder.parameters():
```
jxmsML, line 842: interesting I thought these are already there. 
thanks for the fix!

Issue #3662, opened 2021-05-21
MultiWorld behaves nondeterministically on `:ordered`


stephenroller:
Trying to do 

```
while not multiworld.epoch_done():
     multiworld.parley()
```

Results in sampling-style randomized examples, possibly with some repeated.

Paul points out:
> I wondered if this might be the problem https://github.com/facebookresearch/ParlAI/blob/master/parlai/core/worlds.py#L667

Which comes from
https://github.com/facebookresearch/ParlAI/blob/157f416745dd62824230f4b41b2156ace255cc63/parlai/core/worlds.py#L556
and
https://github.com/facebookresearch/ParlAI/blob/157f416745dd62824230f4b41b2156ace255cc63/parlai/utils/data.py#L80

probably need to update the self.is_training to be more like self.should_shuffle
https://github.com/facebookresearch/ParlAI/blob/157f416745dd62824230f4b41b2156ace255cc63/parlai/utils/data.py#L50

PR #4425, opened 2022-03-16
use should_shuffle instead of is_training to determine whether to ran…

juderoque:
…domly sample from worlds

**Patch description**
Addresses #3662 by using should_shuffle instead of is_training to determing whether to randomly sample between worlds or to use the round-robin world selection.

**Testing steps**
`pytest tests/test_multiworld.py` (test_with_ordered)

**Other information**
When should shuffle is False, multitask weights are effectively ignored. We warn the user if they choose a datatype that doesn't shuffle and also specify multitask weights.

### DIFF
--- a/parlai/core/worlds.py
+++ b/parlai/core/worlds.py
@@ -53,7 +53,7 @@ from parlai.core.opt import Opt
 from parlai.core.params import ParlaiParser
 from parlai.core.teachers import Teacher, create_task_agent_from_taskname
 from parlai.utils.data import DatatypeHelper
-from parlai.utils.misc import Timer, display_messages
+from parlai.utils.misc import Timer, display_messages, warn_once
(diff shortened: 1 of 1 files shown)


stephenroller (approved):
lgtm, let's see what @skiingpacman says

skiingpacman:
LGTM. One question (more a clarification) what's being addressed by "added warning about using datatype that doesn't shuffle with multitask weights"?  It's not clear to me what this is saying? That "weighting" of tasks isn't respected or something else?

skiingpacman (approved):

juderoque:
> LGTM. One question (more a clarification) what's being addressed by "added warning about using datatype that doesn't shuffle with multitask weights"? It's not clear to me what this is saying? That "weighting" of tasks isn't respected or something else?

Yep! warning that the weighing of tasks is not respected

PR #4426, opened 2022-03-16
Updating to Mephisto 1.0

JackUrb:
**Patch description**
First set of steps that get `crowdsourcing` tests _running_ (no longer breaking the newer mephisto conventions), but still not **passing**. More work is to be done there, leaving this open as a starting point for others to make comments and take over.

**Testing steps**
```
pytest tests/crowdsourcing
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,7 +159,7 @@ commands:
       - run:
             name: Install Mephisto
             command: |
-              pip install mephisto==1.0.0b1
+              pip install mephisto==1.0.0b2
               echo | mephisto check
 
   setupcuda:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,7 +159,7 @@ commands:
       - run:
             name: Install Mephisto
             command: |
-              pip install mephisto=1.0.0b1
+              pip install mephisto==1.0.0b1
               echo | mephisto check
 
   setupcuda:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,12 +159,7 @@ commands:
       - run:
             name: Install Mephisto
             command: |
-              cd ..
-              git clone git@github.com:facebookresearch/Mephisto.git Mephisto
-              cd Mephisto; git checkout v0.4.2 -b stable
-              pip install -r requirements.txt
-              python setup.py develop
-              # `echo` so that ENTER will be pressed at the prompt
+              pip install mephisto=1.0.0b1
               echo | mephisto check
 
   setupcuda:
--- a/parlai/crowdsourcing/utils/tests.py
+++ b/parlai/crowdsourcing/utils/tests.py
@@ -93,6 +93,7 @@ class AbstractCrowdsourcingTest:
                     f'mephisto/provider=mock',
                     f'+task_dir={task_directory}',
                     f'+current_time={int(time.time())}',
+                    f"+mephisto.task.submission_timout=10",
                 ]
                 + overrides,
             )
@@ -111,13 +112,30 @@ class AbstractCrowdsourcingTest:
         self.operator.validate_and_run_config(
             self.config.mephisto, shared_state=shared_state
         )
-        self.server = self._get_channel_info().job.architect.server
+        self.server = self._get_server()
+
+    def _get_live_run(self):
+        """
+        Get the LiveTaskRun from this job
+        """
+        live_runs = list(self.operator.get_running_task_runs().values())
+        if len(live_runs) == 0:
+            raise ValueError('No live runs present')
+        return live_runs[0]
+
+    def _get_server(self):
+        """
+        Return the MockArchitect's server associated with this run
+        """
+        live_run = self._get_live_run()
+        return live_run.architect.server
 
     def _get_channel_info(self):
         """
         Return channel info for the currently running job.
         """
-        channels = list(self.operator.supervisor.channels.values())
+        live_run = self._get_live_run()
+        channels = list(live_run.client_io.channels.values())
         if len(channels) > 0:
             return channels[0]
         else:
@@ -135,22 +153,28 @@ class AbstractCrowdsourcingTest:
 
         for idx in range(num_agents):
 
-            mock_worker_name = f"MOCK_WORKER_{idx:d}"
-            max_num_tries = 6
+            mock_worker_registration_name = f"MOCK_WORKER_{idx:d}"
+            mock_worker_name = f"{mock_worker_registration_name}_sandbox"
+            max_num_tries = 3
             initial_wait_time = 0.5  # In seconds
             num_tries = 0
             wait_time = initial_wait_time
             while num_tries < max_num_tries:
                 try:
-
-                    # Register the worker
-                    self.server.register_mock_worker(mock_worker_name)
-                    workers = self.db.find_workers(worker_name=mock_worker_name)
-                    worker_id = workers[0].db_id
-
                     # Register the agent
                     mock_agent_details = f"FAKE_ASSIGNMENT_{idx:d}"
-                    self.server.register_mock_agent(worker_id, mock_agent_details)
+                    self.server.register_mock_agent(
+                        mock_worker_registration_name, mock_agent_details
+                    )
+                    self.assert_sandbox_worker_created(mock_worker_name)
+                    workers = self.db.find_workers(worker_name=mock_worker_name)
+                    print(
+                        f"try num: {num_tries}:",
+                        workers,
+                        self.db.find_workers(),
+                        [w.worker_name for w in self.db.find_workers()],
+                    )
+                    worker_id = workers[0].db_id
 
                     if assume_onboarding:
                         # Submit onboarding from the agent
@@ -159,6 +183,9 @@ class AbstractCrowdsourcingTest:
                         self.server.register_mock_agent_after_onboarding(
                             worker_id, onboard_agents[0].get_agent_id(), onboard_data
                         )
+                        self.await_channel_requests()
+                    self.await_channel_requests()
+                    print(f"try num: {num_tries}:", self.db.find_agents())
                     _ = self.db.find_agents()[idx]
                     # Make sure the agent can be found, or else raise an IndexError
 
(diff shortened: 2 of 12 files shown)


parlai/crowdsourcing/utils/tests.py
```diff
@@ -135,22 +153,28 @@ def _register_mock_agents(
 
         for idx in range(num_agents):
 
-            mock_worker_name = f"MOCK_WORKER_{idx:d}"
-            max_num_tries = 6
+            mock_worker_registration_name = f"MOCK_WORKER_{idx:d}"
+            mock_worker_name = f"{mock_worker_registration_name}_sandbox"
+            max_num_tries = 3
```
JackUrb, line 158: As an aside, retries should no longer be necessary with `assert_sandbox_worker_created` and `await_channel_requests` which run the async loop until pending things are processed.

parlai/crowdsourcing/utils/tests.py
```diff
@@ -186,6 +213,25 @@ def _register_mock_agents(
...
+        self.assertTrue(  # type: ignore
+            self.operator._run_loop_until(
+                lambda: len(tracked_run.client_io.request_id_to_channel_id) == 0,
+                timeout,
+            ),
+            f"Channeled requests not processed in time!",
+        )
+
+    def assert_sandbox_worker_created(self, worker_name, timeout=2) -> None:
+        self.assertTrue(  # type: ignore
```
JackUrb, line 228: Note, this is failing on `in_flight_qa` tests because it's not extending a test case presumably? You may want to rewrite these two helpers to just throw an assertion error instead.

parlai/crowdsourcing/utils/tests.py
```diff
@@ -308,22 +349,18 @@ def _test_agent_states(
...
-            )
+            self.server.submit_mock_unit(agent_id, {'final_data': {}})
+            self.await_channel_requests()
 
         # # Check that the inputs and outputs are as expected
 
         # Wait until all messages have arrived
         wait_time = 5.0  # In seconds
-        max_num_tries = 30  # max_num_tries * wait_time is the max time to wait
+        max_num_tries = 1  # max_num_tries * wait_time is the max time to wait
```
EricMichaelSmith, line 363: Hmm - so we should just be able to get rid of `max_num_tries` and the retry code altogether, right?

tests/crowdsourcing/tasks/test_chat_demo.py
```diff
@@ -386,6 +386,7 @@ def test_base_task(self):
                 '++mephisto.task.maximum_units_per_worker=0',
                 '++num_turns=3',
                 '++turn_timeout=300',
+                '++mephisto.task.submission_timout=10',
```
EricMichaelSmith, line 390: Hmm what's the reason for this change?

EricMichaelSmith:
@JackUrb Most of the crowdsourcing tests don't seem to be running currently due to `ImportError`s - disabling the `try/except` blocks around them to see what's going on

parlai/crowdsourcing/utils/tests.py
```diff
@@ -135,22 +153,28 @@ def _register_mock_agents(
...
                     # Register the agent
                     mock_agent_details = f"FAKE_ASSIGNMENT_{idx:d}"
-                    self.server.register_mock_agent(worker_id, mock_agent_details)
+                    self.server.register_mock_agent(
+                        mock_worker_registration_name, mock_agent_details
+                    )
+                    self.assert_sandbox_worker_created(mock_worker_name)
+                    workers = self.db.find_workers(worker_name=mock_worker_name)
+                    print(
+                        f"try num: {num_tries}:",
```
EricMichaelSmith, line 172: Is printing the try num useful here if we no longer need retries?

parlai/crowdsourcing/utils/tests.py
EricMichaelSmith: Interesting, yeah perhaps

EricMichaelSmith (commented):
Thanks for this PR and for doing these refactors - yeah, it looks like it's useful to have a few pieces of boilerplate code abstracted away, and not needing to do retries would be useful. Am trying to get all 43 crowdsourcing tests to run to get a better sense of what needs to be done with this

tests/crowdsourcing/tasks/test_chat_demo.py
JackUrb: we added a new config option here that I wasn't able to get hydra's configure to pick up the default of. This was the first place I could find that restored it.

parlai/crowdsourcing/utils/tests.py
JackUrb: Almost definitely yes, but I was focused on minimal changes.

tests/crowdsourcing/tasks/test_chat_demo.py
EricMichaelSmith: Fair enough (nit: typo in the var name I believe)

JackUrb: > Fair enough (nit: typo in the var name I believe)

_that_ was the real issue, thanks for the catch! Pushing this through the 1.0 release branch right now

EricMichaelSmith:
Okay, I got all 43 crowdsourcing checks to run so that we can debug them. @JackUrb 27 of them, the Fast-Acute ones, are currently failing with a "No live runs present" error due to no `LiveTaskRun`s being found when calling `Operator.get_running_task_runs()` within ParlAI's `AbstractCrowdsourcingTest._get_live_run()`. Do you know what might cause no task runs to be found?

tests/crowdsourcing/tasks/test_chat_demo.py
EricMichaelSmith: Ah great - hmm in that case, presumably we don't still need this line in the PR once the Hydra flag name is changed?

JackUrb:
Hadn't noticed that the blueprint was broken under-the-hood, can make a quick change for this.

tests/crowdsourcing/tasks/test_chat_demo.py
JackUrb: We'll see once I get 1.0 out and we're pulling from there instead.

EricMichaelSmith:
> Hadn't noticed that the blueprint was broken under-the-hood, leading to a blueprint launch error (and thus no running tasks), can make a quick change for this. (running locally identified the issue in logs)

Great, thanks! Hmm, now I'm seeing an `AgentTimeoutError` in the CI check logs when cleaning up the test unit: https://app.circleci.com/pipelines/github/facebookresearch/ParlAI/11125/workflows/8f87deb8-70aa-47d3-86a7-dd9b474a900b/jobs/91214?invite=true#step-111-5068

JackUrb:
That would imply to me that the unit was still running when it was shutdown, and thus the shutdown waited for the timeout. You may need to examine this locally to see what was coming through the agent and what wasn't, as I'm unclear why this happened from the given info.

EricMichaelSmith:
> That would imply to me that the unit was still running when it was shutdown, and thus the shutdown waited for the timeout. You may need to examine this locally to see what was coming through the agent and what wasn't, as I'm unclear why this happened from the given info.

Hmm, on my devfair, it looks like this issue with the unit being left hanging came from a `ValueError` due to the test unit not being registered in the data browser correctly:
```
self = <parlai.crowdsourcing.tasks.acute_eval.analysis.AcuteAnalyzer object at 0x7fc491f6e670>

    def _extract_to_dataframe(self) -> pd.DataFrame:
        """
        Extract the data from the run to a pandas dataframe.
        """
        units = self.mephisto_data_browser.get_units_for_task_name(self.run_id)
        responses: List[Dict[str, Any]] = []
        for unit in units:
            unit_details = self._parse_unit(unit)
            if unit_details is None:
                continue
            for idx in range(len(unit_details['data'])):
                response = self._extract_response_by_index(unit_details, idx)
                if response is not None:
                    responses.append(response)

        if len(responses) == 0:
>           raise ValueError('No valid results found!')
E           ValueError: No valid results found!

parlai/crowdsourcing/tasks/acute_eval/analysis.py:251: ValueError
```
So I suppose the question now is (1) whether the unit got saved in the data browser correctly, and if so, (2) why it's not being loaded back in

JackUrb:
> So I suppose the question now is (1) whether the unit got saved in the data browser correctly, and if so, (2) why it's not being loaded back in

My bet is still not have been marked as completed, which would happen in another thread (and I imagine if this script is launched before the unit is completed, you won't get result data). I expect this to more likely be (1) than (2). You'd likely want to dig into the `TaskRunner` to be sure that your `TaskRunner`'s `run_unit` function completes.

.circleci/config.yml
```diff
@@ -159,12 +159,7 @@ commands:
       - run:
             name: Install Mephisto
             command: |
-              cd ..
-              git clone git@github.com:facebookresearch/Mephisto.git Mephisto
-              cd Mephisto; git checkout v0.4.2 -b stable
-              pip install -r requirements.txt
-              python setup.py develop
-              # `echo` so that ENTER will be pressed at the prompt
+              pip install mephisto==1.0.0b2
```
stephenroller, line 162: why b2?

JackUrb: 1.0.0 full isn't out until later today, but wanted to get started on this flow sooner.

parlai/crowdsourcing/utils/tests.py
```diff
@@ -308,61 +344,38 @@ def _test_agent_states(
...
-        else:
-            actual_num_messages = sum(
-                len(state['outputs']['messages']) for state in actual_states
-            )
-            print(f'\nPrinting all {actual_num_messages:d} messages received:')
-            for state in actual_states:
-                for message in state['outputs']['messages']:
-                    print(message)
+        # Get and filter actual messages
+        time.sleep(self.message_sleep_time)
```
EricMichaelSmith, line 357: @JackUrb Adding this in because, for the model image chat task, the messages only arrive in the agent state a few seconds after the calls to `self.await_channel_requests()`. This is probably because one of the messages contains a large encoded image file

tests/crowdsourcing/tasks/test_chat_demo.py
```diff
@@ -18,142 +18,73 @@
     "outputs": {
         "messages": [
             {
-                "packet_type": "update_status",
-                "sender_id": "mephisto",
-                "receiver_id": "1",
-                "data": {"state": {"agent_display_name": "Chat Agent 1"}},
-                "timestamp": 1604343628.610868,
+                "text": "Hi! How are you?",
```
EricMichaelSmith, line 21: @meganung It looks like this change fixes TestChatDemo, so you may not need to take any further action on this PR

parlai/crowdsourcing/utils/tests.py
EricMichaelSmith: These checks seem to be passing now

tests/crowdsourcing/tasks/test_chat_demo.py
EricMichaelSmith: Just removed these lines

parlai/crowdsourcing/utils/tests.py
EricMichaelSmith: Just removed this retry loop without breaking the tests

PR #4427, opened 2022-03-16
More flexible token metadata logging

c-flaherty:
**Patch description**

This is a followup PR to https://github.com/facebookresearch/ParlAI/pull/4169. In that PR, I added support to log token probabilities and token ranks for outputs of ParlAI models. However, after using it, it became clear that we would like to log additional token-level metadata, such as top 10 tokens and top ranked token (relevant for sampling-based decoding methods). 

Rather, than add these features directly, I am instead making the token-level metadata object more flexible. In this PR, each token has associated with it a typed dictionary `_PathSelectionTokenDetails` that contains the `token_score` and `token_rank` of the relevant token. No code outside the `TreeSearch:select_paths` method implementations and this typed dictionaries' definition make any reference to the specific fields in this dictionary. This makes it easy to override this dictionary's definition and a `TreeSearch:select_paths` implementation to add more verbose metadata. Since different research use-cases may want to generate different data token-level metadata, this approach will be more future-proof.

Additionally, I make a small change to how token probabilities are logged in nucleus sampling. Instead of logging token probs from the truncated (nucleus) distribution, we will now log token probas from the non-truncated distribution.

Finally, I also return normalized probabilities instead of logprobs for token-level probabilities.

**Testing steps**
`pytest tests/test_tga.py`

`parlai dm --model-file zoo:unittest/transformer_generator2/model --truncate 1024 -v --task integration_tests:multiturn_nocandidate -ne 1 --inference beam --beam-size 3`

Screen Shot 2022-03-24 at 6 07 35 PM

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -17,6 +17,7 @@ Contains the following utilities:
 * Beam class which provides some generic beam functionality for classes to use
 """
 
+from typing_extensions import TypedDict
 from parlai.core.params import ParlaiParser
 from abc import ABC, abstractmethod
 from typing import TypeVar, List, Dict, Optional, Tuple, Set, Iterable
@@ -467,7 +468,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
         self.beam_block_full_context = opt.get('beam_block_full_context', False)
         self.temperature = opt.get('temperature', 1.0)
         assert self.temperature > 0, '--temperature must be greater than 0'
-        self.output_token_losses = opt.get(
+        self.show_token_details = opt.get(
             'verbose', False
         ) or 'token_losses' in opt.get('display_add_fields', '')
         self.compute_tokenized_bleu = opt.get('compute_tokenized_bleu', False)
@@ -779,11 +780,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
 
     def _construct_generated_token_details(self, tokens, tokens_metadata):
         tokens_as_txt = [self.dict[int(token)] for token in tokens]
-
-        details_lists = [tokens_as_txt, tokens_metadata["logprobs"].tolist()]
-        if "ranks" in tokens_metadata:
-            details_lists.append(tokens_metadata["ranks"].tolist())
-        return list(zip(*details_lists))
+        return list(zip(tokens_as_txt, tokens_metadata))
 
     def _compute_fairseq_bleu(self, batch: Batch, preds):
         """
@@ -873,7 +870,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
         if batch.label_vec is not None:
             # calculate loss on targets with teacher forcing
             loss, model_output = self.compute_loss(batch, return_output=True)
-            if self.output_token_losses:
+            if self.show_token_details:
                 token_losses = self._construct_label_token_losses(
                     batch.label_vec, model_output
                 )
@@ -896,12 +893,12 @@ class TorchGeneratorAgent(TorchAgent, ABC):
             beam_texts_token_info: List[List[List[Tuple]]] = []
             for beam in beams:
                 beam_texts.append([])
-                if self.output_token_losses:
+                if self.show_token_details:
                     beam_texts_token_info.append([])
 
                 for tokens, score, token_metadata in beam.get_rescored_finished():
                     try:
-                        if self.output_token_losses:
+                        if self.show_token_details:
                             beam_texts_token_info[-1].append(
                                 self._construct_generated_token_details(
                                     tokens, token_metadata
@@ -923,7 +920,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
             else None
         )
 
-        if self.output_token_losses and beam_preds_scores is not None:
+        if self.show_token_details and beam_preds_scores is not None:
             text_token_info = []
             for beam_text_token_info in beam_texts_token_info:
                 text_token_info.append(beam_text_token_info[0])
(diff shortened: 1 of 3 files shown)


emilydinan (commented):
nice @c-flaherty! this will make this code much more flexible!!!

high level comment -- can we call `token_scores` `token_probs` throughout?

also flagging for @stephenroller  to take a look before merging

parlai/core/torch_generator_agent.py
```diff
@@ -1750,27 +1730,33 @@ def select_paths(self, logprobs, prior_scores, current_length) -> _PathSelection
...
                 .argsort(1)
                 .view(-1)
                 .gather(0, best_idxs)
             )
 
+            token_details = []
+
+            for score, rank in zip(tok_scores.cpu().numpy(), tok_ranks.cpu().numpy()):
+                token_details.append(
+                    {"token_score": score.item(), "token_rank": int(rank.item())}
```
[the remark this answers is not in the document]
c-flaherty: To my understanding, TypedDict's are really just dicts that the Python typing system has some extra information about (i.e. you implement them just as dicts), and I typed `token_details` on line 1733. 

From the PEP:
> Methods are not allowed, since the runtime type of a TypedDict object will always be just dict (it is never a subclass of dict).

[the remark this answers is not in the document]
c-flaherty: This looks to be the correct approach from the PEP, and the type system is happy. From the PEP:

> Another example where a type checker should treat a dictionary display as a TypedDict is in an assignment to a variable with a previously declared TypedDict type
```
class Movie(TypedDict):
    name: str
    year: int

movie: Movie
...
movie = {'name': 'Blade Runner', 'year': 1982}

File: parlai/core/torch_generator_agent.py [hunk 2]
```diff
@@ -779,11 +780,7 @@ def _construct_label_token_losses(self, labels, model_output):
 
     def _construct_generated_token_details(self, tokens, tokens_metadata):
         tokens_as_txt = [self.dict[int(token)] for token in tokens]
-
-        details_lists = [tokens_as_txt, tokens_metadata["logprobs"].tolist()]
-        if "ranks" in tokens_metadata:
-            details_lists.append(tokens_metadata["ranks"].tolist())
-        return list(zip(*details_lists))
+        return list(zip(tokens_as_txt, tokens_metadata))
```
stephenroller, line 783: Much better

parlai/core/torch_generator_agent.py
```diff
@@ -1246,15 +1243,19 @@ class _HypothesisTail(object):
...
         self.timestep = timestep
         self.hypid = hypid
         self.score = score
         self.tokenid = tokenid
-        self.token_score = token_score
-        self.token_rank = token_rank
+        self.token_details = token_details
+
+
+class _PathSelectionTokenDetails(TypedDict, total=False):
```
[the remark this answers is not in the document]
stephenroller: Do we need to set total falze now? This seems roughly like it would negate the `__slots__` style speedup?

parlai/core/torch_generator_agent.py
```diff
@@ -1339,16 +1345,12 @@ def __init__(
         ]
 
         self.verbose = verbose
-        self.token_scores, self.token_ranks = None, None
+        self.token_details: Optional[List[List[_PathSelectionTokenDetails]]] = None
```
stephenroller, line 1348: Can you leave a comment explaining this type?

parlai/core/torch_generator_agent.py
```diff
@@ -1750,27 +1728,35 @@ def select_paths(self, logprobs, prior_scores, current_length) -> _PathSelection
...
+                probs.argsort(1, descending=True)
                 .argsort(1)
                 .view(-1)
                 .gather(0, best_idxs)
             )
 
+            token_details = []
+
+            for tok_prob, tok_rank in zip(
+                tok_probs.cpu().numpy(), tok_ranks.cpu().numpy()
```
stephenroller, line 1749: .numpy May implicitly cpu?

stephenroller (approved):
Lgtm

parlai/core/torch_generator_agent.py
c-flaherty: Seems to be a best practice to ensure behavior is the same when tensor may or may not be on GPU: https://discuss.pytorch.org/t/should-it-really-be-necessary-to-do-var-detach-cpu-numpy/35489/6

parlai/core/torch_generator_agent.py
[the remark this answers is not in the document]
c-flaherty: I don't think we get `__slots__` speeds up with typed dicts by default anyways. In fact, this really just changes how the typing system reads this code. From PEP 589:
```
...the runtime type of a TypedDict object will always be just dict (it is never a subclass of dict).
```

tests/test_tga.py
```diff
@@ -190,17 +190,17 @@ def test_token_level_loss_logging(self):
         gold_data = {
             'beam': {
                 'text_token_info': [
-                    ('__start__', 0.0, 1.0),
-                    ('5', -2.5510462364763953e-05, 0.0),
-                    ('__end__', -1.1920922133867862e-06, 0.0),
+                    ('__start__', {"token_prob": 0.0, "token_rank": 1}),
+                    ('5', {"token_prob": 0.999, "token_rank": 0}),
```
stephenroller, line 194: i'd kinda prefer to keep them in logprobs if possible

stephenroller (approved):

tests/test_tga.py
c-flaherty: @emilydinan wanted to do probs not logprobs. Either is good for me though

stephenroller: Logprobs are significantly more accurate...

emilydinan: Logprobs are fine but we were using un-normalized logprobs before. So if we want to use logprobs it will be better to normalize them first

c-flaherty: Okay, I'll do normalized logprobs then!

PR #4428, opened 2022-03-17
[BlenderBot2] Fix ObservationEchoRetriever

jxmsML:
**Patch description**
This is to fix some corner cases in `ObservationEchoRetriever` and `BlenderBot2WizIntGoldDocRetrieverFiDAgent`:
1.
-  the `show_observation_to_echo_retriever` is called twice in `BlenderBot2WizIntGoldDocRetrieverFiDAgent._set_qyert_vec`, which means the `self.model_api.retriever.add_retrieve_doc(observation[self._query_key], retrieved_docs)` is also called twice, which would mess up with the query → idx → doc mapping in `ObservationEchoRetriever`, and  the same docs might be returned for distinct queries in a single batch.
- For example when batch size = 2, in  `add_retrieve_doc`: `new_idx = 0`,  `_query_ids = {query0 : 0}` ,` _saved_docs={0 : docs0}`
- When the same  `show_observation_to_echo_retriever` is called the 2nd time, `new_idx = 1`, `_query_ids={query0 : 1}` , `_saved_docs={1 : docs0}` ; 
- When the 2nd query in the same batch arrives, `new_idx = 1`,  `_query_ids={query0 : 1, query1: 1}` ,` _saved_docs={1: docs1 }`. **_notice that 1: docs0 was overriden_**. After the 2nd call, `new_idx = 2, _query_ids={query0 : 1, query1: 2} , _saved_docs={1 : docs1, 2: docs1}` .
- In this case, the 2 queries share the same docs `docs1` since the new_idx for the new query1 isn't unique when created.

Another issue is that in ObservationEchoRetriever: the `_query_ids` dictionary can grow linearly with the number of unique queries it see during training. 

The fix is to 
- create a new new_idx and make sure it's **unseen**.
- clear the `_query_ids` and `_saved_doc` dictionaries after batch retrieval is done.

**Testing steps**

**Other information**

### DIFF
--- a/parlai/agents/rag/retrievers.py
+++ b/parlai/agents/rag/retrievers.py
@@ -1305,10 +1305,17 @@ class ObservationEchoRetriever(RagRetriever):
         self.n_docs = opt['n_docs']
         self._query_ids = dict()
         self._saved_docs = dict()
+        self._largest_seen_idx = -1
         super().__init__(opt, dictionary, shared=shared)
 
     def add_retrieve_doc(self, query: str, retrieved_docs: List[Document]):
-        new_idx = len(self._query_ids)
+        self._largest_seen_idx += 1
+        new_idx = max(len(self._query_ids), self._largest_seen_idx)
+        if new_idx in self._query_ids.values() or new_idx in self._saved_docs:
+            raise RuntimeError(
+                "Nonunique new_idx created in add_retrieve_doc in ObservationEchoRetriever \n"
+                "this might return the same set of docs for two distinct queries"
+            )
         self._query_ids[query] = new_idx
         self._saved_docs[new_idx] = retrieved_docs or [
             BLANK_DOC for _ in range(self.n_docs)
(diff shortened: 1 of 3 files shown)


parlai/agents/rag/retrievers.py
```diff
@@ -1305,10 +1305,17 @@ def __init__(self, opt: Opt, dictionary: DictionaryAgent, shared: TShared = None
         self.n_docs = opt['n_docs']
         self._query_ids = dict()
         self._saved_docs = dict()
+        self._largest_seen_idx = -1
         super().__init__(opt, dictionary, shared=shared)
 
     def add_retrieve_doc(self, query: str, retrieved_docs: List[Document]):
-        new_idx = len(self._query_ids)
+        self._largest_seen_idx += 1
+        new_idx = max(len(self._query_ids), self._largest_seen_idx)
```
mojtaba-komeili, line 1313: If we are using `self._largest_seen_idx` as a counter that gets reset afterward, why do we still need `len(self._query_ids)`?

jxmsML: ah ok u r right, we don't. fixxing that

mojtaba-komeili (approved):

PR #4439, opened 2022-03-23
Report final evaluation scores mid-way.

stephenroller:
**Patch description**
For risky multitask evals that might crash, it's convenient to have printed the final results for a dataset before the very final report at the end.

**Testing steps**
`parlai em -t blended_skill_talk,empathetic_dialogues` on a private model.


EricMichaelSmith (approved):
Seems reasonable, approving

PR #4447, opened 2022-03-25
SeeKeR

klshuster:
# Patch description
Project code for [SeeKeR](). This PR includes the following:

### SeeKeR Dialogue
- `projects.seeker.agents.seeker:ComboFidAgent`: This agent is used to **train** SeeKeR models, as it is the core agent that handles all of the various functionalities simultaneously.
- `projects.seeker.agents.seeker:SeekerAgent`: The SeeKeR agent itself. See the provided opt preset, `gen/seeker_dialogue`, for an example invocation.

### GPT2 SeeKeR
- `projects.seeker.agents.gpt2_seeker:GPT2WithRetrieverAgent`: This agent packs in the retrieved documents and input context as one big "prompt" to the language model. Utilizes `FidAgent` for retriever components
- `projects.seeker.agents.gpt2_seeker:GPT2ComboAgent`: Agent that can both retrieve for some contexts, and not retrieve for others. This agent is used to *train* GPT2 SeeKeR models.
- `projects.seeker.agents.gpt2_seeker:GPT2SeekerAgent`: Full SeeKeR agent, that handles search query, knowledge, and response generation

## Tasks
- I've included manually constructed teachers for all 30 task variants used to train the SeeKeR dialogue model. They are found  in `projects.seeker.agents.tasks.<dialogue/search_query/search_decision/knowledge>`

### Scripts
- `generate_lm_data.py`: Script used for generating the LM data for `seeker_lm` models.

### Opt Presets
- `arch/r2c2_base_3B`: architecture for R2C2 3B model
- `arch/r2c2_base_400M`: architecture for R2C2 400M model
- `gen/seeker_dialogue`: generation parameters for a SeeKeR Dialogue model
- `gen/seeker_lm`: generation parameters for a SeeKeR (GPT2) LM 

### General Task Additions/Fixes
- Fix the ConvAI2 Normalized teacher to allow the `no_cands` option
- Add `NaturalQuestionsOpenTeacher`

# Testing Steps
Included CI to test the most important functionality.


jaseweston (approved):

projects/seeker/agents/gpt2_seeker.py
```diff
@@ -0,0 +1,323 @@
...
+        opt: Opt, dictionary: DictionaryAgent
+    ) -> "GPT2WithRetrieverModel":
+        return GPT2WithRetrieverModel(opt, dictionary)
+
+
+class GPT2WithRetrieverAgent(FidAgent, GPT2WithRetrieverAgentBase):
+    """
+    GPT2 with Retriever agent.
+
+    This agent packs in the retrieved documents and input context
```
stephenroller, line 49: lint

projects/seeker/scripts/generate_lm_data.py
```diff
@@ -0,0 +1,346 @@
...
+```shell
+$ parlai dd -t projects.seeker.tasks.lm:ResponseTeacher --root-dir /path/to/directory/for/log/saving/valid_split_0.5_f1_overlap
+$ parlai dd -t projects.seeker.tasks.lm:KnowledgeTeacher --root-dir /path/to/directory/for/log/saving/valid_split_0.5_f1_overlap
+$ parlai dd -t projects.seeker.tasks.lm:SearchQueryTeacher --root-dir /path/to/directory/for/log/saving/valid_split_0.5_f1_overlap
+```
+"""
+import nltk
+import os
+import random
+import time
```
stephenroller, line 37: nit: strip unused import

projects/seeker/scripts/generate_lm_data.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def run(self):
+        datatype = self.opt['datatype'].split(':')[0]
+        self.opt['world_logs'] = os.path.join(
+            self.opt['save_dir'],
+            f"{datatype}_split_{self.opt['f1_overlap_threshold']}_f1_overlap",
+        )
+        try:
+            self.generate_data()
+        except:
```
stephenroller, line 237: nit lint

stephenroller (approved):
incredible

projects/seeker/agents/gpt2_seeker_modules.py
```diff
@@ -0,0 +1,553 @@
...
+            scors_i = []
+            # Change this debug later
+            logging.debug(
+                f'URLS:\n{self._display_urls(search_results)}'
+            )  # type: ignore
+            label_text = self.dict.vec2txt(
+                self.label_vec[batch_id, :-1]
+            )  # type: ignore
+            for i, doc in enumerate(search_results):
+                url = doc['url']
```
mojtaba-komeili, line 94: nit: thinking about the variation of the search module used, often they may not return or title. And the KeyError generated here may not be very informative for someone who uses it but doesn't know it's internal. How about checking if all these exist and having an assert that tells user what was expected, or a load warning and then replacing the missing entries with a deafult?

projects/seeker/agents/gpt2_seeker_modules.py
```diff
@@ -0,0 +1,553 @@
...
+):
+    pass
+
+
+#######################################
+# Special Modules/Agents for GPT2 FiD #
+#######################################
+
+
+class IdentityLayer(torch.nn.Module):
```
mojtaba-komeili, line 153: I remember Spencer had added an IdentityLayer to ParlAI recently. Is that possible to use that one instead of the custom one here?

projects/seeker/agents/gpt2_seeker_modules.py
```diff
@@ -0,0 +1,553 @@
...
+
+        if input_turns_cnt is not None:
+            # Input Turns is a tensor of dim [bsz]
+            input = input.repeat_interleave(input_turns_cnt, dim=0)  # type: ignore
+        doc_starts = (enc_out == self.pad_idx).sum(dim=1)  # left padded
+        doc_lens = (seq_len - input_lengths.repeat_interleave(n_docs)) - doc_starts
+        # if no padding, docs are assumed to be min doc length long
+        doc_lens[doc_lens.le(0)] = self.min_doc_len
+        new_enc_out = enc_out.clone()
+        # BEFORE:
```
mojtaba-komeili, line 303: Question: we had this before and after for FiD because we needed to encode the doc with the input. In this module that only does string operations, is it still a reason for doing this way other than using pre-existing methods for making masks etc.?

projects/seeker/agents/gpt2_seeker_modules.py
klshuster: that's not a bad idea; I took this function literally verbatim from the one in `parlai/agents/rag/retrievers.py`; the only change is line 104 to 106. if we want to change that we should go directly to the function there

projects/seeker/agents/gpt2_seeker_modules.py
klshuster: well we're actually still doing operations on tensors, which is why it's a bit confusing. This mess of logic is a result of me trying to format a GPT2 decoder-style model into a FiD setup, rather than the other way around

projects/seeker/agents/seeker.py
```diff
@@ -0,0 +1,905 @@
...
+            agent.history.update_history(
+                observation,
+                temp_history=self.dialogue_agent.get_temp_history(observation),
+            )
+        if (
+            self.search_decision_agent
+            and self.search_decision is SearchDecision.COMPUTE
+        ):
+            self.search_decision_agent.self_observe(self_message)
+            self.search_decision_agent.history.reset()
```
mojtaba-komeili, line 894: Does the decision agent need to self_observe with the history getting reset right after?

projects/seeker/scripts/generate_lm_data.py
```diff
@@ -0,0 +1,346 @@
...
+            except IndexError:
+                # Something's up with the NLTK Sentence tokenizer here.
+                docs += [(i, s) for s in d.split('.')]
+        f1s, inds = torch.FloatTensor(
+            [F1Metric.compute(labels[0], [d]).value() for _, d in docs]
+        ).topk(len(docs))
+        best_doc = None
+        best_doc_idx = None
+        best_f1 = None
+        for f1, ind in zip(f1s, inds):
```
mojtaba-komeili, line 157: The function name says best doc. Why does it looks like it picks the first doc that is above the threshold?

projects/seeker/agents/seeker.py
klshuster: i'm calling `history.reset` directly, so the agent still needs a self observe.

This may be old code though; what i've realized is now I can just call `self.search_decision_agent.reset()`, which will reset the invariants

projects/seeker/scripts/generate_lm_data.py
klshuster: ahh because we assume that the docs are retrieved in order of their quality. so presumably the first doc that is above the threshold is in fact the best doc

mojtaba-komeili (approved):

projects/seeker/scripts/generate_lm_data.py
mojtaba-komeili: They are, but the metric used by the search engine might be different by the F1 metric we use here.

klshuster: agreed, we're also placing our trust more so in the search engine than in a heuristic word overlap metric

Issue #4451, opened 2022-03-26
Using Seeker in chat service


louis030195:
Hi, congratulation for the new model [Seeker](https://parl.ai/projects/seeker/)!

I am trying to use it in Python, but I'm not sure whether I did something wrong or there is an issue, tried to dive into ParlAI code to debug it without success yet, I was hoping maybe you could help me on this?

```bash
# from ParlAI@master root
python3 ./parlai_internal/run.py --config_path parlai_internal/tasks/ava/seeker.yaml --port 8082
```

```py
from parlai.core.params import ParlaiParser
from parlai_internal.services import CustomWebsocketManager
import parlai.chat_service.utils.config as config_utils


SERVICE_NAME = 'websocket'


def setup_args():
    """
    Set up args.
    """
    parser = ParlaiParser(False, False)
    parser.add_parlai_data_path()
    parser.add_websockets_args()
    parser.add_model_args()
    return parser.parse_args()


def run(opt):
    """
    Run MessengerManager.
    """
    opt['service'] = SERVICE_NAME
    manager = CustomWebsocketManager(opt)
    try:
        manager.start_task()
    except BaseException:
        raise
    finally:
        manager.shutdown()


if __name__ == '__main__':
    opt = setup_args()
    config_path = opt.get('config_path')
    config = config_utils.parse_configuration_file(config_path)
    opt.update(config['world_opt'])
    opt['config'] = config
    run(opt)

```

```yaml
tasks:
  default:
    onboard_world: MyOnboardWorld
    task_world: MyTaskWorld
    timeout: 1800
    agents_required: 1
task_name: mytask
world_module: parlai_internal.tasks.mytask.worlds
overworld: MyMessengerOverworld
max_workers: 5
opt:
  models:
    seeker:
      model: projects.seeker.agents.seeker:SeekerAgent
      model_file: zoo:seeker/seeker_dialogue_3B/model
      override:
        no_cuda: True
        search_server: 0.0.0.0:8081
```

```py
from parlai.chat_service.services.websocket.websocket_manager import WebsocketManager
from parlai.chat_service.services.websocket.sockets import MessageSocketHandler
import tornado
from tornado.options import options

class CustomWebsocketManager(WebsocketManager):

    def _make_app(self):
        """
        Starts the tornado application.
        """
        message_callback = self._on_new_message

        options['log_to_stderr'] = True
        tornado.options.parse_command_line([])

        return tornado.web.Application(
            [
                (
                    r"/",
[...]

PR #4458, opened 2022-03-28
[SeeKeR] Opt construction for Python 3.8

klshuster:
**Patch description**
Fixes issue in #4451 

The opt construction for sub-opts for each agent takes the final option string in the parser; for python < 3.8, this would *always* be the `--` option when a `-` was also present. In Python >=3.8, this seems to be non-deterministic.

The new logic ensures we get the long option every time.

**Testing steps**
Tested sample commands from #4451

### DIFF
--- a/projects/seeker/agents/seeker.py
+++ b/projects/seeker/agents/seeker.py
@@ -201,7 +201,7 @@ class SeekerAgent(Agent):
         """
         additional_agent_parser = cls.get_additional_agent_args()
         for action in additional_agent_parser._actions:
-            key = action.option_strings[-1]
+            key = sorted(action.option_strings, key=lambda x: len(x))[-1]
             type = action.type
 
             for prefix in ['krm', 'drm', 'sqm', 'sdm']:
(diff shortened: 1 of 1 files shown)


projects/seeker/agents/seeker.py
```diff
@@ -201,7 +201,7 @@ def add_additional_subagent_args(cls, parser: ParlaiParser) -> ParlaiParser:
         """
         additional_agent_parser = cls.get_additional_agent_args()
         for action in additional_agent_parser._actions:
-            key = action.option_strings[-1]
+            key = sorted(action.option_strings, key=lambda x: len(x))[-1]
```
mojtaba-komeili, line 204: nit: do you think using `max` makes it slightly shorter? You wouldn't need the `[-1]`.

klshuster: oh yeah good point

klshuster: since we're dealing with strings not numbers

stephenroller (commented):
Max instead of sorted...

projects/seeker/agents/seeker.py
mojtaba-komeili: I don't understand, the behavior shouldn't change between the max or sort here.

PR #4468, opened 2022-03-31
train_model: set default   --eval-dynamic-batching to off

jaseweston:
**Patch description**
'eval-dynamic-batching' shuffles the order of the validation set which when using --validation-max-exs may give different results to not using dynamic batching, so this should be by default off I think, to maintain the original order of the validation set.

**Testing steps**
Ran scripts on cluster to verify when eval-dynamic-batching is off, validation numbers are reproducible on commandline without dynamic batching using the original validation order (otherwise they are not)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,26 +222,26 @@ commands:
             - setupcuda
       - fixgit
       - restore_cache:
-          key: deps-20220331-<< parameters.cachename >>-{{ checksum "requirements.txt" }}
+          key: deps-20220328-<< parameters.cachename >>-{{ checksum "requirements.txt" }}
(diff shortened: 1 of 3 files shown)


klshuster:
@stephenroller any idea what's going on with pre-commit here? I've been having my own issues with this `click` and `typer` nonsense with `black` on my own machine (just fully installed parlai as py38)

stephenroller:
Oo saw this in fairseq. Click had its first release in a year. Pinning to 7 fixes.

klshuster:
I can't figure this out...

stephenroller (changes requested):
I don't think we should do this...

PR #4474, opened 2022-04-01
T5 - Add option for using TGA's generation

moyapchen:
Using T5's `_generate()` code is fine, but I want to muck around with some internals of how T5 functions. Using TGA's generation is easier for that.

While there, also do a minor noop change about how we move tensors to device, since it doesn't quite exactly matter. 

Tested by runnning T5 with and without this flag, verified that the generations (using greedy) are the same.


stephenroller (approved):
J/w why would we want to switch between HF and parlai gen?

moyapchen:
To muck with internals of generation without having to modify open source code. 

In hindsight, I maybe could've just used a wrapper class that inherited from T5 and overwritten it there.. but this seemed less inheritance-tree obnoxious.

parlai/agents/hugging_face/t5.py
```diff
@@ -106,7 +106,13 @@ def add_cmdline_args(
...
                 'translation_en_to_ro',
             ],
-            help='Task specific generation config for T5',
+            help='Task specific generation config for T5. Cannot be used if `--tf-hf-generation` is set to False.',
+        )
+        group.add_argument(
+            '--t5-hf-generation',
+            type=bool,
+            default=True,
+            help='Use implementation of generation from Hugging Face or TorchGeneratorAgent',
```
stephenroller, line 115: Clarify which is true

stephenroller: Better yet make it an enum

parlai/agents/hugging_face/t5.py
```diff
@@ -238,10 +247,7 @@ def forward(
             self.stack.parallelize()
         mask = input != self.padding_idx
         outputs = self.stack(input, attention_mask=mask, output_hidden_states=False)
-        for k in outputs:
-            if torch.is_tensor(outputs[k]):
-                outputs[k] = outputs[k].to(input.device)
-        return outputs[0], mask
+        return outputs[0].to(input.device), mask
```
stephenroller, line 250: Cool. Maybe leave a comment saying future uses of outer outputs should also to() the tensors.

stephenroller (approved):

PR #4492, opened 2022-04-09
Removed empty key/value in validate_onboarding

evelynkyl:
I removed `submitted_data` in `validate_onboarding`, as `data` was not in `status_message` and caused an empty value in the `final_status`

**Patch description**
In per_turn_eval's `worlds.py`, the function `validate_onboarding` was trying to get the variable `data` in the dict of `status_message`.But `data` does not exist in the `status_message` and it is causing `final_status` to always be None, therefore blocking all workers. 

Here's the issue https://github.com/facebookresearch/Mephisto/issues/748  where I wanted to reset onboarding qualification, I used a different worker id but found out `validate_onboarding` ended up keep blocking me because the `final_status` is always None. 

**Testing steps**
Launch `run.py` and do onboarding. Perform onboarding correctly and incorrectly to test qualified and disqualified workers and check if `data` ever exists in `status_message` for this task.

### DIFF
--- a/parlai/crowdsourcing/tasks/pairwise_per_turn_eval/worlds.py
+++ b/parlai/crowdsourcing/tasks/pairwise_per_turn_eval/worlds.py
@@ -517,10 +517,7 @@ def validate_onboarding(data):
     status_message = messages[-2]
     if status_message is None:
         return False
-    submitted_data = status_message.get('data')
-    if submitted_data is None:
-        return False
-    final_status = submitted_data.get('final_status')
+    final_status = status_message.get('final_status')
     return final_status == ONBOARD_SUCCESS
 
 


JackUrb:
@EricMichaelSmith I'm surprised that per-turn eval was passing here before - does the current test not include onboarding?

Something we'll be looking at more as we develop an E2E testing framework for Mephisto tasks.

EricMichaelSmith (commented):
Yeah, I think this fix is right, thanks @evelynkyl for adding - if you look at https://github.com/facebookresearch/ParlAI/pull/4426/files, this same fix was done for `model_chat` as part of the Mephisto 1.0 upgrade. However, it looks like a bunch of CI checks are failing now, for reasons that seem to be unrelated to this PR - can you maybe try doing a `git merge` from the main branch on `facebookresearch` to see if the breakages are just related to not having the most recent commits?

@JackUrb I had been afraid of this - unfortunately we haven't had time to put a unit test on per-turn eval yet. I'll ask around to see if anyone else has bandwidth for this currently

JackUrb:
@EricMichaelSmith Feel free to defer for now - Mephisto shouldn't have any breaking changes prior to the E2E integration testing PR, so work fixing now would likely just be easier after that framework is provided.

JackUrb:
@EricMichaelSmith We should probably approve and merge this! Failing tests here seem unrelated to these changes.

EricMichaelSmith:
> @EricMichaelSmith We should probably approve and merge this! Failing tests here seem unrelated to these changes.

@JackUrb Yeah fair enough - I just created a new PR for this at https://github.com/facebookresearch/ParlAI/pull/4560 and tested that it fixes the issue. Closing this PR given that new one

PR #4497, opened 2022-04-10
Add TB logging to parlai eval_model script

DrMatters:
**Patch description**
This patch adds the capability to report evaluation results to the tensorboard when using `parlai train_model` script.

**Testing steps**
Ran
```parlai eval_model -t blended_skill_talk,empathetic_dialogues -mf zoo:blender/blender_90M/model --skip-generation True --tensorboard-log True --tensorboard-logdir ~/tb_reporting_test/ --bs 32 --datatype valid``` and with `--datatype test`

**Other information**
I was fine-tuning a blenderbot when WandB suddenly failed (probably due to the unintended launch of the second fine-tuning loop on the same model). After that, the tuning process worked fine and finished properly. **BUT** more important thing is that the final test-set evaluation was never done (some WandB exception caused the failure). So I decided to evaluate the model on the test set using the dedicated `eval_model` script and then use `wandb sync` on the generated Tensoflow report.

### CODE REVIEW

**Comment by DrMatters** [comment]
#4495

**Comment by DrMatters** [comment]
@stephenroller 
I've made a patch as you asked here: https://github.com/facebookresearch/ParlAI/issues/4495#issuecomment-1094338292

**Comment by stephenroller** [comment]
How does this render? As a single data point in the time series?

**Comment by DrMatters** [comment]
> How does this render? As a single data point in the time series?

Yup!
This is still useful as it allows seamlessly `wandb sync` the tensorboard report into the existing run of wandb

image

**Review by stephenroller — APPROVED** [review]

File: parlai/scripts/eval_model.py [hunk 1]
```diff
@@ -144,6 +144,33 @@ def get_task_world_logs(task, world_logs, is_multitask=False):
         return f'{base_outfile}_{task}{extension}'
 
 
+def prepare_tb_logger(opt):
+    if opt['tensorboard_log'] and is_primary_worker():
+        tb_logger = TensorboardLogger(opt)
+
```
stephenroller, line 150:
```suggestion
```

stephenroller (approved):

DrMatters:
What about failed tests?

stephenroller:
> What about failed tests?

Pretty sure fixed in main.

PR #4503, opened 2022-04-14
[GoldDocFidAgent] clear mapping at the end of each eval_step rather than retrieve step

jxmsML:
**Patch description**
When running `parlai em -m BlenderBot2WizIntGoldDocRetrieverFiDAgent --skip-generation False` the `model.encoder` will be called twice, however the `_saved_docs` and `_query_ids` mapping in the observation_echo_retriever will get reset after the retrieval step in the 1st `model.encoder`, and one will run into KeyError when trying to access the `_saved_docs` again during the 2nd `model.encoder`.

The fix is to clear the two mappings at the end of the batch_act

**Testing steps**

**Other information**

### DIFF
--- a/parlai/agents/fid/fid.py
+++ b/parlai/agents/fid/fid.py
@@ -376,6 +376,12 @@ class GoldDocRetrieverFiDAgent(SearchQueryFiDAgent):
         self.show_observation_to_echo_retriever(observation)
         super()._set_query_vec(observation)
 
+    def eval_step(self, batch):
+        output = super().eval_step(batch)
+        if hasattr(self.model_api.retriever, 'clear_mapping'):
+            self.model_api.retriever.clear_mapping()
+        return output
+
 
 class WizIntGoldDocRetrieverFiDAgent(GoldDocRetrieverFiDAgent):
     """
--- a/parlai/agents/fid/fid.py
+++ b/parlai/agents/fid/fid.py
@@ -348,9 +348,9 @@ class GoldDocRetrieverFiDAgent(SearchQueryFiDAgent):
                 'GoldDocRetrieverFiDAgent only works with `rag_retriever_query` being `"full_history"`. '
                 f'Changing opt value for `rag_retriever_query`: `"{prev_sel}"` -> `"full_history"`'
             )
-        if (
-            opt['dynamic_batching'] == 'full'
-            or opt.get('eval_dynamic_batching') == 'full'
+        if not (
+            opt['dynamic_batching'] in [None, 'off']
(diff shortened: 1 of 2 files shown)


parlai/agents/fid/fid.py
```diff
@@ -376,6 +376,12 @@ def _set_query_vec(self, observation: Message) -> Message:
         self.show_observation_to_echo_retriever(observation)
         super()._set_query_vec(observation)
 
+    def eval_step(self, batch):
+        output = super().eval_step(batch)
+        if hasattr(self.model_api.retriever, 'clear_mapping'):
```
mojtaba-komeili, line 381: This looks like a good solution. But what about cleaning on batchify that you suggested?

mojtaba-komeili (approved):

parlai/agents/fid/fid.py
jxmsML: I realize I can't do that in batchify, since model act flow is observe (insert k,v pairs to the `_saved_doc` and `_query_ids`) -> batchify -> act;
if I clear up the mapping during batchify then that literally clear up all the newly inserted kv pairs that are from the same batch.

Also notice another problem with this change (it won't reset the mapping in train step), let me update the pr and ping you guys again after it's ready.

mojtaba-komeili: What if we clear after act/batch_act from the FiD agent?

parlai/agents/fid/fid.py
```diff
@@ -348,7 +348,13 @@ def __init__(self, opt: Opt, shared: TShared = None):
...
                 f'Changing opt value for `rag_retriever_query`: `"{prev_sel}"` -> `"full_history"`'
             )
-
+        if (
+            opt['dynamic_batching'] == 'full'
+            or opt.get('eval_dynamic_batching') == 'full'
+        ):
+            raise RuntimeError(
+                "For now dynamic batching doesn't work with ObservationEchoRetriever as it cleans up _saved_docs mapping after each batch act."
+            )
```
klshuster, line 357: note that `dynamic_batching` can also be `'batchsort'`

klshuster (approved):
this looks great, thanks!

parlai/agents/fid/fid.py
jxmsML: Thanks for pointing that out. fixed

parlai/agents/fid/fid.py
jxmsML: yea that's my sol rn.

PR #4505, opened 2022-04-15
[T5] T5 with DDP backend

klshuster:
**Patch description**
Handle t5 DDP training with `--ddp-backend ddp`

**Testing steps**
CI that includes generation during eval, where this originally broke


stephenroller (approved):

Rebecca-Qian:
Thanks for this fix!

PR #4508, opened 2022-04-23
Fix default tokenizer for HuggingFace Agents

Rebecca-Qian:
**Patch description**
Updates the HuggingFace agent to over-ride `is_prebuilt` to return True, since HuggingFace agents build their own dictionaries.

Using the default tokenizer setting with T5 caused the exception `ValueError('Dictionaries should be pre-built before distributed train.')`

**Testing steps**
```
parlai mp_train -m parlai.agents.hugging_face.t5:T5Agent -mf /tmp/model_file -t jsonfile --jsonfile-datapath $PATH --fp16 true --t5-model-parallel False --ddp-backend zero2 --jsonfile-datatype-extension True
```


klshuster (approved):
thank you for this!!

stephenroller (approved):
yay

PR #4513, opened 2022-04-26
Delete BST aux teachers

stephenroller:
**Patch description**
After discussion as a group, we've decided to retire these teachers instead of updating them.

**Testing steps**
CI


EricMichaelSmith (approved):
Approved, assuming that this is all of the references to the PersonaTopicifier teachers, and conditional on CI checks passing of course :P I see some lint errors for unused imports, for instance

PR #4514, opened 2022-04-26
Archive detectron dependencies

stephenroller:
**Patch description**
As per discussion in the group, deprecate the multimodal blenderbot due to its unweidly mask r-cnn dependency.

**Testing steps**
CI


klshuster (approved):
looks like there are some imports you can remove from `image_featurizers.py` to appease lint

PR #4515, opened 2022-04-26
add new test set to saferdialogues task

meganung:
**Patch description**
add the newly collected test set that is parallel with the BAD test set.

### DIFF
--- a/parlai/tasks/saferdialogues/README.md
+++ b/parlai/tasks/saferdialogues/README.md
@@ -16,4 +16,6 @@ Note: The dataset is flattened, so there is one example per episode.
 
 If the `--recovery` flag is set to `false` (`true` by default) then the recovery response is omitted and the labels contains the signaling message and the text contains the context lines before that.
 
+Use the `--bad` flag and set to `true` (`false` by default) to use a different test set that is paralell to the BAD dataset's test split.
+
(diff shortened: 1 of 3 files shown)


parlai/tasks/saferdialogues/README.md
```diff
@@ -16,4 +16,6 @@ Note: The dataset is flattened, so there is one example per episode.
 
 If the `--recovery` flag is set to `false` (`true` by default) then the recovery response is omitted and the labels contains the signaling message and the text contains the context lines before that.
 
+Use the `--bad` flag and set to `true` (`false` by default) to use a different test set that is paralell to the BAD dataset's test split.
```
jxmsML, line 19: is the new addition just for test split, or for both valid and test?

meganung: just for test split

stephenroller, line 19: Spelling

parlai/tasks/saferdialogues/agents.py
```diff
@@ -35,6 +37,12 @@ def add_cmdline_args(
             default=True,
             help="Whether or not to include the recovery utterance",
         )
+        agent.add_argument(
+            '--bad',
```
stephenroller, line 41: I kinda suggest we make a new teacher instead of hiding this behind a flag

PR #4516, opened 2022-04-26
[tasks/huggingface] Way to add to message but not "text"

moyapchen:
I want to be able to add my own mutators to grab information from other fields out of a huggingface dataset. Since I don't want it to show up in the text itself, I do it in a new attribute.

I also delete some code that was written at some point but isn't used anywhere while I'm here.

Test Plan:
Run `parlai dd -t` with a new dataset that I'm working on that uses this.

Also run `parlai dd -t glue; parlai dd -t superglue` to validate that I don't break anything that already exists. (Former dies, but it's cause of a dataset issues on the HF side where the size of the datasets have changed, so not worrying about it)

### DIFF
--- a/parlai/tasks/huggingface/agents.py
+++ b/parlai/tasks/huggingface/agents.py
@@ -15,12 +15,6 @@ import os
 import datasets
 
 
-class SplitsMappingDict(TypedDict):
-    train: str
-    valid: str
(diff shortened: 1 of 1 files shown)


meganung (approved):
lgtm !!

PR #4532, opened 2022-05-08
remove package-lock file 

meganung:
Remove unnecessary package-lock file- it causes a bunch of vulnerability warnings and also should be automatically created from building the package.


EricMichaelSmith (approved):
Oof, yes, agreed

JackUrb:
You'll want to add these to `.gitignore` or they may inevitably be re-added in the future.

stephenroller (approved):
Tyyy

PR #4536, opened 2022-05-10
Fix end time update rule for TimerMetric

zhangmozhi:
**Patch description**
Fix end time update rule for TimerMetric.

**Testing steps**

**Other information**
This doesn't seem to cause any bug at the moment, but it seems like a typo.


stephenroller (approved):
ty

PR #4554, opened 2022-05-19
[torchscript] use device type when concatenating tensors

maryamdaneshi:
**Patch description**

there were errors on some predictor gpu machines for device mismatch so explicitly using the device type when concatenating the tensors

**Testing steps**
Fixing the error
before

281281826_325898053026523_7071148804019578843_n

after:
Screen Shot 2022-05-18 at 4 44 11 PM

**Other information**


stephenroller (approved):
What a wild ride

PR #4562, opened 2022-05-25
fixing cache keys

Golovneva:
**Patch description**
Updating cache keys


klshuster (approved):
approving assuming tests pass

PR #4565, opened 2022-06-01
cache api calls and responses on setup

macota:
**Patch description**
Cache api calls to responses dictionary on setup for multiwoz_v22

**Testing steps**
Unit tests all pass
image

Train dataset output validation:
Original output on left, new output on right
image

Test dataset output:
image

Valid dataset output:
image

**Other information**
Unit tests run in 1/4 the time with caching vs without
~63 seconds vs ~257 seconds

### DIFF
--- a/parlai/tasks/multiwoz_v22/agents.py
+++ b/parlai/tasks/multiwoz_v22/agents.py
@@ -15,7 +15,6 @@ import pandas as pd
 from parlai.core.opt import Opt
 import parlai.core.tod.tod_core as tod
 import json
-import pickle
 from typing import Optional
 from parlai.utils.data import DatatypeHelper
 from parlai.utils.io import PathManager
@@ -265,8 +264,7 @@ class MultiwozV22Parser(tod_agents.TodStructuredDataParser):
                                     continue
 
                             call = maybe_call
-                            call_key = frozenset(call.items())
-
+                            call_key = str(call)
                             if call_key not in self.call_response_cache:
                                 resp = self._get_find_api_response(
                                     intent,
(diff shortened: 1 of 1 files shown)


parlai/tasks/multiwoz_v22/agents.py
```diff
@@ -300,8 +311,23 @@ def setup_episodes(self, fold):
         """
         Parses into TodStructuredEpisode.
         """
-        self.dbs = self.load_dbs()
         self.schemas = self.load_schemas()
+        valid_path = self.dpath + "/valid_call_response_cache"
+        train_path = self.dpath + "/train_call_response_cache"
+        test_path = self.dpath + "/test_call_response_cache"
+        if fold == "valid" and os.path.isfile(valid_path):
```
moyapchen, line 318: Nits here - 

1. Would use `PathManager` rather than `os` for opening + closing files (there's a couple of extra filesystem safety checks with `PathManager`) 
2. Additionally, would use `os.path.join` for handling file names that join paths, rather than using string manipulation directly (also would put in a file extension :) )
3. Since your fold is directly in the naming of the cache file, I'd get rid of some of the cases in this (and the other) if block by leveraging this. (Ex. doing something like `path = os.path.join(self.dpath, f"{fold}_call_response_cache.pickle")`)

moyapchen (changes requested):
Logic looks right based on what we chatted about; couple of coding/safety nits. 

Out of curiosity, are the numbers in the Summary from the first time loading the dataset or using the cache for subsequent calls?

moyapchen:
..and which datatype is that?

macota:
> ..and which datatype is that?
unit tests cover all three looks like

moyapchen:
Ahhh brainfart there; eyes somehow skipped "pytest" there in "first pytest run"

parlai/tasks/multiwoz_v22/agents.py
```diff
@@ -345,6 +363,11 @@ def setup_episodes(self, fold):
                 rounds=rounds,
             )
             episodes.append(episode)
+
+        with PathManager.open(cache_path, 'w') as f:
+            json_cache_data = json.dumps(self.call_response_cache)
```
moyapchen, line 368: nit - `json.dump()` will let you do a one-liner without a separate `f.write()`

macota: good catch, that was a vestige from something I was playing with after having to change `call_key` from a frozenset to a string since json didn't like the former

moyapchen (approved):
Thanks!

PR #4575, opened 2022-06-07
Pearlli model chat analysis refactor

pearlli98:
**Patch description**
1. refactor crowdsourcing model_chat analysis script to use mephisto databrowser's `get_data_from_unit()` function instead of loading from a specific location on disk. 
2. Changes on CSV: (1) no longer contains a folder column (2) worker_id is changed to numeric result from `task_unit['worker_id']` instead of the string placeholder "--NOT-MTURK-AGENT-x_sandbox"

**Testing steps**

**Other information**


parlai/crowdsourcing/tasks/model_chat/analysis/compile_results.py
```diff
@@ -15,6 +15,7 @@
 
 from parlai.crowdsourcing.utils.acceptability import AcceptabilityChecker
 from parlai.crowdsourcing.utils.analysis import AbstractTurnAnnotationResultsCompiler
+from parlai.crowdsourcing.tasks.model_chat.model_chat_blueprint import BLUEPRINT_TYPE
```
EricMichaelSmith, line 18: To get rid of this lint error, you can add the `_ = BLUEPRINT_TYPE` line that exists in the per-turn eval analysis script

pearlli98: why do we need `--task-name`? Where is this used?

EricMichaelSmith: Hmm - you need to specify it for use with https://github.com/facebookresearch/ParlAI/blob/7660197631c06ee41c0ce25c5de37a74b11ecb77/parlai/crowdsourcing/utils/analysis.py#L107 , which your code now calls via `self.get_task_data()`

PR #4577, opened 2022-06-07
Pearlli model chat analysis refactor

pearlli98:
**Patch description**

1. refactor crowdsourcing model_chat analysis script to use mephisto databrowser's get_data_from_unit() function instead of loading from a specific location on disk.
2. Changes on CSV: (1) no longer contains a folder column (2) worker_id is changed to numeric result from task_unit['worker_id'] instead of the string placeholder "--NOT-MTURK-AGENT-x_sandbox"
3. change unit test to include a dummy class that inherits the `ModelChatResultsCompiler` class such that its `get_task_data()` mimics databrowser outputs. 
4. update test data
5. remove `results-folders` flag from `AbstractTurnAnnotationResultsCompiler` and added to the subclasses that still need it (`ModelChatResultsCompiler` and `TurnAnnotationsStaticResultsCompiler`)

**Testing steps**
`pytest ~/ParlAI/tests/crowdsourcing/tasks/model_chat/test_model_chat_analysis.py`


parlai/crowdsourcing/tasks/model_chat/analysis/compile_results.py
```diff
@@ -15,6 +13,9 @@
 
 from parlai.crowdsourcing.utils.acceptability import AcceptabilityChecker
 from parlai.crowdsourcing.utils.analysis import AbstractTurnAnnotationResultsCompiler
+from parlai.crowdsourcing.tasks.model_chat.model_chat_blueprint import BLUEPRINT_TYPE
+
+_ = BLUEPRINT_TYPE
```
EricMichaelSmith, line 18: Probably good to add a comment here saying that you're just importing BLUEPRINT_TYPE to force registration of the blueprint, but that you're not using this var itself

parlai/crowdsourcing/tasks/model_chat/analysis/compile_results.py
```diff
@@ -137,262 +109,255 @@ def compile_results(self) -> pd.DataFrame:
 
         conversation_idx = 0
         conversation_dfs = []
-        for read_folder in read_folders:
-            read_folder_name = os.path.split(read_folder)[-1]
-            for file_name in sorted(os.listdir(read_folder)):
-                if file_name in self.hit_block_list:
```
EricMichaelSmith, line 143: ooof - since we're not using `self.hit_block_list` anymore in this script, I think we should remove the `--hit-block-list` flag from this script and move it into `parlai/crowdsourcing/projects/multisession_chat/human_eval/compile_results.py`, which has the only subclass of this script `ModelChatResultsCompiler` that will still need to use `self.hit_block_list`

tests/crowdsourcing/tasks/model_chat/test_model_chat_analysis.py
```diff
@@ -23,6 +26,42 @@
     )
     from parlai.crowdsourcing.utils.tests import check_stdout
 
+    class TestModelChatResultsCompiler(ModelChatResultsCompiler):
```
EricMichaelSmith, line 29: ya this is great - and I love the typing of the output :)

tests/crowdsourcing/tasks/model_chat/test_model_chat_analysis.py
```diff
@@ -23,6 +26,42 @@
...
+        def get_task_data(self) -> List[Dict[str, Any]]:
+            fake_jsons = []
+            read_folders = []
+            date_strings = []
```
EricMichaelSmith, line 33: super minor, but it'd be cleaner to skip the initialization of these here - date_strings is re-initialized below, and `read_folders` is effectively set to be equal to `folders`

tests/crowdsourcing/tasks/model_chat/test_model_chat_analysis.py
```diff
@@ -23,6 +26,42 @@
...
+            read_folders.extend(folders)
+
+            for read_folder in read_folders:
+                for file_name in sorted(os.listdir(read_folder)):
+                    # Read in file
+                    with open(os.path.join(read_folder, file_name), 'rb') as f:
+                        data = json.load(f)
+                        worker_id = data['workers'][0]
+                        assignment_id = data['assignment_ids'][0]
+                        fake_jsons.append(
```
EricMichaelSmith, line 54: Very effective faking! 🎉

tests/crowdsourcing/tasks/model_chat/test_model_chat_analysis.py
```diff
@@ -62,14 +101,12 @@ def setup_teardown(self):
...
---output-folder {tmpdir} \
-{flag_string}
-"""
-                        parser_ = ModelChatResultsCompiler.setup_args()
+                        arg_string = f"""--output-folder {tmpdir} {flag_string}"""
+                        parser_ = TestModelChatResultsCompiler.setup_args()
                         args_ = parser_.parse_args(arg_string.split())
-                        ModelChatResultsCompiler(vars(args_)).compile_and_save_results()
+                        compiler = TestModelChatResultsCompiler(vars(args_))
+                        compiler.results_folder = analysis_samples_folder
```
EricMichaelSmith, line 108: 👍

EricMichaelSmith (commented):
Yeah this looks great! Ha, you handled these edge cases and got the tests to pass much more smoothly than what I was worried about :) I have a few comments but nothing too major, just some cleanups

@JackUrb @pringshia This is working towards the goal of eventually getting all of ParlAI's tasks onto `DataBrowser` and away from a brittle loading of raw outputs from disk

PR #4582, opened 2022-06-10
Pearlli streamline model chat unit test

pearlli98:
**Patch description**

1. removed all logic inside [._check_final_chat_data()](https://github.com/facebookresearch/ParlAI/blob/989e29ff8d7a9606e2bbc7db7290b58fe9b49017/parlai/crowdsourcing/tasks/model_chat/utils.py#L398) and [._check_output_key()](https://github.com/facebookresearch/ParlAI/blob/989e29ff8d7a9606e2bbc7db7290b58fe9b49017/parlai/crowdsourcing/tasks/model_chat/utils.py#L385) in the class `AbstractModelChatTest` into [._remove_non_deterministic_keys()](https://github.com/facebookresearch/ParlAI/blob/989e29ff8d7a9606e2bbc7db7290b58fe9b49017/parlai/crowdsourcing/tasks/model_chat/utils.py#L338).
2. changed 4 unit tests `test_model_chat.py`, `test_model_image_chat.py`, `test_demo_chat.py`, and `test_qa_data_collection.py` to use pytest regressions, which replaces `._check_output_key()` in `AbstractParlAIChatTest`
3. deleted expected_states for the above unit tests and added .yml file generated by pytest regressions
4. update `_test_agent_states()` signature to no longer include `expected_states` because it is not needed

Tests:

1. `pytest ~/ParlAI/tests/crowdsourcing/tasks/qa_data_collection/test_qa_data_collection.py`
2. `pytest ~/ParlAI/tests/crowdsourcing/tasks/test_chat_demo.py`
3. `pytest ~/ParlAI/tests/crowdsourcing/tasks/model_chat/test_model_chat.py`
4. `pytest ~/ParlAI/tests/crowdsourcing/tasks/model_chat/test_model_image_chat.py`


EricMichaelSmith:
Hmm could you add the commands to run the 4 unit tests that you've ported over?

EricMichaelSmith:
Ah, make sure to delete the unused imports inside `parlai/crowdsourcing/utils/tests.py` :)

tests/crowdsourcing/tasks/model_chat/expected_states/state.json
```diff
@@ -272,6 +272,7 @@
                         "annotations_config": "{\n    \"config\": {\n        \"bucket_0\": {\n            \"name\": \"Bucket 0\",\n            \"description\": \"this response implies something...0\"\n        },\n        \"bucket_1\": {\n            \"name\": \"Bucket 1\",\n            \"description\": \"this response implies something...1\"\n        },\n        \"bucket_2\": {\n            \"name\": \"Bucket 2\",\n            \"description\": \"this response implies something...2\"\n        },\n        \"bucket_3\": {\n            \"name\": \"Bucket 3\",\n            \"description\": \"this response implies something...3\"\n        },\n        \"bucket_4\": {\n            \"name\": \"Bucket 4\",\n            \"description\": \"this response implies something...4\"\n        },\n        \"none_all_good\": {\n            \"name\": \"None, all good\",\n            \"description\": \"This response implies that there are no problems with the data\"\n        }\n    },\n    \"show_line_breaks\": false\n}\n",
                         "model_nickname": "fixed_response",
                         "model_opt": {
+                            "is_debug": false,
```
EricMichaelSmith, line 275: Hmm - this file and state__image_chat.json can also be deleted because they've been supplanted by the outputs from pytest regressions, right?

tests/crowdsourcing/tasks/model_chat/test_model_chat.py
```diff
@@ -132,20 +133,8 @@ def test_base_task(self):
...
-                with open(expected_chat_data_path) as f:
-                    expected_chat_data = json.load(f)
-                results_path = list(
-                    glob.glob(os.path.join(chat_data_folder, '*/*_*_*_sandbox.json'))
-                )[0]
-                with open(results_path) as f:
-                    actual_chat_data = json.load(f)
-                self._check_final_chat_data(
-                    actual_value=actual_chat_data, expected_value=expected_chat_data
+                    data_regression=data_regression,
```
EricMichaelSmith, line 137: It looks like `expected_state` on line 129 is never used, because of course we're using data_regression here - I'd get rid of that

tests/crowdsourcing/tasks/model_chat/test_model_chat/test_base_task.yml
```diff
@@ -0,0 +1,386 @@
+- inputs: {}
```
EricMichaelSmith, line 1: Awesome :)

tests/crowdsourcing/tasks/model_chat/test_model_chat/test_base_task.yml
```diff
@@ -0,0 +1,386 @@
...
+            is_debug: false
+            loglevel: info
+            model: fixed_response
+            model_parallel: true
+            multitask_weights:
+            - 1
+            task: null
+            verbose: false
+        workers:
+        - --NOT-MTURK-AGENT-MOCK_WORKER_0_sandbox
```
EricMichaelSmith, line 239: oh ha technically we should change this given your other PR :P

tests/crowdsourcing/tasks/qa_data_collection/test_qa_data_collection/test_base_task.yml
```diff
@@ -0,0 +1,31 @@
+- inputs: {}
```
EricMichaelSmith, line 1: great! But we can remove `expected_states/` now because we have this file instead

EricMichaelSmith (commented):
Awesome! There's a ton of great work here :) I know you're still working on stuff, so I'll await your other changes

EricMichaelSmith:
Hmm, I'll try to look into this on Friday if I have time - the PR is all ready to go except for a crowdsourcing check failing on the CI box but not on a devfair, and I want to understand why this is

klshuster:
Any updates on this PR?

EricMichaelSmith:
> Any updates on this PR?

Not yet, will merge from main and look into this again later today, thanks for the ping

JackUrb:
I'm having a hard time parsing how the above is occurring. It would appear two agents are being established, however I would need more log info to debug, like the actual agent that's present in the above print call. I haven't really seen this type of behavior though, considering the world is actually playing out I find it really strange that data isn't getting saved here, especially if this isn't reproducible on devfair and only occurs on CI.

EricMichaelSmith:
> I'm having a hard time parsing how the above is occurring. It would appear two agents are being established, however I would need more log info to debug, like the actual agent that's present in the above print call. I haven't really seen this type of behavior though, considering the world is actually playing out I find it really strange that data isn't getting saved here, especially if this isn't reproducible on devfair and only occurs on CI.

Yeah, it *sometimes* but not always occurs when I run on devfair, and it always occurs on CI. Hmm - @JackUrb do you happen to know offhand if there's another way to find agents other than `self.db.find_agents()`, to tell if/how two different agents are being established? If not, no worries - I'll have to dig into this a lot more when I can. @klshuster if I'm too slow on fixing this CI test, it's fine with me if you want to close the PR for cleanliness purposes - Pearl worked hard on this and the streamlining that she made to the unit test will make it much easier to maintain these tests in the future, but since it's a relatively lopri fix I might not be able to work on it more immediately

parlai/crowdsourcing/utils/tests.py
```diff
@@ -299,7 +300,7 @@ def _test_agent_states(
                 agent_task_data.append([{}] * len(message_round))
 
         # Set up the mock human agents
-        agent_ids = self._register_mock_agents(num_agents=num_agents)
+        agents, agent_ids = self._register_mock_agents(num_agents=num_agents)
```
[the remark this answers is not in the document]
EricMichaelSmith: Ah! Cool, that makes sense, and that would explain why model_image_chat, which has a 20-second hold time for the image to arrive, is broken but model_chat isn't. Will try that, thanks :)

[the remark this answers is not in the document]
EricMichaelSmith: Okay, @JackUrb this is pretty strange - I've added a lot of print statements to track when the messages get saved to and loaded from disk, and it seems like the issue is that the messages are getting saved in the test class's `_teardown()` method, presumably during `self.operator.force_shutdown()` or `self.server.shutdown_mock()`. You can see this in the full logs of https://app.circleci.com/pipelines/github/facebookresearch/ParlAI/11871/workflows/96af4bb1-1b17-430d-95ae-b74ff45ccac0/jobs/96304/parallel-runs/0/steps/0-111 - the test tried to read in the state file at `2022-08-19 14:43:00`, but the per-turn messages didn't saved until 5 seconds later, at `2022-08-19 14:43:05`, in the stdout teardown.

Does this sound familiar, that the state might not save messages until operator/server shutdown? If so, I wonder if we should be calling shutdown during the test itself, not during test teardown, but I'm not sure if that'll have side effects...

klshuster:
@JackUrb @EricMichaelSmith is there any update here?

EricMichaelSmith:
> @JackUrb @EricMichaelSmith is there any update here?

@JackUrb I'm finding that the test fails because the messages aren't getting saved until the test class's `._teardown() `method, presumably during `self.operator.force_shutdown()` or `self.server.shutdown_mock()`, and thus the part of the test that reads the messages back in can't see the messages because they aren't saved yet. Is this familiar to you at all?

@klshuster Urrg... looking at my timeline I realistically may not have time to debug this further until mid-October-ish, given ICLR, BlenderBot integrity, EMNLP revisions, PTO, etc. Feel free to close it if you'd like, and I have a note to come back to it later when I have time!

JackUrb:
> I'm finding that the test fails because the messages aren't getting saved until the test class's `._teardown()` method, presumably during `self.operator.force_shutdown()` or `self.server.shutdown_mock()`, and thus the part of the test that reads the messages back in can't see the messages because they aren't saved yet. Is this familiar to you at all?

Not directly familiar, but feels a little reminiscent of some of the bugs where I had to update tests to run things in async contexts to be sure that the background (async) thread was able to execute the pending jobs. [Here's](https://github.com/facebookresearch/Mephisto/blob/main/test/core/test_live_runs.py#L145-L212) an example of some of the helpers we've put together for this type of test on the Mephisto side, though they're not in the public api yet. 

Admittedly this was prior to me learning how to make `asyncio` test fixtures, so perhaps the longer-term solution will be to expose an `operator._await_pending_activity()` (closest to [this one](https://github.com/facebookresearch/Mephisto/blob/48c0c23f79f4ea0d94d500831b44b39c7d168b17/test/core/test_live_runs.py#L196-L201), but with actual task filtering) that could be used in tests, and then that could be used here.

PR #4593, opened 2022-06-15
fixing minor issues rebased

Golovneva:
**Patch description**
Adding/fixing comments. Fixing main to make it pass tests.

**Testing steps**

**Other information**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,26 +208,26 @@ commands:
             - setupcuda
       - fixgit
       - restore_cache:
-          key: deps-20220519-<< parameters.cachename >>-{{ checksum "requirements.txt" }}
+          key: deps-20220615-<< parameters.cachename >>-{{ checksum "requirements.txt" }}
       - setup
       - installdeps
       - << parameters.more_installs >>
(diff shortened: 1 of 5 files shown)


klshuster:
We removed the `tensorboard` from `requirements.txt` yesterday, and I think this fixes the `cleaninstall` issues; thus, I'm not sure this PR is necessary anymore

Golovneva:
> We removed the `tensorboard` from `requirements.txt` yesterday, and I think this fixes the `cleaninstall` issues; thus, I'm not sure this PR is necessary anymore

Not sure what do you mean. Main is failing. You can also look at previous runs in this commit thread on latest main, they are all failing. `tensorboard` removal introduced new breakages

klshuster:
i think those are spurious errors; it seems that #4590 had passing tests when it was landed

Golovneva:
Will create a PR with text changes only

klshuster (approved):
approving given everything passes. Thanks!

Golovneva:
Huh, still failed in mainline. Is it possible that some other changes are required in circleci config?

PR #4615, opened 2022-06-18
trying prev version

Golovneva:
**Patch description**
Reverted setuptools to the version that didn't fail 
Reverted removal for the tensorflow and protobuf dependencies that didn't help to resolve the failure, but somehow happened to clear tests

In [v62.4.0 version of setuptools](https://setuptools.pypa.io/en/latest/history.html) they [added a new command](https://github.com/pypa/setuptools/pull/3256) "setuptools.command.build" to match distutils.command.build . For some reason, our pipeline fails to find corresponding scripts. With this commit, including temporary workaround that reverts setuptools to its previous state. Ideally we would need to figure why circleCI can't see updates, although logs indicate that newest version was installed.

Failures:
Screen Shot 2022-06-17 at 11 08 16 PM

Corresponding version:
Screen Shot 2022-06-17 at 11 08 54 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,7 +91,7 @@ commands:
       - run:
           name: Installs basic dependencies
           command: |
-            python -m pip install --progress-bar off pip setuptools==62.3.2
+            python -m pip install --progress-bar off pip setuptools==62.3.4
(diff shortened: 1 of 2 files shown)


stephenroller (approved):
tyvm

PR #4623, opened 2022-06-21
SaferDialogues parallel test and valid set with BAD dataset

meganung:
Adding a new teacher `SaferDialoguesBAD` that has a different test and valid set that is parallel with the BAD dataset (train set is the same amongst both teachers) 

Added test for the new teacher, pytest passes.

### DIFF
--- a/parlai/tasks/saferdialogues/agents.py
+++ b/parlai/tasks/saferdialogues/agents.py
@@ -17,7 +17,6 @@ def _path(opt):
     # Build the data if it doesn't exist.
     build(opt)
     dt = opt['datatype'].split(':')[0]
-    print("BOOO")
     return os.path.join(
         opt['datapath'], 'saferdialogues', 'saferdialogues_dataset', dt + '.txt'
     )
@@ -27,8 +26,6 @@ def _bad_path(opt):
     # Build the data if it doesn't exist.
     build(opt)
     dt = opt['datatype'].split(':')[0]
-    print("YAYY")
-
     if dt == 'valid' or dt == 'test':
         dt += '_bad'
     return os.path.join(
--- a/parlai/tasks/saferdialogues/agents.py
+++ b/parlai/tasks/saferdialogues/agents.py
@@ -17,8 +17,20 @@ def _path(opt):
     # Build the data if it doesn't exist.
     build(opt)
     dt = opt['datatype'].split(':')[0]
-    if opt['bad'] and dt == 'test':
-        dt = 'test_bad'
+    print("BOOO")
+    return os.path.join(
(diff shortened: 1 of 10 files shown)


parlai/tasks/saferdialogues/build.py
```diff
@@ -10,13 +10,13 @@
...
 
-SAFERDIALOGUES_VERSION = "v0.1"
+SAFERDIALOGUES_VERSION = "v0.3"
 
 RESOURCES = [
     DownloadableFile(
         f'http://parl.ai/downloads/saferdialogues/saferdialogues_{SAFERDIALOGUES_VERSION}.tar.gz',
         f'saferdialogues_{SAFERDIALOGUES_VERSION}.tar.gz',
-        '7ecf062fdf3e0ddfd16c52998341f543331e0dbaa1fcdb27026e8b0b80446fed',
+        '3d1bc731fb0c63d9f61a52b3d5f1aab0911cda1ca38f49822c79c18a2dc8e834',
```
jxmsML, line 19: will the new teacher be overriding the previous version (the one with the old valid and test split), I wonder if it would be nicer to keep both splits (one for reproducibility in your paper, and one here)

parlai/tasks/saferdialogues/agents.py
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
 
-# Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright (c) Facebook, Inc. and its affiliates.
```
klshuster, line 3: do we need to update our copyrights @stephenroller ?

parlai/tasks/saferdialogues/test/saferdialogues_test.yml
```diff
@@ -1,86 +1,66 @@
 acts:
 - - episode_done: true
     eval_labels:
-    - I agree with that, they're kids after all.
+    - i am also a fun maker
```
klshuster, line 4: why did the train/val/test values change in the original dataset?

klshuster (commented):
looks great just one q

parlai/tasks/saferdialogues/agents.py
stephenroller: yes we do

stephenroller:
Can you rebase/merge main in?

parlai/tasks/saferdialogues/build.py
```diff
@@ -10,13 +10,13 @@
 import parlai.core.build_data as build_data
 import os
 
-SAFERDIALOGUES_VERSION = "v0.1"
+SAFERDIALOGUES_VERSION = "v0.4"
```
jxmsML, line 13: LOL what happened to version 0.2 ~ 0.3

PR #4631, opened 2022-06-27
Search query API timeout and retry

mojtaba-komeili:
# Patch description
Right now, the `SearchEngineRetriever` agent sends a single search request to the search server and waits on that. Without timeout, this may mean that the agent will be blocked until receiving a response (which may never happen). Also upon failure on request, it does not retry. This patch solve this issue. 
Two more opt are added the to the agent:
* `search_server_timeout` that controls timeout in seconds
* `max_num_retries` for the number of times to retry until there is a 200 response from the server.

### DIFF
--- a/parlai/agents/rag/retrieve_api.py
+++ b/parlai/agents/rag/retrieve_api.py
@@ -18,6 +18,7 @@ from parlai.utils import logging
 
 
 CONTENT = 'content'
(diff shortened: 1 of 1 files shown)


mojtaba-komeili:
@klshuster I am working on making requests async. Don't review it yet. I will ping you again when it was ready.

klshuster (approved):

PR #4637, opened 2022-06-28
Early stopping and finetuned optimizer state

lightvector:
**Patch description**
Adds some additional flags that allow some more flexible uses of ParlAI for running experiments when finetuning models and starting/stopping runs to change between similar datasets or different training parameters with the same model.

`--preserve-optimizer-for-finetuning` - during a finetuning run, preserve the optimizer state from the prior run. This can be important for ADAM to not take erroneous steps when starting training again from the model from a prior run.
`--early-stop-at-n-steps` - Stop training early at n steps. This can be convenient for launching a run in several parts, swapping out the dataset or parameters in the middle of a run after n steps, etc, while still obeying the learning rate schedule and/or any other settings that would be used in a longer run.

**Testing steps**
Added a short integration test for --early-stop-at-n-steps to test_lr_schedulers.py.
For `--preserve-optimizer-for-finetuning` I'm a little bit at a loss as to how to test it in a more automated way. I used essentially this change for my own project, but someone has any suggestions how one would more standardly test within the test framework for this library, or even what dataset/model/command would be a good baseline, please feel free to comment.

### DIFF
--- a/parlai/core/torch_agent.py
+++ b/parlai/core/torch_agent.py
@@ -560,6 +560,12 @@ class TorchAgent(ABC, Agent):
             ),
         )
         optim_group.add_argument(
+            "--preserve-optimizer-for-finetuning",
+            default=False,
+            type='bool',
+            help='When starting a finetuning run, preserve the optimizer state from the prior run'
+        )
+        optim_group.add_argument(
             '-lr', '--learningrate', type=float, default=1, help='Learning rate'
         )
         optim_group.add_argument(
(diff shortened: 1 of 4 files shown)


lightvector:
@c-flaherty

parlai/core/torch_agent.py
```diff
@@ -1071,9 +1077,16 @@ def init_optim(
                     self.optimizer, should_sync_gradnorm(opt)
                 )
 
+        is_finetune_preserving_optimizer = False
         if is_finetune:
-            logging.warning('Detected a fine-tune run. Resetting the optimizer.')
-            return True
+            if opt.get("preserve_optimizer_for_finetuning", False):
+                logging.warning(
+                    'Detected a fine-tune run. But NOT resetting the optimizer since -preserve-optimizer-for-finetuning was set.'
```
klshuster, line 1084: nit: `--preserver-optimizer...`

klshuster (commented):
I think this looks great!

For testing preserving the optimizer - I would suggest training a dummy model on a dummy task for X amount of steps, with a certain optimizer, and then training a new model initializing from the old model both with and without your proposed flag, and then testing to see that 1) state change/no change is correct; 2) LR change/no change is correct

e.g. a [setup like this](https://github.com/facebookresearch/ParlAI/blob/75a58707fd35c6e75d143b051e9455cb01862f80/tests/test_train_model.py#L373-L383) to train the first model

c-flaherty:
@lightvector  Note for later: Turning these flags on may break requeue-ability (because lr scheduler state is destroyed on job initialization). We need to look into that before merging this

PR #4648, opened 2022-07-06
Fixed T5 and GPT2 related bugs in the hugging_face agents

michaelsaxon:
**Patch description**
Previously T5 dictionary was undefined and elements of the huggingface dictionary agent were uninitialized leading to immediate exceptions in the train step for T5 and at the beginning of the validation step for GPT2 when training using train_model.py and the latest versions of HuggingFace

**Testing steps**
Verify that `python parlai/scripts/train_model.py --m hugging_face/t5 --t multiwoz_v22 --mf ~/t5test --truncate 512 --vstep 150 --max-train-steps 300` and  `python parlai/scripts/train_model.py --m hugging_face/gpt2 --t multiwoz_v22 --mf ~/gpt2test --truncate 512 --vstep 150 --max-train-steps 300` successfully terminate without exceptions

**Other information**
As discussed with Kurt and Andrew previously


klshuster (commented):
Thank you for the fix! 

I think this all looks great. However, prior to approving, I do think that your original error actually uncovered the real issue; namely, that we don't define an `unk` token for GPT2. Could you please additionally modify the `Gpt2DictionaryAgent` to define `self.unk_token`?

Presumably you'd just [override some of the functionality here](https://github.com/facebookresearch/ParlAI/blob/03ca319f440cdffafe7068c0e56c49f568d6de83/parlai/agents/hugging_face/dict.py#L178-L205)

[Looking at huggingface](https://huggingface.co/docs/transformers/model_doc/gpt2), i think we can just set it to `<|endoftext|>` like the other special tokens.

klshuster:
Is there any update here?

klshuster:
closing in favor of #5016

PR #4650, opened 2022-07-06
[TGA] Delayed nucleus

Jsiewierski11:
**Patch description**
I have a new search method for the generator agent that I wanted to share. The new method, which I call nucleusbeam search provides more variation that traditional beam search while still maintaining coherence.
The seq2seq model can be igonored, that wasn't meant to be in the pull request.


klshuster:
Thank you for your contribution! Could you please remove the `seqseq` model from the PR, if it is not meant to be there?

How would you feel about naming it `DelayedNucleusBeamSearch`, as it is quite similar to the `DelayedBeamSearch` except for Nucleus instead of Top-K?

Jsiewierski11:
I think that name is more fitting. Just removed the seqseq agent and changed the name.

klshuster:
Thanks! One more note - you'll most likely need to update the [command line argument here](https://github.com/facebookresearch/ParlAI/blob/f706af2a1db3746359b5eeb2d8fa0ae7df64a176/parlai/core/torch_generator_agent.py#L428) to accept `delayednucleusbeam` as an option

Jsiewierski11:
Added the command line argument in my latest commit. Is there anything else that needs to be done before it can be merged? I am seeing that the long_gpu_tests are failing, do you know what might be causing that?

klshuster (approved):
everything looks good! the test failure is a known issue, I wouldn't worry about it - I'll go ahead and merge, thanks again for your contribution!

PR #4652, opened 2022-07-06
Add Async Multiparty Chat to ParlAI

chiehminwei:
**Patch description**
This patch adds basic async multiparty chat functionality to ParlAI, following the specification outlined in the [design document](https://docs.google.com/document/d/1USR-8ywI80VqcoOuaZA2FbGe-64xuYsQp9ZJjRStnn8/edit#).

During training, we can still use a two-agent world, even for multiparty chat. This is achieved by making one agent the teacher, one agent the student, and modifying the relevant `Teacher` to format the context correctly.

To support multiparty chat, though, we extend the `TorchAgent` class to have the ability to handle more than two speakers, and enable it to have a notion of time and a notion of silence.

These are enabled by setting the following CLI flags:
1) `--timestep-field`
2) `--speaker-field`
3) `--record-silence`
4) `--silence-token`

For example, for an `observation`:

`observation = { 'speaker': 'Groot', 'text': 'I am Groot.', 'timestep': '00:00:00', }`

The `--timestep-field` would be `timestep`, and the `--speaker-field` would be `speaker`.
The `timestep` would get saved in the `History` class in new fields related to timesteps, and `speaker` would replace the `__P1__` token to get saved in the conversation history.

To stop the agent from recording silence in its own output, specify `--record-silence=True` to stop agent recording it during its call to `self_observe`.

**Testing steps**
Additional unit tests have been added to the test suite.
Run `pytest tests/pytest tests/test_torch_agent.py`


parlai/core/torch_agent.py
```diff
@@ -284,6 +284,8 @@ def __init__(
         self.add_p1_after_newln = opt.get('add_p1_after_newln', False)
         self.p1_token = p1_token
```
mojtaba-komeili, line 285: Given this is the others, is it still going to be a single token, not a list?

parlai/core/torch_agent.py
```diff
@@ -284,6 +284,8 @@ def __init__(
...
         self.p2_token = p2_token
+        self.speaker_field = opt.get('speaker_field', None)
+        self.speakers = set([p1_token, p2_token])
```
mojtaba-komeili, line 288: Same as above, for multiple users.

parlai/core/torch_agent.py
```diff
@@ -783,15 +813,7 @@ def __init__(self, opt: Opt, shared=None):
...
-                # Volta cores revert to FP32 hardware if tensors are not multiples
-                # of 8 in all dimensions. This INCLUDES the embeddings layer! As
-                # such, we need some extra magic to ensure the dictionary is padded
-                # with extra tokens to make it a multiple of 8.
-                from parlai.utils.torch import FP16_PAD_SIZE
-
-                if len(self.dict) % FP16_PAD_SIZE != 0:
-                    for i in range(FP16_PAD_SIZE - len(self.dict) % FP16_PAD_SIZE):
-                        self.dict['__FP16_PAD_{}__'.format(i)] = 1
+                self._pad_dictionary_fp16()
```
mojtaba-komeili, line 816: Good job refactoring this part.

parlai/core/torch_agent.py
```diff
@@ -1930,19 +1996,23 @@ def self_observe(self, self_message: Message) -> None:
 
         raise RuntimeError("Unexpected case in self_observe.")
 
-    def _validate_observe_invariants(self):
+    def _validate_observe_invariants(self, observation: Message = None):
         """
         Check that we properly called self_observe after the last batch_act.
         """
-        if self.__expecting_to_reply:
+        if not self.allow_multiple_observe and self.__expecting_to_reply:
```
mojtaba-komeili, line 2003: Look like this should take care of backward compatibility. But have tests that make sure it is (if there aren't any already).

parlai/core/torch_agent.py
```diff
@@ -1930,19 +1996,23 @@ def self_observe(self, self_message: Message) -> None:
...
             raise RuntimeError(
                 "Last observe() had a label, but no call to self_observe ever "
                 "happened. You are likely making multiple observe() calls without "
                 "a corresponding act(). This was changed in #2043. File a GitHub "
                 "issue if you require assistance."
             )
 
-        if self.__expecting_clear_history:
+        if (not self.allow_multiple_observe and self.__expecting_clear_history) or (
```
mojtaba-komeili, line 2011: Same as above.

mojtaba-komeili (commented):
Looks great. You added the minimal amount of code needed and refactored well. This is good example of adding on top of high level code. Just make sure that you have enough test coverage for backward compatibility.
I am guessing some of the GPU test failures might be related to the changes to the Torch Agent.

parlai/core/torch_agent.py
chiehminwei: Sounds great I'll add unit tests once I have silence and teachers, etc. figured out

chiehminwei:
> Looks great. You added the minimal amount of code needed and refactored well. This is good example of adding on top of high level code. Just make sure that you have enough test coverage for backward compatibility. I am guessing some of the GPU test failures might be related to the changes to the Torch Agent.

Awesome I'll add unit tests once I figure out silence and teachers, etc.

I took a look and it seems like the GPU test failures are from broken main branch, not from this PR: https://fb.workplace.com/groups/770643789812374/permalink/1997298873813520/

parlai/core/worlds.py
```diff
@@ -481,7 +481,7 @@ def __init__(self, opt: Opt, agents, shared=None):
             # Add passed in agents directly.
             self.agents = agents
         self.acts = [None] * len(self.agents)
-        self.parley_mode = opt['parley_mode']
+        self.parley_mode = opt.get('parley_mode', 'round-robin')
```
mojtaba-komeili, line 484: If this is in the flag default there is no need to add it here too. In fact, it might be a good check to see the code crash if the value is not set correctly.

chiehminwei:
Another question I had after reading through the code again is whether we want to modify the how we handle observations in `TorchAgent`.

Currently, `act` is called with the observation `self.observation` that was stored from the last call to `observe(observation)`.

This `observation` (or `batch` of `observation`s) is then passed to the `train_step`.

https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L2278

This behavior is fine if we only have two agents parley to each other. The agent is always in an `observe-act` loop so `self.observation` is always fresh.

Things are different if we introduce a tick-based system and allow multiple `observe` before calling `act`. When the agent `act`, only the latest `observation` gets stored in `self.observation` and gets sent to `train_step`.

For example, in a world with 3 agents, for Agent A:
```
observe(action from Agent B)
observe(action from Agent C)
act
```
Although both `action from Agent B` and `action from Agent C` will get stored in the history (by appending to the `self.history.history_vecs` array, etc.) when `observe` is called, only `action from Agent C` will be passed to `train_step(action from Agent C)`.

This may be fine since previous observations are all stored in the history, but I wonder if _all observations from this time tick_ should be somehow grouped together and sent to `batch_act` and then to `train_step(batch)`, maybe by modifying `batch` so that it's not just a list of `Message`, but rather a list of list of `Message`, e.g.:

```
# batch of size 2
batch = [
  [Message_from_A, Message_from_B, Message_from_C],
  [Message_from_A, Message_from_B, Message_from_C],
]
```

This will require changes in a lot of places though (the `validate` function in `words.py`, and changing function signatures for `batch_act`, `train_step`, and `eval_step`), so I'm not sure if it's worth doing.

parlai/core/worlds.py
chiehminwei: I actually added it here after the tests crash (it also crashed for all the other options I added). I'll take a closer to at the unit tests to see what's going on.

chiehminwei:
> Another question I had after reading through the code again is whether we want to modify the how we handle observations in [TorchAgent](https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L417).
> 
> Currently, [act](https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L2142) calls [batch_act](https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L2152) with the observation `self.observation` that was stored from the last call to [observe(observation)](https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L1822).
> 
> https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L1822
> 
> https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L1842-L1843
> 
> https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L2142-L2148
> 
> This `observation` (or `batch` of `observation`s) is then passed to the `train_step` (or `eval_step`).
> 
> https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L2152
> 
> 
> https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L2169
> 
> 
> https://github.com/facebookresearch/ParlAI/blob/dff9aabb5024c30c81e146cebffbc88bc6431b61/parlai/core/torch_agent.py#L2239
> 
> This behavior is fine if we only have two agents parley to each other. The agent is always in an `observe-act` loop so `self.observation` is always fresh.
> 
> Things are different if we introduce a tick-based system and allow multiple `observe` before calling `act`. When the agent `act`, only the latest `observation` gets stored in `self.observation` and gets sent to `train_step`.
> 
> For example, in a world with 3 agents, for Agent A:
> 
> ```
> agentA.observe(action from Agent B)
> agentA.observe(action from Agent C)
> agentA.act()
> ```
> 
> Although both `action from Agent B` and `action from Agent C` will get stored in the history (by appending to the `self.history.history_vecs` array, etc.) when `observe` is called, only `action from Agent C` will be passed to `train_step(action from Agent C)`.
> 
> This may be fine since previous observations are all stored in the history, but I wonder if _all observations from this time tick_ should be somehow grouped together and sent to `batch_act` and then to `train_step(batch)`, maybe by modifying `batch` so that it's not just a list of `Message`, but rather a list of list of `Message`, e.g.:
> 
> ```
> # batch of size 2
> batch = [
>   [Message_from_A, Message_from_B, Message_from_C],
>   [Message_from_A, Message_from_B, Message_from_C],
> ]
> ```
> 
> This will require changes in a lot of places though (the `validate` function in `words.py`, and changing function signatures for `batch_act`, `train_step`, and `eval_step`), so I'm not sure if it's worth doing.

After Jul 11 Multiparty Chat meeting, we decide to go against allowing multiple "observe" in a row. Interactive world will be handling this stuff to make sure all agents act in observe-act loop (even when new observation comes in the middle of an act, etc.)

chiehminwei:
From the failing tests it looks like adding `silence_token` to the dictionary will break models trained without `silence_token` when they restore from checkpoints due to mismatch in embedding size.

We could just remove `silence_token` from the `dict.py` and instead use a command line flag that saves to `self.silence_token`. The only reason we need a `silence_token` is so that we can skip it during `self_observe`.

But this feels a little unclean. Let me know what you think.

chiehminwei:
> From the failing tests it looks like adding `silence_token` to the dictionary will break models trained without `silence_token` when they restore from checkpoints due to mismatch in embedding size.
> 
> We could just remove `silence_token` from the `dict.py` and instead use a command line flag that saves to `self.silence_token`. The only reason we need a `silence_token` is so that we can skip it during `self_observe`.
> 
> But this feels a little unclean. Let me know what you think.



> From the failing tests it looks like adding `silence_token` to the dictionary will break models trained without `silence_token` when they restore from checkpoints due to mismatch in embedding size.
> 
> We could just remove `silence_token` from the `dict.py` and instead use a command line flag that saves to `self.silence_token`. The only reason we need a `silence_token` is so that we can skip it during `self_observe`.
> 
> But this feels a little unclean. Let me know what you think.

I went ahead and changed to using the command line flag `--silence-token` for `TorchAgent`.
This still feels a little hacky to me, but I guess it doesn't make sense to have an unused silence token in the dictionary and grow the embedding size by one for agents that don't care about the silence token.

parlai/core/torch_agent.py
```diff
@@ -328,8 +363,18 @@ def add_reply(self, text):
         self._update_strings(text)
         # update history vecs
         self._update_vecs(text)
+        if self.timestep_field is not None:
```
klshuster, line 366: should this be checking `if self.timestep is not None` rather than self.timestep_field?

parlai/core/torch_agent.py
```diff
@@ -346,18 +391,36 @@ def update_history(self, obs: Message, temp_history: Optional[str] = None):
...
+                        self.speakers.add(speaker)
                     text = self._add_person_tokens(
-                        obs[self.field], self.p1_token, self.add_p1_after_newln
+                        obs[self.field], speaker, self.add_p1_after_newln
                     )
                 # update history string
                 self._update_strings(text)
                 # update history vecs
                 self._update_vecs(text)
+                if self.timestep_field is not None:
```
klshuster, line 416: ditto here

parlai/core/torch_agent.py
```diff
@@ -402,6 +465,51 @@ def get_history_vec_list(self):
...
 
+    def get_history_str_with_timesteps(self) -> Optional[str]:
+        """
+        Return the string version of the history, including timesteps.
+        """
+        if len(self.history_strings) > 0:
+            history = self.history_strings[:]
+            timesteps = self.history_timesteps[:]
+            history = self.delimiter.join(
+                [f"{h}{self.delimiter}{t}" for h, t in zip(history, timesteps)]
```
klshuster, line 476: could we have a separate flag for the timestep delimiter? since we may not want to separate by newline

parlai/core/torch_agent.py
```diff
@@ -666,7 +786,27 @@ def add_cmdline_args(
...
+        )
+        agent.add_argument(
+            '-spkr',
+            '--speaker-field',
+            type='nonestr',
+            default=None,
+            help="field in the observation to track indicating the id of the speaker "
+            "over the course of the episode (defaults to None); `observation['speaker_field']` "
+            "overrides `__p1__` in `--person-tokens` if field is not None and is present in the observation.",
+        )
```
klshuster, line 801: so we have the `id` field already in Messages - what do we think about just defaulting to that?

parlai/core/torch_agent.py
```diff
@@ -666,7 +786,27 @@ def add_cmdline_args(
...
+        agent.add_argument(
+            '-tstep',
```
klshuster, line 803: i would remove this `'-tstep'`; we have training args `-lstep` etc. for indicating steps during training

parlai/core/torch_agent.py
```diff
@@ -666,7 +786,27 @@ def add_cmdline_args(
...
+            '--timestep-field',
+            type='nonestr',
+            default=None,
```
klshuster, line 806: we could also default this to `'timestep'` to have some sort of standardization

parlai/core/torch_agent.py
```diff
@@ -501,6 +609,18 @@ def add_cmdline_args(
...
             ' e.g. eval scripts.',
         )
+        agent.add_argument(
+            '--record-silence',
+            type='bool',
+            default=True,
+            help='Whether to record silence token in the agent history when the agent self-observes. Defaults to True for backward-compatability.',
+        )
+        agent.add_argument(
+            '--silence-token',
```
klshuster, line 619: 1. could we default this to `'__SILENCE__'`? It's the only one we ever use
2. could you specify in the help that this is a string, just to differentiate from e.g. tokenized strings (which are ints)

tests/test_torch_agent.py
```diff
@@ -263,26 +263,29 @@ def test_batchify(self):
         """
         Make sure the batchify function sets up the right fields.
         """
-        agent = get_agent(rank_candidates=True)
+        agent = get_agent(rank_candidates=True, timestep_field='timestep')
```
klshuster, line 266: could we alternate messages with and without timesteps? to test both cases?

klshuster (commented):
love the tests. looking great so far! a few comments (apologies for the late review!!)

parlai/core/torch_agent.py
chiehminwei: since we're changing the default for `timestep_field` to `timestep`, I removed this check and make `timestep` default to the empty string instead of `None` so that the tokenizer doesn't break

parlai/core/torch_agent.py
chiehminwei: I've added a new flag `--timestep-delimiter` that defaults to `\t`.
For line 492 I've kept `delimiter` though, since I want `history_timestep_vec` to match `timestep_vec`

tests/test_torch_agent.py
chiehminwei: Added a test case in line 933 of the test file

parlai/core/torch_agent.py
```diff
@@ -239,11 +244,11 @@ class History(object):
         sets the maximum number of tunrs
 
     :param p1_token:
-        token indicating 'person 1'; opt must have 'person_tokens' set to True
+        token indicating 'person 1' (teacher/ others); opt must have 'person_tokens' set to True
         for this to be added
 
-    :param p1_token:
```
mojtaba-komeili, line 245: hmm, was it used to be p1_token for both? Was this a mistake or am I missing something here?

parlai/core/torch_agent.py
```diff
@@ -277,27 +283,45 @@ def __init__(
         self.history_strings = []
         self.history_raw_strings = []
         self.history_vecs = []
+        self.history_timesteps = []
```
mojtaba-komeili, line 286: Could you add a comment about the relation between this and other history variables here?

parlai/core/torch_agent.py
```diff
@@ -277,27 +283,45 @@ def __init__(
...
+        self.timestep_field = opt.get('timestep_field', None)
+        self.speakers = set([p1_token, p2_token])
 
     def parse(self, text):
         """
         Tokenize text with the given dictionary.
         """
         return self.dict.txt2vec(text)
 
+    def parse_timestep(self, timestep):
```
mojtaba-komeili, line 306: why does the name say *parse* when what it does is *tokenizing*?

mojtaba-komeili: Ok, I see why after reading the full docstring. Never mind ;)

parlai/core/torch_agent.py
```diff
@@ -317,6 +341,18 @@ def _update_vecs(self, text):
                 self.history_vecs.pop(0)
         self.history_vecs.append(self.parse(text))
 
+    def _update_timesteps(self, timestep):
+        if self.size > 0:
+            while len(self.history_timesteps) >= self.size:
+                self.history_timesteps.pop(0)
+        self.history_timesteps.append(timestep)
+
+    def _update_timestep_vecs(self, timestep):
```
mojtaba-komeili, line 350: I don't see any situation that you only call the `_update_timestep_vecs` function. And it looks a lot like the `_update_timesteps` function. Is it possible to merge these two into one function? They seem to be doing very similar things and they are called together.

parlai/core/torch_agent.py
```diff
@@ -1693,6 +1871,17 @@ def batchify(self, obs_batch, sort=False):
                     x_lens, xs, x_lens, valid_inds, exs, descending=True
                 )
 
+        # TIME_STEPS
+        ts = t_lens = None
+        if any(ex.get('timestep_vec') is not None for ex in exs):
+            _ts = [ex.get('timestep_vec', self.EMPTY) for ex in exs]
+            ts, t_lens = self._pad_tensor(_ts)
+            if sort:
```
mojtaba-komeili, line 1879: Is it possible that this gets affected by the `sort` change in [line 1869]( https://github.com/facebookresearch/ParlAI/blob/8271c8520d11514a39ef38bbe1b1ee9ed9b2d3e2/parlai/core/torch_agent.py#L1869)

mojtaba-komeili: I am not even sure we need this sort here. The sort on batching is important for text, but may not be relevant for the timestamps, which shouldn't have wildly varying lengths.

tests/test_torch_agent.py
```diff
@@ -794,6 +825,299 @@ def test_temp_history_observe(self):
         assert obs['full_text'] == '1 2 3'
         agent.act()  # get that self-observe in
 
+    def test_history_timestep(self):
+        """
+        Test that history handles timestep properly.
+        """
+        obs = {
+            'speaker': 'Groot',
```
mojtaba-komeili, line 833: 😆

mojtaba-komeili (commented):
Looks good thank you.  I had a few small comments on the code. Also an extra major comment about the tests: it looks like a lot of your tests are a combinations of multiple tests with many resets and repeats. This makes them very long and hard to debug. If possible, break your tests into multiple smaller ones. I see you have commented on the test parts, why now having them with functions that have proper names coming from those comments?

chiehminwei:
> Looks good thank you. I had a few small comments on the code. Also an extra major comment about the tests: it looks like a lot of your tests are a combinations of multiple tests with many resets and repeats. This makes them very long and hard to debug. If possible, break your tests into multiple smaller ones. I see you have commented on the test parts, why now having them with functions that have proper names coming from those comments?

I wrote the tests based on[ the existing one that we had](https://github.com/facebookresearch/ParlAI/blob/8271c8520d11514a39ef38bbe1b1ee9ed9b2d3e2/tests/test_torch_agent.py#L619) and thought that was the style we were going for.
Should I break things up?

parlai/core/torch_agent.py
chiehminwei: Yea I think it's gonna mess up the `sort` change earlier in text, and I agree it's really not necessary to have sort by length for timesteps. I've gone ahead and removed the sort for timesteps.

PR #4655, opened 2022-07-07
fix frame heights in annotation task

dexterju27:
**Patch description**
The current "frame_height": '100%', will cause the annotation task to fail. The proposed change fixed this.


klshuster:
Can you please elaborate on what the error is here?

Also, I think maybe @EricMichaelSmith might be a good reviewer for this

EricMichaelSmith:
Hi @dexterju27 it looks like Pratik already fixed this in `main`: https://github.com/facebookresearch/ParlAI/pull/4625

dexterju27:
Interesting! then why does it still let me open this PR against main? Glad to know this is already fixed.

EricMichaelSmith:
> Interesting! then why does it still let me open this PR against main? Glad to know this is already fixed.

Hmm I think if you do a `git merge main` this branch will become identical to main

PR #4656, opened 2022-07-07
Added DST Teacher for multiwoz_v22 task

prajjwal1:
The DST Teacher for `multiwoz_v22` is not present. Adding this teacher would allow `multiwoz_v22` to be usable for Dialog State Tracking task.

cc @chinnadhurai


parlai/tasks/multiwoz_v22/agents.py
```diff
@@ -373,6 +379,212 @@ def get_id_task_prefix(self):
         return "MultiwozV22"
 
 
+class MultiWOZv22DSTTeacher(tod_agents.TodUserSimulatorTeacher):
```
chinnadhurai, line 382: might be better to add a comment saying this teacher is needed for reproducing the Joint Goal Accuracy values reported in the simpleTOD & SOLOIST papers.

prajjwal1: Is it better now ? @chinnadhurai

moyapchen:
Looks okay code-wise, though I haven't done a through review. Should this be in ParlAI-Internal rather than public ParlAI though?

parlai/tasks/multiwoz_v22/agents.py
```diff
@@ -393,5 +605,5 @@ class SingleApiSchemaAgent(MultiwozV22Parser, tod_agents.TodSingleApiSchemaAgent
     pass
 
 
-class DefaultTeacher(SystemTeacher):
```
chinnadhurai, line 396: would be better to keep the default to SystemTeacher as is.

chinnadhurai:
> Looks okay code-wise, though I haven't done a through review. Should this be in ParlAI-Internal rather than public ParlAI though?

we need this teacher for the actual DST task in Multiwoz and reproduce JGA reported in the SOTA models. So, figured it might make sense to push to the public one.

parlai/tasks/multiwoz_v22/agents.py
prajjwal1: Fixed this.

prajjwal1:
@stephenroller Can you please look into this ? It has been here for sometime now.

stephenroller (commented):
I'll defer to @moyapchen ans @klshuster here. Don't see any issue though

klshuster (commented):
deferring to @moyapchen for final say but looks OK to me

could we [generate test files](http://parl.ai.s3-website.us-east-2.amazonaws.com/docs/tutorial_task.html#part-5-contributing-upstream) for it as well?

prajjwal1:
@moyapchen Can you please see this ?

moyapchen (approved):
lgtm

PR #4662, opened 2022-07-12
set_seed in train_model

prajjwal1:
**Patch description**
Adding a function that allows user to set seed for training. User just has to pass the flag --set_seed with existing arguments to train_model.py in case they want to use a seed other than 42.

**Testing steps**

**Other information**

### DIFF
--- a/parlai/core/tod/tod_agents.py
+++ b/parlai/core/tod/tod_agents.py
@@ -40,7 +40,7 @@ class TodStructuredDataParser(Agent):
 
     Inherit from this class and implement `setup_episodes()` to implement the intermediate representation for a specific dataset. Use multiple inheritence with classes that implement an `act()` below to use.
 
-    For example, if we have a `MyDataset_DataParser(TodStructuredDataParser)` and wanted to make a teacher to train a model togenerate User Utterances based on a goal prompt, we would do so by defining `class MyDatasetUserSimulatorTeacher(MyDataset_DataParser, TodUserSimulatorTeacher)`.
+    For example, if we have a `MyDataset_DataParser(TodStructuredDataParser)` and wanted to make a teacher to train a model to generate User Utterances based on a goal prompt, we would do so by defining `class MyDatasetUserSimulatorTeacher(MyDataset_DataParser, TodUserSimulatorTeacher)`.
     """
 
     @classmethod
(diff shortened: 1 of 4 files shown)


prajjwal1:
@stephenroller Can you please look into this ? It has been here for sometime now.

prajjwal1:
I was thinking of using the functionality brought from this to teachers also. Teacher files have `SEED` manually hardcoded to 42. How is a user supposed to set the seed in `Teacher`. This will allow user to bring this functionality to control `rng` in teacher.

stephenroller (commented):
I'm fine with this, but I'm pretty sure the existing code already sets the seed. Can we double check?

prajjwal1:
I did check, i didn't find any way to setting the seed universally. I plan to send one more PR for `multiwoz_v22`, which will set the `rng` through `opt`. Current `opt` dict doesn't have it, so where else is `seed` parameter supposed to go ?

klshuster:
> I'm fine with this, but I'm pretty sure the existing code already sets the seed. Can we double check?

interestingly I can't seem to find it either. we set seeds in nearly every other script except for train model

pure non-determinism! In that case, can we have the default seed be `None` and only set seed if it's not `None`?

prajjwal1:
Default way is to set the seed to 42 right ? Can I set it to 42, and this can be overridden by the user by passing in `--seed` parameter ? By not setting the seed, users are likely to observe variability in their training runs.

stephenroller:
> > I'm fine with this, but I'm pretty sure the existing code already sets the seed. Can we double check?
> 
> interestingly I can't seem to find it either. we set seeds in nearly every other script except for train model
> 
> pure non-determinism! In that case, can we have the default seed be `None` and only set seed if it's not `None`?

ah right this is what i misremembered

stephenroller:
> Default way is to set the seed to 42 right ? Can I set it to 42, and this can be overridden by the user by passing in `--seed` parameter ? By not setting the seed, users are likely to observe variability in their training runs.

historically we've preferred backwards compatibility

klshuster:
I agree, let's keep it to None. We've advocated for variability from the start

prajjwal1:
@klshuster Is it fine now ?

stephenroller (approved):
lgtm, will leave Kurt for final approval and merge

klshuster (approved):
thank you!

PR #4664, opened 2022-07-13
Pass working directory path in TurnAnnotationStatic task

basilnajjar:
This will help us avoid code duplication when building a new task built on top of the TurnAnnotationStatic task.

### DIFF
--- a/parlai/crowdsourcing/tasks/turn_annotations_static/run.py
+++ b/parlai/crowdsourcing/tasks/turn_annotations_static/run.py
@@ -25,6 +25,8 @@ crowdsourcing tasks with this script.
 _ = STATIC_BLUEPRINT_TYPE
 
 TASK_DIRECTORY = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


EricMichaelSmith (commented):
Hmm - I think this flag could be good to have, but is it used in this task? If it's only used in a subclass of this task, it'd be best to put the flag there. Or is there a benefit to having it in the base task?

basilnajjar:
> Hmm - I think this flag could be good to have, but is it used in this task? If it's only used in a subclass of this task, it'd be best to put the flag there. Or is there a benefit to having it in the base task?
@EricMichaelSmith I think it's necessary to have this flag here to get the relevant paths which are in a different repo: [sample.yaml](https://github.com/fairinternal/ParlAI-Internal/pull/3324/commits/bf0bc684af8c8f5bb8d0fe308ff6947e4b2194f5#diff-7035af73b4ee85795945601187133f853ebd08bb4d9be3d91d675a160632c82a)

PR #4666, opened 2022-07-13
add emoji input functionality to model chat crowdsourcing task

basilnajjar:
Users can now add emojis to the text box by using the emoji picker. To show the emoji picker, you need to set the emoji_picker flag to true (it's set to false by default).

# Test plan:
`npm install react-input-emoji`
`python parlai/crowdsourcing/tasks/model_chat/run.py mephisto.blueprint.emoji_picker=true`

Screen Shot 2022-07-13 at 5 00 21 PM

### DIFF
--- a/parlai/crowdsourcing/tasks/model_chat/frontend/components/response_panes.jsx
+++ b/parlai/crowdsourcing/tasks/model_chat/frontend/components/response_panes.jsx
@@ -161,7 +161,7 @@ function FinalSurvey({ taskConfig, onMessageSend, active, currentCheckboxes }) {
   }
 }
 
-function CheckboxTextResponse({ onMessageSend, active, currentCheckboxes }) {
+function CheckboxTextResponse({ onMessageSend, taskConfig, active, currentCheckboxes }) {
   const [textValue, setTextValue] = React.useState("");
   const [sending, setSending] = React.useState(false);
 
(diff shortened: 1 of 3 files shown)


EricMichaelSmith:
Hmm - can you run `python parlai/crowdsourcing/tasks/model_chat/run.py`, make sure it still works, and then add that to the PR description as part of the test plan? Since we don't have any frontend integration tests currently, this manual step is a good check to make sure that the task still works as expected :)

basilnajjar:
> Hmm - I'm a bit nervous about changing the model chat task to _always_ show the emoji picker, because many of our models can't handle emojis, and so this might mean that the future conversations that we collect will contain largely unparseable emojis. Could you try adding a Hydra flag to conditionally show the emoji picker and default it to False? Another issue is that it looks like this change means that developers will always need to run `npm install react-input-emoji` to run the model chat task, and I'm loath to require that if most people won't need to be using the emoji picker

See updated description above

EricMichaelSmith (approved):
Hmm - the one thing I'm still a bit worried about is that people will now *have* to install the emoji picker tool in order to use model chat, even if the emoji picker flag is False. But since we always have to install a bunch of packages anyway to run any task, I'm not convinced that one more will be a huge problem. Thus, I'm willing to go ahead and approve this - if this extra package becomes a huge problem, we can reassess later. Great addition!

tianxie-sf:
Hi, I am a new user of ParlAI. I tried to run the model chat task, but I got this error:

```
ERROR in ./src/components/response_panes.jsx 23:0-43
Module not found: Error: Can't resolve 'react-input-emoji' in xxx/xxx
```

Could you please try to create a new environment, and setup ParlAI, then run this task?

Thank you so much!

PR #4669, opened 2022-07-20
Add k-fold cross validation and noise mutators

dexterju27:
**Patch description**
This PR adds some mutators:
1. k-fold mutators, where it withholds a certain fold in the train set and later uses that in the validation.
2. Flip labels / random label mutators, they worked in binary classification case, will introduce artificial noise to the data.  
**Testing steps**
```
 parlai display_data -t dialogue_safety:standard --mutators flip_label,k_fold_withhold_on_train  --k-fold 3  --held-fold 1 -dt train -n 8000 --noise-level 1.0
```

### DIFF
--- a/parlai/mutators/flip_labels.py
+++ b/parlai/mutators/flip_labels.py
@@ -36,6 +36,12 @@ class FlipClassificationLabelMutator(MessageMutator):
             type=float,
             help='The probability of label flipping',
         )
+        agent.add_argument(
+            '--classification-label-to-mutate',
+            default='__ok__,__notok__',
+            type=str,
+            help='The binary labels in the data for flipping / random assignment. seprated by ,',
+        )
         return parser
 
     def __init__(self, opt: Opt):
@@ -46,6 +52,7 @@ class FlipClassificationLabelMutator(MessageMutator):
             logging.info('No noise, teacher should be the same as before.')
         if opt['noise_level'] == 1:
             logging.info('Labels are exact the opposite.')
+        self.labels = opt['classification_label_to_flip'].split(',')
         self.noise_level = opt['noise_level']
         logging.info(f'Flipping with noise level {self.noise_level}')
         self.is_training = DatatypeHelper.is_training(opt['datatype'])
@@ -67,12 +74,14 @@ class FlipClassificationLabelMutator(MessageMutator):
                 # deterministic flipping for each example
                 if flip:
                     self.flip_cnt += 1
-                    if labels[0] == '__ok__':
-                        new_labels = ['__notok__']
-                    elif labels[0] == '__notok__':
-                        new_labels = ['__ok__']
+                    if labels[0] == self.labels[0]:
+                        new_labels = [self.labels[1]]
+                    elif labels[0] == self.labels[1]:
+                        new_labels = [self.labels[0]]
                     else:
-                        raise ValueError("labels must be '__ok__' and '__notok__'")
+                        raise ValueError(
+                            f'labels must be binary and the same as in {self.labels}'
+                        )
                     assert len(new_labels) == 1
             if flip:
                 message.force_set('labels', new_labels)
@@ -97,7 +106,7 @@ class FlipClassificationLabelTrainOnlyMutator(FlipClassificationLabelMutator):
             return super().message_mutation(message)
 
 
-@register_mutator("flip_label_valid_only")
+@register_mutator("flip_classification_label_valid_only")
 class FlipClassificationLabelValidOnlyMutator(FlipClassificationLabelMutator):
     def __init__(self, opt: Opt):
         super().__init__(opt)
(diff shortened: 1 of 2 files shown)


klshuster (commented):
think everything looks good here, just a few questions - could you please fix lint errors too :)

parlai/mutators/flip_labels.py
```diff
@@ -0,0 +1,143 @@
...
+        return
+
+    def message_mutation(self, message: Message) -> Message:
+        if self.is_training:
+            return message
+        else:
+            return super().message_mutation(message)
+
+
+@register_mutator("random_label")
```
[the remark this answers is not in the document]
dexterju27: You can have this via setting noise level to 1.0?

dexterju27:
Addressed all the comments, fixed the lint issues.

parlai/mutators/flip_labels.py
```diff
@@ -0,0 +1,140 @@
...
+            if len(labels) != 1:
+                raise ValueError(
+                    f'{type(self).__name__} can only be used with one label!'
+                )
+            else:
+                # deterministic flipping for each example
+                if flip:
+                    self.flip_cnt += 1
+                    if labels[0] == '__ok__':
+                        new_labels = ['__notok__']
```
jxmsML, line 71: I wonder if there is a way to set the `__ok__` and `__notok__` as argument instead of hardcoded here.

parlai/mutators/flip_labels.py
```diff
@@ -0,0 +1,140 @@
...
+                    elif labels[0] == '__notok__':
+                        new_labels = ['__ok__']
+                    else:
+                        raise ValueError("labels must be '__ok__' and '__notok__'")
```
jxmsML, line 75: I wonder if there is a way to set the __ok__ and __notok__ as argument instead of hardcoded here.

parlai/mutators/flip_labels.py
```diff
@@ -0,0 +1,140 @@
...
+        return
+
+    def message_mutation(self, message: Message) -> Message:
+        if not self.is_training:
+            return message
+        else:
+            return super().message_mutation(message)
+
+
+@register_mutator("flip_label_valid_only")
```
jxmsML, line 100: maybe rename it as "flip_classification_label_valid_only" to be consistent with L85

parlai/mutators/flip_labels.py
```diff
@@ -0,0 +1,140 @@
...
+                        new_labels = ['__ok__']
+                    else:
+                        raise ValueError("labels must be '__ok__' and '__notok__'")
+                    assert len(new_labels) == 1
+            if flip:
+                message.force_set('labels', new_labels)
+        return message
+
+    def __del__(self):
+        logging.info(f'Flipped {self.flip_cnt}')
```
jxmsML, line 82: curious in which case will the `self.flip_cnt` being reset? when the `task_agent` is reset?

parlai/mutators/flip_labels.py
```diff
@@ -0,0 +1,140 @@
...
+                    f'{type(self).__name__} can only be used with one label!'
+                )
+            else:
+                # deterministic flipping for each example
+                if flip:
+                    self.flip_cnt += 1
+                    if hashlib.md5(text.encode('utf8')).digest()[1] / 256 > 0.5:
+                        new_labels = ['__notok__']
+                    else:
+                        new_labels = ['__ok__']
```
jxmsML, line 135: similar to above, maybe change this hard-coded conversion logic

parlai/mutators/k_fold.py
```diff
@@ -0,0 +1,96 @@
...
+        agent = parser.add_argument_group('KFoldWithholdOnTrainMutator args')
+        agent.add_argument(
+            '--k-fold',
+            type=int,
+            help='the number of k-fold',
+        )
+        agent.add_argument(
+            '--held-fold',
+            type=int,
+            help='the curent fold reserved from training',
```
jxmsML, line 43: nit: curent -> current

parlai/mutators/k_fold.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+        return parser
+
+    def __init__(self, opt: Opt):
+        super().__init__(opt)
+        self.k_fold = opt['k_fold']
+        self.held_fold = opt['held_fold']
```
jxmsML, line 50: need to assert that 0<=held_fold<k_fold

parlai/mutators/flip_labels.py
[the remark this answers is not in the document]
klshuster: good point, thanks

parlai/mutators/flip_labels.py
dexterju27: I don't think this will be necessary? bc the mutator works for binary classification case anyway? __ok__ and __notok__are the default options for parlai classifier agent.
1.  I'm not sure if there are such use cases, eg binary classification with other labels?
2. It might make sense for people with other use cases to modify this code or write a mutator extending this to avoid unitended behaviours?

parlai/mutators/flip_labels.py
dexterju27: This will not reset, we log the self.flip_cnt when the agent is destroyed. I take this line as a sanity check for the mutator. I don't think there will be a use case where we really need those acurate numbers.

dexterju27:
Addressed all the comments, could we have a stamp?

klshuster (approved):

jxmsML (approved):
thanks! LGTM

jxmsML:
maybe try fixing the test failures if possible before merging?

PR #4670, opened 2022-07-20
Expose Precision and Recall to user via TeacherMetrics

prajjwal1:
**Patch description**
Considering that other metrics such as `ctpb`, `ctrunc`, `ctrunclen` (less frequently used by user) are exposed, it makes a lot of sense to expose `precision` and `recall` as they are internally being computed.


stephenroller (commented):
This is cool but I think it will be slow from recomputing these values?

prajjwal1:
The way this function is setup, extracting `f`, `p` and `r` all at once will break other functionalities. I've been using it for sometime now, doesn't seem to add much overhead. Most of the time `bs` is set to <256, so should be fine. I can make it return all three things with one run, but then some other things will have to be changed. Do you have suggestions ?

prajjwal1:
Or I can make "f1" default. If the user wants, they can print `precision` and `recall` themselves by passing the mode to `output` parameter.

prajjwal1:
@stephenroller Can you please provide what you think based on the reply ?

stephenroller:
Sorry; compute and return all 3 metrics in a generalization of the existing utility method. The caller will be responsible for separating and accumulating them.

prajjwal1:
@stephenroller Is it fine now ? We only compute `p`, `r` and `f1` only once now. So runtime is not affected.

prajjwal1:
@klshuster Can you please take a look ?

klshuster (commented):
this looks great. could you please update the metrics master list (https://github.com/facebookresearch/ParlAI/blob/main/parlai/core/metrics.py#L53) to have the new metrics, and also please rename 'precision' and 'recall' to refer to the fact that they are computed based on unigrams (e.g., `uni_prec`)

klshuster (approved):

PR #4677, opened 2022-07-22
Upgrade Mephisto, Remove old deps

JackUrb:
**Patch description**
I updated Mephisto here to use our new poetry-prepared version, which should resolve some dependency mismatch issues that had surfaced before between Mephisto and ParlAI. (Poetry is great by the way! May be worth it to adopt).

I also dropped out some dependencies from ParlAI that used to exist for the parlai-mturk implementation. Now that all of the correct requirements are installed when installing Mephisto, they don't belong here.

**Testing steps**
Crowdsourcing tests are passing

### DIFF
--- a/parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
+++ b/parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
@@ -22,6 +22,10 @@ class TurnAnnotationsStaticResultsCompiler(AbstractTurnAnnotationResultsCompiler
 
     Change PROBLEM_BUCKETS in task_config/annotations_config.json to be the buckets that
     you are asking crowdsource workers to annotate with.
+
+    NOTE this script directly accesses Mephisto files rather than using the DataBrowser.
+    This makes it fragile and not a great example for extension. It is a good candidate
+    for refactor.
     """
 
     NUM_SUBTASKS = 7
@@ -232,9 +236,21 @@ class TurnAnnotationsStaticResultsCompiler(AbstractTurnAnnotationResultsCompiler
                 continue
 
             # HIT-level metric of HIT completion time has to be done here for now
-            task_completion_time_seconds = (
-                data['times']['task_end'] - data['times']['task_start']
-            )
+            try:
+                task_completion_time_seconds = (
+                    data['times']['task_end'] - data['times']['task_start']
+                )
+            except KeyError:
+                # We've been relying on non-mephisto API, and in 1.0.2 'times' was deprecated
+                # so this uses the new location
+                metadata_path = os.path.join(os.path.dirname(dp), 'agent_meta.json')
+                with open(metadata_path, 'rb') as f:
+                    metadata = json.load(f)
+
+                # HIT-level metric of HIT completion time has to be done here for now
+                task_completion_time_seconds = (
+                    metadata['task_end'] - metadata['task_start']
+                )
             print(task_completion_time_seconds)
 
             subtasks = data['outputs']['final_data']
--- a/parlai/crowdsourcing/utils/tests.py
+++ b/parlai/crowdsourcing/utils/tests.py
@@ -261,7 +261,6 @@ class AbstractOneTurnCrowdsourcingTest(AbstractCrowdsourcingTest):
         """
         Given an agent state, test that it is as expected.
         """
-        del state['times']  # Delete variable timestamps
         data_regression.check(state)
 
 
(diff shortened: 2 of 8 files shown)


klshuster (approved):
thank you! can you please [bump the cache keys](https://github.com/facebookresearch/ParlAI/blob/75a58707fd35c6e75d143b051e9455cb01862f80/.circleci/config.yml#L211) in the circle config?

JackUrb:
Happy to - what do I need to do for that though?

klshuster:
Ahh, wherever there's a cache key (e.g., `data-20220615`), just update to a new date -- we cache the installs and need to bump the cache whenever we change `requirements.txt`

parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
```diff
@@ -22,6 +22,10 @@ class TurnAnnotationsStaticResultsCompiler(AbstractTurnAnnotationResultsCompiler
 
     Change PROBLEM_BUCKETS in task_config/annotations_config.json to be the buckets that
     you are asking crowdsource workers to annotate with.
+
+    NOTE this script directly accesses Mephisto files rather than using the DataBrowser.
```
EricMichaelSmith, line 26: good to flag - I have some outstanding that ports this over to DataBrowser, but some work will need to be done first to fully test it, convert the unit tests over, handle all old code that relies on this, etc.

parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
```diff
@@ -232,9 +236,21 @@ def compile_initial_results(self, results_folders) -> list:
...
             # HIT-level metric of HIT completion time has to be done here for now
-            task_completion_time_seconds = (
-                data['times']['task_end'] - data['times']['task_start']
-            )
+            try:
+                task_completion_time_seconds = (
+                    data['times']['task_end'] - data['times']['task_start']
+                )
+            except KeyError:
+                # We've been relying on non-mephisto API, and in 1.0.2 'times' was deprecated
```
EricMichaelSmith, line 244: Hmm - doing a simple search, it looks like `parlai/crowdsourcing/tasks/acute_eval/acute_eval_agent_state.py` and `parlai/crowdsourcing/tasks/dialcrowd/webapp-results/src/components/requesters/template/category/CategoryQuality.js` use this `'times'` key as well. I assume these will also need to be ported over? Or is the change in this file only a result of the fact that turn_annotations_static is still not using DataBrowser?

parlai/crowdsourcing/utils/tests.py
```diff
@@ -431,6 +430,16 @@ def _send_agent_message(
         self.await_channel_requests()
 
 
+def collapse_whitespace(in_string: str):
```
EricMichaelSmith, line 433: oh ha, this is great to have, I'm sure you had to deal with some frustrating table output check issues :/

EricMichaelSmith (commented):
This is great, thanks! Hmm, the one question I have is about the `'times'` key deprecation and whether this means that the remaining references to this key need to be removed (see comment)

parlai/crowdsourcing/utils/tests.py
JackUrb: Haha it wasn't too bad - I saw a few tests fail, copied and pasted the lines, and saw it was only whitespace. Probably just using a different version of `tabulate`

parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
JackUrb: Of course - it's a bit of work for sure but I wanted to make sure that anyone stumbling across this in the meantime knows what's up.

parlai/crowdsourcing/tasks/turn_annotations_static/analysis/compile_results.py
JackUrb: Good callout! I missed the `AcuteEvalAgentState`. We can drop the `times` reference here as Mephisto now does this under-the-hood as part of the `AgentState.metadata` object.

For `results/src/components/requesters/template/category/CategoryQuality.js` (and the `dialcrowd` code in general) I was having a hard time understanding what is and is not important to change. As it stands, I'm pretty sure it doesn't function quite correctly as the version of Mephisto their frontend uses predates `v1.0.0`. 

In the interest of not breaking anything _more_ though I'll add a fix for this too.

EricMichaelSmith: Okay, great, thanks! Yeah, in the absence of frontend testing, it won't be easy to tell how much `dialcrowd` is becoming out of date, especially since we have no in-house maintainer of this code. @klshuster happy to hear your thoughts on this as well - maybe we should just mark this task as working on Mephisto v1.0.0 and untested on later versions?

Also @JackUrb it looks like there's still a lint error?

JackUrb: Lint error doesn't appear locally, or two commits ago, and is on a file this PR doesn't touch. Not sure how to resolve (should I change that file myself?)

EricMichaelSmith: ha good catch :/ Okay, I'm approving this now just because that shouldn't block you

EricMichaelSmith (approved):
Great to have these fixes!

PR #4685, opened 2022-07-26
Removing k-fold teacher from main

dexterju27:
**Patch description**
This mutator will stay in project folder.


dexterju27:
Hey, could you two give a stamp on this?  I'll move this mutator to the project folder instead. Thanks!

klshuster (approved):

PR #4686, opened 2022-07-26
updated acute eval docs to use ec2 architect instead of heroku

juderoque:
**Patch description**
Updated acute eval docs to use ec2 architect instead of heroku.


klshuster (approved):
thanks!

PR #4689, opened 2022-07-28
multiparty BB2

chiehminwei:
## Summary
After reading through the entire `BB2` codebase, it looks like to support multiparty chat, the only changes necessary are disabling adding the prefixes (`“partner’s persona: “`, `“your persona: “`) in the `_extract_from_raw_memories` method of `MemoryDecoder`, and disabling the removal of `__P1__` and `__P2__` tokens in `_remove_person_tokens`.

`BB2` removes `__P1__` and `__P2__` tokens from `obs[‘full_text’]` before adding the prefixes; if `opt[‘add_person_tokens’]` is `True` in `TorchAgent`, then `BB2` calls `_remove_person_tokens` before vectorizing the text and saves them in `obs[‘query_vec’]`, `obs[‘query_generator_vec’]`, and `obs[‘memory_decoder_vec’]`.

Since [the multiparty teacher](https://github.com/fairinternal/ParlAI-Internal/pull/3312) has the speaker labels in `obs[‘full_text’]` already,  we can skip `_remove_person_tokens` and also skip adding the prefixes. Since no distinction in memory is made for different personas in `LongtermMemory`, its methods (`write_memory`, `score_memories`, and `retrieve_and_score`) or relevant methods (e.g. `access_long_term_memory` in `BlenderBot2RagModel`), just having the correct speaker labels for each persona would be sufficient to support multiparty chat.

## Relevant `opt`

* `—-add-person-tokens`
	* Change behavior of calling `_remove_person_tokens` when `--add-person-tokens` is `True`
* `--retriever-ignore-phrase`
	* Defaults to `"persona: "`.
	* No need to change. Not gonna ignore anything, and `“persona: “` is not in my teacher’s `obs['"full_text”]`
* `--memory-key`
	* Set to `personas`
* `--memory-extractor-phrase`
	* Set this to `None` so it doesn’t filter out everything
* `--memory-reader-model`
	* Choose from  `['bert', 'bert_from_parlai_rag', 'dropout_poly']`

## Proposed Changes  

Add an `--use-multiparty` (`bool`) that disables calling `_remove_person_tokens` and disables adding prefixes

### DIFF
--- a/projects/blenderbot2/agents/blenderbot2.py
+++ b/projects/blenderbot2/agents/blenderbot2.py
@@ -247,6 +247,15 @@ class BlenderBot2RagAgent(RagAgent):
             'none => do not access any knowledge.\n',
         )
         bb2_group.add_argument(
+            '--use-multiparty',
+            type=bool,
+            default=False,
+            help='Whether to enable multiparty mode by disabling calls to _remove_person_tokens and '
+            "disabling adding prefixes ('your persona: ', 'partner's persona: ') to text. "
+            'Assumes the speaker label is already in the text. '
+            'Defaults to False.',
+        )
+        bb2_group.add_argument(
             '--memory-key',
             type=str,
             default='full_text',
@@ -290,7 +299,7 @@ class BlenderBot2RagAgent(RagAgent):
         )
         bb2_group.add_argument(
             '--memory-extractor-phrase',
-            type=str,
+            type='nonestr',
             default='persona:',
             help="phrase used to extract memories from `--memory-key` in the observation. "
             "For example, set to 'your persona:' to limit memories to only lines that "
@@ -576,7 +585,7 @@ class BlenderBot2RagAgent(RagAgent):
                 self.opt['retriever_ignore_phrase'],
                 delimiter=self.opt['retriever_delimiter'],
             )
-        if self.add_person_tokens:
+        if self.add_person_tokens and not self.opt['use_multiparty']:
             query_str = self._remove_person_tokens(query_str)
         observation['query_vec'] = self.model_api.tokenize_query(query_str)
         return observation
@@ -645,7 +654,7 @@ class BlenderBot2RagAgent(RagAgent):
                     self.opt['query_generator_ignore_phrase'],
                     self.opt['query_generator_delimiter'],
                 )
-            if self.add_person_tokens:
+            if self.add_person_tokens and not self.opt['use_multiparty']:
                 query_generator_input = self._remove_person_tokens(
                     query_generator_input
                 )
@@ -742,7 +751,7 @@ class BlenderBot2RagAgent(RagAgent):
                     self.opt['memory_decoder_ignore_phrase'],
                     self.opt['memory_decoder_delimiter'],
                 )
-            if self.add_person_tokens:
+            if self.add_person_tokens and not self.opt['use_multiparty']:
                 memory_decoder_input = self._remove_person_tokens(memory_decoder_input)
             conv_lines = [
                 t
--- a/projects/blenderbot2/agents/sub_modules.py
+++ b/projects/blenderbot2/agents/sub_modules.py
@@ -14,6 +14,7 @@ import time
 import torch
 import torch.nn
 from typing import List, Tuple, Dict, Optional, Any
+from collections import defaultdict
 
 from parlai.agents.rag.retrievers import clean_vec
 from parlai.core.agents import create_agent_from_model_file, create_agent_from_shared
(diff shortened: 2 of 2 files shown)


projects/blenderbot2/agents/blenderbot2.py
```diff
@@ -290,7 +299,7 @@ def add_cmdline_args(
         )
         bb2_group.add_argument(
             '--memory-extractor-phrase',
-            type=str,
+            type='nonestr',
```
mojtaba-komeili, line 302: I haven't seen this before. Does it mean none or str?

projects/blenderbot2/agents/sub_modules.py
```diff
@@ -347,6 +348,56 @@ def _extract_from_raw_memories(self, raw_memories: List[str]) -> List[str]:
...
 
+        :param raw_memories:
+            raw memory generations. sometimes we need skip the memories because
+            nothing was generated
+
+        :return memories:
+            return filtered memories
+        """
+        if self.opt['use_multiparty']:
+            return self._extract_from_raw_memories_multi_speaker(raw_memories)
```
mojtaba-komeili, line 359: Good job separating them logically into two functions.

projects/blenderbot2/agents/blenderbot2.py
```diff
@@ -742,7 +751,7 @@ def _set_memory_decoder_vec(self, observation: Message) -> Message:
                     self.opt['memory_decoder_ignore_phrase'],
                     self.opt['memory_decoder_delimiter'],
                 )
-            if self.add_person_tokens:
+            if self.add_person_tokens and not self.opt['use_multiparty']:
```
mojtaba-komeili, line 754: Does this mean we keep the persona tokens for the multiparty chat? If I have it correctly,`memory_decoder_input` is going into the memory decoder (am I right). Don't we need to remove all the persona tokens from it and then rematch them with the memories later?

mojtaba-komeili (commented):
Could you also check the input and output to the memory decoder, maybe via `logging.debug` to make sure we are passing them to memory decoder correctly and then we receive and match them to the memories right afterwards?

projects/blenderbot2/agents/blenderbot2.py
chiehminwei: Yes it means none or str: https://github.com/facebookresearch/ParlAI/blob/304fe1c61fc135a76c03555b45f04df0f4550805/parlai/core/torch_agent.py#L705

projects/blenderbot2/agents/blenderbot2.py
chiehminwei: Yes, `memory_decoder_input` goes into the memory decoder.

So`memory_decoder_input` (which defaults to `observation['full_text']`) first gets tokenized and saved as `observation['memory_decoder_vec']`:
https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/blenderbot2.py#L752-L757

Afterwards, `observation['mememory_decoder_vec]` is then passed to `self.memory_decoder.generate_memories`
https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/modules.py#L359-L364

Which then goes through each example in the batch, 
https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/sub_modules.py#L327-L337

calls `self._batch_generate`, which observes the input and adds a reply as the output
https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/sub_modules.py#L140-L148

and finally calls `self. _extract_from_raw_memories`, which adds the persona tokens.
https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/sub_modules.py#L365-L375

These are all done before the `generated_memoreis` get passed to the `MemoryDecoder` in `self.access_longterm_memory`:
https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/modules.py#L407-L414

Finally, `access_longterm_memory` then takes care of writing the `generated_memories` (which comes from memory decoder's replies to `observation['full_text']`), along with `memory_vec ` from `observation['personas']`, to memory.
https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/modules.py#L588-L609

https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/modules.py#L639-L649

chiehminwei: So I think you're right, if `self._batch_generate` doesn't want persona tokens, we should remove them first and restore them later

https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/sub_modules.py#L335

I think this means I'll need to modify the teacher and remove speaker labels in `observation['text']`, add a `speakers=['A', 'B', 'C']` field in `observation`.

So the teacher `text` would change from this:

```
text:
    Knight: Hello
    Queen: Hi
    King:

label:
    Hola
```

to this:
```
text:
    Hello
    Hi

label:
    Hola

speakers:
    Knight
    Queen
    King
   
```

I will also need to extend a custom `Batch` class that includes `batch.speakers` (if we don't want to touch `TorchAgent`), and maybe also add `batch.speakers` the `_model_input`:

https://github.com/facebookresearch/ParlAI/blob/5f13a2d805131688aaf5253d4e29adc8b4ad6020/projects/blenderbot2/agents/blenderbot2.py#L873-L909

It seems like I won't need to vectorize the `speakers` in the `batch`, since I only need the raw text.

Thoughts?

chiehminwei:
Closed as moving things to internal

PR #4693, opened 2022-07-29
Friends Dataset Teacher Add `speakers` field and flag to exclude speaker labels in `text`

chiehminwei:
**Patch Description**
This patch adds a `speakers` field in the message generated from the teacher for the Friends dataset.

It also adds an option to exclude speaker ids from the `text`. This option is named `--include-speaker-in-context`. The original `--include-speaker-in-context` flag, which determines whether speaker labels are added at the end of `text`, is renamed to `--add-speaker-to-context-end`.

Together, these changes make it convenient to feed the data into downstream models such as BlenderBot2 that require cleaned `text` input without speaker labels, but need to restore the speaker labels later.

**Sample Output**
`parlai dd -t friends -n 2 --verbose --include-speaker-in-context True --add-speaker-to-context-end True`
This is the default behavior.
image

`parlai dd -t friends -n 2 --verbose --include-speaker-in-context True --add-speaker-to-context-end False`
image

`parlai dd -t friends -n 2 --verbose --include-speaker-in-context False --add-speaker-to-context-end True`
Notice the empty line which marks an empty sentence, and notice how the current speaker label is also added to the `speakers` field.
image

`parlai dd -t friends -n 2 --verbose --include-speaker-in-context False --add-speaker-to-context-end False`
image

### DIFF
--- a/parlai/tasks/friends/agents.py
+++ b/parlai/tasks/friends/agents.py
@@ -92,7 +92,6 @@ class DefaultTeacher(DialogTeacher):
                         continue
 
                 speaker = utterance['speaker']
-                speakers.append(speaker)
 
                 # Replace newline character by whitespace so that the data
                 # format plays nicely with BB2, which splits each utterance
@@ -106,7 +105,7 @@ class DefaultTeacher(DialogTeacher):
                 if self.include_speaker_in_context:
                     context += self.utterance_delimiter + f'{speaker}: {text}'
                 else:
-                    context += self.utterance_delimiter + {text}
+                    context += self.utterance_delimiter + text
 
                 isConversationDone = index == last_utterance_index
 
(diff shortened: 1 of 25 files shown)


chiehminwei:
I took a look at the failing tests, and they don't seem to be related to this PR.
Something else on the `main` branch is broken.
For the gpu tests it's the same old `sqlite3.OperationalError: database or disk is full`.
For the teacher tests it's the `parlai/tasks/multiwoz_v22` teacher that's failing.

parlai/tasks/friends/agents.py
```diff
@@ -89,23 +100,80 @@ def setup_data(self, datafile):
                 if (
                     self.character == 'All' and speaker in self.characters
                 ) or speaker == self.character:
+                    text, label, speakers, hasAddedSpeaker = self._get_message_fields(
+                        text, speaker, speakers, prev_context
+                    )
+                    _speakers = speakers[:]
```
mojtaba-komeili, line 106: curious, why do you do this?

mojtaba-komeili (approved):
Thanks!

parlai/tasks/friends/agents.py
chiehminwei: Pass by reference vs pass by value
If I don't do this then every example becomes the same as the last example, containing all the speakers in the entire episode  (I ran into this bug)

PR #4698, opened 2022-08-01
Add DST teacher for GoogleSGD

jacksukk:
Adding this teacher would allow google_sgd to be usable for Dialog State Tracking task.
cc @chinnadhurai


klshuster:
could you please run `pytest --force-regen parlai/tasks/google_sgd`?

jacksukk:
> could you please run `pytest --force-regen parlai/tasks/google_sgd/test.py` and then push the updated/autogenerated test files?

Done, thanks for reminding! Let me know if there's anything I can do. :)

jacksukk:
> could you please run `pytest --force-regen parlai/tasks/google_sgd/test.py` and then push the updated/autogenerated test files?

Done, thanks for reminding! Let me know if there's anything I can do

klshuster:
Ahh yes, one more thing: as you've added a new package to `requirements.txt`, you'll need to update the cache key in our [circle config file](https://github.com/facebookresearch/ParlAI/blob/5b2fc1d9c5fb74d4ccb641138cf21765557d78e1/.circleci/config.yml#L210). You can just do a string replace of `'20220805'`

jacksukk:
> Ahh yes, one more thing: as you've added a new package to `requirements.txt`, you'll need to update the cache key in our [circle config file](https://github.com/facebookresearch/ParlAI/blob/5b2fc1d9c5fb74d4ccb641138cf21765557d78e1/.circleci/config.yml#L210). You can just do a string replace of `'20220805'` for each entry to something like `'20220819'`

Done, Thanks!!

klshuster (approved):
all looks good on my end, thanks!

PR #4707, opened 2022-08-04
Add FlashAttention Kernel in Triton

pearlli98:
**Tldr: Add implementations of [FlashAttention](https://arxiv.org/abs/2205.14135) using OpenAI's triton language.**

**Major Changes:**
1. add FlashAttention forward pass to parlai.
2. add custom masking and dropout to FlashAttention code.
3. replace encoder self-attention with FlashAttention.
4. add unit test for self-attention and encoder/decoder attention functionality to ensure correctness within acceptable epsilon(0.01).
---

**Background**:
- FlashAttention: an IO-aware exact attention algorithm that uses tiling to reduce the number of memory reads/writes, 15% end-to-end speedup on BERT-large compared to the MLPerf 1.1, 3× speedup on GPT-2, and 2.4× speedup on long-range arena.
-  [Triton](https://openai.com/blog/triton/): Python-like programming language to write highly efficient GPU code. 
---

**Important Notes**: 
I referenced Triton's implementation of FlashAttention [here](https://github.com/openai/triton/blob/master/python/tutorials/06-fused-attention.py). However, I ran into the following problems and made some modifications accordingly:
   1. triton's fused attention example does not work with arbitrary sequence length. The inputs need to be  padded before the calculations.
   2. triton did not implement custom masking and dropout. I added both to the kernel. Masking works properly. The dropout implementation I used, however, does not match `nn.Dropout`, which is used by Parlai. Unit tests on dropout do not pass at the moment. 
   3. triton's implementation only works on A100 because of a compiler bug they have. The issues are explained [here](https://github.com/openai/triton/issues/616). 
   4. Currently, setting masked items to `float("-inf")` causes the outputs to contain `Nan`. Parlai seems to have experienced a similar problem according to @stephenroller and thus, uses a near infinity value (`-1e20` for float32). The newly added triton kernel uses that same value to work around the Nan problems. 
---

**Evaluations:**
1. unit test runtime in ms for self-attention with masking (no dropout). bsz, n_head, head_dim = 3, 36, 64. To run the benchmark: `python tests/test_triton_mha.py`.

|  N_CTX | Triton Modified |   ParlAI  | Speedup | Vanilla Triton |
|:------:|:---------------:|:---------:|:-------:|----------------|
| 512.0  | 0.333863        | 0.696174  | 2.09x   | 0.077230       |
| 1024.0 | 1.005227        | 2.410752  | 2.41x   | 0.191713       |
| 2048.0 | 3.513344        | 9.326592  | 2.66x   | 0.596668       |
| 4096.0 | 13.148160       | 37.028866 | 2.81x   | 2.096564       |

3. convai2 teacher 
    - runtime: I **do not see an obvious runtime difference** between parlai implementation and the new triton kernel. I suspect that padding the inputs and slicing them in the end, which is required for the current triton kernel, are making the operations slower and muting the potential advantages we saw for the unit tests. 
    - memory usage: no significant changes observed 
    - comparison of outputs
        - quantitatively, the average of `(parlai_out - triton_out).abs().sum(dim=-1)` is 0.0318, which is reflected in the output utterances. 268 out of 7801 sentences generated by the two models differ, which is about 3.4%. 
---

**Testing Steps:**
1. unit tests: `pytest tests/test_triton_mha.py`
   - add `-k test_self_attention` to test self attention
   - add `-k test_enc_dec_attention` for enc/dec attention
   - add `-k test_dropout` for self attention with dropout, current do not pass ❌
3. to run convai2 teacher task, `parlai eval_model --model-file "zoo:tutorial_transformer_generator/model" -t convai2 --optimizer adam -dt valid --bs 64 --gpu-beam-blocking True --flash-attention True`


stephenroller:
wow this heroic change set

klshuster:
is this ready for review or still a draft?

dexterju27:
@klshuster  I don't think the code will ever be mergeble given Triton's experimental nature. I talked with some people worked on Flash attention, it seems Triton's implementation only work with specific head dim. 
I asked @pearlli98  to open a PR in order to save her WIP, and we could review the code if we want to.

klshuster:
perhaps we can merge some of the results to an internal project directory?

pearlli98:
> perhaps we can merge some of the results to an internal project directory?

@klshuster 
I can push the results to parlai-internal. The problem is that we don't have the transformers modules files, which i made changes to, on the internal repo. Do you want me to separate the changes in two different places (code files here and results on internal) ?

pearlli98:
> yes that would be great

@klshuster I have removed the results folder and move it to ParlAI-internal under this [PR](https://github.com/fairinternal/ParlAI-Internal/pull/3604).

PR #4720, opened 2022-08-06
expose retain_graph to user within TorchAgent

prajjwal1:
**Patch description**
Although `loss.backward()` is the norm in usual workflow, there are many instances in which a user might want to pass in `retain_graph=True` within `loss`. Considering that parlai doesn't expose the loss api to user, since `TorchAgent` expects user to call `self.backward()`, it would be a good addition to let the user pass in `retain_graph=True` if they want to. User can also pass in `create_graph=False`, `inputs=None` params if they want to, to `backward()`.

**Testing steps**

**Other information**


klshuster (approved):
lgtm

PR #4727, opened 2022-08-08
Remove TGA._get_context()

juderoque:
**Patch description**
- Deleted `TGA._get_context()` as it is not used anywhere and was replaced with `TGA._get_batch_context()`
- Added warning incase users try to implement `TGA._get_context()`

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -1057,7 +1057,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
         if hasattr(self, '_get_context'):
             # Warn users that have subclassed with '_get_gontext
             warn_once(
-                "WARNING: TGA._get_context() is deprecated, use TGA.get_batch_context() instead"
+                "WARNING: TGA._get_context() has been removed, use TGA.get_batch_context() instead"
             )
 
         if self.beam_context_block_ngram <= 0:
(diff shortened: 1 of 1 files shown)


klshuster (commented):
I think this is used in the [RAG Agent](https://github.com/facebookresearch/ParlAI/blob/1d38dae7be3a7260cb2706bb305f124655348c4a/parlai/agents/rag/rag.py#L650-L660)?

does that mean that that function was never correctly being used?

juderoque:
> I think this is used in the [RAG Agent](https://github.com/facebookresearch/ParlAI/blob/1d38dae7be3a7260cb2706bb305f124655348c4a/parlai/agents/rag/rag.py#L650-L660)?
> 
> does that mean that that function was never correctly being used?

Yep seems that way.. seems that the function is never actually called even if one was to override it

klshuster:
> > I think this is used in the [RAG Agent](https://github.com/facebookresearch/ParlAI/blob/1d38dae7be3a7260cb2706bb305f124655348c4a/parlai/agents/rag/rag.py#L650-L660)?
> > does that mean that that function was never correctly being used?
> 
> Yep seems that way.. seems that the function is never actually called even if one was to override it

Very interesting. I'm ok with this then

klshuster (approved):

PR #4729, opened 2022-08-09
modify ROUGEMetric to return any of precision, recall, f1-score in metrics.py

pranav-maddali:
allows user to determine which measure of `rouge.Rouge` (precision, recall, f1-score) to use 

**Patch description**
Although recall may be the most used rouge score measure, it is important to allow the user to determine which measure to return while invoking `RougeMetric.compute_many()`. User can continue to not pass `measure` as a parameter and the function defaults to `recall 'r'`.

**Testing steps**

**Other information**


klshuster (approved):
thanks!

PR #4730, opened 2022-08-09
[Mutators] Check if partial_opts is None

spencerp:
**Patch description**
`partial_opts` is optional and can therefore be None, in which case `partial_opt.get('mutators')` will throw an exception.


mojtaba-komeili (approved):
LGTM

spencerp:
So many tests broken right now :(

Comparing with other PRs, though, it doesn't appear that this change introduces any new failures.

PR #4732, opened 2022-08-09
Added data filter to the BotAdversarialDialogueTeacher to remove repeating "Hey do you want to talk about something else" episodes.

leox1v:
In the data ~7k episodes end with "Hey do you want to talk about something else". With this flag we can skip these episodes.


jxmsML:
wonder if we could replace the flag with a "mutator"

jaseweston (approved):

leox1v:
@jxmsML thanks for the comment! I changed the filter to a mutator. What do you think?

klshuster (approved):

PR #4736, opened 2022-08-10
Fix seed from opt in rng sampler of multiwoz_v22

prajjwal1:
**Patch description**
Since the support for setting seed in `train_model.py` has been added, we can let user fix the seed in dataset sampler also.

**Testing steps**

**Other information**


parlai/tasks/multiwoz_v22/agents.py
```diff
@@ -416,14 +416,16 @@ class MultiWOZv22DSTTeacher(tod_agents.TodUserSimulatorTeacher):
...
 
-    rng = np.random.RandomState(SEED)
-
     def __init__(self, opt: Opt, shared=None, *args, **kwargs):
         self.opt = opt
         self.fold = opt["datatype"].split(":")[0]
         opt["datafile"] = self.fold
         self.dpath = os.path.join(opt["datapath"], "multiwoz_v22")
         self.id = "multiwoz_v22"
+        if self.opt["seed"]:
```
moyapchen, line 425: Would recommend using `self.opt.get("seed")` to be safe, but otherwise looks fine.

moyapchen (approved):
See inline comment; Accepting under assumption recommendation there taken

moyapchen:
Actually thinking about this slightly more - nothing actually adds a flag for "seed" to the argparser. How are you passing the "seed" argument down?

prajjwal1:
@moyapchen I now perform indexing with `.get()`. Even I realized the issue you raised, I thought it would come from `train_model.py`, but seems like it has to come from `Teacher`, so the argument for `seed` is propagated from `Teacher`. Is this fine now ?

parlai/core/teachers.py
```diff
@@ -163,6 +163,7 @@ class Teacher(Agent):
     def add_cmdline_args(
         cls, parser: ParlaiParser, partial_opt: Optional[Opt] = None
     ) -> ParlaiParser:
+        parser.add_argument('--seed', default=None)
```
klshuster, line 166: can we change this to `--teacher-seed`? to differentiate from the `--seed` already in train_model.py?

parlai/core/teachers.py
```diff
@@ -163,6 +163,7 @@ class Teacher(Agent):
     def add_cmdline_args(
         cls, parser: ParlaiParser, partial_opt: Optional[Opt] = None
     ) -> ParlaiParser:
+        parser.add_argument('--teacher_seed', default=None)
```
klshuster, line 166: could we also enforce this to have `type=float`?

klshuster (approved):

PR #4742, opened 2022-08-12
[LIGHT] Who Am I - Preventing doubleloading model

JackUrb:
**Patch description**
Made a small update to the PACER model plugin that allows for it to be added to other agents (like the `ExpandedAttentionDecoderAndPacerAgent`) without doubly loading the base model by instead checking to see if the reranking model has been initialized in some other form.

**Testing steps**
Was consuming 1.2Gb more GPU space than expected and observed logs loading the reranker model twice before this change.

Model works correctly (as tested in the LIGHT game) after this change.

**Other information**


klshuster (approved):
thanks for upstreaming this!

PR #4746, opened 2022-08-15
[BB3] Agents README

klshuster:
**Patch description**
Providing some more information about how BB3 internals work.


projects/bb3/agents/README.md
```diff
@@ -0,0 +1,413 @@
...
+```
+
+Note: The search documents here are passed in _directly in the context_, with the appropriate prefixes.
+
+##### Expected Output
+Knowledge sentence(s) on which to ground a dialogue response.
+
+###  `CRM`: Contextual Response Module
+
+Given an extracted entity from the context, generate a dialogue response. Looks at the whole dialogue context. Default inference uses beam search in the 3B model and factual nucleus decoding in the 30B/175B models.
```
mojtaba-komeili, line 250: Could you add a reference to factual nucleus decoding, please?

mojtaba-komeili (approved):
My understanding was that in some places (such as search submodule) there was no batching? Is that right? Should we add that as note if so?

klshuster:
> My understanding was that in some places (such as search submodule) there was no batching? Is that right? Should we add that as note if so?

hmm not sure what you mean by that; all of these should support batching

there was a batching issue was on the metaseq side, but that has been resolved

PR #4749, opened 2022-08-17
[GPU] Fix Tests

klshuster:
**Patch description**
1. #4670 introduced an error in several places that used `F1Metric.compute`
2. fixing an iterable issue in the bb3 tests

**Testing steps**
CI


jxmsML (approved):
LGTM

PR #4753, opened 2022-08-18
[BB3] Option to ignore in session memories

klshuster:
**Patch description**
Provide an option to ignore in-session memories when generating a dialogue response. An "in session memory" is any memory that is generated during the current dialogue session; a "session" is the life-cycle of the BB3 agent. 

If set to `True`, *only* memories seeded in the first observation (with the correct `Opening:` prefix) will be used throughout a conversation.

**Testing steps**
Several added tests. Note: this depends on #4749 , so tests will fail until that is merged

### DIFF
--- a/projects/bb3/agents/r2c2_bb3_agent.py
+++ b/projects/bb3/agents/r2c2_bb3_agent.py
@@ -421,6 +421,7 @@ class BlenderBot3Agent(ModularAgentMixin):
             self.agents[Module.SEARCH_KNOWLEDGE] = agent
 
         self.memories = []
+        self.in_session_memories = set()
         self.search_knowledge_responses = ['__SILENCE__']
         self.memory_knowledge_responses = ['__SILENCE__']
         self.contextual_knowledge_responses = ['__SILENCE__']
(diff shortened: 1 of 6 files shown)


projects/bb3/agents/r2c2_bb3_agent.py
```diff
@@ -421,6 +421,7 @@ def __init__(self, opt, shared=None):
             self.agents[Module.SEARCH_KNOWLEDGE] = agent
 
         self.memories = []
+        self.in_session_memories = set()
```
mojtaba-komeili, line 424: Do we need to have this as a set? Why not relying on the memory deduplication algorithm we had before?

klshuster: the `set` is just for quick O(1) checking in the `get_available_memory` function for memory utils

memory in list --> O(n)
memory in set --> O(1)

mojtaba-komeili: OK, I see. I thought you had that for deduplication. That makes sense.

mojtaba-komeili (approved):

PR #4757, opened 2022-08-19
Fix Friends Teacher regression tests

chiehminwei:
This patch updates the test fixtures for the Friends Teacher


klshuster (approved):
thanks!

PR #4759, opened 2022-08-19
Bump Requirements

klshuster:
**Patch description**
1. Numpy is from a dependabot alert
2. tomli is from our build failing (`error: tomli 1.2.3 is installed but tomli<3.0.0,>=2.0.0 is required by {'docformatter'}`)

**Testing steps**
Let's see if CI works!


mojtaba-komeili (approved):
As far as I can tell there is nothing wrong with it.

PR #4761, opened 2022-08-22
[Tests] Fix autotests

klshuster:
**Patch description**
fixing autotests for two teachers (google sgd, multiwoz v22)

**Testing steps**
pytest


JackUrb (approved):
✅

PR #4762, opened 2022-08-22
[Test] Fix MultiWozv22 testing

klshuster:
**Patch description**
Turns out that multiwoz shuffles the data *always*. This is fine, but the auto-generated test files were different on my machine vs. circle CI due to different random number generators. This PR forces determinism when in a continuous integration state

**Testing steps**
CI PLEASE


EricMichaelSmith (approved):
Stamping, but others can speak to the test content more

PR #4763, opened 2022-08-22
Added shared embedding option to director model.

leox1v:
Added an option (with flag --director-use-shared-embedding) to share the generation and classification heads of the director. Only two additional parameters are introduced to scale the logits before applying the sigmoid (for the classification).

### DIFF
--- a/projects/director/director_agent.py
+++ b/projects/director/director_agent.py
@@ -26,19 +26,38 @@ from parlai.core.torch_generator_agent import PPLMetric
 import parlai.utils.logging as logging
 
 
+class ScalarLayer(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.params = nn.Parameter(torch.Tensor([1.0, 0.0]))
+
+    def forward(self, input: torch.Tensor):
+        return input * self.params[0].expand_as(input) + self.params[1].expand_as(input)
+
+
 class DirectorModel(TransformerGeneratorModel):
     """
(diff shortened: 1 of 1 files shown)


jaseweston (approved):

projects/director/director_agent.py
```diff
@@ -56,6 +75,9 @@ def classifier_output(self, input: torch.Tensor):
         if self.freeze_decoder:
             input = input.detach()
 
+        if self.use_shared_embedding:
```
jxmsML, line 78: should this come before the L75 `input = input.detach()`

projects/director/director_agent.py
```diff
@@ -26,19 +26,38 @@
...
+        use_shared_embedding=False,
+        **kwargs,
+    ):
         super().__init__(opt, dictionary, **kwargs)
 
         vocabulary_size = len(dictionary)
 
         decoder_output_dim = self.decoder.out_dim
-        self.classifier_heads = nn.Linear(decoder_output_dim, vocabulary_size)
+        self.use_shared_embedding = use_shared_embedding
```
jxmsML, line 56: if `use_shared_embedding` is passing from opt, then you can make sure of that  `self.use_shared_embedding = opt.get('use_shared_embedding', False)?

leox1v: Thanks for the comment. I changed it.

jxmsML (approved):

PR #4765, opened 2022-08-22
[BB3/SeeKeR/BB2] Simplify Retriever Building

klshuster:
**Patch description**
I've abstracted building the retriever/search engine object such that we no longer require hacky opt-setting to make sure that the correct retriever is built for an agent. Any model that subclasses `RagModel` (i.e., BB2, BB3, SeeKeR, FiD, etc.) can now specify a `build_retriever` method.

**Testing steps**
Relying on existing CI

### DIFF
--- a/parlai/agents/rag/modules.py
+++ b/parlai/agents/rag/modules.py
@@ -28,7 +28,7 @@ from parlai.core.opt import Opt
 from parlai.core.torch_generator_agent import TorchGeneratorModel
 from parlai.utils.torch import padded_tensor
 
-from parlai.agents.rag.retrievers import retriever_factory, Document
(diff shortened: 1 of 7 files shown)


mojtaba-komeili (approved):
Do we need custom building for each under retriever factory or as long as they are subclass of the major ones (eg RAG, FiD, etc.) we are OK.

klshuster:
> Do we need custom building for each under retriever factory or as long as they are subclass of the major ones (eg RAG, FiD, etc.) we are OK.

As long as we subclass a major one it's all good. Everything stems from `RagModel` and that defines the standard `build_retriever` so it's totally fine to not define one

PR #4768, opened 2022-08-23
[torch_agent]return serializable score list

wyshi:
**Patch description**

**Testing steps**

**Other information**


klshuster (commented):
could you maybe do a quick search through the repo to make sure that there are no assumptions about `sorted_scores` being a tensor?

wyshi:
> could you maybe do a quick search through the repo to make sure that there are no assumptions about `sorted_scores` being a tensor?

sure, will do later, I did the change mainly because tensor is not serializable and cannot be saved to a json

klshuster:
@jxmsML

PR #4770, opened 2022-08-24
[BB3] Memory Heuristics

klshuster:
**Patch description**
Implement the following memory heuristics when selecting what memories to show the bot when crafting a response:

1. `memory_overlap_threshold` --> if set to > 0, we block any memories that do not have this F1 overlap with the incoming text
2. `memory_hard_block_for_n_turns` --> if set to > 0, we block memories that have been used in the last n turns
3. `memory_soft_block_decay_factor` --> if set to > 0, we block memories according the following probability: `pr(block) = random.random() < decay_factor ** n_turns_since_used`

This required an overhaul of how `memories` are stored in the agent; rather than a list of memories, we now use a dictionary mapping `memory: n_turns_since_use`

**Testing steps**
Several more tests added

### DIFF
--- a/projects/bb3/agents/opt_bb3_agent.py
+++ b/projects/bb3/agents/opt_bb3_agent.py
@@ -285,16 +285,7 @@ class BlenderBot3Agent(R2C2Agent):
             prefixed_memories,
             set(),
             dictionary=self.dictionary,
-            ignore_in_session_memories=self.opt.get(
-                'ignore_in_session_memories_mkm', False
-            ),
-            memory_overlap_threshold=self.opt.get('memory_overlap_threshold', 0.0),
-            memory_hard_block_for_n_turns=self.opt.get(
-                'memory_hard_block_for_n_turns', 0
-            ),
-            memory_soft_block_decay_factor=self.opt.get(
-                'memory_soft_block_decay_factor', 0.0
-            ),
+            **self._get_memory_heuristic_values(),
         )
         if not memories_to_use:
             # we need at least one memory to open with...
@@ -314,11 +305,11 @@ class BlenderBot3Agent(R2C2Agent):
         This function is designed to handle legacy cases where memories are
         presented as a list, rather than a dict.
 
-        :param observation:
-            incoming message
+        :param memories:
+            memories from the opening message
 
         :return opening_memories:
-            return the set of opening memories
+            return the set of true opening memories
         """
         opening_memories = None
         if memories:
--- a/projects/bb3/agents/opt_bb3_agent.py
+++ b/projects/bb3/agents/opt_bb3_agent.py
@@ -153,13 +153,6 @@ class BlenderBot3Agent(R2C2Agent):
             help='Number of times to retry on API request failures (< 0 for unlimited retry).',
         )
         parser.add_argument('--metaseq-server-timeout', default=20.0, type=float)
-        parser.add_argument(
-            '--ignore-in-session-memories-mkm',
-            type='bool',
-            default=False,
-            help='If true, we do not look at the in-session memories when '
-            'generating from the MKM',
-        )
         return parser
 
     def __init__(self, opt, shared=None):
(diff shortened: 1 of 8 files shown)


projects/bb3/agents/opt_bb3_agent.py
```diff
@@ -285,27 +274,65 @@ def get_orm_observation(
...
+            set(),
+            dictionary=self.dictionary,
+            **self._get_memory_heuristic_values(),
         )
-        new_obs.force_set('memories', opening_memories)
+        if not memories_to_use:
+            # we need at least one memory to open with...
+            memories_to_use = random.sample(list(prefixed_memories.keys()), 1)
+
+        new_obs.force_set('text', self._check_and_limit_len('\n'.join(memories_to_use)))
```
mojtaba-komeili, line 294: Why are we setting `text` here?

projects/bb3/agents/opt_bb3_agent.py
```diff
@@ -285,27 +274,65 @@ def get_orm_observation(
...
         # handle passed memories as well
         observation = Message(observation)
-        opening_memories = observation.pop('memories', None)
+        opening_memories = self.get_opening_memories(observation.pop('memories', None))
         observations = super().observe(observation)
         for m in Module.dialogue_modules():
             ag_obs = copy.deepcopy(observation)
             observations[m] = self.agents[m].observe(ag_obs)
         if is_opener(observation['text'], opening_memories):
+            assert opening_memories
```
mojtaba-komeili, line 335: Question: having no memory does it make it generate an opening message anyway? I mean, if it doesn't crash the model, it may be a legitimate case to have here (?).

projects/bb3/agents/opt_bb3_agent.py
```diff
@@ -701,7 +725,15 @@ def _failed_messages(replies):
             retries += 1
             n_mems = [min(1, len(obs['memories']) // 3) for obs in opening_obs]
             for i, o in enumerate(opening_obs):
-                o.force_set('memories', random.sample(o['memories'], n_mems[i]))
+                mem_indices = random.sample(range(len(o['memories'])), n_mems[i])
+                o.force_set(
```
mojtaba-komeili, line 729: Does this purge the existing memories if the current item in the batch doesn't have memory access this round?

projects/bb3/agents/utils.py
```diff
@@ -361,25 +373,95 @@ def maybe_reduce_memories(
...
+            # check if we should ignore in session memories
+            if ignore_in_session_memories and memory in in_session_memories:
+                continue
+            # check overlap
+            non_stopword_memory = ' '.join(
+                normal_tokenizer(memory.split(':')[-1], include_pronouns=True)
+            )
+            non_stopword_text = ' '.join(normal_tokenizer(text, include_pronouns=True))
+            if (
+                F1Metric.compute(non_stopword_memory, [non_stopword_text]).value()
```
mojtaba-komeili, line 408: nit: maybe only checking if `memory_overlap_threshold > 0`? Just saving a bit on F1 compute.

mojtaba-komeili (approved):

projects/bb3/agents/opt_bb3_agent.py
klshuster: no, we're just playing around with the memories sent in the opening message. This is legacy logic for if the opening message is not meaningful (back when our models couldn't handle the opener)

projects/bb3/agents/utils.py
klshuster: yeah good idea

projects/bb3/agents/utils.py
```diff
@@ -400,15 +400,18 @@ def get_available_memories(
...
-            non_stopword_memory = ' '.join(
-                normal_tokenizer(memory.split(':')[-1], include_pronouns=True)
-            )
-            non_stopword_text = ' '.join(normal_tokenizer(text, include_pronouns=True))
-            if (
-                F1Metric.compute(non_stopword_memory, [non_stopword_text]).value()
-                < memory_overlap_threshold
-            ):
-                continue
+            if memory_overlap_threshold > 0:
```
mojtaba-komeili, line 403: 👍

projects/bb3/agents/opt_bb3_agent.py
klshuster: we're rebuilding the observation for the opening model; `'text'` is what the model is seeing

PR #4779, opened 2022-09-01
[cuda ext] Protect Cuda Extension Loading

klshuster:
**Patch description**
Wrap loading of our cuda extension for GPU ngram blocking in a try/catch.

**Testing steps**
YOLO


mojtaba-komeili (approved):
Thanks for fixing it.

stephenroller (approved):

PR #4789, opened 2022-09-15
[BB3] Handle non-ints in memories dict

klshuster:
**Patch description**
@JoshDLane pointed out that our memory soft blocking would not work if `turns_since_used` was not an integer; this fix ensures that it is.

**Testing steps**
Updated CI


mojtaba-komeili (approved):
LGTM

JoshDLane (approved):

PR #4798, opened 2022-09-23
[BB3] Fix Module Level Tasks

klshuster:
**Patch description**
The module level tasks are all multi-task teachers, so a lot were missing the opt keys from their individual component teachers. This PR fixes that.

This patch also ensure that the mutators are properly loaded for teachers.

**Testing steps**
Tested locally that each module level task displays properly


mojtaba-komeili (approved):
LGTM

PR #4819, opened 2022-10-04
Fix the api call and api response mismatch issue

jacksukk:
This PR is to 
1.  Fixed a bug: also consider cum_api_call when processing the goal_calls
2. In taskmaster2 data structure, sometime there's api response in system without user api call. And it will cause problems when people want to implement lookup table DB in the simulation scenario. So, using len(cum_api_call) > 0 and len(api_resp) > 0 as a constraint.


klshuster:
cc @moyapchen unless you think someone else should review?

PR #4825, opened 2022-10-12
Generalized validation metric to support product or sum function of multiple criteria.

leox1v:
You can provide a comma separated list as validation metric and it will use the product or sum of those as final validation metric.


parlai/scripts/train_model.py
```diff
@@ -221,6 +221,15 @@ def setup_args(parser=None) -> ParlaiParser:
         choices=['max', 'min'],
         help='the direction in which to optimize the validation metric, i.e. maximize or minimize',
     )
+    train.add_argument(
+        '-vmf',
+        '--validation-metric-fn',
```
klshuster, line 226: could we change the long version to be `--validation-metric-aggregation-function`? as that's more descriptive (I was confused at first)

PR #4836, opened 2022-10-21
[Names Bias] Code for creating sets of names

EricMichaelSmith:
Release the code used to process and produce the two sets of names used in https://arxiv.org/pdf/2109.03300.pdf , given the lists of names in the original papers that we used to create these sets.


projects/dialogue_bias/util.py
```diff
@@ -0,0 +1,216 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import os
+from collections import defaultdict
+from typing import Dict, List, Tuple
+
+import pandas as pd
+
+
+RACES_ETHNICITIES = ['hispanic', 'white', 'black', 'api', 'aian', '2prace']
```
davides, line 14: Are the spellings of the last two items here intentional?

projects/dialogue_bias/util.py
```diff
@@ -0,0 +1,216 @@
...
+    # Split original name lists by gender
+    orig_name_lists = get_tzioumis_name_lists(
+        tzioumis_data_path=tzioumis_data_path,
+        race_gender_name_lists=race_gender_name_lists,
+    )
+    print('\nSplitting original name lists by gender.')
+    names_to_new_lists = {}
+    for name_list, names in orig_name_lists.items():
+        for name in names:
+            proc_name = name.replace('-', '')
```
davides, line 56: It seems like this will concatenate hyphenated names. Why is that required?

projects/dialogue_bias/util.py
```diff
@@ -0,0 +1,216 @@
...
+    max_percent_series = percent_df[percent_columns].max(axis=1)
+    percent_plurality_names_df = (
+        percent_df[lambda df: df[this_ethnicity_column] == max_percent_series]
+        .assign(
+            obs_of_this_ethnicity=lambda df: df['obs'] * df[this_ethnicity_column] / 100
+        )
+        .sort_values('obs_of_this_ethnicity', ascending=False)
+    )
+    tzioumis_plurality_names = percent_plurality_names_df.iloc[
+        :200
```
davides, line 184: Not sure I follow, why 200? Can you please add a comment to clarify?

projects/dialogue_bias/util.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # Combine these names with the Caliskan+ race+gender names and deduplicate
+    if race_ethnicity in tzioumis_to_race_gender_mapping:
+        mapped_ethnicity = tzioumis_to_race_gender_mapping[race_ethnicity]
+        female_race_gender_name_list = race_gender_name_lists[
+            f'{mapped_ethnicity}_female'
+        ]
+        if mapped_ethnicity == 'aa':
+            # Avoid the same name in two lists by removing it from this one
+            female_race_gender_name_list.remove('Yolanda')
```
davides, line 195: Is there a programmatic way to detect duplicates? That way it would just work for updated versions of the source datasets (assuming the schema stays the same).

davides (commented):
LGTM overall, just a few minor things to clarify.

projects/dialogue_bias/util.py
EricMichaelSmith: Yeah, that's how Tzioumis spelled them - adding a comment for this

projects/dialogue_bias/util.py
EricMichaelSmith: Semi-arbitrary, to prevent the lists of names from being too big - adding a comment

projects/dialogue_bias/util.py
EricMichaelSmith: good call - haha this was obviously a hack. Adding a TODO here

davides (approved):

PR #4843, opened 2022-10-28
fixing version to make tests pass

Golovneva:
**Patch description**
Currently tests are broken because `protobuf` version is incompatible

See passing tests below


moyapchen (approved):
Thanks!

PR #4853, opened 2022-10-31
replace fuzzywuzzy with rapidfuzz

maxbachmann:
**Patch description**
Fuzzywuzzy is GPL licensed, so this PR replaces it with rapidfuzz which is MIT licensed. As a side effect this is much faster, but I do not know whether this is of any relevance in this code section.

**Testing steps**
The behavior should not change.

**Other information**


klshuster:
Thanks for the PR, could you also please update the [cache keys for our CI](https://github.com/facebookresearch/ParlAI/blob/8b18a15b71d148b7d2a108196115bcc4f025d655/.circleci/config.yml#L234)? 

CC @moyapchen as this is TOD-related (lmk if someone else is better suited for review)

maxbachmann:
@klshuster @moyapchen Is there anything still needed to do from my side?

maxbachmann:
@klshuster @moyapchen is there still interest in this?

klshuster:
yes! could you please update the cache keys one more time?

moyapchen:
Whoops - well, I apparently completely and utterly failed to notice this PR's existence. 

This code is was being used for some of the applied research teams, but I'm pretty sure they've moved far beyond using this teacher a while back, so happy to accept.

moyapchen (approved):
CC: @skiingpacman just in case

PR #4862, opened 2022-11-02
upgrade protobuf

Golovneva:
**Patch description**
Fixing dependencies

Protobuf is limited by TensorboardX. Looks like they are working on bumping the version anytime soon, can be tracked here: https://github.com/lanpa/tensorboardX/blob/master/test-requirements.txt
and PR here: https://github.com/lanpa/tensorboardX/pull/675


klshuster (approved):
awesome, thanks!

PR #4868, opened 2022-11-08
Parlai container

mojtaba-komeili:
# Patch description
The Dockerfile for building a working image for ParlAI.

## Note
A tutorial on using this Docker image will follow.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+# Copyright (c) Meta, Inc. and its affiliates.
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+FROM nvidia/cuda:11.1.1-cudnn8-devel-ubuntu18.04 AS install_parlai_prereqs
+
+# Installing the required packages
+RUN apt update -y
+RUN apt install -y git curl
+
+# Installing Anaconda
+WORKDIR /root
+RUN curl https://repo.anaconda.com/archive/Anaconda3-2021.05-Linux-x86_64.sh -o anaconda_installer.sh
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN apt install -y git curl
+
+# Installing Anaconda
+WORKDIR /root
+RUN curl https://repo.anaconda.com/archive/Anaconda3-2021.05-Linux-x86_64.sh -o anaconda_installer.sh
+RUN bash anaconda_installer.sh -b -p
+ENV PATH="/root/anaconda3/bin:$PATH"
+
+# Installing recommmended pre-requirements
+RUN conda install pytorch torchvision torchaudio -c pytorch-lts -c nvidia
```
klshuster, line 17: should we specify torch version? or just assume we work with highest torch number?

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# Download the ParlAI Github repo
+RUN git clone https://github.com/facebookresearch/ParlAI.git ~/ParlAI
+
+# Running ParlAI install
+RUN cd ~/ParlAI && \
+    pip install -r requirements.txt && \
+    python setup.py develop
+
+CMD ["parlai", "party"]
```
klshuster, line 33: 😍

klshuster (approved):
awesome, just small comment about torch versioning

Dockerfile
mojtaba-komeili: Yeah, good point. Should we set it to the same version as the install requirements?

PR #4873, opened 2022-11-10
SKR project page  typo

mojtaba-komeili:
**Patch description**
Ran into a problem trying to fine-tune the R2C2 SKR model. Turns out the problem was because I copied a typo from the readme file of project.


klshuster:
good catch

klshuster (approved):

PR #4875, opened 2022-11-11
Allow flan-t5 models in ParlAI with fp16 improvment

dexterju27:
**Patch description**
The current setting does allow flan-t5 model to run under parlai hugging face agent seamlessly. Adding Flan-t5 options in the agent. 

**Testing steps**
```
parlai train_model -t convai2  -dt train --batchsize 1  --fp16 True  --gradient-clip 1.0 --label-truncate 256 --text-truncate 512 --log-every-n-secs 30 --lr-scheduler reduceonplateau --max-train-time 169344.0 --model hugging_face/t5 --save-after-valid True --skip-generation True --optimizer adam -lr 1e-05 -veps 0.15 -vmm min -vmt ppl -vp 15 -tblog True --t5-model-arch google/flan-t5-large  --t5-model-parallel True --dict-tokenizer gpt2
```

### DIFF
--- a/parlai/agents/hugging_face/t5.py
+++ b/parlai/agents/hugging_face/t5.py
@@ -86,7 +86,18 @@ class T5Agent(TorchGeneratorAgent):
             '--t5-model-arch',
             type=str,
             default='t5-base',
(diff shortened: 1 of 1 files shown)


klshuster (approved):
thanks!

PR #4883, opened 2022-11-16
fixing reasoning perturbation issues

Golovneva:
**Patch description**
Fixing some issues with perturbations:
1. Math selecting empty steps to insert as hallucination. Ideally, we would want to remove this empty steps at tokenization, but that would change all our MATH data and affect all scores. So instead this patch.
2. In proofwriter, use only those hallucinated steps that are not part of the context
3. In Grammar, compare with tokenized step, to avoid situation where chains are identical everywhere except some random spaces
4. Fix the name of the finetuned BART model to align with what was uploaded to S3

### DIFF
--- a/parlai/tasks/math_dataset/agents.py
+++ b/parlai/tasks/math_dataset/agents.py
@@ -109,7 +109,10 @@ class MathDatasetStepByStepReasoningTeacher(MWPStepsReasoningTeacher):
                 rand_steps = self._clean_steps(
                     self.math_random.choice(data)["solution"]
                 ).split(". ")
-                random_step = self.math_random.choice(rand_steps)
+                random_step = ""
+                # find non-empty random step
+                while random_step == "" or random_step == " ":
+                    random_step = self.math_random.choice(rand_steps)
             if convert:
                 question = self._latex_conversion(question)
                 final_answer = self._latex_conversion(final_answer)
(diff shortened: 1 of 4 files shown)


parlai/tasks/math_dataset/agents.py
```diff
@@ -109,7 +109,10 @@ def get_data_for_fold(self, fold):
                 rand_steps = self._clean_steps(
                     self.math_random.choice(data)["solution"]
                 ).split(". ")
-                random_step = self.math_random.choice(rand_steps)
+                random_step = ""
+                # find non-empty random step
+                while random_step == "" or random_step == " ":
```
moyapchen, line 114: Any chance this could infinite loop (cause all steps are "", in case someone tries to mix perturbations)?

parlai/tasks/proof_writer/agents.py
```diff
@@ -328,8 +328,11 @@ def get_data_for_fold(self, fold):
 
         for m in messages:
             if extrinsic_step:
-                rand_steps = self.proofwriter_random.choice(messages)["steps"]
-                random_step = self.proofwriter_random.choice(rand_steps)
+                random_step = None
+                # make sure new step is from a different context
+                while not random_step or random_step in m["question"]:
+                    rand_steps = self.proofwriter_random.choice(messages)["steps"]
+                    random_step = self.proofwriter_random.choice(rand_steps)
```
moyapchen, line 335: ditto to question above about infinite loops

moyapchen (approved):
LGTM

PR #4886, opened 2022-11-18
[Core] Delay Loading NGram-Blocking on GPU 

klshuster:
**Patch description**
Loading the ngram blocking on GPU not only fails for several builds, but can also be quite time consuming; I'm shifting to a lazy instantiation, such that loading does not occur unless the user explicitly asks for ngram blocking on GPU.

**Testing steps**
Local testing


klshuster:
cc @dexterju27

dexterju27 (approved):
LGTM! Thanks!

PR #4890, opened 2022-11-18
[TGA] Factual Nucleus Sampling

klshuster:
**Patch description**
[Factual nucleus](https://arxiv.org/abs/2206.04624) is a decoding method that decays the nucleus sampling `p` value over time according to a constant factor `lambda_decay`. The `p` value resets when encountering a full-stop (when `p_reset` is `True`), and there is a lower bound of `omega_bound` to what `p` can decay to. 

**Testing steps**
Added Factual Nucleus to the existing tree search tests.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -1636,7 +1636,7 @@ class TreeSearch(object):
             hyp_device = self.partial_hyps.get_device()
         self.partial_hyps = torch.cat(
             (
-                self.partial_hyps[path_selection.hypothesis_ids.long().to(hyp_device)],
+                self.partial_hyps[path_selection.hypothesis_ids.long()],
                 path_selection.token_ids.view(path_selection.token_ids.shape[0], -1).to(
                     hyp_device
                 ),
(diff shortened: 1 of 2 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -1777,7 +1817,7 @@ def __init__(self, *args, **kwargs):
             raise ValueError('Greedy search can only be run with beam size 1.')
 
     def select_paths(self, logprobs, prior_scores, current_length) -> _PathSelection:
-        tok_scores, tok_ids = logprobs.max(1)
+        tok_scores, tok_ids = logprobs
```
mojtaba-komeili, line 1820: I don't know this code well. I don't understand why did you have to change this part? isn't this for greedy? And I don't find any indication that the general code was changed to justify changing this part.

mojtaba-komeili (approved):
Seems good to me. Thanks for adding.

parlai/core/torch_generator_agent.py
klshuster: good catch, this is an artifact, will fix

parlai/core/torch_generator_agent.py
```diff
@@ -1047,6 +1066,24 @@ def _treesearch_factory(self, device, verbose=False):
...
+                block_ngram=self.beam_block_ngram,
+                context_block_ngram=self.beam_context_block_ngram,
+                length_penalty=self.opt.get('beam_length_penalty', 0.65),
+                padding_token=self.NULL_IDX,
+                bos_token=self.START_IDX,
+                eos_token=self.END_IDX,
+                device=device,
+                verbose=verbose,
+                dict=self.dict,
+            )
```
jxmsML, line 1086: is `gpu_beam_blocking` needed as in other sampling methods?

klshuster: yes, will add thanks

Issue #3411, opened 2021-01-26
New Metric.from_mask helper method


stephenroller:
We have quite a few instances where we have some per-token losses/metrics along with a corresponding mask

```python
metric_per_token # torch.Tensor of shape (batchsize, num_tokens)
mask # torch.BoolTensor of shape (batchsize, num_tokens)
```
And we want a per-batch-example average:
```python
tokens_per_ex = mask.long().sum(dim=-1)
metric_per_ex = (metric_per_token * mask).sum(dim=-1)
metrics: List[MyMetric] = MyMetric.many(metric_per_ex, tokens_per_ex)
self.record_local_metric('metric_name', metrics)
```

I'd like us to have a helper classmethod in `Metric` called `from_mask`:
```python
class Metric:
    @classmethod
    def from_mask(cls, metric_per_token, token_mask):
        # returns the equivalent of the "metrics" object above
```

Once this is done, add unit tests for this (test AverageMetric and PPLMetric directly). Checkpoint there.

After you've implemented this, upgrade TorchGeneratorAgent to use your new helper, upgrading the code for loss, ppl, and token_acc.

Example:
https://github.com/facebookresearch/ParlAI/blob/67433e376fc361dee5aa045cb6bb2b68d3faa478/parlai/core/torch_generator_agent.py#L711-L727

See if you can find at least one other place who can benefit from upgrading this pattern.

poojasethi:
Hi @klshuster and @stephenroller! I've just submitted a PR for this issue: https://github.com/facebookresearch/ParlAI/pull/4894

PR #4894, opened 2022-11-23
Add Metric.from_mask helper method (#3411)

poojasethi:
**Patch description**
This change introduces a new `from_mask` helper function in the `Metric` class. It also refactors the `compute_loss` function in `torch_generator_agent.py` to call the `from_mask` helper when computing the *loss*, *ppl*, and *token_acc*.

**Testing steps**
`pytest -v tests/test_metrics.py` 
Note that I added two new unit tests,  `test_average_metric_from_mask` and `test_ppl_metric_from_mask`.

**Other information**
* I believe some of these [files](https://github.com/facebookresearch/ParlAI/search?q=.long%28%29.sum%28dim%3D-1%29&type=code) could potentially be refactored to use the new helper function as well. Happy to submit another PR, if helpful!
* I tried running `autoformat.sh` but it was running quite slowly... Tips on how to use these script would be helpful :)

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -24,6 +24,7 @@ from typing import (
     Optional,
     Set,
     Tuple,
+    Type,
     Union,
 )
 
@@ -272,7 +273,7 @@ class Metric(ABC):
         """
         Construct many of a Metric from the base parts.
 
-        Useful if you separately compute numerators and denomenators, etc.
+        Useful if you separately compute numerators and denominators, etc.
         """
         lengths = [len(o) for o in objs]
         objs = list(objs)  # convert from tuple for inplace modification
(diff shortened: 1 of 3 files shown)


parlai/core/metrics.py
```diff
@@ -286,6 +287,27 @@ def many(cls, *objs: List[TVector]) -> List[Metric]:
...
+        :param mask:
+            a (batchsize x num_tokens) Tensor to mask out tokens that should *not* be considered in the aggregate metric calculation.
+        :param MyMetric:
+            a subclass of Metric
+        :return:
+            a (batchsize) Tensor
+        """
+        tokens_per_ex = mask.long().sum(dim=-1)
+        metric_per_ex = (metric_per_token * mask).sum(dim=-1)
+        metrics = MyMetric.many(metric_per_ex, tokens_per_ex)
```
stephenroller, line 308: Can we use `cls.many` instead and get away without passing MyMetric as an extra parameter?

stephenroller (commented):
This looks like it fixes that old issue! Actually @mojtaba-komeili looks like he may be able to use this.

Leaving one comment to see if we can go even one step cleaner

parlai/core/metrics.py
poojasethi: Ooh, yes! Good idea.

stephenroller (approved):
lgtm!

poojasethi:
Sweet! @stephenroller and @klshuster should I wait for / help dig into the lint and CircleCI failures, or go ahead and land the PR?

poojasethi:
Landed! (Discussed with @klshuster offline and it seems that the lint and CircleCI failures are not related to this PR)

PR #4897, opened 2022-11-29
[T5] Fix T5 Init Bug

klshuster:
**Patch description**
Our CI has been failing since earlier versions of HF transformers do not have the `torch_dtype` option for the T5 model. This patch wraps the init in a try/catch to handle backwards compatibility.

**Testing steps**
CI


mojtaba-komeili (approved):
Thanks for patching.

PR #4899, opened 2022-11-30
Dropping the none probabilities

mojtaba-komeili:
**Patch description**
Running some evaluations that were using the OPT model as their backend server, I noticed that sometime they were crashing while calculating the PPL. Digging deeper it turned out there is a `None` value in the `token_logprobs` returned from the server. Adding this filter resolved the crashing issue.

Screen Shot 2022-11-30 at 12 47 51 PM

**PS**: I am not sure if this has any unwanted consequences on the validity of the metric.

### DIFF
--- a/projects/bb3/agents/utils.py
+++ b/projects/bb3/agents/utils.py
@@ -650,7 +650,7 @@ class APIUtils:
             start_label = [i for i, off in enumerate(text_off) if off <= prompt_len]
             assert len(start_label) > 0
             start_label = start_label[-1]
-            all_log_probs = result['choices'][0]['logprobs']['token_logprobs']
+            all_log_probs = [lp for lp in result['choices'][0]['logprobs']['token_logprobs'] if lp]
             if not all(l <= 0 for l in all_log_probs):
                 logging.warning(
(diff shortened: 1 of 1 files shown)


klshuster:
uhh cc @stephenroller why would we be getting `None` back?

stephenroller:
the first token doesn't have a probability associated with it. we did this to be compatible with openai

projects/bb3/agents/utils.py
```diff
@@ -650,7 +650,9 @@ def compute_perplexities(
             start_label = [i for i, off in enumerate(text_off) if off <= prompt_len]
             assert len(start_label) > 0
             start_label = start_label[-1]
-            all_log_probs = result['choices'][0]['logprobs']['token_logprobs']
+            all_log_probs = [
+                lp for lp in result['choices'][0]['logprobs']['token_logprobs'] if lp
```
klshuster, line 654: could we change to `if lp is not None`, as `lp==0` could happen

mojtaba-komeili: It shouldn't be an issue in practice given it comes from the softmax. But who know what we will be using in the future, so added it lol.

klshuster (approved):
thanks for the patch!

PR #4901, opened 2022-11-30
torch concat backward compatibility

mojtaba-komeili:
**Patch description**
Backward compatibility for `torch.concat` function in cringe loss.

### DIFF
--- a/projects/cringe/cringe_loss.py
+++ b/projects/cringe/cringe_loss.py
@@ -88,9 +88,15 @@ class ContrastiveCrossEntropyLoss(CrossEntropyLoss):
 
         # concatenate the logits of the preds with the actual label's logits
         x_target = x[torch.arange(x.shape[0]), y]
-        x_ct = torch.concat(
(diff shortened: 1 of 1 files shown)


projects/cringe/cringe_loss.py
```diff
@@ -88,9 +88,15 @@ def __call__(self, x, y, classifier_labels=None, **kwargs):
...
-        )
+        if hasattr(torch, 'concat'):
+            # torch > 1.10
+            x_ct = torch.concat(
+                [x_target.unsqueeze(1), sample_preds_values.unsqueeze(1)], -1
+            )
+        else:
+            x_ct = torch.cat(
+                [x_target.unsqueeze(1), sample_preds_values.unsqueeze(1)], -1
+            )
```
klshuster, line 99: we could just do `torch.cat` to avoid the check? pretty sure it's just an alias

stephenroller (approved):

klshuster (approved):

PR #4902, opened 2022-12-01
None metrics

mojtaba-komeili:
**Patch description**
Running the `projects.cringe.cringe_loss:ContrastiveTransformerGeneratorAgent`, noticed some issues about calculating metrics for some `None` values; this was crashing the model during the training.
This was due to dropping some of the values in the Generator Agent ([example](https://github.com/facebookresearch/ParlAI/blob/4c96c8867061fba34fdd03a1f87481f8c3623439/projects/cringe/cringe_loss.py#L285-L293)).
This patch fixes that by ignoring the `None` values in an observation.

### DIFF
--- a/parlai/core/metrics.py
+++ b/parlai/core/metrics.py
@@ -292,7 +292,8 @@ class Metric(ABC):
         cls, metric_per_token: torch.Tensor, mask: torch.Tensor
     ) -> List[Metric]:
         """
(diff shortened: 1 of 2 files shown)


klshuster:
i think this is ok? Maybe CC @stephenroller --> is there a canonical way to handle averages of metrics where in some cases one of the `Metric`s is `None`?

klshuster (approved):
i think this is ok

Issue #3753, opened 2021-06-28
Add support for Zero3 FSDP


stephenroller:
In #3740, we added support for FullyShardedDataParallel, but limited implementation to that of Zero2, not Zero3. Zero3 results in substantial decreases of memory usage compared with Zero2 while bringing speed back in line with vanilla DDP.

We have already added support for this (via manual calls to `wrap`) within the Transformer modules, but we still cannot support Zero3. The main issue is that Zero3 assumes that every worker calls `forward` the exact same number of times, and performs a parameter-transfer during this forward (moving the sharded parameters to each worker just in time). ParlAI cannot provide this guarantee though because:
- During validation, each worker sees a variable number of examples. This is okay in itself, but it is problematic (hang) if it results in any worker having extra batches.
- During generation, workers will have a variable number of forwards due to the variable sequence length. While everything stays happy for a while, if one worker ends the run with needing more generations than the others, we will get hangs.

It seems far too difficult (and ugly) to try to force this equality in worlds.py or in our dataset sharding. So our best future bet is to implement something like `.join()` in vanilla DDP. It would work roughly as follows:
- Every worker in forward tries to synchronize a `True` boolean saying "Am I doing a true forward?"
- Upon `__exit__` of the context, workers enter an infinite loop where they sync a `False` boolean. As long as any worker is providing a `True` value, they participate in a dummy batch forward.
- When all workers agree on the `False` boolean, we can end the infinite loop.

This feature makes the most sense to implement upstream in Fairscale, and then integrate into ParlAI.

blefaudeux:
>> During validation, each worker sees a variable number of examples. This is okay in itself, but it is problematic (hang) if it results in any worker having extra batches.

Pytorch distributed has a wrapper for that, I've tried to look it up to no avail (maybe not public yet). Not sure how applicable that would be, just a heads up

seed93:
So any updates here? We are really looking forward to using ZERO3 for boosting.

blefaudeux:
> > > During validation, each worker sees a variable number of examples. This is okay in itself, but it is problematic (hang) if it results in any worker having extra batches.
> 
> Pytorch distributed has a wrapper for that, I've tried to look it up to no avail (maybe not public yet). Not sure how applicable that would be, just a heads up

https://pytorch.org/tutorials/advanced/generic_join.html 
@stephenroller would that help ? cc @min-xu-ai

PR #4903, opened 2022-12-01
[FSDP] Zero 3 Optimization Support

klshuster:
# Patch description
This PR adds support for zero3 fsdp optimization. Specify via `--ddp-backend zero3`. Zero3 optimization shards not only the optimizer and gradients but also the model weights. This improves memory pressure, and is especially useful for larger models. Note that this can increase latency as there is more communication cost, and as such for smaller models there may be a slight hit to speed compared to zero2 (and model parallel, of course).

Addresses #3753

**NOTE: This requires pytorch >= 1.12 for use**

# Testing steps
- Enabled the zero3 tests already in CI (thanks Stephen!)
- Confirmed that zero3 works when the number of validation examples is not divisible by the number of GPUs
- Confirmed that zero3 works when `--skip-generation False` during training. 
- Conducted a comprehensive analysis of a variety of "staple" models within ParlAI under different distributed settings; see screenshots below

The main conclusions from each of the models below are:
- **Performance** remains roughly the same regardless of model parallel, zero2, or zero3
- **Speed**: for BART, zero2 remains faster. For the other models, we see varying results; ultimately, everything is faster than model parallel
- **GPU Memory**: zero3 

## BART-Large (400M)
Screen Shot 2022-12-01 at 2 53 15 PM

## T5-Large (770M)
Screen Shot 2022-12-01 at 2 51 32 PM

## Reddit 2.7B (base of BlenderBot) 
Screen Shot 2022-12-01 at 2 51 51 PM

## GPT2-XL (1.5B)
Screen Shot 2022-12-01 at 2 52 16 PM

## R2C2 2.7B
Screen Shot 2022-12-01 at 2 52 31 PM

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -526,20 +526,8 @@ class TorchGeneratorAgent(TorchAgent, ABC):
                     states = {}
                 return states
 
-            init_model_loaded = False
-            self.model = self.build_model()
-            if (
-                fsdp_utils.should_use_fsdp(opt)
-                and opt.get('ddp_backend') == 'zero3'
-                and opt.get('accelerate_load', False)
-            ):
-                # load the model params, FIRST
-                states = load_init_model()
-                init_model_loaded = True
-                self.model.cpu()
-
-            with fsdp_utils.maybe_fsdp_wrap(opt, self.model):
-                self.model = fsdp_utils.fsdp_wrap(self.model)
+            with fsdp_utils.maybe_fsdp_wrap(opt):
+                self.model = fsdp_utils.fsdp_wrap(self.build_model())
                 if self.fp16 and not fsdp_utils.delay_halving(opt):
                     self.model = self.model.half()
 
@@ -568,11 +556,11 @@ class TorchGeneratorAgent(TorchAgent, ABC):
                 f"Total parameters: {total_params:,d} ({train_params:,d} trainable)"
             )
 
-            if init_model is not None and not init_model_loaded:
+            if init_model is not None:
                 # load model parameters if available
                 logging.info(f'Loading existing model params from {init_model}')
                 states = self.load(init_model)
-            elif not init_model_loaded:
+            else:
                 states = {}
 
         if shared is not None:
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -20,7 +20,7 @@ Contains the following utilities:
 from typing_extensions import TypedDict
 from parlai.core.params import ParlaiParser
 from abc import ABC, abstractmethod
-from typing import TypeVar, List, Dict, Optional, Tuple, Set, Iterable, Any
+from typing import TypeVar, List, Dict, Optional, Tuple, Set, Iterable
 import math
 from operator import attrgetter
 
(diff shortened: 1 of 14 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -516,6 +516,16 @@ def __init__(self, opt: Opt, shared=None):
         else:
             # this is not a shared instance of this class, so do full init
             self.criterion = self.build_criterion()
+
+            def load_init_model() -> Dict[str, Any]:
```
stephenroller, line 520: how is this used?

parlai/scripts/distributed_train.py
```diff
@@ -51,8 +52,9 @@ def setup_args(cls):
 
     def run(self):
         with distributed_utils.slurm_distributed_context(self.opt) as opt:
-            self.train_loop = single_train.TrainLoop(opt)
-            return self.train_loop.train()
+            self.train_loop = fsdp_utils.JoinableTrainLoop(opt)
+            with fsdp_utils.fsdp_join(self.train_loop):
```
stephenroller, line 56: do you need to do this for distributed_eval too?

parlai/utils/fsdp.py
```diff
@@ -7,22 +7,46 @@
...
+    )
 
+    PYTORCH_FSDP_AVAILABLE = True
     FSDP_AVAILABLE = True
 except ImportError:
-    FSDP_AVAILABLE = False
+    PYTORCH_FSDP_AVAILABLE = False
+    try:
+        from fairscale.nn.wrap.auto_wrap import wrap, enable_wrap
+        from fairscale.nn.data_parallel import FullyShardedDataParallel as FSDP
```
stephenroller, line 42: why keep around fairscale support?

stephenroller (commented):
nice, pretty smooth implementation. i might recommend just axing support for fairscale but you're the boss

stephenroller (approved):

parlai/core/torch_generator_agent.py
klshuster: ahh it's not, that's an artifact. will delete

parlai/scripts/distributed_train.py
klshuster: yes, forgot about that. also multiprocessing eval

parlai/utils/fsdp.py
klshuster: so getting rid of fairscale would force anyone who wants to use distributed training to be on pytorch >=1.12. Is that a reasonable ask, you think?

stephenroller: That's your call but pytorch 1.13 is out (and I'm using it successfully)

klshuster: i've removed fairscale FSDP

klshuster:
Note: There is currently a bug in pytorch 1.13 that doesn't allow specifying your own pickle module for loading; this is breaking some tests (`test_apex.py`); the fix is merged in https://github.com/pytorch/pytorch/pull/88570, but we'll need to wait for pytorch to patch to allow us to pass tests with pytorch 1.13

for now, will have requirements specify latest 1.12 version

PR #4905, opened 2022-12-02
Custom metrics from `compute_loss` in TGA

mojtaba-komeili:
**Patch description**
The `compute_loss` function in TGA generates the `loss_per_token` values but they are not accessible from outside this method. This PR adds a handle for calculating custom metrics that require `loss_per_token`. The added method (`custom_loss_metrics`) can be overridden by its children to work with the `loss_per_token` and `batch` for generating custom metrics.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -34,7 +34,7 @@ from parlai.core.torch_agent import TorchAgent, Batch, Output, DictionaryAgent
 from parlai.utils.misc import warn_once
 from parlai.utils.io import PathManager
 import parlai.utils.logging as logging
-from parlai.core.metrics import Metric, SumMetric, AverageMetric, FairseqBleuMetric
+from parlai.core.metrics import SumMetric, AverageMetric, FairseqBleuMetric
 from parlai.utils.fp16 import FP16SafeCrossEntropy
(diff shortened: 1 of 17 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -34,7 +34,7 @@
 from parlai.utils.misc import warn_once
 from parlai.utils.io import PathManager
 import parlai.utils.logging as logging
-from parlai.core.metrics import Metric, SumMetric, AverageMetric, FairseqBleuMetric
+from parlai.core.metrics import SumMetric, AverageMetric, FairseqBleuMetric
```
mojtaba-komeili, line 37: This is just for the lint error.

parlai/core/torch_generator_agent.py
```diff
@@ -715,6 +715,12 @@ def _encoder_input(self, batch):
         """
         return self._model_input(batch)
 
+    def custom_loss_metrics(self, batch, loss_per_token):
```
klshuster, line 718: super nit: can we call this something more related to what it's doing?

e.g., `compute_per_token_metrics`?

mojtaba-komeili:
Rebase added some unwanted changes to this PR. Closing it and opening a new [clean one](https://github.com/facebookresearch/ParlAI/pull/4913).

PR #4913, opened 2022-12-15
Custom metrics from `compute_loss` in TGA (2nd try)

mojtaba-komeili:
**Patch description**
The `compute_loss` function in TGA generates the `loss_per_token` values but they are not accessible from outside this method. This PR adds a handle for calculating custom metrics that require `loss_per_token`. The added method (`custom_loss_metrics`) can be overridden by its children to work with the `loss_per_token` and `batch` for generating custom metrics. 

**NOTE**: this is cleaned up version of [4905](https://github.com/facebookresearch/ParlAI/pull/4905) which became corrupted with other commits after some bad rebase and merge.

### DIFF
--- a/parlai/core/torch_generator_agent.py
+++ b/parlai/core/torch_generator_agent.py
@@ -717,7 +717,7 @@ class TorchGeneratorAgent(TorchAgent, ABC):
         """
         return self._model_input(batch)
 
-    def custom_loss_metrics(self, batch, loss_per_token):
+    def record_per_token_metrics(self, batch, loss_per_token):
         """
         Override this method for custom loss values that require loss_per_token.
         """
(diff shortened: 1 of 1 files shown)


parlai/core/torch_generator_agent.py
```diff
@@ -752,6 +758,7 @@ def compute_loss(self, batch, return_output=False):
         self.record_local_metric(
             'token_em', AverageMetric.many(num_tokens_correct == num_target_tokens)
         )
+        self.custom_loss_metrics(batch, loss_per_token)
```
klshuster, line 761: nit: could we rename to `record_per_token_metrics` or something similar? I want to make sure people understand that this is function is specifically provided the `loss_per_token`

mojtaba-komeili:
@klshuster reminding you of this old PR.

klshuster (approved):
thanks for the changes!

PR #4934, opened 2023-01-19
Reverting emoji response patch in model chat

mojtaba-komeili:
**Patch description**

Revert "add emoji input functionality to model chat crowdsourcing task (#4666)"

This reverts commit 2f42a5b90bdbbd4fc67970d4c051d4955787445c.

Ran into some issue running the the human-model chat. After consulting with @pringshia it was related to the changes in this PR. @meganung confirms that there is not going to be major side effects reverting this.


meganung (approved):
lgtm ! -- we can reimplement / refer back if this functionality is needed later down the line.

EricMichaelSmith (approved):
Agreed with Megan, seems fine to remove for now!

PR #4962, opened 2023-02-19
fix datatype

heya5:
**Patch description**
Fix a bug of wizard_of_internet task. This bug led to a wrong number of examples when use "train:evalmode" as datatype.
The problem is like this:
image

**Testing steps**
To test this PR, you can run `parlai convert_to_json -t wizard_of_internet --datatype train:evalmode`
Expect to get the correct number of samples.
image

### DIFF
--- a/parlai/tasks/wizard_of_internet/agents.py
+++ b/parlai/tasks/wizard_of_internet/agents.py
@@ -150,7 +150,7 @@ class WizardOfInternetBaseTeacher(DialogTeacher):
 
     def __init__(self, opt: Opt, shared=None):
         opt = deepcopy(opt)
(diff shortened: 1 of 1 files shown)


mojtaba-komeili:
Thanks for reporting the bug. I don't see how this fixes it though. Unless I missed something, `get_dtype` function and the `DatatypeHelper` are doing the same thing?

mojtaba-komeili (approved):

heya5:
Hi, I see a Linting problem, but it doesn't seem to be related to this PR.

mojtaba-komeili:
Don't worry we had some issues relate to the tests recently.

PR #4978, opened 2023-03-10
Now multiple checkpoints will be saved after using -stopk option in train_model.py

### LINKED ISSUE #4970: Keep multiple checkpoints during training
Hi, 
Is there an option or a way to keep for example 5 best checkpoints in the training process by using [train_model.py](../blob/main/parlai/scripts/train_model.py) and not just the best model checkpoint?
As far as I understood, there isn't any option to keep multiple checkpoints in  [train_model.py](../blob/main/parlai/scripts/train_model.py), and by any option group, the new checkpoint will be overwritten on the only last checkpoint that has been saved.
Should I add this feature and if it's required create a pull request or is there a logic behind the way [train_model.py](../blob/main/parlai/scripts/train_model.py) keeps the checkpoints?

hamjam:
**Patch description**

In this PR, the [train_model.py](../blob/main/parlai/scripts/train_model.py) script has the option to keep and store multiple checkpoints based on validation metrics. Some checkpoints that have best validation metrics during training will be kept in the directory specified by `--model-file`. Resolves #4970 

**Testing steps**

To test this PR, you can run something like `parlai train_model --task babi:task10k:1 --model seq2seq --model-file seq2seq/babi_task10k --batchsize 32 --validation-every-n-secs 30 -stopk 5 -vstep 50`
Note that `-vstep` option must be exist to save and store 5 top checkpoints in this example. For me, the list of files in the `-model-file` directory after 1350 steps of training is like this:
2023-03-10_16h39_48


parlai/scripts/train_model.py
```diff
@@ -606,13 +620,45 @@ def validate(self):
...
+            self._modify_next_rank_checkpoints(model_rank)
+            if (
+                opt['validation_metric_mode'] == 'max'
+                and self.best_k_models[-1][1] >= opt['validation_cutoff']
+            ) or (
+                opt['validation_metric_mode'] == 'min'
+                and self.best_k_models[-1][1] <= opt['validation_cutoff']
+            ):
+                logging.info('task solved! stopping.')
+                return True
```
klshuster, line 647: i think this check requires that we look at `self.best_k_models[0]`, right? since we're looking at the best metric

klshuster (commented):
thanks for adding this! Could you please include a test?

parlai/scripts/train_model.py
hamjam: My idea here was that we want the last saved model's metric to be better than `validation_cutoff` so I checked the last one with `self.best_k_models[-1]`

PR #4993, opened 2023-03-24
Multi model chat

mojtaba-komeili:
**Patch description**
An extension of the model chat which runs in non-turn-based multi-party mode. In the current form it runs a chat between one human and two other agents controlled by the model.


parlai/crowdsourcing/tasks/multi_model_chat/task_config/onboard_task_data.json
```diff
@@ -0,0 +1,67 @@
+{
```
klshuster, line 1: do we generally open-source our onboarding ?

klshuster (approved):
the python code looks good to me! deferring to other reviewers for .jsx

parlai/crowdsourcing/tasks/multi_model_chat/task_config/onboard_task_data.json
mojtaba-komeili: We had it at least for the Wizard of Internet as far as I remember.

PR #4999, opened 2023-03-31
More tests

mojtaba-komeili:
**Patch description**
* Another try for turning them green.
* Upgraded the deprecated CircleCI image that was running our tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,11 @@ commands:
             bash
             conda deactivate
             conda activate conda_parlai
-            conda install pytorch==1.13.0 torchvision==0.14.0 torchaudio==0.13.0 pytorch-cuda=11.6 -c pytorch -c nvidia
+            for package in pytorch==1.13.0 torchvision==0.14.0 torchaudio==0.13.0 pytorch-cuda=11.6 -c pytorch -c nvidia
+            do
+              echo "Conda installing ${package}."
+              conda install --quiet --yes "${package}"
+            done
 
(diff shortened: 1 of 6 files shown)


klshuster:
which tests should this be turning green? i.e., cleaninstall and unittests_38 still red?

mojtaba-komeili:
> which tests should this be turning green? i.e., cleaninstall and unittests_38 still red?

Yeah, I think `cleaninstall` is what causes the issue with the others too. There is some version conflict betweeen torch, torchvision and torchtext which looks impossible to solve.

klshuster:
> Yeah, I think `cleaninstall` is what causes the issue with the others too. There is some version conflict betweeen torch, torchvision and torchtext which looks impossible to solve.

There is some logic in how to assign versions 

can you try these? this is my current env:

```
torch                    1.13.1
torchaudio               0.13.1
torchvision              0.14.1
```

mojtaba-komeili:
OK, thanks. That was the magic combination. It cleaned up the clean install. Now four unit test are failing apex (1) and starspace (3). I would say let's merge these and work on those in another PR?

klshuster:
yep, so i already know the apex issue

pytorch 1.13.1 has an issue where `torch.load` doesn't respect custom pickle modules, so it's breaking when a model is loaded that we trained a while ago when apex was installed

for now, we may need to just mark that test as skippable if the pytorch version is 1.13.1

klshuster (approved):

PR #5005, opened 2023-04-11
Upgrading to torch 2.0

mojtaba-komeili:
# Patch description
* Trying to upgrade to torch 2.0 while passing all the relevant torch related tests.
* REMOVED the codecov that was reporting our code coverage. The package is deprecated ([link](https://github.com/codecov/codecov-python)) and removed from pip. 
* Changed the tolerances on some of the GPU tests (within a reasonable range) to pass the tests.
# Testing
1. The GPU and torch related tests are passing.
2. Installed ParlAI in a new conda environment and ran an eval with the WoW model on WizInt dataset.


mojtaba-komeili:
On the first push it broke a lot less tests than I was expecting. Really suspicious now.
Screenshot 2023-04-11 at 5 16 02 PM

mojtaba-komeili:
@klshuster I think it is ready for review. The failing test is from crowdsourcing.

.github/workflows/lint.yml
```diff
@@ -89,7 +89,7 @@ jobs:
           pip install -q -r requirements.txt
           pip install -q mypy mypy-extensions
           pip install -q git+https://github.com/numpy/numpy-stubs.git
-          pip install -q torch==1.5.1+cpu torchvision==0.6.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
+          pip install -q torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
```
klshuster, line 92: we may want to pin this to a torch version (2.0.0?)

klshuster (approved):
amazing, thanks!

two small nits but otherwise looks great

.github/workflows/lint.yml
mojtaba-komeili: I tried copying directly from PyTorch Install instruction. This seems to be the recommended way now. We could either mend it if/when it broke later, or have it as is. The problem with pinning to a special version has been haunting our tests for a long time. Let's see what happens with keeping them to  lower level for a while?

PR #5006, opened 2023-04-11
update onboarding with autofail and handle radio button checking

meganung:
- update `handleOnboardingSubmit` function to handle the `must_pass` option in the onboarding config, if `true` then the user must get that question correct else they automatically fail the onboarding test (can retry but fail that try). 
- if radio buttons are being used, user isn't penalized twice for 1 question incorrect (since you can only select one answer), prior to this fix, if a user selected the wrong answer and the type was radio buttons, the incorrect count would be 2 for that 1 question since they checked the incorrect answer and didn't check the correct answer (checkbox logic).


EricMichaelSmith (approved):
Yeah this looks great, thanks! Definitely appreciate noticing the issue with counting every single wrong radio button press as an individual incorrect answer @__@

PR #5014, opened 2023-04-18
Rework Installation Instructions

Bengt:
**Patch description**

- Add some structure to clearly distinguish texts regarding the operating system, Python interpreter, and virtual environment
- Reformat end user and developer installation headings to use ATX-style headers for cleaner use of markdown
- Extend instructions to use a virtual environment as recommended in the text

**Testing steps**

- Run the instructions on a supported operating system and observe that they work flawlessly as-is.

**Other information**

These fixes are largely stylistic, but the instructions are now more complete and actually work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,22 +58,35 @@ For those who want to start with ParlAI now, you can try our [Colab Tutorial](ht
 
 ## Installing ParlAI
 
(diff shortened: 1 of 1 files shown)


Bengt:
I just signed the CLA.

PR #5015, opened 2023-04-18
Rework Installation Instructions

Bengt:
**Patch description**

- Add some structure to clearly distinguish texts regarding the operating system, Python interpreter, and virtual environment
- Reformat end user and developer installation headings to use ATX-style headers for cleaner use of markdown
- Extend instructions to use a virtual environment as recommended in the text

**Testing steps**

- Run the instructions on a supported operating system and observe that they work flawlessly as-is.

**Other information**

These fixes are largely stylistic, but the instructions are now more complete and actually work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,22 +58,35 @@ For those who want to start with ParlAI now, you can try our [Colab Tutorial](ht
 
 ## Installing ParlAI
 
-ParlAI currently requires Python3.8+ and [Pytorch](https://pytorch.org) 1.6 or higher.
-Dependencies of the core modules are listed in [`requirements.txt`](https://github.com/facebookresearch/ParlAI/blob/main/requirements.txt). Some
-models included (in [`parlai/agents`](https://github.com/facebookresearch/ParlAI/tree/main/parlai/agents)) have additional requirements.
-We *strongly* recommend you install ParlAI in a [venv](https://docs.python.org/3/library/venv.html) or [conda](https://www.anaconda.com/) environment.
+### Operating System
 
-We do not support Windows at this time, but many users [report success on Windows using Python 3.8](https://github.com/facebookresearch/ParlAI/issues/3989) and issues with Python 3.9. We are happy to accept patches that improve Windows support.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,30 +58,46 @@ For those who want to start with ParlAI now, you can try our [Colab Tutorial](ht
...
+### End User Installation
 
 If you want to use ParlAI without modifications, you can install it with:
 
 ```bash
-pip install parlai
+cd /path/to/your/parlai-app
+python3.8 -m venv venv
+venv/bin/pip install --upgrade pip setuptools wheel
+venv/bin/pip install parlai
```
klshuster, line 86: This looks unnecessarily complex when all we are asking is for a `pip install`

README.md
```diff
@@ -58,30 +58,46 @@ For those who want to start with ParlAI now, you can try our [Colab Tutorial](ht
...
 environment, run the following commands to clone the repository and install
 ParlAI:
 
 ```bash
 git clone https://github.com/facebookresearch/ParlAI.git ~/ParlAI
-cd ~/ParlAI; python setup.py develop
+cd ~/ParlAI
+python3.8 -m venv venv
+venv/bin/pip install --upgrade pip setuptools wheel
+venv/bin/python setup.py develop
```
klshuster, line 100: ditto here; does this fail if you do not specify `venv/bin/python`?

Bengt: Same as before: `python` is the system Python interpreter, because the venv was never activated, which I suppose is the more portable option. Consider that users of the readme are usually copying these commands, so there is a tradeoff to be made between the legibility of the command and readability of the instructions. I mean, when there are options for many shells listed, users get easily confused. This portable version should work for all of them and is still easy enough to follow, I feel like.

Bengt:
@klshuster Thanks for having a look at this. I would be glad to contribute. I was in need of better installation instructions, and I hope I can help the next person in my position.

klshuster:
Could you please rebase onto the main branch? then I think we can go ahead and merge this

klshuster (approved):

PR #5016, opened 2023-04-18
T5 and GPT2 fixes

klshuster:
**Patch description**
Proper fixes that were introduced in #4648 . 

**Testing steps**
Same testing plan as #4648


parlai/agents/hugging_face/t5.py
```diff
@@ -149,6 +149,15 @@ def build_dictionary(self):
         """
         return T5DictionaryAgent(self.opt)
 
+    @staticmethod
+    def dictionary_class():
+        """
+        Return the dictionary class that this agent expects to use.
+
+        Can be overridden if a more complex dictionary is required.
```
jxmsML, line 157: if overriden then do we need to change L150 to be overrideable?

jxmsML (approved):

parlai/agents/hugging_face/t5.py
klshuster: ahh i had just copied from the original PR, that docstring is out of date let me fix it

PR #5018, opened 2023-04-20
Sweeping the JS lint errors under the rug

mojtaba-komeili:
**Patch description**
Now that our JS lint is working, we have so many legacy JS code in our crowdsourcing. This PR fixes the ones that it can and adds lint ignore commands on those that are not trivial to resolve.


mojtaba-komeili:
@klshuster sorry for the number of changes. But there is no changes to the logic of how codes work here. It is either fixing the lint issues in JS or adding skip conditions on them.

klshuster (approved):

PR #5021, opened 2023-04-28
Add option to collect inference metrics from an OPT server

davides:
**Patch description**
Adding an option to collect resource utilization metrics from an OPT server.

Depends on: https://github.com/facebookresearch/metaseq/pull/718

**Testing steps**
Ran a metaseq eval that passes this flag to the agent and records the metrics.


ellenxtan (approved):
LGTM! thanks! A small comment: is it possible to split the lines that are too long?

davides:
> Think this looks good to me. Could you please rebase onto (or merge) main to make sure tests pass?

Do you mind if I do that when we either bring the feature branch (davides/bb3_eval) back to main or rebase that? I think it'll get noisy to do it in this merge.

